PR #3, opened 2023-02-20
[mempool] `mempool.Item` interface + testing

samliok:
-- added an interface `mempool.Item` that is used by the mempool package
-- added unit testing to the mempool package
-- added comments to the mempool package

### DIFF
--- a/chain/dependencies.go
+++ b/chain/dependencies.go
@@ -61,7 +61,7 @@ type VM interface {
 
 type Mempool interface {
 	Len(context.Context) int
-	Add(context.Context, []mempool.MempoolItem)
+	Add(context.Context, []mempool.Item)
 	Build(
 		context.Context,
 		func(context.Context, *Transaction) (bool /* continue */, bool /* restore */, bool /* remove account */, error),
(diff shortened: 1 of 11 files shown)


chain/transaction.go
```diff
@@ -89,6 +89,8 @@ func (t *Transaction) ID() ids.ID { return t.id }
 
 func (t *Transaction) Expiry() int64 { return t.Base.Timestamp }
 
+func (t *Transaction) Value() uint64 { return t.Base.UnitPrice }
```
patrick-ogrady, line 92: I don't like this name. We should use `UnitPrice`

chain/dependencies.go
```diff
@@ -60,7 +61,7 @@ type VM interface {
 
 type Mempool interface {
 	Len(context.Context) int
-	Add(context.Context, []*Transaction)
+	Add(context.Context, []mempool.MempoolItem)
```
patrick-ogrady, line 64: Usually you should try to avoid the repeat of "mempool.Mempool" -> it is probably safe to say `mempool.Item`.

PR #14, opened 2023-02-21
[mempool] Use generics instead of `mempool.Item`

patrick-ogrady:
Using `mempool.Item` in `mempool.Add` prevents us from adding an array of transactions at once (can't add array of interfaces). Using generics helps us get around this constraint without giving up performance or generality.

### DIFF
--- a/mempool/sorted_mempool.go
+++ b/mempool/sorted_mempool.go
@@ -87,7 +87,7 @@ func (sm *SortedMempool[T]) SetMinVal(val uint64) []T {
 // PeekMin returns the minimum value in sm.
 func (sm *SortedMempool[T]) PeekMin() (T, bool) {
 	if sm.minHeap.Len() == 0 {
-		return *new(T), false
+		return *new(T), false //nolint:gocritic
 	}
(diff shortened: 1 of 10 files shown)


mempool/sorted_mempool.go
```diff
@@ -84,48 +84,48 @@ func (sm *SortedMempool) SetMinVal(val uint64) []Item {
...
-	return sm.minHeap.items[0].tx
+	return sm.minHeap.items[0].item, true
 }
 
 // PopMin removes the minimum value in sm.
-func (sm *SortedMempool) PopMin() Item {
+func (sm *SortedMempool[T]) PopMin() (T, bool) {
 	if sm.minHeap.Len() == 0 {
-		return nil
+		return *new(T), false
```
patrick-ogrady, line 98: One of the annoying things about generics is you can't do a `nil` check against T

patrick-ogrady: Otherwise you need to do this (but you need to assume `T` is a pointer, which it may not be.
image

PR #18, opened 2023-02-22
Add `TokenVM` example

patrick-ogrady:
Mint and swap arbitrary fungible/non-fungible tokens

## TODO
- [x] Port code from `IndexVM`
- [x] Remove all unneeded functionality
- [x] Get existing tests to pass
- [x] Add Integration/E2E Tests to CI
- [x] Add `Mint`
  - [x] Rotate `Owner` (can revoke if set to empty public key) - #28 
  - [x] Add test to mint after initial mint - #28 
  - [x] Count total supply on object/error if surpass MaxUint64 - #28 
  - [x] Add token Metadata (make updatable like minting) - #28 
  - [x] Use `TxID` as `AssetID` - #28 
- [x] Add `Swap`
  - [x] refund gas if fill order fails
  - [x] Add integration tests for all order-related actions
- [x] Add cache for open orders
  - [x] Add client/RPC for open orders
- [x] Update `HyperSDK` README
  - [x] Add `TokenVM` example
  - [x] Specify `StateKeys` are for DB keys, not hashes that frequently change
- [x] Update `TokenVM` README
  - [x] Add header image for `TokenVM`
  - [x] Add badges for `TokenVM`
  - [x] Add demo using CLI - #32
- [x] Return rate when orders are returned (instead of just in/out tick)

## Future Work
* Cleanup `E2E` and `Integration` tests (create standard components other HyperVMs can import to reduce duplicate code) - #31 
* Add simple order expiry (don't return in order book but still must be closed) 

## Breaking Changes
* Signature for `Actions.StateKeys` now includes `TxID`
* Signature for `UnpackPublicKey` now includes `Required` - #28


mempool/uint64_heap.go
```diff
@@ -9,90 +9,89 @@ import (
...
+	Val  uint64 // Value to be prioritized
 
-	index int // index of the entry in heap
+	Index int // Index of the entry in heap
 }
 
-// uint64Heap[T] is used to track pending transactions by [val]
-type uint64Heap[T Item] struct {
+// Uint64Heap[T] is used to track pending transactions by [Val]
+type Uint64Heap[T any] struct {
```
patrick-ogrady, line 21: We still use generics here so we don't have to cast the returned Item

PR #21, opened 2023-02-23
Optimizing `crypto/ed25519.go`

rafael-abuawad:
1. **Avoid unnecessary copying of byte arrays**, it is not necessary to make a copy of the byte array in these cases since the byte arrays are the same size and have the same underlying structure. 

2. **Use of constants instead of literals** To make it more declarative`const CompressedPublicKeySize = 33` instead `if len(paddr) != 33`

### DIFF
--- a/crypto/ed25519.go
+++ b/crypto/ed25519.go
@@ -21,6 +21,7 @@ type (
 )
 
 const (
+	CompressedPublicKeySize = 33
 	PublicKeyLen  = ed25519.PublicKeySize
 	PrivateKeyLen = ed25519.PrivateKeySize
 	SignatureLen  = ed25519.SignatureSize
@@ -51,7 +52,6 @@ func ParseAddress(hrp, saddr string) (PublicKey, error) {
 	if phrp != hrp {
 		return EmptyPublicKey, ErrIncorrectHrp
 	}
-	const CompressedPublicKeySize = 33
 	if len(paddr) != CompressedPublicKeySize {
 		return EmptyPublicKey, ErrInvalidPublicKey
 	}
@@ -75,7 +75,7 @@ func GeneratePrivateKey() (PrivateKey, error) {
 // PublicKey returns a PublicKey associated with the Ed25519 PrivateKey p.
 // The PublicKey is the last 32 bytes of p.
 func (p PrivateKey) PublicKey() PublicKey {
-	return PublicKey(p[32:])
+	return PublicKey(p[PublicKeyLen:])
 }
 
 // ToHex converts a PrivateKey to a hex string.
--- a/crypto/ed25519.go
+++ b/crypto/ed25519.go
@@ -24,7 +24,7 @@ const (
 	PublicKeyCompressedSize = 33
 	PublicKeyLen            = ed25519.PublicKeySize
 	PrivateKeyLen           = ed25519.PrivateKeySize
-	PrivateKeySeedSize      = ed25519.SeedSize
+	PrivateKeySeedLen       = ed25519.SeedSize
 	SignatureLen            = ed25519.SignatureSize
 )
 
@@ -57,7 +57,7 @@ func ParseAddress(hrp, saddr string) (PublicKey, error) {
 		return EmptyPublicKey, ErrInvalidPublicKey
 	}
 	var p PublicKey
-	copy(p[:], paddr[:32])
+	copy(p[:], paddr[:PublicKeyLen])
 	return p, nil
 }
 
@@ -76,7 +76,7 @@ func GeneratePrivateKey() (PrivateKey, error) {
 // PublicKey returns a PublicKey associated with the Ed25519 PrivateKey p.
 // The PublicKey is the last 32 bytes of p.
 func (p PrivateKey) PublicKey() PublicKey {
-	rpk := p[PrivateKeyLen:] // privateKey == private|public
+	rpk := p[PublicKeyLen:] // privateKey == private|public
 	var pk PublicKey
 	copy(pk[:], rpk)
 	return pk
--- a/crypto/ed25519.go
+++ b/crypto/ed25519.go
@@ -51,7 +51,8 @@ func ParseAddress(hrp, saddr string) (PublicKey, error) {
 	if phrp != hrp {
 		return EmptyPublicKey, ErrIncorrectHrp
 	}
-	if len(paddr) != 33 /* compressed public key size */ {
+	const CompressedPublicKeySize = 33
+	if len(paddr) != CompressedPublicKeySize {
 		return EmptyPublicKey, ErrInvalidPublicKey
 	}
 	var p PublicKey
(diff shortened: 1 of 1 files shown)


crypto/ed25519.go
```diff
@@ -51,7 +51,8 @@ func ParseAddress(hrp, saddr string) (PublicKey, error) {
 	if phrp != hrp {
 		return EmptyPublicKey, ErrIncorrectHrp
 	}
-	if len(paddr) != 33 /* compressed public key size */ {
+	const CompressedPublicKeySize = 33
```
patrick-ogrady, line 54: You should define this at the top of the file instead of inline (so we only need to define it once), not each call of `ParseAddress`.

crypto/ed25519.go
```diff
@@ -74,10 +75,7 @@ func GeneratePrivateKey() (PrivateKey, error) {
 // PublicKey returns a PublicKey associated with the Ed25519 PrivateKey p.
 // The PublicKey is the last 32 bytes of p.
 func (p PrivateKey) PublicKey() PublicKey {
-	rpk := p[32:] // privateKey == private|public
-	var pk PublicKey
-	copy(pk[:], rpk)
-	return pk
+	return PublicKey(p[32:])
```
patrick-ogrady, line 78: Very nice...not sure why I didn't do this in the first place.

You should use `PublicKeyLen` here instead of 32. I also think you should keep the commend that explains why we go 32 in.

rafael-abuawad: Done! Thank you!

crypto/ed25519.go
rafael-abuawad: I thought of that, I don't know why I left it there. Thanks.

crypto/ed25519.go
```diff
@@ -74,10 +75,7 @@ func GeneratePrivateKey() (PrivateKey, error) {
 // PublicKey returns a PublicKey associated with the Ed25519 PrivateKey p.
 // The PublicKey is the last 32 bytes of p.
 func (p PrivateKey) PublicKey() PublicKey {
-	rpk := p[32:] // privateKey == private|public
-	var pk PublicKey
-	copy(pk[:], rpk)
-	return pk
+	return PublicKey(p[PublicKeyLen:])
```
patrick-ogrady, line 78: Not to be too pedantic (this is the last change), I would use `PrivateKeyLen` here because that is what we are actually skipping.

I think I suggested the wrong change so this is on me.

patrick-ogrady:
Ahhh that's why I didn't use it. Is this feature new in go1.20?
```
Error: crypto/ed25519.go:78:20: cannot convert p[PublicKeyLen:] (value of type []byte) to type PublicKey
```

crypto/ed25519.go
```diff
@@ -75,7 +75,10 @@ func GeneratePrivateKey() (PrivateKey, error) {
 // PublicKey returns a PublicKey associated with the Ed25519 PrivateKey p.
 // The PublicKey is the last 32 bytes of p.
 func (p PrivateKey) PublicKey() PublicKey {
-	return PublicKey(p[PublicKeyLen:])
+	rpk := p[PrivateKeyLen:] // privateKey == private|public
```
patrick-ogrady, line 78: You can put a TODO here for #23 so we make sure to catch it.

patrick-ogrady: Or just leave it. LGTM now.

rafael-abuawad: Added the `PrivateKeyLen` variable to be more explicit, this MR is just some minor clean-ups.

patrick-ogrady:
🤦 lol forgot about this...sigh.

image

So question is do we want a separate const for "private key component" of privatekey. Looking in the `crypto` codebase to see if they have a name for this component.

patrick-ogrady:
Ahh they call this the `SeedSize`: https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/crypto/ed25519/ed25519.go;l=204

So let's add a constant called `PrivateKeySeedSize = ed25519.SeedSize` and use that.

rafael-abuawad:
> Ahhh that's why I didn't use it. Is this feature new in go1.20 (tests don't run on that yet)?
> 
> ```
> Error: crypto/ed25519.go:78:20: cannot convert p[PublicKeyLen:] (value of type []byte) to type PublicKey
> ```
> 
> I agree much cleaner (nice find)...may just need to update the versions rq first.
> 
> _Update: yeah looks like this added it [golang/go#46505](https://github.com/golang/go/issues/46505)._
> 
> _Update 2: we'll be able to remove a TON of memory allocations using this. creating an issue now to apply this change all over the codebase (#23)._

Yeah sorry about that, I'm fairly new with Go, and I didn't check for backward compatibility.

patrick-ogrady:
> > Ahhh that's why I didn't use it. Is this feature new in go1.20 (tests don't run on that yet)?
> > ```
> > Error: crypto/ed25519.go:78:20: cannot convert p[PublicKeyLen:] (value of type []byte) to type PublicKey
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I agree much cleaner (nice find)...may just need to update the versions rq first.
> > _Update: yeah looks like this added it [golang/go#46505](https://github.com/golang/go/issues/46505)._
> > _Update 2: we'll be able to remove a TON of memory allocations using this. creating an issue now to apply this change all over the codebase (#23)._
> 
> Yeah sorry about that, I'm fairly new with Go, and I didn't check for backward compatibility.

You've actually been able to do this since go1.17 but the old style was super gross, so I stayed away. But 1.20 makes it MUCH better: 
image

rafael-abuawad:
Should I define a `PrivateKeySeedSize` instead of `CompressedPublicKeySize`? Wouldn't that be 1 byte short?

patrick-ogrady:
> Should I define a `PrivateKeySeedSize` instead of `CompressedPublicKeySize`? Wouldn't that be 1 byte short?

You should define `PrivateKeySeedSize` in addition to `CompressedPublicKeySize`. 👍

patrick-ogrady:
Note the `Len` above just to be consistent.

rafael-abuawad:
@patrick-ogrady I don't understand why are we defining `PrivateKeySeedSize`?  😕

crypto/ed25519.go
```diff
@@ -74,7 +76,7 @@ func GeneratePrivateKey() (PrivateKey, error) {
 // PublicKey returns a PublicKey associated with the Ed25519 PrivateKey p.
 // The PublicKey is the last 32 bytes of p.
 func (p PrivateKey) PublicKey() PublicKey {
-	rpk := p[32:] // privateKey == private|public
```
patrick-ogrady, line 77: Put `PrivateKeySeedLen` here ^^

patrick-ogrady:
Found a few erroneous comments and am going to do a pass on this after I merge it. So, don't worry about getting it perfect rn.

crypto/ed25519.go
rafael-abuawad: Thank you!

patrick-ogrady:
@rafael-abuawad Going to merge this and make the lint fixes myself. Thanks for putting this up!

patrick-ogrady (approved):

patrick-ogrady:
Continues here: https://github.com/ava-labs/hypersdk/pull/26

patrick-ogrady:
Congrats on being the first external contributor! https://twitter.com/_patrickogrady/status/1628870022360895488

PR #48, opened 2023-02-25
Generic Heap

patrick-ogrady:
* Implement a single min/max heap instead of adding one per type (use generics)
* Add wrapper functions for `heap.Push`, `heap.Remove`, and `heap.Pop` so users of this package don't need to use `containers.Heap` as well
  * We only make the "incorrect" functions directly accessible on the `innerHeap`, which cannot be called outside of the package.


patrick-ogrady:
Going to wrap the inner heap struct so I can use Push/Pop instead of Add/Remove

heap/heap.go
```diff
@@ -0,0 +1,81 @@
...
+// which makes using this heap less error-prone.
+func (h *Heap[I, V]) Push(e *Entry[I, V]) {
+	heap.Push(h.ih, e)
+}
+
+// Pop can be called by external users to remove an object from the heap at
+// a specific index instead of using `containers.heap`,
+// which makes using this heap less error-prone.
+func (h *Heap[I, V]) Pop() *Entry[I, V] {
+	if len(h.ih.items) == 0 {
```
patrick-ogrady, line 56: We directly call into `ih` to avoid the overhead of another function call here

samliok:
could probably change bucketHeap in [emap] to use this instead?

PR #55, opened 2023-02-26
Update EMap to use Heap Package

samliok:
Change EMap to reuse the heap package( #48 ) for its binaryHeap struct.
Updated testing to reflect changes.

### DIFF
--- a/emap/emap.go
+++ b/emap/emap.go
@@ -4,12 +4,11 @@
 package emap
 
 import (
-	"container/heap"
-	"fmt"
 	"sync"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/hypersdk/heap"
 )
 
 type bucket struct {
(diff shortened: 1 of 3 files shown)


emap/emap.go
```diff
@@ -20,54 +19,9 @@ type bucket struct {
...
-
-// Peek returns the first bucket in bh. If bh has no
-// buckets, Peek returns nil.
-func (bh *bucketHeap) Peek() *bucket {
-	if len(bh.buckets) == 0 {
-		return nil
-	}
-	return bh.buckets[0]
-}
+// type bucketHeap struct {
```
patrick-ogrady, line 22: I think you meant to remove this?

emap/emap.go
```diff
@@ -91,9 +45,7 @@ func NewEMap[T Item]() *EMap[T] {
 	return &EMap[T]{
 		seen:  set.Set[ids.ID]{},
 		times: make(map[int64]*bucket),
-		bh: &bucketHeap{
-			buckets: []*bucket{},
-		},
+		bh:    heap.New[*bucket, int64](0, true),
```
patrick-ogrady, line 48: I think we should pre-allocate 120 here instead of 0.

emap/emap.go
patrick-ogrady: (maybe the comment block above too)?

patrick-ogrady (approved):

PR #57, opened 2023-02-26
Add Avalanche Warp Messaging Support

patrick-ogrady:
## TODO
- [x] block verify with context (ShouldVerifyWithContext && VerifyWithContext)
  - [x] check if block must have context (https://github.com/ava-labs/xsvm/blob/master/execute/expects_context.go#L13)
  - [x] skip verify if bootstrapping (getting rollback of validator state may slow down node)
- [x] async generate message signature and make queryable over P2P by anyone
- [x] allow actions to define a "warpable" payload: https://github.com/ava-labs/xsvm/blob/cbeff25ecbe3330cb3dc6effe065b1541c3bb77b/execute/tx.go#L88-L95
- [x] verify warp signature: https://github.com/ava-labs/xsvm/blob/cbeff25ecbe3330cb3dc6effe065b1541c3bb77b/execute/tx.go#L159-L181
- [x] every node is a relayer => query validators for their signatures and store on-disk
  - [ ] make this mode optional if you know you won't be using the node to generate new warp messages via the CLI
- [x] fetch collected signatures from API and aggregate in CLI: https://github.com/ava-labs/xsvm/blob/cbeff25ecbe3330cb3dc6effe065b1541c3bb77b/cmd/issue/importtx/cmd.go
- [x] verify multi-signature async and check at end of block (like with signatures)
- [x] add multi-plexed network router  - #59 
  - [x] append handler id at end to avoid possible array copy
- [x] Run verify with context
- [x] Add logging to networking failures
- [x] Add quick verification check before BLS sig check that each is allowed
- [x] Don't make app request directly from failed
- [x] https://github.com/ava-labs/hypersdk/pull/62
  - [x] allow transfer to optionally reward tx submitter (relayer)
  - [x] charge more for an action if it has more signers (blocked on `warp.Signature.NumSigners` method
  - [x] protect against warp message replay by storing accepted warp IDs in state (cannot be based on signature, as this is mutable)
  - [x] track balance exported to a given region and don't let more come back (https://github.com/ava-labs/xsvm/blob/cbeff25ecbe3330cb3dc6effe065b1541c3bb77b/state/storage.go#L133-L200)
  - [x] set assetID on other chains as ID(sourceChainID, assetID) -> put raw info in metadata with empty owner
  - [x] assets take on a new identity when they flow through a subnet (README) -> recommended point-to-point
  - [x] handle fund return (when importing, track with existing assetID rather than a new one)
  - [x] must make imported asset special to avoid people copying pattern and exporting unimported balances
  - [x] should scope import/export to a specific chain so can contain a bug
  - [x] store warp messages in state so they can be signed by new validators
  - [x] (not needed) ~pad bitset if last X elements are not signers (length of signers in signature will be < validators)?: https://github.com/ava-labs/avalanchego/blob/d755f872a4bf0de12297b3994b729ea684f78519/vms/platformvm/warp/validator.go#L97-L105~
  - [x] charge fee in VM based on num signers: https://github.com/ava-labs/hypersdk/pull/71
  - [x] check wrap TxID is not duplicate before performing any validation
  - [x] we must mark the tx as fee-paying if the warp message is not valid, rather than marking the block as invalid (DoS)
  - [x] implement build block with context...if given a block without this, make sure not to include any warp messages
  - [x] improve name for "StateMapping"
  - [x] add provide warp message to action unmarshal (so can parse inner message and provide state keys)
  - [x] codify warp support in registry so don't have to have a check in every unmarshal
  - [x] add warp E2E transfer test (native and non-native round trip): https://github.com/ava-labs/hypersdk/pull/77
  - [x] ensure we only persist warpID if tx successful
  - [x] make sure warpID just includes payload (otherwise could do duplicate with anycast and chainID)
  - [x] add import/export to registry and metrics
  - [x] handle no signer case/empty payload/incorrect signature
    - [x] test odd warp messages in integration test
  - [x] block non-return export of imported assets (force point-to-point)
- [x] scope requestID to NodeID to avoid overflow with course tracking (avalanchego is ok as long as you don't reuse nodeID+requestID)
- [x] re-trigger wrap signature fetch when calling GetSignatures for any validators that have not yet sent sigs (and that aren't pending) -> protects against shutdowns and validator rotation
- [x] if signature requested from us and we haven't generated yet, generate and store (may have state synced past tx)
- [x] add a "trade-in" option where a relayed message can specify a max rate it will pay for a trade to get the native currency (relayer can then link an order that will execute upon tx acceptance) -> should just force orderIDs willing to accept + input (if none valid, just process with no swap) - https://github.com/ava-labs/hypersdk/pull/80
- [ ] ~add warp message cleanup refund (if user clears item from state, get some % of original fee back -> user-driven ack) -> may be simpler if we just say each user can have at most 1 pending warp message at a time? -> just limit outstanding per account -> lockup + refund (unrelated to current network fee, so no gas token)~
- [x] add network/account switch to token-cli (so can easily transfer between known networks/use stored APIs) - https://github.com/ava-labs/hypersdk/pull/86
  - [x] persist default on shutdown and show each run
  - [x] must manually add networks instead of defaulting to local
  - [x] add clear defaults in case of restart
  - [x] if no default set yet, request on startup (if not genesis function)
  - [x] add warp support to explorer
  - [x] add multiple endpoints per network
- [x] make `warpID` just be the hash of the payload (store by sourceChainID) - https://github.com/ava-labs/hypersdk/pull/85
- [x] add GitHub action timeouts in case of infinite loop bug
- [ ] ~only enable BLS_PORTABLE on macos/windows~
- [x] Add BLS support to ANR: (https://github.com/ava-labs/hypersdk/pull/82 + https://github.com/ava-labs/avalanche-network-runner/pull/467)
  - [x] should be able to get down to a single restart (add nodes, create subnets, add blockchain configs, restart, add subnet validators, create chains)
- [x] warp result should be boolean - https://github.com/ava-labs/hypersdk/pull/92
  - [x] attach AWM verification results to block for bootstrapping/lite clients
- [x] update TokenVM demo with a warp swap (move to `Subnet B` to swap a minted token) - https://github.com/ava-labs/hypersdk/pull/93
- [x] Update docs - https://github.com/ava-labs/hypersdk/pull/93
- [x] drop warp messages from builder if expired - https://github.com/ava-labs/hypersdk/pull/93


patrick-ogrady:
Update to `v1.9.10-rc.2` to get support for fetching BLS public key from snowCtx

PR #70, opened 2023-03-01
[workers] Testing + Comments

samliok:
Added testing for the workers package.


workers/workers_test.go
```diff
@@ -131,3 +136,147 @@ func BenchmarkSerial(b *testing.B) {
...
+		require.NoError(err)
+		for j := 0; j < 10; j++ {
+			task := func() error {
+				atomic.AddInt64(&vals, 1)
+				return nil
+			}
+			job.Go(task)
+		}
+		// TODO: Better way to do this?
+		time.Sleep(time.Millisecond * 2)
```
samliok, line 240: not sure why leaving this out fails the test

docs/index.html
```diff
@@ -1,4 +1,9 @@
 <!DOCTYPE html>
+<!--
```
patrick-ogrady, line 2: Oooo nice

workers/workers.go
```diff
@@ -97,7 +102,7 @@ func (w *Workers) startWorker() {
 				w.stoppedWorkers <- struct{}{}
 				return
 			case j := <-w.tasks:
-				// Check if we should even do the work
+				// Check if we should even do the work\
```
patrick-ogrady, line 105: remove the `\`

workers/workers.go
```diff
@@ -142,25 +149,33 @@ type Job struct {
...
 	j.tasks <- f
 }
 
+// Done closes the tasks channel, then waits for the job's completed channel
+// to be closed. Calls [f] after the tasks have been completed if [f] is not null.
 func (j *Job) Done(f func()) {
 	close(j.tasks)
 	if f != nil {
-		// Callback when completed (useful for tracing)
+		// Callback  when completed (useful for tracing)
```
patrick-ogrady, line 162: extra space?

workers/workers_test.go
patrick-ogrady: What is the fail that we see?

patrick-ogrady: That is odd

patrick-ogrady: May be indicative of some race between `Go` and `Done`.

patrick-ogrady: ooooo I think I know why.

You need to call `job.Wait()` in another loop somewhere. Just calling `Done()` does not block for things to finish.

patrick-ogrady: Calling `w.Stop()` before all jobs have been waited on means that you could exit before processing everything.

workers/workers.go
```diff
@@ -75,6 +79,7 @@ func (w *Workers) processQueue() {
 			w.lock.Lock()
 			close(j.completed)
 			j.result <- w.err
+
```
patrick-ogrady, line 82: why `\n`?

workers/workers.go
```diff
@@ -157,10 +170,15 @@ func (j *Job) Done(f func()) {
 	}
 }
 
+// Wait returns the value received by the j.result channel.
 func (j *Job) Wait() error {
 	return <-j.result
 }
 
+// NewJob creates a new job and adds it to the workers' queue. [taskBacklog]
+// specifies the maximum number of  tasks that can be added to the created job
```
patrick-ogrady, line 179: unnecessary space

workers/workers.go
```diff
@@ -157,10 +170,15 @@ func (j *Job) Done(f func()) {
 	}
 }
 
+// Wait returns the value received by the j.result channel.
```
patrick-ogrady, line 173: Should indicate this only occurs once all tasks for a job are completed and after `Done` has been called.

workers/workers_test.go
```diff
@@ -131,3 +135,180 @@ func BenchmarkSerial(b *testing.B) {
...
+func TestJobWait(t *testing.T) {
+	require := require.New(t)
+	job := Job{
+		tasks:     make(chan func() error),
+		completed: make(chan struct{}),
+		result:    make(chan error),
+	}
+	testError := errors.New("TestError")
+	go func() {
+		job.result <- testError
```
patrick-ogrady, line 148: recommend putting a sleep before this

workers/workers_test.go
```diff
@@ -131,3 +135,180 @@ func BenchmarkSerial(b *testing.B) {
 		})
 	}
 }
+
+func TestJobWait(t *testing.T) {
```
patrick-ogrady, line 139: This test is very tuned to the internals of the implementation, usually you would not do this to test a package (makes refactor tough to do).

patrick-ogrady: Your "gut check" for this sort of thing should be: "am I using the provided initializer or am I creating the struct from scratch to get a test to run"

workers/workers_test.go
```diff
@@ -131,3 +135,180 @@ func BenchmarkSerial(b *testing.B) {
...
+			job.Go(task)
+		}
+		go func() {
+			defer wg.Done()
+			err := job.Wait()
+			require.NoError(err, "Error waiting on job.")
+		}()
+		job.Done(nil)
+	}
+	finished := make(chan bool)
```
patrick-ogrady, line 256: You should enqueue the jobs you created and just wait on each one out here

workers/workers_test.go
```diff
@@ -131,3 +135,180 @@ func BenchmarkSerial(b *testing.B) {
...
+		job, err := w.NewJob(10)
+		require.NoError(err)
+		for j := 0; j < 10; j++ {
+			task := func() error {
+				valLock.Lock()
+				defer valLock.Unlock()
+				val += 1
+				return nil
+			}
+			job.Go(task)
```
patrick-ogrady, line 247: usually you would do:
```
job.Go(func() error {
valLock.Lock()
				defer valLock.Unlock()
				val += 1
				return nil
})
```

workers/workers_test.go
patrick-ogrady: something like:
```
for _, j := range jobs {
  require.NoError(j.Wait(), "Error waiting on job.")
}
```

patrick-ogrady: You can then get rid of all this "finish" logic

you can call `w.Stop` after ensuring all jobs finished

workers/workers_test.go
samliok: so Job is an internal struct used by Workers. Basically you're saying we never use Job by itself so theres no point writing tests for it?

patrick-ogrady: Not that there isn't a "point per se" but generally people avoid writing tests for internal structs because it makes refactors much more difficult (as you usually try to maintain the same functional interface with changes behind the scenes -> AKA invariants/expectations typically only apply to the public interface not anything internal).

In this case, we can keep it because it is so simple but just a heads up.

patrick-ogrady (approved):

PR #87, opened 2023-03-09
[Codec] Testing + Comments

### LINKED ISSUE #61: Ensure optional packer asserts 0 values for all non-registered indices

samliok:
-- Added testing and relevant comments for the codec package
-- fixes #61


codec/optional_packer.go
```diff
@@ -69,12 +88,18 @@ func (o *OptionalPacker) PackID(id ids.ID) {
 	o.setBit()
 }
 
+// UnpackID unpacks an id into [dest] if the bitmask is set at the current offset.
+// Increments offset regardless.
 func (o *OptionalPacker) UnpackID(dest *ids.ID) {
 	if o.checkBit() {
 		o.ip.UnpackID(true, dest)
+	} else {
```
samliok, line 96: #61 was referring to this?

codec/packer.go
```diff
@@ -104,21 +120,24 @@ func (p *Packer) PackPublicKey(src crypto.PublicKey) {
...
 	}
 }
 
 func (p *Packer) PackSignature(src crypto.Signature) {
 	p.p.PackFixedBytes(src[:])
 }
 
+// UnpackPublicKey crypto.Signature into [dest].
+// TODO: should add required param?
 func (p *Packer) UnpackSignature(dest *crypto.Signature) {
```
samliok, line 137: some functions do not have required, should we be consistent and add required to all?

codec/optional_packer_test.go
```diff
@@ -0,0 +1,142 @@
...
+		var unpackedPubkey crypto.PublicKey
+		// Unpack
+		opr.UnpackPublicKey(&unpackedPubkey)
+		require.Equal(crypto.EmptyPublicKey[:], unpackedPubkey[:], "PublicKey unpacked correctly")
+		opr.UnpackPublicKey(&unpackedPubkey)
+		require.Equal(pubKey, unpackedPubkey, "PublicKey unpacked correctly")
+	})
+}
+
+func TestOptionalPackerID(t *testing.T) {
```
samliok, line 83: Im only testing new Pack/Unpack methods not in wrappers.Packer. Could add tests for all pack/unpack methods but wasn't sure if that was necessary.

codec/optional_packer.go
```diff
@@ -9,21 +9,31 @@ import (
 	"github.com/ava-labs/hypersdk/crypto"
 )
 
+// Represents a bitmask.
 // Inspired by: https://yourbasic.org/golang/bitmask-flag-set-clear
 type bits uint8
```
patrick-ogrady, line 14: We should probably just replace this with `BitSet`: https://github.com/ava-labs/avalanchego/blob/c5d8cd3b17747a6dfb3eae71de711a764ebd69f5/utils/set/bits.go#L12-L102

Would be great if we did that migration here.

patrick-ogrady: We can check that an item is properly represented by `b.Bytes() == bytes`

patrick-ogrady: We can also check that the BitLen is never greater than the registered bits.

codec/optional_packer.go
```diff
@@ -4,62 +4,94 @@
...
+	ip       *Packer
 }
 
-func NewOptionalWriter() *OptionalPacker {
+// NewOptionalWriter returns an instance of OptionalPacker that includes
+// a new Packer instance with MaxSize set to the maximum size. The maximum items
+// OptionalPacker can hold is set to [size]. If [size] > MaxItems sets
+// OptionalPackers MaxItems to MaxItems
+func NewOptionalWriter(size uint8) *OptionalPacker {
+	if size > MaxItems {
```
samliok, line 35: should we include a check here?

codec/optional_packer.go
samliok: updated

codec/optional_packer.go
```diff
@@ -4,62 +4,94 @@
...
 	"github.com/ava-labs/hypersdk/consts"
 	"github.com/ava-labs/hypersdk/crypto"
 )
 
-// Inspired by: https://yourbasic.org/golang/bitmask-flag-set-clear
-type bits uint8
+// MaxItems stores the maximum items an optional packer can hold.
+// Since a bitset can range from [0, INT_MAX] the max number of
+// flags in a bitset is log2(INT_MAX).
+var MaxItems = uint8(math.Log2(float64(consts.MaxInt)))
```
patrick-ogrady, line 18: Let's make this a much smaller number. How about 64?

patrick-ogrady: I'd also be ok to remove this entirely now

samliok: MaxItems is actually 63 in this case

PR #120, opened 2023-03-24
[Pubsub] Adding a pubsub package for server/client connections

samliok:
Adding a generic pubsub package to the hypersdk
to be used by decesionsRPCServer and blocksRPCServer and others

Why WS over HTTP/2 (outside of AvalancheGo-compatibility): https://www.rn.inf.tu-dresden.de/hara/arbeiten/SCC_HS_SS2017_Fietze_-_HTTP-2%20Streams_-_Is_the_Future_of_WebSockets_decided.pdf


patrick-ogrady:
```
to be used by decesionsRPCServer and blocksRPCServer and others
```

My original thought was to just have 1 server after this. We'd then multiplex traffic over the single connection based on what the client subscribed to.

pubsub/connection.go
```diff
@@ -0,0 +1,166 @@
...
+	s *Server
+
+	// The websocket connection.
+	conn *websocket.Conn
+
+	// Buffered channel of outbound messages.
+	send chan interface{}
+
+	// Represents if the connection can receive new messages.
+	active uint32
```
patrick-ogrady, line 39: We should just use an `atomic.Bool` here.

pubsub/connection.go
```diff
@@ -0,0 +1,166 @@
...
+
+// connection is a representation of the websocket connection.
+type Connection struct {
+	s *Server
+
+	// The websocket connection.
+	conn *websocket.Conn
+
+	// Buffered channel of outbound messages.
+	send chan interface{}
```
patrick-ogrady, line 36: this should be `send chan []byte`

patrick-ogrady: We should not impose any marshaling/unmarshaling concerns on the pubsub.

pubsub/connection.go
```diff
@@ -0,0 +1,166 @@
...
+
+// Send sends [msg] to c's send channel and returns whether the message was sent.
+func (c *Connection) Send(msg interface{}) bool {
+	if !c.isActive() {
+		return false
+	}
+	select {
+	case c.send <- msg:
+		return true
+	default:
```
patrick-ogrady, line 64: You should log if we drop it

pubsub/server.go
```diff
@@ -0,0 +1,98 @@
...
+		return
+	}
+	conn := &Connection{
+		s:         s,
+		conn:      wsConn,
+		send:      make(chan interface{}, maxPendingMessages),
+		active:    1,
+		rCallback: s.rCallback,
+	}
+	s.addConnection(conn)
```
patrick-ogrady, line 65: Don't create a new var `conn`. Just instantiate this here like:
```golang
s.addConnection(&Connection{
		s:         s,
		conn:      wsConn,
		send:      make(chan interface{}, maxPendingMessages),
		active:    1,
		rCallback: s.rCallback,
	})
```

pubsub/server.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Publish sends msg from [s] to [toConns].
+func (s *Server) Publish(msg interface{}, toConns *Connections) {
+	for _, conn := range toConns.Conns() {
+		// check server has connection O(1)
+		if !s.conns.conns.Contains(conn) {
```
patrick-ogrady, line 72: use the exported `Conns()` instead of `conns`.

pubsub/server_test.go
```diff
@@ -0,0 +1,296 @@
...
+	// Go routine that listens on dummyAddress for connections
+	go func() {
+		defer close(serverDone)
+		err := srv.ListenAndServe()
+		require.ErrorIs(err, http.ErrServerClosed, "Incorrect error closing server.")
+	}()
+	// Connect to pubsub server
+	u := url.URL{Scheme: "ws", Host: *dummyAddr}
+	webCon, resp, err := websocket.DefaultDialer.Dial(u.String(), nil)
+	defer resp.Body.Close()
```
patrick-ogrady, line 76: Usually you only close after checking `err` as `resp` may be nil if `err!=nil`

pubsub/consts.go
```diff
@@ -0,0 +1,33 @@
...
+
+package pubsub
+
+import (
+	"time"
+
+	"github.com/ava-labs/avalanchego/utils/units"
+)
+
+const (
```
patrick-ogrady, line 12: You need to make these configurable when constructing the server. Many of these are insufficient.

pubsub/connection.go
```diff
@@ -0,0 +1,166 @@
...
+				)
+				return
+			}
+			if !ok {
+				// The hub closed the channel. Attempt to close the connection
+				// gracefully.
+				_ = c.conn.WriteMessage(websocket.CloseMessage, []byte{})
+				return
+			}
+			if err := c.conn.WriteJSON(message); err != nil {
```
patrick-ogrady, line 150: This won't be JSON. It'll be bytes.

pubsub/connection.go
```diff
@@ -0,0 +1,166 @@
...
+	ErrFilterNotInitialized = errors.New("filter not initialized")
+	ErrAddressLimit         = errors.New("address limit exceeded")
+	ErrInvalidFilterParam   = errors.New("invalid bloom filter params")
+	ErrInvalidCommand       = errors.New("invalid command")
+)
+
+// Callback type is used as a callback function for the
+// WebSocket server to process incoming messages.
+// Accepts a byte message, the connection and any additional information.
+type Callback func([]byte, *Connection, []interface{}) []byte
```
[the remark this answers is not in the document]
patrick-ogrady: We will wrap this in a `registry` that decodes the inputs and routes them to the write `server` so they can send to the connection when needed.

pubsub/connection.go
```diff
@@ -0,0 +1,166 @@
...
+				websocket.CloseGoingAway,
+				websocket.CloseAbnormalClosure,
+			) {
+				c.s.log.Debug("unexpected close in websockets",
+					zap.Error(err),
+				)
+			}
+			break
+		}
+		if c.rCallback != nil {
```
patrick-ogrady, line 106: You should avoid invoking `callback` if the message is a `ping`

patrick-ogrady: Actually, I think that is handled for you with:
```
	c.conn.SetPongHandler(func(string) error {
		return c.conn.SetReadDeadline(time.Now().Add(pongWait))
	})
```

pubsub/server.go
```diff
@@ -0,0 +1,98 @@
...
+var upgrader = websocket.Upgrader{
+	ReadBufferSize:  readBufferSize,
+	WriteBufferSize: writeBufferSize,
+	CheckOrigin: func(*http.Request) bool {
+		return true
+	},
+}
+
+// Server maintains the set of active clients and sends messages to the clients.
+type Server struct {
```
patrick-ogrady, line 26: I recommend adding a small comment about how you would dial this server. I think you just use an included `gorilla` client?

pubsub/server.go
```diff
@@ -0,0 +1,98 @@
...
+	conns Connections
+	// Callback function when server receives a message
+	rCallback Callback
+	//  Supplemental info for the rCallback function
+	e []interface{}
+}
+
+// New returns a new Server instance with the logger set to [log]. The callback
+// function [f] is called by the server in response to messages if not nil.
+func New(log logging.Logger, r Callback, e []interface{}) *Server {
```
patrick-ogrady, line 39: One really nice feature would be to allow us to register handlers directly with this constructor. Then handlers we wouldn't need a root handler later that strips the first byte and routes elsewhere.

If you don't know what I mean, I can do this.

pubsub/server.go
```diff
@@ -0,0 +1,98 @@
...
+	return &Server{
+		log:       log,
+		rCallback: r,
+		conns:     *newConnections(),
```
patrick-ogrady, line 43: This is usually a sign that the var type you have in your struct is suboptimal. As I commented elsewhere, this should be `*Connections`

pubsub/server.go
```diff
@@ -0,0 +1,98 @@
...
+		e:         e,
+	}
+}
+
+// ServeHTTP adds a connection to the server, and starts go routines for
+// reading and writing.
+func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	wsConn, err := upgrader.Upgrade(w, r, nil)
```
patrick-ogrady, line 51: nit: add a comment why this is nil

pubsub/server.go
```diff
@@ -0,0 +1,161 @@
...
+	// The address to listen on
+	addr string
+	log  logging.Logger
+	lock sync.RWMutex
+	// conns a set of all our connections
+	conns *Connections
+	// Callback function when server receives a message
+	rCallback Callback
+	// Size of the ws read buffer
+	readBufferSize int
```
patrick-ogrady, line 41: You should put all of these in a `Config` struct instead of storing them directly in the struct. This also makes passing to `New` much easier.

Do something like: https://github.com/ava-labs/hypersdk/blob/a73396d63b87501b8423fff0187ebcb69d88ba07/pebble/db.go#L34-L53

pubsub/server.go
```diff
@@ -0,0 +1,161 @@
...
+	"time"
+
+	"github.com/gorilla/websocket"
+
+	"go.uber.org/zap"
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+)
+
+var upgrader = websocket.Upgrader{
```
patrick-ogrady, line 20: You should just initialize this in `New` and set the following there:
```
	upgrader.ReadBufferSize = s.readBufferSize
	upgrader.WriteBufferSize = s.writeBufferSize
```

pubsub/server.go
```diff
@@ -17,6 +17,38 @@ import (
 	"github.com/ava-labs/avalanchego/utils/logging"
 )
 
+type ServerConfig struct {
```
patrick-ogrady, line 20: These should be uppercase (exported) so they can be set outside the package.

pubsub/server.go
```diff
@@ -37,22 +69,8 @@ type Server struct {
 	conns *Connections
 	// Callback function when server receives a message
 	rCallback Callback
```
patrick-ogrady, line 71: this should be `callback` instead of `rCallback`

pubsub/server.go
```diff
@@ -37,22 +69,8 @@ type Server struct {
...
-	// Time allowed to write a message to the peer.
-	writeWait time.Duration
-	// Time allowed to read the next pong message from the peer.
-	pongWait time.Duration
-	// Send pings to peer with this period. Must be less than pongWait.
-	pingPeriod time.Duration
-	// ReadHeaderTimeout is the maximum duration for reading a request.
-	readHeaderTimeout time.Duration
+	// Config variables
+	config ServerConfig
```
patrick-ogrady, line 73: make this `*ServerConfig`

pubsub/server_test.go
```diff
@@ -52,10 +52,7 @@ func TestServerPublish(t *testing.T) {
 	// Create a new logger for the test
 	logger := logging.NoLog{}
 	// Create a new pubsub server
-	server := New(*dummyAddr, nil, logger, readBufferSize,
-		writeBufferSize, maxPendingMessages, maxMessageSize, writeWait,
-		pongWait, readHeaderTimeout,
-	)
+	server := New(*dummyAddr, nil, logger, NewDefaultServerConfig())
```
patrick-ogrady, line 55: why not just provide "localhost:8080" here directly? Why is the input a `flag.String`?

pubsub/connection.go
```diff
@@ -100,14 +100,14 @@ func (c *Connection) readPump() {
 			}
 			break
 		}
-		if c.s.rCallback != nil {
+		if c.s.callback != nil {
```
patrick-ogrady, line 103: I think this is still "correct" but I don't think there should be a case where the callback is empty (why would we even run the server then)

pubsub/server.go
```diff
@@ -19,33 +19,33 @@ import (
...
+		ReadBufferSize:     ReadBufferSize,
+		WriteBufferSize:    WriteBufferSize,
+		MaxPendingMessages: MaxPendingMessages,
+		MaxMessageSize:     MaxMessageSize,
+		WriteWait:          WriteWait,
+		PongWait:           PongWait,
+		PingPeriod:         (9 * PongWait) / 10,
+		ReadHeaderTimeout:  ReadHeaderTimeout,
 	}
 }
```
patrick-ogrady, line 51: I think I could argue that `CheckOrigin` should eventually be in the config but it is good for now.

patrick-ogrady (approved):

Issue #64, opened 2023-02-28
Update `vm/streaming` to properly maintain WS connections (read/write timeouts)


patrick-ogrady:

patrick-ogrady:
Recommend taking a look at how AvalancheGo does this: https://github.com/ava-labs/avalanchego/tree/c5d8cd3b17747a6dfb3eae71de711a764ebd69f5/pubsub

PR #137, opened 2023-04-12
Integrating new pubsub package into vm streaming

samliok:
Related: #64 
Requires: #158


vm/mock_vm
patrick-ogrady, line 1: This looks like an accidental commit.

vm/streaming.go
```diff
@@ -280,185 +77,86 @@ func (d *DecisionRPCClient) Close() error {
...
-	conn net.Conn
+	conn *websocket.Conn
 
 	ll sync.Mutex
 	cl sync.Once
 }
 
+// NewBlockRPCClient creates a new client for the blocks rpc server.
+// Dials into the server at [uri] and returns a client.
 func NewBlockRPCClient(uri string) (*BlockRPCClient, error) {
```
patrick-ogrady, line 129: We should unify these clients I think. Seems odd to require a separate client/port for each function.

examples/tokenvm/config/config.go
```diff
@@ -101,6 +100,8 @@ func (c *Config) setDefault() {
 	c.MempoolVerifyBalances = defaultMempoolVerifyBalances
 	c.StateSyncServerDelay = c.Config.GetStateSyncServerDelay()
 	c.StreamingBacklogSize = c.Config.GetStreamingBacklogSize()
+	// TODO: hardcoded for testing, idk why gorilla doesn't like port 0.
+	c.StreamingPort = 4000
```
patrick-ogrady, line 104: I'll unify this with the default server so that we don't even have an extra port after this.

listeners/listeners.go
```diff
@@ -25,115 +29,180 @@ type Transaction struct {
...
+	p.UnpackByte()
+	// read the txID from packer
+	var txID ids.ID
+	p.UnpackID(true, &txID)
+	// didn't unpack id correctly
+	if p.Err() != nil {
+		return ids.Empty, nil, nil, p.Err()
+	}
+
+	// TODO: from original Listen(), but can we receive a result and a decision error?
```
patrick-ogrady, line 146: Please remove this TODO

listeners/listeners.go
```diff
@@ -25,115 +29,180 @@ type Transaction struct {
...
+
+// Unpacks a tx message
+func UnpackTxMessage(msg []byte) (ids.ID, error, *chain.Result, error) {
+	p := codec.NewReader(msg, consts.MaxInt)
+	p.UnpackByte()
+	// read the txID from packer
+	var txID ids.ID
+	p.UnpackID(true, &txID)
+	// didn't unpack id correctly
+	if p.Err() != nil {
```
patrick-ogrady, line 142: You can wait to check for error until the end (one time)

listeners/listeners.go
```diff
@@ -25,115 +29,180 @@ type Transaction struct {
...
+			listeners,
+		)
 		delete(w.txListeners, txID)
 		// [expiringTxs] will be cleared eventually (does not support removal)
 	}
 }
+
+// Could be a better place for these methods
+// Packs an accepted block message
+func PackAcceptedTxMessage(p *codec.Packer, txID ids.ID, result *chain.Result) {
```
patrick-ogrady, line 119: You should check for packer error at the end of these IMO.

listeners/listeners.go
```diff
@@ -25,115 +29,180 @@ type Transaction struct {
...
-func (w *Listeners) RemoveBlockListener(id ids.ID) {
-	w.blockL.Lock()
-	defer w.blockL.Unlock()
-
-	delete(w.blockListeners, id)
-}
-
 // If never possible for a tx to enter mempool, call this
-func (w *Listeners) RemoveTx(txID ids.ID, err error) {
+func (w *Listeners) RemoveTx(txID ids.ID, err error, streamingServer *pubsub.Server) {
```
patrick-ogrady, line 58: Is there a reason you don't just set the streaming server in `New()`?

examples/tokenvm/config/config.go
patrick-ogrady: If you want to make this change yourself, all you do is this: https://github.com/ava-labs/coreth/blob/61baf6dc01755513c195bd54c218b942d2d113d7/plugin/evm/vm.go#L1176-L1184

patrick-ogrady: Create one of these: https://github.com/ava-labs/hypersdk/blob/0907bf7c016c3ab569952201270e37cdfb8592b1/utils/utils.go#L35-L59

But call it `NewWebsocketHandler`. Change the name of the existing one to `NewRPCHandler`: https://github.com/ava-labs/hypersdk/blob/0907bf7c016c3ab569952201270e37cdfb8592b1/utils/utils.go#L35-L59

patrick-ogrady: Add it as an endpoint here under a new vm config called `vm.WebsocketEndpoint`: https://github.com/ava-labs/hypersdk/blob/0907bf7c016c3ab569952201270e37cdfb8592b1/examples/tokenvm/controller/controller.go#L128-L135

Then you can just use the standard port.

listeners/listeners.go
samliok: added for unpacking, checking during pack might be redundant since unpack checks for errors anyway?

vm/streaming.go
```diff
@@ -5,460 +5,131 @@ package vm
...
+	conn *websocket.Conn
+	wl   sync.Mutex
+	dll  sync.Mutex
+	bll  sync.Mutex
+	cl   sync.Once
+}
+
+// NewDecisionRPCClient creates a new client for the decision rpc server.
+// Dials into the server at [uri] and returns a client.
+func NewStreamingClient(uri string) (*Client, error) {
```
patrick-ogrady, line 37: TODO: move to `listeners`

PR #140, opened 2023-04-14
Avoid `tstate` copy + distinguish between empty/non-existent values

patrick-ogrady:


chain/processor.go
```diff
@@ -52,15 +51,16 @@ func (p *Processor) Prefetch(ctx context.Context, db Database) {
...
 		// Store required keys for each set
-		alreadyFetched := set.Set[string]{}
+		alreadyFetched := map[string][]byte{}
 		for _, tx := range p.blk.GetTxs() {
 			storage := map[string][]byte{}
 			for _, k := range tx.StateKeys(sm) {
 				sk := string(k)
-				if alreadyFetched.Contains(sk) {
+				if v, ok := alreadyFetched[sk]; ok {
+					storage[sk] = v
```
patrick-ogrady, line 60: Always populate storage as if we fetched keys

PR #144, opened 2023-04-16
TokenVM: added prompt for initial supply and if present, mint the tokens immediately

itinance:
This PR enhances the `create-asset` command of the TokenVM to prompt an initial supply (can still be 0 or empty), and if a value were given, the corresponding amount would be minted immediately to the current actor.

Also, the `assetID` will now be logged at the end. Without passing a value as a supply, the last printed `txID` will be equal to the `AssetID`; in the other case, we will have two different `txID`: one for the creation of the asset and one for minting. However, in both cases, it feels only suitable to have a line stating the AssetID.

Last but not least, I reduced some duplicated code in the `asset`-cmd for submitting and waiting for a transaction, which was shared by almost every command in that file and consisted of the same repeated lines.

Motivation:
=========
While studying the TokenVm source code and playing around a bit, I thought this could be an excellent candidate to walk through an HyperSDK-example and learn the basics. 

Learnings
========
In my first attempt, I tried to put the value of the `Supply`-prompt into the `CreateAsset`-struct, including necessary marshaling in order to perform the minting-transaction in the Execute-function of the `create-asset`-action. But in this case, the error `ErrKeyNotSpecified` will be thrown because the asset has not been created successfully at this point. So this definitely was a good thing to learn by practicing that such a second operation must happen outside of an Action after the first action was already finalized on-chain.


examples/tokenvm/cmd/token-cli/cmd/action.go
```diff
@@ -117,12 +119,39 @@ var createAssetCmd = &cobra.Command{
 			return err
 		}
 
+		promptSupply := promptui.Prompt{
```
patrick-ogrady, line 122: Please use `promptInt` for this: https://github.com/ava-labs/hypersdk/blob/d7e8d760f7768c45e3d927877616b5f75cae5640/examples/tokenvm/cmd/token-cli/cmd/utils.go#L141-L166

I think we should probably change this to `promptUint64` (if you want to do that too)

examples/tokenvm/cmd/token-cli/cmd/action.go
```diff
@@ -117,12 +119,39 @@ var createAssetCmd = &cobra.Command{
...
+
 		// Confirm action
 		cont, err := promptContinue()
 		if !cont || err != nil {
 			return err
 		}
 
+		var supply uint64 = 0
+
+		if supplyText != "" {
```
patrick-ogrady, line 148: Won't be necessary after migrating to `promptInt`

examples/tokenvm/cmd/token-cli/cmd/action.go
```diff
@@ -138,6 +167,17 @@ var createAssetCmd = &cobra.Command{
 			return err
 		}
 		printStatus(tx.ID(), success)
+
+		if supply > 0 {
+			// Generate transaction
+			err = executeAndWaitForMintAssetTransaction(ctx, cli, tx.ID(), actor.PublicKey(), supply, factory)
```
patrick-ogrady, line 173: This error is ignored

examples/tokenvm/cmd/token-cli/cmd/action.go
```diff
@@ -798,14 +793,8 @@ var exportAssetCmd = &cobra.Command{
...
-		if err := submit(ctx); err != nil {
-			return err
-		}
-		success, err := cli.WaitForTransaction(ctx, tx.ID())
-		if err != nil {
-			return err
-		}
-		printStatus(tx.ID(), success)
+
+		_ = waitForTransactionAndPrintStatus(ctx, cli, submit, tx)
```
patrick-ogrady, line 797: You should not ignore this error

examples/tokenvm/cmd/token-cli/cmd/action.go
```diff
@@ -833,3 +822,30 @@ var exportAssetCmd = &cobra.Command{
...
 	},
 }
+
+func executeAndWaitForMintAssetTransaction(ctx context.Context, cli *client.Client, assetID ids.ID, recipient crypto.PublicKey, amount uint64, factory *auth.ED25519Factory) error {
+	submit, tx, _, err := cli.GenerateTransaction(ctx, nil, &actions.MintAsset{
+		Asset: assetID,
+		To:    recipient,
+		Value: amount,
+	}, factory)
+
```
patrick-ogrady, line 832: remove this `\n`

examples/tokenvm/cmd/token-cli/cmd/action.go
itinance: shouldnt we then also change the return type of `PromptInt` into uint64? I have already prepared a commit for that  @patrick-ogrady

examples/tokenvm/cmd/token-cli/cmd/action.go
itinance: done

examples/tokenvm/cmd/token-cli/cmd/action.go
itinance: done

examples/tokenvm/cmd/token-cli/cmd/action.go
itinance: done

patrick-ogrady: Yes, you are correct.

examples/tokenvm/cmd/token-cli/cmd/utils.go
```diff
@@ -138,7 +138,7 @@ func promptAmount(
 	return amount, err
 }
 
-func promptInt(
+func promptUint64(
 	label string,
 ) (int, error) {
```
patrick-ogrady, line 143: This should return `uint64` if you change this.

examples/tokenvm/cmd/token-cli/cmd/utils.go
```diff
@@ -138,7 +138,7 @@ func promptAmount(
 	return amount, err
 }
 
-func promptInt(
+func promptUint64(
```
patrick-ogrady, line 141: I recommend keeping both `promptInt` and `promptUint64` (in some cases, we just want an int)

PR #145, opened 2023-04-16
Example tokenvm shell script using tmp var

itinance:
The run-shell-script puts everything in the "/tmp"-directory (3-4 files and 2 folders). If one wants to change that, multiple places must be touched. I put the path for the temp-directory into a shell variable and found it helpful, to give it a dedicated subfolder called "hypersdk".

I also found it helpful to provide another shellscript which acts as the opposite of "run.sh": `stop.sh`


patrick-ogrady:
Nice touch with `stop.sh`. This will come in handy if we add more services that we need to tear down.

patrick-ogrady:
Recommend changing this print logs (and to actually call ./scripts/stop.sh` on exit): 
image

itinance:
> Recommend changing this print logs (and to actually call ./scripts/stop.sh` on exit): image

really good idea. It's done :)

PR #158, opened 2023-04-25
Add WS support

patrick-ogrady:
## TODO
- [x]  Move streaming client into listeners
- [x] Add WS server to 9650 port server
- [x] Move all RPC/WS server/client logic to a single package called `rpc`
- [x] Add errors to packers
- [x] Handler->Server
- [x] Add separate endpoint/client for controller APIs (may be easier to reason about than embedding...then don't need to override core from controller -> can be much easier to reason about)
- [x] cleanup paths 
- [x] add missing servers (both WS and RPC to integration)
- [x] add path to URIs (and `ws://` for websockets)


examples/tokenvm/tests/integration/integration_test.go
```diff
@@ -624,11 +627,8 @@ var _ = ginkgo.Describe("[Tx Processing]", func() {
 
 	ginkgo.It("processes valid index transactions (w/streaming verification)", func() {
 		// Create streaming client
-		streamingPort, err := instances[0].cli.StreamingPort(context.TODO())
-		gomega.Ω(err).Should(gomega.BeNil())
-		gomega.Ω(streamingPort).Should(gomega.Not(gomega.Equal(0)))
-		tcpURI := fmt.Sprintf("127.0.0.1:%d", streamingPort)
-		cli, err := vm.NewStreamingClient(tcpURI)
+		// TODO: cleanup URL creation
+		cli, err := listeners.NewStreamingClient("ws://" + strings.TrimPrefix(instances[0].WebSocketServer.URL, "http://") + "/ws")
```
patrick-ogrady, line 631: TODO: cleanup this formatting when initializing client

listeners/client.go
```diff
@@ -0,0 +1,80 @@
...
+	conn *websocket.Conn
+	wl   sync.Mutex
+	dll  sync.Mutex
+	bll  sync.Mutex
+	cl   sync.Once
+}
+
+// NewDecisionRPCClient creates a new client for the decision rpc server.
+// Dials into the server at [uri] and returns a client.
+func NewStreamingClient(uri string) (*Client, error) {
```
patrick-ogrady, line 22: How do we make customizable like the RPC client?

listeners/dependencies.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"context"
+
+	"github.com/ava-labs/avalanchego/trace"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/hypersdk/chain"
+)
+
+type VM interface {
```
patrick-ogrady, line 11: Move server RPC handlers in here too

listeners/client.go
patrick-ogrady: For example, it would be nice to stream orders?

PR #167, opened 2023-05-02
Add support for workspaces via go.work

### LINKED ISSUE #58: Error loading workspace: You have opened a nested module.
This is probably not a big deal, but thought I'd mention it.

I am getting the following warning in VS Code:
```
Error loading workspace: You have opened a nested module. To work on multiple modules at once, 
please use a go.work file. See https://github.com/golang/tools/blob/master/gopls/doc/workspace.md 
for more information on using workspaces.
```

hexfusion:
This PR adds workspaces separating hypersdk and tokenvm

```
Error loading workspace: You have opened a nested module. To work on multiple modules at once, 
please use a go.work file. See https://github.com/golang/tools/blob/master/gopls/doc/workspace.md 
for more information on using workspaces.
```

Fixes #58 

@EricForgy can you please test I just ran into this myself fixes me locally.


patrick-ogrady:
We should be able to remove `replace github.com/ava-labs/hypersdk => ../../` from the `tokenvm` go.mod with this?

hexfusion:
Good point will explore

hexfusion:
There is some question as to IF we should be checking go.work in to version control.

```
We could recommend never checking them in, but I think that's a bad idea: in some particular scenarios it can make perfect sense to check them in, such as monorepos holding multiple modules.
```

[1] https://github.com/golang/go/issues/53502

PR #170, opened 2023-05-04
[tstate] change map[key] from string to fixed size byte array

hexfusion:
This is an experimentation with replacing the current internal maps from `map[string][]byte` to `map[[65]byte][]byte`. In benchmarks I found that using a byte array was much faster than using string as a key.

```
goos: linux
goarch: amd64
pkg: byte-slice-indexed-maps/byte-slice-indexed-maps
=== RUN   BenchmarkArrayKeyed
BenchmarkArrayKeyed
BenchmarkArrayKeyed-12
 4196601               313.0 ns/op             0 B/op          0 allocs/op
=== RUN   BenchmarkOptimizedStringKeyed
BenchmarkOptimizedStringKeyed
BenchmarkOptimizedStringKeyed-12
 3048964               453.4 ns/op            16 B/op          1 allocs/op
=== RUN   BenchmarkStringKeyed
BenchmarkStringKeyed
BenchmarkStringKeyed-12                  2209490               505.6 ns/op            48 B/op          3 allocs/op
=== RUN   BenchmarkHexKeyed
BenchmarkHexKeyed
BenchmarkHexKeyed-12                     1739541               778.4 ns/op            96 B/op          4 allocs/op
```

ref. https://gitlab.com/pthevenet/go-small-answers/-/blob/29c03407006a269f8f6ba27fa3123fb9742ddf7c/byte-slice-indexed-maps/benchmark_test.go

This reduces the allocations for `FetchAndSetScope` from 9 to 2. and provides much cleaner profile (will upload soon).

```
goos: linux
goarch: amd64
pkg: github.com/ava-labs/hypersdk/tstate
BenchmarkFetchAndSetScope/fetch_and_set_scope_100_keys-12         	  192068	      2374 ns/op	     816 B/op	       2 allocs/op
BenchmarkFetchAndSetScope/fetch_and_set_scope_1000_keys-12        	  221142	      2540 ns/op	     816 B/op	       2 allocs/op
BenchmarkFetchAndSetScope/fetch_and_set_scope_10000_keys-12       	  195196	      2321 ns/op	     816 B/op	       2 allocs/op
PASS
ok  	github.com/ava-labs/hypersdk/tstate	4.474s
```

### DIFF
--- a/chain/processor.go
+++ b/chain/processor.go
@@ -20,7 +20,7 @@ type fetchData struct {
 
 type txData struct {
 	tx      *Transaction
-	storage map[string][]byte
+	storage map[[65]byte][]byte
 }
 
 type Processor struct {
@@ -49,26 +49,27 @@ func (p *Processor) Prefetch(ctx context.Context, db Database) {
 		defer span.End()
 
 		// Store required keys for each set
-		alreadyFetched := make(map[string]*fetchData, len(p.blk.GetTxs()))
+		alreadyFetched := make(map[[65]byte]*fetchData, len(p.blk.GetTxs()))
 		for _, tx := range p.blk.GetTxs() {
-			storage := map[string][]byte{}
+			storage := map[[65]byte][]byte{}
 			for _, k := range tx.StateKeys(sm) {
-				sk := string(k)
-				if v, ok := alreadyFetched[sk]; ok {
+				var ary [65]byte
+				copy(ary[:], k)
+				if v, ok := alreadyFetched[ary]; ok {
 					if v.exists {
-						storage[sk] = v.v
+						storage[ary] = v.v
 					}
 					continue
 				}
 				v, err := db.GetValue(ctx, k)
 				if errors.Is(err, database.ErrNotFound) {
-					alreadyFetched[sk] = &fetchData{nil, false}
+					alreadyFetched[ary] = &fetchData{nil, false}
 					continue
 				} else if err != nil {
 					panic(err)
 				}
-				alreadyFetched[sk] = &fetchData{v, true}
-				storage[sk] = v
+				alreadyFetched[ary] = &fetchData{v, true}
+				storage[ary] = v
 			}
 			p.readyTxs <- &txData{tx, storage}
 		}
--- a/tstate/tstate_test.go
+++ b/tstate/tstate_test.go
@@ -9,23 +9,14 @@ import (
 	"testing"
 
 	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/hypersdk/trace"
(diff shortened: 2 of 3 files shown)


chain/processor.go
```diff
@@ -20,7 +20,7 @@ type fetchData struct {
 
 type txData struct {
 	tx      *Transaction
-	storage map[string][]byte
+	storage map[[tstate.MapKeyLength]byte][]byte
```
darioush, line 23: nit: could we define a type such as `Key` or `MapKey`?

tstate/tstate_test.go
```diff
@@ -4,11 +4,13 @@ package tstate
...
 import (
 	"context"
+	"crypto/rand"
+	"fmt"
 	"testing"
 
 	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/hypersdk/trace"
-
+
```
darioush, line 13: nit: unneeded wspace

tstate/tstate_test.go
```diff
@@ -323,3 +325,46 @@ func TestWriteChanges(t *testing.T) {
...
+	require := require.New(b)
+	ts := New(10)
+	db := NewTestDB()
+	ctx := context.TODO()
+
+	keys := [][]byte{}
+	vals := [][]byte{}
+
+	// each k/v is unique to simulate worst case
+	for range "0..size" {
```
darioush, line 347: I don't think this does what you think it does :)

hexfusion: haha

patrick-ogrady:
You should make a similar change here: https://github.com/ava-labs/hypersdk/blob/a61157d9337fc658f3a6082befc32425641f2bb3/chain/processor.go#L44-L79

chain/processor.go
```diff
@@ -49,11 +49,11 @@ func (p *Processor) Prefetch(ctx context.Context, db Database) {
...
 
 		// Store required keys for each set
-		alreadyFetched := make(map[string]*fetchData, len(p.blk.GetTxs()))
+		alreadyFetched := make(map[tstate.Key]*fetchData, len(p.blk.GetTxs()))
 		for _, tx := range p.blk.GetTxs() {
-			storage := map[string][]byte{}
+			storage := map[tstate.Key][]byte{}
 			for _, k := range tx.StateKeys(sm) {
-				sk := string(k)
+				sk := tstate.ToStateKeyArray(k)
```
patrick-ogrady, line 56: I wonder if the best middle-ground is just to use the approach in `BenchmarkOptimizedStringKeyed`?

https://pthevenet.com/posts/programming/go/bytesliceindexedmaps/

We then get most of the benefit while retaining flexibility 🤷 .

Issue #52, opened 2023-02-25
Add large block support


patrick-ogrady:
To send blocks larger than 2MB, allow a block to include hashes for other chunks that can be downloaded over the P2P layer before verification (and after parse).

If all chunks can't be received, then the block will simply fail verification.

We should probably do this with some form of erasure coding.

While we could simply just "send more blocks", this allows us to perform fewer merkle root calculations and consensus events.

We will need to be very thoughtful about introducing protection mechanisms so a malicious producer can't force us to fetch a bunch of useless data.

rafael-abuawad:
How can this be implemented efficiently?

aaronbuchwald:
Closing in favor of the Vryx design

PR #136, opened 2023-04-06
Store transactions in `nanoBlocks`

patrick-ogrady:
Resolves: #130 
- [x] start broadcasting chunk ownership before block building is done
- [x] use expiring map to clear chunks no longer being used (where accept height is the key)
- [x] add metrics for chunks produced, fetched, and chunk fetch errors
- [x] add chunk cache when persisting to disk
- [x] Must be able to respond with ShouldVerifyWithContext in `ParseBlock`, `GetBlock`, `BuildBlock`: https://github.com/ava-labs/avalanchego/blob/7d73b59cb4838d304387ea680b9cc4053b72620c/vms/rpcchainvm/vm_server.go#L457-L464
- [x] Make chunk push dynamic
- [ ] push chunks after creation to some number of hosts (to avoid surge of requests on a single node) -> only keep in memory if send by preferred producer
- [x] send processing gossip immediately after building (instead of waiting normal 100ms)
- [ ] add pruning mechanism
  - [ ] convert txID storage to a cache (avoid persisting on disk)
  - [ ] remove old chunks
- [ ] if bootstrapping, store chunks on disk
- [ ] consider generating root async (as soon as verify is called the first time) so engine just really "polls" -> should have no processing consensus messages but should still bubble votes
  - [ ] will need to handle case where root underlying verification changes
  - [ ] may be an easier stopgap to just generate root async and include state hash of block we are building on
- [x] allow for configurable gossip frequency 
- [x] only cache chunk if it was non-empty
- [x] don't append handler to chunks using append (will change ID)
- [x] allow for only spinning up a single E2E network
- [ ] if don't have chunk, return list of people that do
- [ ] exit handle AppGossip as soon as possible (does handling block? If so, tx handling in app gossip should also be separated) -> async hardcoded at 3 per VM
  - [ ] so, we are handling 3 concurrently and the whole thing will end up dead-locking on handling
  - [ ] consider just allowing network router to set own concurrency limit? tx app gossip is probably almost always fully used...
- [x] run builder async -> can end up being a long-time offline per second in a small network (stuck verifying)
- [x] add metric for time waiting on chunks
- [x] increase async threads allowed in avalanche-ops config
- [ ] advertise optimistically fetched chunks
- [x] fetch optimistic chunks concurrently
- [x] gossip chunk info to all connected peers
- [x] delete chunks from optimistic if accepted block (then add to just an ID cache to avoid re-fetching optimistic chunks)
- [ ] store last X blocks of chunks in a map in storage instead of using LRU
[...]

PR #173, opened 2023-05-12
Introduce `TxBlock` and `RootBlock`

patrick-ogrady:
Closes: #52 
Replaces: #136 

TODO
- [ ] Add issue to remove tx gossip between validators, just rely on API/Client to distribute
- [x] invariant: tx block never needs to interact with root block
- [ ] add height to txBlock hash to avoid massive compaction when storing/deleting
- [x] create genesis txBlock to avoid empty parent checks (height == 0)
- [x] resolve txBlocks from disk
- [ ] avoid using multiple layers of MerkleDB on TxBlocks (will generate IDs on each layer) -> should use `VersionDB`
  - [ ] @dboehm-avalabs is fixing this and we won't need to change it 
- [x] Add metric for verification delay (time between request verification and verification ready to proceed)
- [ ] Use streamed blocks to ack txs (or test larger write buffer) or batch acks based on what is available
  - [x] also batch submit
- [x] Add metric for build time of blocks issued to consensus
- [x] Ensure verify delay time exits immediately after checking blocks (don't allow redos)
- [x] Handle closed Chan 
- [ ] Carry view forward across all TxBlocks (don't allow forks to be processed?)
  - [ ] can assert each node gets 1 txBlock at each height
- [ ] Verify signatures during `Last` so we don't block `RootBlock.Verify`
- [ ] Use millisecond timestamps (`UnixMilli`) + limit 1 block per X ms (avoid surge of seconds)
- [ ] recover txs from dropped txBlocks (into mempool)
- [ ] only gossip to next validators, non-stake can fetch chunks


patrick-ogrady:
**Need to pull off of disk in verify**

patrick-ogrady:
Current issue:
```
panic: the trie this view was based on has changed, rendering this view invalid

goroutine 54593 [running]:
github.com/ava-labs/hypersdk/chain.(*Processor).Prefetch.func1()
	/Users/patrickogrady/go/src/hypersdk/chain/processor.go:68 +0x494
created by github.com/ava-labs/hypersdk/chain.(*Processor).Prefetch
	/Users/patrickogrady/go/src/hypersdk/chain/processor.go:48 +0x19c
[05-16|23:01:19.577] INFO <ETcrYiJxrtLVPKtUMZKPKXS1wn8fo8PMjGja7qjRB2cVxWPFr Chain> subprocess/runtime.go:111 stdout collector shutdown
[05-16|23:01:19.577] INFO <ETcrYiJxrtLVPKtUMZKPKXS1wn8fo8PMjGja7qjRB2cVxWPFr Chain> subprocess/runtime.go:124 stderr collector shutdown
```

patrick-ogrady:
There should be no tx blocks missing: 
image

patrick-ogrady:
gzip compression should be disabled? 
image

patrick-ogrady:
Amount of state changes we can process rapidly degrades: 
image

patrick-ogrady:
Verify and build time explode: 
image

PR #186, opened 2023-05-23
Fix RequestID for VM2 requests + Create `network` package

felipemadero:
Closes https://github.com/ava-labs/hypersdk/issues/191 https://github.com/ava-labs/hypersdk/issues/192


patrick-ogrady:
Can you elaborate a bit more on this invariant and where you found it?

felipemadero:
> Can you elaborate a bit more on where this came up? I think it is right but I'd like to double-check.

I observed that the current code will try to get from n.requesters a nodeIDRequester for ids.EmptyNodeID:
	n.handleSharedRequestID(ids.EmptyNodeID, requestID)
When the network manager receives either a CrossChainAppResponse or a CrossChainAppRequestFailed.
I believe no such ids.EmptyNodeID is added anywere. So those incoming responses are never going to be handled.

The current way to add to n.requesters is by using getSharedRequestID, currently only being called by w.SendAppRequest.

But creating the map from a newRequestID to the requestID+handler should be made in SendCrossChainAppRequest for this case.

So, a possible way could be to use a ids.EmptyNodeID trick right there, eg:
	newRequestID := w.n.getSharedRequestID(w.handler, ids.EmptyNodeID, requestID)

But instead of that it seemed to me more reasonable to also create a map from chainID to requesters, forcing separated
increasing numbers for each chain, same as it is being done for nodes.

patrick-ogrady:
> > Can you elaborate a bit more on where this came up? I think it is right but I'd like to double-check.
> 
> I observed that the current code will try to get from n.requesters a nodeIDRequester for ids.EmptyNodeID: n.handleSharedRequestID(ids.EmptyNodeID, requestID) When the network manager receives either a CrossChainAppResponse or a CrossChainAppRequestFailed. I believe no such ids.EmptyNodeID is added anywere. So those incoming responses are never going to be handled.
> 
> The current way to add to n.requesters is by using getSharedRequestID, currently only being called by w.SendAppRequest.
> 
> But creating the map from a newRequestID to the requestID+handler should be made in SendCrossChainAppRequest for this case.
> 
> So, a possible way could be to use a ids.EmptyNodeID trick right there, eg: newRequestID := w.n.getSharedRequestID(w.handler, ids.EmptyNodeID, requestID)
> 
> But instead of that it seemed to me more reasonable to also create a map from chainID to requesters, forcing separated increasing numbers for each chain, same as it is being done for nodes.

Thanks for clarifying. Will confirm and look into this.

vm/vm.go
```diff
@@ -156,7 +157,7 @@ func (vm *VM) Initialize(
 	}
 	vm.metrics = metrics
 	vm.proposerMonitor = NewProposerMonitor(vm)
-	vm.networkManager = NewNetworkManager(vm, appSender)
+	vm.networkManager = network.NewNetworkManager(vm.snowCtx.Log, appSender)
```
patrick-ogrady, line 160: This stutters...we should change `NewNetworkManager` to `New/NewManager`

vm/vm.go
```diff
@@ -108,7 +109,7 @@ type VM struct {
 	warpManager *WarpManager
 
 	// Network manager routes p2p messages to pre-registered handlers
-	networkManager *NetworkManager
+	networkManager *network.NetworkManager
```
patrick-ogrady, line 112: This stutters...we should change this to `Manager`

network/network.go
```diff
@@ -26,24 +27,26 @@ type request struct {
 }
 
 type NetworkManager struct {
-	vm     *VM
+	log    logging.Logger
```
patrick-ogrady, line 30: nice change

patrick-ogrady:
I can confirm, there definitely is a bug here: https://github.com/ava-labs/hypersdk/blob/96e794f5e65d1fabc110c611a063ec2fca35dd4b/vm/network.go#L418-L419

We need to register this requestID as you mentioned.

patrick-ogrady:
> > > Can you elaborate a bit more on where this came up? I think it is right but I'd like to double-check.
> > 
> > 
> > I observed that the current code will try to get from n.requesters a nodeIDRequester for ids.EmptyNodeID: n.handleSharedRequestID(ids.EmptyNodeID, requestID) When the network manager receives either a CrossChainAppResponse or a CrossChainAppRequestFailed. I believe no such ids.EmptyNodeID is added anywere. So those incoming responses are never going to be handled.
> > The current way to add to n.requesters is by using getSharedRequestID, currently only being called by w.SendAppRequest.
> > But creating the map from a newRequestID to the requestID+handler should be made in SendCrossChainAppRequest for this case.
> > So, a possible way could be to use a ids.EmptyNodeID trick right there, eg: newRequestID := w.n.getSharedRequestID(w.handler, ids.EmptyNodeID, requestID)
> > But instead of that it seemed to me more reasonable to also create a map from chainID to requesters, forcing separated increasing numbers for each chain, same as it is being done for nodes.
> 
> Thanks for clarifying. Will confirm and look into this.

Ok....I thought about this. We should keep the single map and revert the changes you did into separate IDs. We should use the node's ID instead of using the emptyID and just register it properly.

vm/vm.go
felipemadero: Will also rename struct NetworkManager to Manager and struct NetworkHandler to Handler

patrick-ogrady (approved):

PR #204, opened 2023-06-20
Move repeat check to deferred verification

patrick-ogrady:
if we aren't checking signatures before accept, we don't need to check repeats

- [ ] update height in block manager at end of block issuance (trigger gossip)
- [ ] track account pending balances in a DA account (one-way transfer) if producer to know if will get fees -> how to originally fund? -> don't need to keep 2 states (need this to avoid DOS'ing proposer mempool with junk txs)
  - [ ] should have a "fee balance" to also make sure can pay for computation units they desire
- [ ] Attempt pull only block fetching (more scalable than complex push/broadcast tree)
- [ ] revert channel wait for each tx in exec loop (will lead to a TON of context switching, which means scheduling locks prevent full CPU usage)
- [ ] check bandwidth throttling on dev-machine (for tx broadcast)

### DIFF
--- a/chain/builder.go
+++ b/chain/builder.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ava-labs/avalanchego/ids"
 	smblock "github.com/ava-labs/avalanchego/snow/engine/snowman/block"
+	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/ava-labs/avalanchego/utils/units"
 	"go.uber.org/zap"
 )
@@ -87,7 +88,8 @@ func BuildBlock(
 
 		start = time.Now()
 
-		// TODO: consider keeping all txs used in a set to avoid re-repeats later
+		// Ensure we don't add duplicate txs to same block (checked in parse)
+		txIDs = set.NewSet[ids.ID](txBatchSize)
 	)
 	b.MinTxHght = txBlock.Hght
 
@@ -106,6 +108,11 @@ func BuildBlock(
 
 			txsAttempted++
 			if next.Base.Timestamp < oldestAllowed {
+				return true, false, nil
+			}
+
+			// Already added tx to block
+			if txIDs.Contains(next.ID()) {
 				return true, false, nil
 			}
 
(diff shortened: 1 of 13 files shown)


patrick-ogrady:
image

Under load, this is still the build bottleneck: 
image
image

patrick-ogrady:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x70 pc=0x1046dcdec]

goroutine 258 [running]:
github.com/ava-labs/hypersdk/chain.(*StatelessTxBlock).CollectRepeats(0x0, {0x104de9588, 0x14000042028}, 0x64921927, {0x140003e20a8, 0x1, 0x1}, 0x14000894688)
	/Users/patrickogrady/go/src/hypersdk/chain/tx_block.go:230 +0x4c
github.com/ava-labs/hypersdk/chain.(*StatelessRootBlock).Execute(0x14005d5e0e0, {0x104de9588?, 0x14000042028}, 0x1400037a0e0)
	/Users/patrickogrady/go/src/hypersdk/chain/root_block.go:627 +0x2b0
github.com/ava-labs/hypersdk/vm.(*VM).processAcceptedBlocks(0x14000122900)
	/Users/patrickogrady/go/src/hypersdk/vm/resolutions.go:161 +0x74
created by github.com/ava-labs/hypersdk/vm.(*VM).Initialize
	/Users/patrickogrady/go/src/hypersdk/vm/vm.go:340 +0x26f0
```

patrick-ogrady:
Memory leak: 
image

patrick-ogrady:
latest on c5.4xlarge cluster (1k accts, 50 tps, 5 clients):
image
image
image
image
image
image
image
image
image
image

patrick-ogrady:
When pushing 100k TPS, start to see an acceptor backlog on one node: 
image

patrick-ogrady:
Slowed execution speed could be from channel waits in prefetch on tx signatures.

patrick-ogrady:
TL;DR consensus/networking is FAR faster than verification...we are accepting blocks so fast verification quickly falls behind at high throughput.

Issue #183, opened 2023-05-21
Fix License Check


patrick-ogrady:
Mirror this change: https://github.com/ava-labs/avalanchego/pull/1492

PR #235, opened 2023-07-17
fix license header test

aphexmunky:
Addresses #183


trace/noop.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2019-2022, Ava Labs, Inc. All rights reserved.
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
```
patrick-ogrady, line 1: I think this should be 2023 🤔

patrick-ogrady: (I think there are a few others like this)

chain/block.go
```diff
@@ -534,7 +534,11 @@ func (b *StatelessBlock) innerVerify(ctx context.Context) (merkledb.TrieView, er
 	processor.Prefetch(ctx, state)
 
 	// Process new transactions
-	unitsConsumed, surplusFee, results, stateChanges, stateOps, err := processor.Execute(ctx, ectx, r)
+	unitsConsumed, surplusFee, results, stateChanges, stateOps, err := processor.Execute(
```
patrick-ogrady, line 537: Can you remove these formatting diffs? I'd prefer to cleanup line length stuff in a separate PR.

patrick-ogrady (approved):
Nice!

Issue #247, opened 2023-07-18
Remove Block Cost


patrick-ogrady:
Blocked by: #244, #182 

Once we have millisecond times and bound the timestamp of a block to 1-2s ahead, we can remove `BlockCost` and replace it with a much simpler `MinTimeDifference` setting that is set in genesis.
* Don't forget to modify block build trigger to do this

Issue #244, opened 2023-07-18
Millisecond Block Time


patrick-ogrady:
ex: https://github.com/ava-labs/hypersdk/pull/177

* Don't forget to chunk the lookback window for repeats to avoid having times at each millisecond.
* Don't forget to modify tx price change logic to use ms instead of s based chunks.

Issue #182, opened 2023-05-21
Change "Future Bound" to 2s


patrick-ogrady:
We don't need a clock synchrony bound of 10s in 2023.

PR #248, opened 2023-07-18
Millisecond Block Time + Reduce Future Bound + Remove Block Cost

patrick-ogrady:
Resolves: #244
Resolves: #182 
Resolves: #247 

## Challenges
* Second-level granularity prevents a number of important data structures from getting too large -> leads to slight over-estimation of windows
* UX for configs easier to reason about in seconds

## TODO
- [x] store ms in block timestamp
- [x] remove blockCost + blockCostWindow
- [x] reduce future bound to 1s
- [x] remove `MaxBlockTxs`
- [x] add min block time to rules
- [x] remove preferred blocks per second from config, just try to build as soon as min block time has surpassed using rules
- [x] reduce tx/block precision for gas cost window + emap
- [x] make block builder dynamic (don't just check every X ms)
- [x] ensure GetValidityWindow is being used correctly
  - [x] convert to ms to avoid multiplying by `consts.MillisecondsPerSecond` everywhere
  - [ ] ~need to wait for n+1 second before exiting during sync (otherwise may not get beginning of modulo range)? (as a simple example, consider a 500ms range -> missing first 500ms but other nodes will use those txs to determine if replay)~
  - [x] ensure oldest allowed lookback in blocks matches behavior in seen emap (non-determinism)
- [x] add replay test where accept, then submit again to see if this is a bug
- [x] [Different Approach] reduce precision of txs instead of truncating inner buckets (can assert as well in tx verification)
- [x] clarify in a comment whether expiry time is inclusive
- [x] Fix hang during E2E builder
  - [x] Confirm that ProposerVM can handle millisecond timestamps (may need to truncate to second-level for it): https://github.com/ava-labs/avalanchego/blob/0ec52a9c6e5b879e367688db01bb10174d70b212/vms/proposervm/pre_fork_block.go#L201
  - [ ] ~We may need to just expose second-level granularity to the proposervm (which wouldn't break any invariants)~


emap/emap.go
```diff
@@ -48,14 +48,16 @@ func (e *EMap[T]) Add(items []T) {
 	defer e.mu.Unlock()
 
 	for _, item := range items {
-		e.add(item.ID(), item.Expiry())
+		e.add(item.ID(), reducePrecision(item.Expiry()))
```
patrick-ogrady, line 51: If we use this approach, we may need whoever calls into emap to ensure they don't re-add below this reduced precision boundary. May make sense for the wrapper to handle this instead?

builder/time.go
```diff
@@ -60,7 +60,7 @@ func (b *Time) shouldBuild(ctx context.Context) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	now := time.Now().Unix()
+	now := time.Now().UnixMilli()
 	since := int(now - preferredBlk.Tmstmp)
 	newRollupWindow, err := window.Roll(preferredBlk.BlockWindow, since)
```
patrick-ogrady, line 65: Need to handle conversion to seconds?

patrick-ogrady:
Current log:
```
./node1-bls/logs/2Kr5MFwVZoAbaXL7u6NQBvMNMahKeokFfxj2An5eZvH3iFFztj.log:[07-19|22:36:54.606] DEBUG <2Kr5MFwVZoAbaXL7u6NQBvMNMahKeokFfxj2An5eZvH3iFFztj Chain> snowman/transitive.go:877 block verification failed {"error": "time must monotonically increase"}
```

PR #253, opened 2023-07-19
docs: Update DEVNETS to reflect latest auth workflow

exdx:
Using avalanche-ops locally has changed significantly due to a new
identity provider for AWS authentication.

Signed-off-by: Dan Sover <dan.sover@avalabs.org>

### DIFF
--- a/examples/tokenvm/DEVNETS.md
+++ b/examples/tokenvm/DEVNETS.md
@@ -29,10 +29,9 @@ chmod +x /tmp/avalancheup-aws
 
 ### Step 2: Install and Configure `aws-cli`
 Next, install the [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html). This is used to
-manipulate CloudFormation.
+authenticate to AWS and manipulate CloudFormation.
 
-Once you've installed the AWS CLI, run `aws configure` to set the access key to
-use while deploying your devnet.
+Once you've installed the AWS CLI, run `aws configure sso` to login to AWS locally. See [the avalanche-ops documentation](https://github.com/ava-labs/avalanche-ops#permissions) for additional details. Set a `profile_name` when logging in, as it will be referenced directly by avalanche-ops commands.
 
 ### Step 3: Install `token-cli`
 ```bash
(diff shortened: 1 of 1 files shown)


examples/tokenvm/DEVNETS.md
```diff
@@ -92,14 +91,17 @@ Now we can spin up a new network of 6 nodes with some defaults:
...
+--profile-name <AWS_PROFILE_NAME>
 ```
 
 The `default-spec` (and `apply`) command only provisions nodes, not Custom VMs.
 The Subnet and Custom VM installation are done via `install-subnet-chain` sub-commands that follow.
 
 *NOTE*: The `--auto-regions 3` flag only lists the pre-defined regions and evenly distributes anchor and non-anchor nodes across regions in the default spec output. It is up to the user to change the regions and spec file based on the regional resource limits (e.g., pick the regions with higher EIP limits). To see what regions available in your AWS account, simply run the [`aws account list-regions`](https://docs.aws.amazon.com/cli/latest/reference/account/list-regions.html) command, or look at the dropdown menu of your AWS console.
 
+*NOTE*: Multiple regions can alternatively be selected by providing a comma-separated list of regions to the `--regions` flag. This will also spin up a dev machine in each region selected, if `--create-dev-machine` is enabled.q
+
```
hexfusion, line 104: Minor typo drop the dangling q `enabled.`

hexfusion (approved):

examples/tokenvm/DEVNETS.md
exdx: vim typo 😅

examples/tokenvm/DEVNETS.md
```diff
@@ -92,14 +91,17 @@ Now we can spin up a new network of 6 nodes with some defaults:
...
+--keys-to-generate 5 \
+--profile-name <AWS_PROFILE_NAME>
 ```
 
 The `default-spec` (and `apply`) command only provisions nodes, not Custom VMs.
 The Subnet and Custom VM installation are done via `install-subnet-chain` sub-commands that follow.
 
 *NOTE*: The `--auto-regions 3` flag only lists the pre-defined regions and evenly distributes anchor and non-anchor nodes across regions in the default spec output. It is up to the user to change the regions and spec file based on the regional resource limits (e.g., pick the regions with higher EIP limits). To see what regions available in your AWS account, simply run the [`aws account list-regions`](https://docs.aws.amazon.com/cli/latest/reference/account/list-regions.html) command, or look at the dropdown menu of your AWS console.
 
+*NOTE*: Multiple regions can alternatively be selected by providing a comma-separated list of regions to the `--regions` flag. This will also spin up a dev machine in each region selected, if `--create-dev-machine` is enabled.
```
patrick-ogrady, line 103: How does this interact with the regions passed into the instance-types map?

exdx: > How does this interact with the regions passed into the instance-types map?

avalancheup uses the `--regions` flag when determining in which regions to spin up dev machines specifically. The dev machine can be further customized using the `dev_machine` field on the spec. There is a `dev_machine.instance_types` field to select specific types. 

The `instance-type` is used for nodes in the subnet specifically. If you use `--regions` only then avalancheup generates an `instance-type` map under the hood using the default EC2 instances in each specific region. 

My impression is that `--regions` and --instance-types` can be used together without conflict.  The dev machines regions should be a subset of the overall regions selected.

PR #297, opened 2023-07-26
create TypeID for Actions/Auths + delete reflection in Registry

### LINKED ISSUE #228: Assign TypeID to Actions/Auths Instead of Using Reflect-Based Registry
Example:
* https://github.com/ava-labs/hypersdk/blob/96e794f5e65d1fabc110c611a063ec2fca35dd4b/chain/transaction.go#L340-L348

najeal:
Closes #228 

This PR adds `GetTypeID` method to the Action & Auth interfaces.

Now we register directly an ID instead of a Type in Registry.
Using a TypeID tied to the Action/Auth struct makes the reflection mechanism "legacy". Therefore it has been cleaned.

@patrick-ogrady is it what you expected?


examples/tokenvm/actions/common.go
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package actions
+
+const (
+	burnAssetID uint8 = iota
```
patrick-ogrady, line 7: Please list these out explicitly. `iota` will do the job but find it is much more delicate.

chain/transaction.go
```diff
@@ -334,22 +329,14 @@ func (t *Transaction) Payer() string {
 
 func (t *Transaction) Marshal(
```
patrick-ogrady, line 330: nit: make this 1 line

chain/transaction.go
```diff
@@ -359,17 +346,15 @@ func (t *Transaction) Marshal(
...
-	p.PackByte(actionByte)
+	p.PackByte(actionID)
 	t.Action.Marshal(p)
-	p.PackByte(authByte)
+	p.PackByte(authID)
 	t.Auth.Marshal(p)
 	return p.Err()
 }
 
 func MarshalTxs(
```
patrick-ogrady, line 356: nit: make this 1 line

examples/tokenvm/actions/common.go
najeal: If there is duplicated id created manually, the registry will return the error 👍

patrick-ogrady: May be worth adding that as a comment just for other contributors to keep in mind.

patrick-ogrady:
Ran the CI to give you a read on what's remaining to change 👍

codec/type_parser.go
```diff
@@ -33,31 +29,17 @@ func NewTypeParser[T any, X any, Y bool]() *TypeParser[T, X, Y] {
...
-	p.indexToDecoder[p.index] = &decoder[T, X, Y]{f, y}
-	p.index++
+	p.indexToDecoder[id] = &decoder[T, X, Y]{f, y}
 	return nil
 }
 
-// LookupType returns the index, decoder function and the success of lookup of
-// type [o] from Typeparser [p].
-func (p *TypeParser[T, X, Y]) LookupType(o T) (uint8, func(*Packer, X) (T, error), Y, bool) {
-	index, ok := p.typeToIndex[fmt.Sprintf("%T", o)]
```
patrick-ogrady, line 53: ❤️

codec/type_parser.go
```diff
@@ -33,31 +29,17 @@ func NewTypeParser[T any, X any, Y bool]() *TypeParser[T, X, Y] {
 // Register registers a new type into TypeParser [p]. Registers the type by using
 // the string representation of [o], and sets the decoder of that index to [f].
 // Returns an error if [o] has already been registered or the TypeParser is full.
-func (p *TypeParser[T, X, Y]) Register(o T, f func(*Packer, X) (T, error), y Y) error {
-	if p.index == consts.MaxUint8 {
+func (p *TypeParser[T, X, Y]) Register(id uint8, f func(*Packer, X) (T, error), y Y) error {
+	if uint8(len(p.indexToDecoder)) == consts.MaxUint8 {
```
patrick-ogrady, line 33: nit: Do we want to enforce items are added monotonically?

I guess we don't need to...so we probably shouldn't 🤔

codec/type_parser.go
```diff
@@ -33,31 +29,17 @@ func NewTypeParser[T any, X any, Y bool]() *TypeParser[T, X, Y] {
 // Register registers a new type into TypeParser [p]. Registers the type by using
 // the string representation of [o], and sets the decoder of that index to [f].
 // Returns an error if [o] has already been registered or the TypeParser is full.
-func (p *TypeParser[T, X, Y]) Register(o T, f func(*Packer, X) (T, error), y Y) error {
-	if p.index == consts.MaxUint8 {
+func (p *TypeParser[T, X, Y]) Register(id uint8, f func(*Packer, X) (T, error), y Y) error {
```
patrick-ogrady, line 32: I don't think `uint8(len(p.indexToDecoder)) == consts.MaxUint8` is broken, but we should probably do `len(p.indexToDecoder) == int(consts.MaxUint8)`

examples/morpheusvm/actions/common.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package actions
+
+const (
+	// IDs explicitly listed. Registry will avoid usage of duplicated IDs by returning an error when occurs
```
patrick-ogrady, line 7: nit (for all):

```
// Note: Registry will error during initialization if a duplicate ID is assigned. We explicitly assign IDs to avoid accidental remapping.
```

patrick-ogrady: You can also put this just above `const`

patrick-ogrady:
@najeal VERY close! just 1-2 small nits. Kicking off full CI rn.

patrick-ogrady:
Would be good future work to convert this to use `TypeID`: https://github.com/ava-labs/hypersdk/blob/51cee4e69deb84f969222c906c01a6a9059a7691/examples/tokenvm/controller/controller.go#L177-L211

codec/type_parser.go
najeal: Hum, if you are pointing the `if` statement. It is not related to a monotically insert.
We just check if the number of stored ID is equal to the maxUint8. When equal, it means we already stored all the possible value represented by a uint8.
By the way it should be `.... == consts.MaxUint8 + 1`

patrick-ogrady:
Looks like full CI passed, as soon as nits are pushed should be good to merge.

codec/type_parser.go
patrick-ogrady: Yep, I left this comment here just to note the change (which used to be enforced by the TypeParser).

It is ok to ignore.

patrick-ogrady (approved):

codec/type_parser.go
patrick-ogrady: Nice catch on the +1.

patrick-ogrady:
(no need to update branch to main)

najeal:
> Would be good future work to convert this to use `TypeID`:
> 
> https://github.com/ava-labs/hypersdk/blob/51cee4e69deb84f969222c906c01a6a9059a7691/examples/tokenvm/controller/controller.go#L177-L211

You want the same switch but on the TypeID instead?

patrick-ogrady:
> > Would be good future work to convert this to use `TypeID`:
> > https://github.com/ava-labs/hypersdk/blob/51cee4e69deb84f969222c906c01a6a9059a7691/examples/tokenvm/controller/controller.go#L177-L211
> 
> You want the same switch but on the TypeID instead?

Yeah, the idea is to remove all use of reflect, especially in the "hot path". Do you want to do that here or in a separate PR?

najeal:
@patrick-ogrady I will do it in another PR. I will also pick another issue in the list.

PR #311, opened 2023-07-27
avoid type assertion on Action/Auth in switchs

najeal:
@patrick-ogrady I am not a big fan of switching from type assertion to `GetTypeID()` usage in switch statement.
1. in cases we often use the type assertion.
2. type assertion is faster than reflection we deleted in #297 
3. it makes code more difficult to read

### DIFF
--- a/examples/morpheusvm/cmd/morpheus-cli/cmd/resolutions.go
+++ b/examples/morpheusvm/cmd/morpheus-cli/cmd/resolutions.go
@@ -52,8 +52,9 @@ func handleTx(tx *chain.Transaction, result *chain.Result) {
 	status := "⚠️"
 	if result.Success {
 		status = "✅"
-		switch action := tx.Action.(type) { //nolint:gocritic
-		case *actions.Transfer:
+		switch tx.Action.GetTypeID() { //nolint:gocritic
+		case (&actions.Transfer{}).GetTypeID():
+			action := tx.Action.(*actions.Transfer)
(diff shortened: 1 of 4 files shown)


examples/morpheusvm/cmd/morpheus-cli/cmd/resolutions.go
```diff
@@ -52,8 +52,9 @@ func handleTx(tx *chain.Transaction, result *chain.Result) {
 	status := "⚠️"
 	if result.Success {
 		status = "✅"
-		switch action := tx.Action.(type) { //nolint:gocritic
-		case *actions.Transfer:
+		switch tx.Action.GetTypeID() { //nolint:gocritic
+		case (&actions.Transfer{}).GetTypeID():
```
patrick-ogrady, line 56: Ah, I was thinking you just would do:
```
case TransferTypeID:
```

examples/morpheusvm/cmd/morpheus-cli/cmd/resolutions.go
```diff
@@ -52,8 +52,9 @@ func handleTx(tx *chain.Transaction, result *chain.Result) {
...
+			action := tx.Action.(*actions.Transfer)
```
patrick-ogrady, line 57: Ah, I didn't think about the need to do this...probably better to just leave for now.

examples/morpheusvm/cmd/morpheus-cli/cmd/resolutions.go
najeal: I thought about doing it but it makes potentially use a declared ID not used by any Type.
The both were possible.

PR #327, opened 2023-08-03
cli: watch command - print block size

### LINKED ISSUE #300: [cli] Emit Block Size in KB in `watch`

najeal:
Closes #300

### DIFF
--- a/chain/block.go
+++ b/chain/block.go
@@ -47,6 +47,12 @@ type StatefulBlock struct {
 	StateRoot     ids.ID     `json:"stateRoot"`
 	UnitsConsumed uint64     `json:"unitsConsumed"`
 	WarpResults   set.Bits64 `json:"warpResults"`
+
+	size int
+}
+
+func (b *StatefulBlock) Size() int {
+	return b.size
 }
 
 // warpJob is used to signal to a listner that a *warp.Message has been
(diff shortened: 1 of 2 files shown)


chain/block.go
```diff
@@ -819,6 +825,7 @@ func UnmarshalBlock(raw []byte, parser Parser) (*StatefulBlock, error) {
 	p.UnpackID(false, &b.StateRoot)
 	b.UnitsConsumed = p.UnpackUint64(false)
 	b.WarpResults = set.Bits64(p.UnpackUint64(false))
+	b.size = len(raw)
```
patrick-ogrady, line 828: Can we also set size after `Marshal`? Or, add a comment indicating it is only populated if `Unmarshaled`?

patrick-ogrady: If you add to `Marshal`, you can get the `Len` of the packer and can just use that.

cli/chain.go
```diff
@@ -239,15 +239,15 @@ func (h *Handler) WatchChain(hideTxs bool, getParser func(string, uint32, ids.ID
 			window.Update(&tpsWindow, window.WindowSliceSize-consts.Uint64Len, uint64(len(blk.Txs)))
 			runningDuration := time.Since(start)
 			tpsDivisor := math.Min(window.WindowSize, runningDuration.Seconds())
-			// TODO: add block size in KB
 			utils.Outf(
-				"{{green}}height:{{/}}%d {{green}}txs:{{/}}%d {{green}}units:{{/}}%d {{green}}root:{{/}}%s {{green}}TPS:{{/}}%.2f {{green}}split:{{/}}%dms\n",
+				"{{green}}height:{{/}}%d {{green}}txs:{{/}}%d {{green}}units:{{/}}%d {{green}}root:{{/}}%s {{green}}TPS:{{/}}%.2f {{green}}split:{{/}}%dms {{green}}size:{{/}}%.2f\n",
```
patrick-ogrady, line 243: Can you add `KB` after the `.2f`?

najeal: done

chain/block.go
najeal: done

Issue #230, opened 2023-07-06
Migrate to Uptrace


patrick-ogrady:
Replace zipkin with https://uptrace.dev/

patrick-ogrady:
Example in AvalancheGo: https://github.com/ava-labs/avalanchego/pull/1727

najeal:
@patrick-ogrady assign me please

PR #330, opened 2023-08-03
trace: replace zipkin with uptrace

najeal:
Related to #230


najeal:
@patrick-ogrady I face one problem when enabling tracing `TRACE=true ./scripts/tests.load.sh`.
On `main` branch and `uptrace` branch test `[It] creates blocks` fails with error message `no transactions`.
My first opinion is some transactions are losts.

Except this data is loaded in uptrace as expected.

examples/morpheusvm/config/config.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 	// Tracing
 	TraceEnabled    bool    `json:"traceEnabled"`
 	TraceSampleRate float64 `json:"traceSampleRate"`
+	DSN             string  `json:"dsn"`
```
patrick-ogrady, line 38: To confirm, this is provided by Uptrace on startup?

patrick-ogrady:
> @patrick-ogrady I face one problem when enabling tracing `TRACE=true ./scripts/tests.load.sh`. On `main` branch and `uptrace` branch test `[It] creates blocks` fails with error message `no transactions`. My first opinion is some transactions are losts.
> 
> Except this data is loaded in uptrace as expected.

Hmmm. Let me look rq.

patrick-ogrady:
It worked for me: 
image

I just had to do this diff (run `go mod tidy` and add a default DSN):
[patch.txt](https://github.com/ava-labs/hypersdk/files/12263271/patch.txt)

patrick-ogrady:
The next iteration here will also be adding the prometheus metrics ingestion to this tool ❤️

examples/morpheusvm/config/config.go
najeal: Yes it is provided by the running uptrace.
It contains a secret to communicate with uptrace, is it safe to use it in this config?

najeal:
@patrick-ogrady the data is loaded in your uptrace but you confirm that the `load tests` are successul with trace enabled and uptrace running?

I will add prometheus in the same PR.

examples/morpheusvm/config/config.go
patrick-ogrady: We can set the default value (like I did in my patch).

patrick-ogrady:
> @patrick-ogrady the data is loaded in your uptrace but you confirm that the `load tests` are successul with trace enabled and uptrace running?
> 
> I will add prometheus in the same PR.

Yep, worked like normal: 
image

What do you see?

najeal:
@patrick-ogrady  I see this:
Screenshot 2023-08-06 at 19 41 01

patrick-ogrady:
🤔 definitely seems like a new bug. This would happen if the calculation for how many times to call build block (based on the fee and any balance is wrong): https://github.com/ava-labs/hypersdk/blob/1eb783683d8324f0ffa9fbd9a52df1abaa4c310a/examples/morpheusvm/tests/load/load_test.go#L373-L381

najeal:
But it works on your side 😩 .
I added some logs
We generate 500 000 txs
Looking at the logs each block from 0 to 109 is created with 4090 txs
the 110th has only 231 txs.
We should have created 122 blocks.

4090 * 110 + 231 = 450 131 txs

50 000 tx have been lost. Is it possible they got removed due to some logic?

najeal:
@patrick-ogrady I wanted to generate prometheus config to export metrics from prometheus into Uptrace using `remote_write`. But this method only work for their saas currently.
If we want to use a local Uptrace we need to use an `OpenTelemetry Collector`.

trace/tracer.go
```diff
@@ -57,26 +57,16 @@ func New(config *Config) (trace.Tracer, error) {
...
-
-	tracerProviderOpts := []sdktrace.TracerProviderOption{
-		sdktrace.WithBatcher(exporter, sdktrace.WithExportTimeout(tracerExportTimeout)),
-		sdktrace.WithResource(
-			resource.NewWithAttributes(
-				semconv.SchemaURL,
-				attribute.String("version", config.Version),
-				semconv.ServiceNameKey.String(config.Agent),
-			),
+	uptrace.ConfigureOpentelemetry(
```
patrick-ogrady, line 60: I'm wondering why we import `uptrace` at all instead of just passing a header into the `sdktrace` options below like AvalancheGo: https://github.com/ava-labs/avalanchego/pull/1727

The appeal there is that we can just use the stock OpenTelemetry?

patrick-ogrady: In the process, we should just migrate to use this package (we don't need a separate `trace` package anymore because `uptrace` can use opentelemtry directly ❤️ ): https://github.com/ava-labs/avalanchego/blob/master/trace/tracer.go

najeal: trace has been updated.

you wanted to do the same with prometheus? delete all reference to prometheus in the code and use opentelemetry metrics instead?

najeal: trace package has been updated. If you want to use `trace` package from avalanchego we need to pass AppName and Version from the Config.

https://github.com/ava-labs/avalanchego/blob/e70a17d9d988b5067f3ef5c4a057f15ae1271ac4/trace/tracer.go#L73C2-L73C2

https://github.com/ava-labs/avalanchego/blob/e70a17d9d988b5067f3ef5c4a057f15ae1271ac4/trace/tracer.go#L74

patrick-ogrady: I think that would be a good change. Can you open an issue for that? We can probably merge that quickly.

patrick-ogrady: I think we should try to do so (use the `trace` package upstream)

najeal: It works 👍

najeal:
@patrick-ogrady issue has been open avalanchego side and PR (https://github.com/ava-labs/avalanchego/pull/1893)

patrick-ogrady:
> @patrick-ogrady issue has been open avalanchego side and PR ([ava-labs/avalanchego#1893](https://github.com/ava-labs/avalanchego/pull/1893))

Bumping this on the platform side

najeal:
is it still relevant? The PR is old and there is a lot of conflicts.

PR #351, opened 2023-08-13
[crypto] Batch verify proposer gossip

patrick-ogrady:
Resolves: #345


chain/block.go
```diff
@@ -159,6 +159,13 @@ func (b *StatelessBlock) populateTxs(ctx context.Context) error {
 	b.sigJob = job
 	batchVerifier := NewAuthBatch(b.vm, b.sigJob, b.authCounts)
 
+	// Make sure to always call [Done], otherwise we will block all future [Workers]
```
patrick-ogrady, line 162: This was likely a bug

Issue #96, opened 2023-03-18
[tokenvm] Overhaul Fees


patrick-ogrady:
Naïve fee pricing/usage targets were added during initial development. We should revisit these and add the updated values to the README.

## Changes
* Charge for keys specified in pre-fetch
* Charge a fee if new state created

patrick-ogrady:
As part of this overhaul, we may find it useful to break apart "units" into specialized units that are independently priced (ex: CPU, State, Size)

patrick-ogrady:
It would also be great if we can find some way to remove the notion of block cost and surplus fee.

We should ideally just refund anything above the required fee.

patrick-ogrady:
With the current parameters, a "full" block can sometimes only be ~200KB: 
<img width="1008" alt="image" src="https://user-images.githubusercontent.com/13023275/228723858-b1cf49e4-fd41-4919-b4a2-7feba73682d1.png">

patrick-ogrady:
We should require a state deposit that is returned when state is freed (minus some penalty) to discourage state bloat attacks.

patrick-ogrady:
Have a few that is always charged and then have a storage fee that if not paid, causes the tx to revert.

Both should be included in the tx result (we should not deduct balance hidden)

Issue #73, opened 2023-03-02
Add cost per state key read


patrick-ogrady:
Each transaction should have some sort of fee based on the number of state keys it requires. 

We should not make this the entirety of the fee because it would not allow us to charge for interesting compute within an action.

PR #352, opened 2023-08-13
[chain] Multidimensional Fees + FIFO Mempool

patrick-ogrady:
Resolves: #96
Resolves: #73 

### TODO
- [ ] ~Allow transactions to emit multiple `Outputs`~: #359 
- [x] Add metric for each unit price as gauge
- [ ] ~Limit auth unit usage to prevent DoS~: #382
- [x] Add multiple fee dimensions
- [x] Store fee data in state instead of in block
- [x] Differentiate between storage read from disk or from cache already loaded form block?: https://github.com/ava-labs/hypersdk/pull/365
- [x] Charge storage by byte: #366 
- [x] add `MaxFee` option to spam
- [x] Make mempool FIFO: https://github.com/ava-labs/hypersdk/pull/383
- [x] add README item for "multi-dimensional fee pricing" to optimize resource usage
- [x] Ensure txs only added to mempool if not too far in advance


patrick-ogrady:
Seems to be some bug with `maxFee`:
```
[Test] accepts transaction after state sync concurrent
/Users/patrickogrady/code/hypersdk/examples/morpheusvm/tests/e2e/e2e_test.go:627
generated transaction maxFee: 3699531850
```

patrick-ogrady:
> Seems to be some bug with `maxFee`:
> 
> ```
> [Test] accepts transaction after state sync concurrent
> /Users/patrickogrady/code/hypersdk/examples/morpheusvm/tests/e2e/e2e_test.go:627
> generated transaction maxFee: 3699531850
> ```

I actually think that this is because of all the creations (we probably ran out long before this)

deuszx:
Hey @patrick-ogrady . I haven't read the whole PR (and its dependants) but how (if at all) do you enforce that every validators uses a FIFO for its mempool? Theoretically, I can fork the code and change my mempool algo to something else than FIFO and it still will work properly?

PR #368, opened 2023-08-18
[x/programs] add rust linter

hexfusion:
This PR adds a github action and fixes for linting rust using clippy via the latest nightly.


scripts/rust/tests.lint.sh
```diff
@@ -0,0 +1,28 @@
...
+# rustup component add rustfmt
+# rustup component add rustfmt --toolchain nightly
+# rustup component add clippy
+# rustup component add clippy --toolchain nightly
+
+rustup default stable
+cargo fmt --all --verbose -- --check
+
+rustup default nightly
+cargo +nightly fmt --all -- --config-path .rustfmt.nightly.toml --verbose --check || true
```
hexfusion, line 22: @samliok if we drop the `|| true` here these will run/error can you see if the output is reasonable if yes update the script? If not add a TODO to enable?

scripts/rust/tests.lint.sh
patrick-ogrady, line 1: Is there a reason this isn't in `x/programs/rust`?

samliok (dismissed):

samliok (commented):
looks good.

patrick-ogrady (approved):

patrick-ogrady (changes requested):
Question on action

.github/workflows/rust-static-analysis.yml
```diff
@@ -0,0 +1,42 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "*"
+  pull_request:
+
+permissions:
```
patrick-ogrady, line 14: Why is this required?

hexfusion: removed

x/programs/rust/wasmlanche_sdk/src/host/mod.rs
```diff
@@ -42,7 +42,12 @@ pub fn init_program_storage() -> ProgramContext {
     unsafe { ProgramContext::from(_init_program()) }
 }
 
-pub fn store_bytes(
+/// # Safety
+/// The caller must ensure that key_ptr + key_len and
+/// value_ptr + value_len point to valid memory locations.
+///
+/// Stores the bytes at value_ptr to the bytes at key ptr on the host.
```
hexfusion, line 49: I believe the proper syntax is the safety notice goes below the docs for the method/fucntion

/// Docs
///
/// # Safety

ref. https://doc.rust-lang.org/src/core/ptr/const_ptr.rs.html#406

hexfusion:
```suggestion
/// Stores the bytes at value_ptr to the bytes at key ptr on the host.
///
/// # Safety
/// The caller must ensure that key_ptr + key_len and
/// value_ptr + value_len point to valid memory locations.
```

patrick-ogrady (approved):

PR #370, opened 2023-08-18
[x/programs] Update init_contract to init_program

samliok:
updates name from init_contract


Cargo.lock
```diff
@@ -0,0 +1,184 @@
+# This file is automatically @generated by Cargo.
```
hexfusion, line 1: can we also add this file to ignore?

hexfusion (approved):

patrick-ogrady (approved):

PR #388, opened 2023-08-22
[x/programs] serialize/deserialize using serde

samliok:
#  Serialization using Serde

## Why?
In order to support more complex data structures moving across the WASM boundary we would need to define our own ways to convert them to/from bytes. This is what `ProgramValues` were used for. `ProgramValue's` implemented the `Store` trait with methods `as_bytes`, `from_bytes` & `as_tag`.  Crates like `Serde` & `serde_bare` manage & implement these methods for us, and for more complex structs. This PR uses the `from_slice` and `to_vec` functions from the `serde_bare` crate to serialize/deserialize objects. 

**Advantages**
- Developers are now limited to `Serde's` serialization [types](https://serde.rs/data-model.html) which are much more expansive than `ProgramValues`
- Program UX is greatly simplified. No need to convert to/from ProgramValues, can just use native types/user defined structs.  See before and after at the bottom. 
- Don't have to program serialization/deserialization which can be a non-trivial task.
- Doesn't increase binary size that much. `TokenProgram`  went from ~228kb to ~222kb.

**Disadvantages**
- At least for now, we still need to define our own serialization for parameters. This includes any type that implements the `Argument` trait. So far these are `ProgramContext`, `i64`, `Bytes32`, and `Address`.  The reason for this is b/c when we invoke an external program, we need to serialize *all* params into one byte array. Then in go we need to deserialize in order to pass on the parameters to the desired external function.
- ~Performance costs see @hexfusion's comment~ see performance section

### Changes

- Remove `ProgramValue`, `Store` and related stuff
- Uses the `from_slice` and `to_vec` functions from the `serde_bare` crate to serialize/deserialize objects.
- `Get` and `Store` functions take in a param that implements `Serialize` or `DeserializedOwned`
- `Argument` trait, used during `program_invoke`
- Added `Pokemon` example to show using `serde` for custom structs 
- Update all program examples
- Changes `ProgramContext` -> `Context`
- Added go tests for `Pokemon` and `Lottery` examples.

## Performance

```
goos: darwin
goarch: arm64
                                        │   old.txt   │              new.txt               │
                                        │   sec/op    │   sec/op     vs base               │
TokenProgram/benchmark_token_program-10   8.948m ± 4%   9.777m ± 3%  +9.27% (p=0.000 n=10)

                                        │   old.txt    │               new.txt               │
                                        │     B/op     │     B/op      vs base               │
TokenProgram/benchmark_token_program-10   4.809Mi ± 0%   4.737Mi ± 0%  -1.50% (p=0.000 n=10)

                                        │   old.txt   │              new.txt               │
                                        │  allocs/op  │  allocs/op   vs base               │
TokenProgram/benchmark_token_program-10   43.32k ± 0%   46.83k ± 0%  +8.09% (p=0.000 n=10)
```

Compiling Rust using 
`./scripts/build.sh`

Running test with 
`go test -v -timeout 30s -run ^TestTokenProgram$ github.com/ava-labs/hypersdk/x/programs/examples --count 1`

## Before 
Screenshot 2023-08-24 at 2 31 07 PM

## After

Screenshot 2023-08-24 at 2 30 55 PM


x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -122,79 +117,84 @@ fn get_field_as_bytes(ctx: &ProgramContext, name: &[u8]) -> Result<Vec<u8>, Stor
...
 }
 
 /// Gets the correct key to in the host storage for a [map_name] and [key] within that map  
-fn get_map_key(map_name: &str, key: &ProgramValue) -> Vec<u8> {
-    [map_name.as_bytes(), &key.as_bytes()].concat()
+fn get_map_key<T>(map_name: &str, key: &T) -> Result<Vec<u8>, StorageError>
+where
+    T: Serialize + std::fmt::Debug,
+{
+    let key_bytes = match to_vec(key) {
```
samliok, line 134: metering does not get passed this match @hexfusion

samliok: weird because this isn't even a host call.

samliok (commented):
comments on bug

x/programs/runtime/map.go
```diff
@@ -23,10 +23,10 @@ type maps map[string][]byte
 
 // Key value store for program data
 type storage struct {
-	// uint64 for simplicity, could be a real hash later
-	state   map[uint64]maps
-	mods    map[uint64]api.Module
-	counter uint64
+	// int64 for simplicity, could be a real hash later
```
samliok, line 26: using int64 to avoid overflow errors(WASM uses ints). 
Also this needs to be changed anyways in the actual vm

x/programs/rust/examples/token/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "token"
 version = "0.1.0"
 edition = "2021"
-
+overflow-checks = true
```
samliok, line 5: nice feature

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -9,36 +9,30 @@ use expose_macro::expose;
...
+    token_program
+        .add_field(String::from("total_supply"), 123456789)
+        .unwrap();
+    token_program.into()
 }
 
 /// Gets total supply or -1 on error.
 #[expose]
 pub fn get_total_supply(ctx: ProgramContext) -> i64 {
     ctx.get_value("total_supply")
```
samliok, line 27: 👀 clean

x/programs/rust/wasmlanche_sdk/src/host/mod.rs
```diff
@@ -8,27 +8,27 @@ extern "C" {
 
     #[link_name = "store_bytes"]
     fn _store_bytes(
-        contractId: u64,
+        contractId: i64,
```
samliok, line 11: overflow prevention

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -1,98 +1,91 @@
...
+    where
+        T: Serialize,
+    {
         let key_bytes = key.as_bytes();
         // Add the tag(u8) to the start of val_bytes
-        store_key_value(self, key_bytes.to_vec(), value)
+        store_key_value(
+            self,
+            key_bytes.to_vec(),
+            to_vec(value).expect("Serialization failed"),
```
samliok, line 27: for larger structs to_vec() increases serialization size by ~4x

x/programs/rust/examples/pokemon/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+expose_macro = { version = "0.1.0", path = "../../expose_macro" }
+wasmlanche_sdk = { version = "0.1.0", path = "../../wasmlanche_sdk" }
+rand = "0.8.5"
+rand_chacha = "0.3.1"
+serde_derive = "1.0.185"
+serde = "1.0.185"
+serde_json = "1.0.105"
```
hexfusion, line 15: If the goal for performance is to minimize the footprint required to manage the stack I am wondering if this is the right move to take these dependencies. Why do we need JSON?

x/programs/rust/wasmlanche_sdk/Cargo.toml
```diff
@@ -7,5 +7,8 @@ edition = "2021"
 
 [dependencies]
 thiserror = "1.0.46"
-
+serde = { version = "1.0.185", features = ["derive"] }
+serde_json = "1.0.105"
```
hexfusion, line 11: I am wondering why the SDK needs a dep on serde_json?

x/programs/examples/lottery.go
```diff
@@ -0,0 +1,178 @@
...
+		return err
+	}
+
+	result, err = lotteryRuntime.Call(ctx, "init_program")
+	if err != nil {
+		return err
+	}
+	lotteryContractId := result[0]
+	t.log.Debug("lottery contract id", zap.Uint64("id", lotteryContractId))
+	runtime.GlobalStorage.Programs[uint32(token_contract_id)] = t.tokenProgramBytes
```
hexfusion, line 134: I can understand that this is easier to manage for the test but we should be creating programs persisting those to state then calling them from invoke_program by id. That can be a followup.

x/programs/examples/lottery_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+	_ "embed"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	//go:embed testdata/target/wasm32-wasi/release/lottery.wasm
```
hexfusion, line 15: this is fine for testing but testdata should just contain wasm blobs

```suggestion
//go:embed testdata/lottery.wasm
```

x/programs/examples/lottery.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return err
+	}
+
+	bobLottoPtr, err := lotteryRuntime.WriteGuestBuffer(ctx, bob_pk[:])
+	if err != nil {
+		return err
+	}
+
+	// set the library contract
```
hexfusion, line 147:
```suggestion
	// set the library program
```

x/programs/examples/pokemon.go
```diff
@@ -0,0 +1,122 @@
...
+		"alloc",
+		"dealloc",
+		"init_program",
+	}
+
+	meter := runtime.NewMeter(t.log, t.maxFee, t.costMap)
+	db := utils.NewTestDB()
+	store := newProgramStorage(db)
+
+	// runtime := runtime.New(t.log, nil, store)
```
hexfusion, line 51: nit

x/programs/examples/pokemon.go
```diff
@@ -0,0 +1,122 @@
...
+	runtime := runtime.New(t.log, meter, store)
+	err := runtime.Initialize(ctx, t.programBytes, functions)
+	if err != nil {
+		return err
+	}
+	fmt.Println("initialized")
```
hexfusion, line 57: drop or use logger

x/programs/examples/lottery.go
```diff
@@ -0,0 +1,178 @@
...
+
+	aliceLottoPtr, err := lotteryRuntime.WriteGuestBuffer(ctx, alice_pk[:])
+	if err != nil {
+		return err
+	}
+
+	bobLottoPtr, err := lotteryRuntime.WriteGuestBuffer(ctx, bob_pk[:])
+	if err != nil {
+		return err
+	}
```
hexfusion, line 145: nit we need to dealloc these

x/programs/examples/pokemon.go
```diff
@@ -0,0 +1,122 @@
...
+	)
+
+	// // get owned
+	// result, err = runtime.Call(ctx, "get_owned", contract_id, alicePtr)
+	// if err != nil {
+	// 	return err
+	// }
+	// t.log.Debug("Result",
+	// 	zap.Int64("Success", int64(result[0])),
+	// )
```
hexfusion, line 119: drop

x/programs/examples/pokemon_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+	_ "embed"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	//go:embed testdata/target/wasm32-wasi/release/pokemon.wasm
```
hexfusion, line 15: The script should copy the wasm file into testdata/
```suggestion
	//go:embed testdata/pokemon.wasm
```

x/programs/examples/token.go
```diff
@@ -167,12 +167,22 @@ func (t *Token) Run(ctx context.Context) error {
...
 
 	pk := priv.PublicKey()
-	return runtime.WriteGuestBuffer(ctx, pk[:])
+	ptr, err := runtime.WriteGuestBuffer(ctx, pk[:])
+	return ptr, pk, err
+}
+
+// writeString writes a string to guest memory and returns the pointer to the string.
+// The string is padded with 0s to fit 32 bytes.
+func writeString(ctx context.Context, runtime runtime.Runtime, str string) (uint64, error) {
```
hexfusion, line 183: nit: check the len of the string and err if > 32

x/programs/examples/token_test.go
```diff
@@ -15,15 +15,15 @@ import (
...
+	//go:embed testdata/target/wasm32-wasi/release/token.wasm
 	tokenProgramBytes []byte
 
 	// example cost map
 	costMap = map[string]uint64{
 		"ConstI32 0x0": 1,
 		"ConstI64 0x0": 2,
 	}
-	maxGas uint64 = 3000
+	maxGas uint64 = 30000
```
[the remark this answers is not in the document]
hexfusion: can we make the gas more realistic I think 13k would be fine.

x/programs/runtime/map.go
```diff
@@ -42,14 +41,18 @@ func NewMapModule(log logging.Logger, meter Meter) *MapModule {
...
 	}
 }
 
+// GlobalStorage is a global variable that holds the state of all programs.
+// This is a placeholder storage system intended to show how a wasm program would access/modify persistent storage.
+// Needs to be gloabl, so state can be persisted across multiple runtime intances.
+var GlobalStorage = storage{
+	state:    make(map[int64]maps),
+	counter:  0,
+	Programs: make(map[uint32][]byte),
```
hexfusion, line 53: as noted above programs should be stored in a db on create and retrieved via db get in `program_invoke`.

x/programs/runtime/map.go
```diff
@@ -42,14 +41,18 @@ func NewMapModule(log logging.Logger, meter Meter) *MapModule {
...
-			mods:    make(map[uint64]api.Module),
-			counter: 0,
-		},
 	}
 }
 
+// GlobalStorage is a global variable that holds the state of all programs.
+// This is a placeholder storage system intended to show how a wasm program would access/modify persistent storage.
+// Needs to be gloabl, so state can be persisted across multiple runtime intances.
+var GlobalStorage = storage{
```
hexfusion, line 50: Its fine for now but this should not be exported as mutation of these maps is a possible security risk and non thread safe.

x/programs/runtime/program_invoke.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	invokeModuleName = "program_invoke"
+	invokeModuleName = "program"
```
hexfusion, line 21: why this change?

x/programs/runtime/program_invoke.go
```diff
@@ -65,23 +64,20 @@ func (m *InvokeModule) programInvokeFn(
 	if !ok {
 		return invokeErr
 	}
-	entryFn := utils.GetGuestFnName(string(entryBuf))
+	entryFn := string(entryBuf)
 
 	// get the program bytes stored in state
-	data, ok, err := m.storage.Get(ctx, uint32(programID))
+	data, ok := GlobalStorage.Programs[uint32(invokeProgramID)]
```
hexfusion, line 70: this means all programs must live in memory vs on disk

x/programs/runtime/program_invoke.go
```diff
@@ -65,23 +64,20 @@ func (m *InvokeModule) programInvokeFn(
...
 	if !ok {
 		return invokeErr
 	}
-	if err != nil {
-		return invokeErr
-	}
```
hexfusion, line 77: nit: since we have a logger should we log the actual error here?

x/programs/runtime/program_invoke.go
```diff
@@ -65,23 +64,20 @@ func (m *InvokeModule) programInvokeFn(
 	if !ok {
 		return invokeErr
 	}
-	entryFn := utils.GetGuestFnName(string(entryBuf))
+	entryFn := string(entryBuf)
 
 	// get the program bytes stored in state
-	data, ok, err := m.storage.Get(ctx, uint32(programID))
+	data, ok := GlobalStorage.Programs[uint32(invokeProgramID)]
 	if !ok {
```
hexfusion, line 71: nit: we have a logger so could we log not found?

x/programs/runtime/program_invoke.go
```diff
@@ -65,23 +64,20 @@ func (m *InvokeModule) programInvokeFn(
...
-	}
 
 	// create new runtime for the program invoke call
 	runtime := New(m.log, m.meter, m.storage)
 
-	// only export the function we are calling
-	exportedFunctions := []string{entryFn}
-	err = runtime.Initialize(ctx, data, exportedFunctions)
+	// also need to export alloc function, so we can write args to guest memory if needed
+	exportedFunctions := []string{entryFn, "alloc"}
```
hexfusion, line 79: +1

hexfusion:
The perf cost of this in non trivial

```
goos: linux
goarch: amd64
                                        │   old.txt   │               new.txt                │
                                        │   sec/op    │   sec/op     vs base                 │
TokenProgram/benchmark_token_program-12   16.62m ± 4%   37.37m ± 2%  +124.88% (p=0.000 n=10)

                                        │   old.txt    │               new.txt                │
                                        │     B/op     │     B/op      vs base                │
TokenProgram/benchmark_token_program-12   4.811Mi ± 0%   7.611Mi ± 0%  +58.20% (p=0.000 n=10)

                                        │   old.txt   │                new.txt                │
                                        │  allocs/op  │  allocs/op    vs base                 │
TokenProgram/benchmark_token_program-12   43.33k ± 0%   116.28k ± 0%  +168.35% (p=0.000 n=10)
```

x/programs/runtime/program_invoke.go
samliok: in case we add more methods to the module

x/programs/rust/wasmlanche_sdk/src/types/mod.rs
```diff
@@ -1,73 +1,142 @@
...
 
-impl From<ProgramValue> for ProgramContext {
-    fn from(value: ProgramValue) -> Self {
-        match value {
-            ProgramValue::ProgramObject(i) => i,
-            _ => panic!("Cannot conver to ProgramContext"),
-        }
+impl From<Context> for Box<dyn Argument> {
+    fn from(value: Context) -> Self {
+        Box::new(value)
```
richardpringle, line 140: All of these are unnecessary. The compiler automatically infers the type when necessary.
```rust
let x: Box<dyn Argument> = Box::new(0i64);
```
The above will just work.

x/programs/rust/README.md
```diff
@@ -2,6 +2,17 @@
 
 The `wasm_guest` folder contains the `wasmlanche_sdk`, macros and example programs
 
+## Build Notes
+
+first install the wasm32-wasi target
+`rustup target add wasm32-wasi`
```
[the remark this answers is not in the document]
samliok: been using println! to debug which require system calls

x/programs/rust/examples/lottery/src/lib.rs
```diff
@@ -10,65 +9,53 @@ static TOKEN_PROGRAM_NAME: &str = "token_contract";
 /// Initializes the program.
 #[expose]
 fn init_program() -> i64 {
-    let lottery_contract = Program::new();
-    lottery_contract.publish().unwrap().into()
+    // Initialize the program with no fields
+    Program::new().into()
```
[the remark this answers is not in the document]
samliok: new() calls a host function that initializes the programs id in persistent storage. The host function returns the `id` which is then returned by `init_program`.  if you don't return the correct `i64`, you wouldn't be able to reference that instance of the program.

 It's an `i64` now. Probably will be a hash/address in the future.

x/programs/rust/examples/pokemon/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+    level: u32,
+    hp: u32,
+    moves: Vec<String>,
+}
+
+// non-string keys are not supported by serde
+type OwnedPokemon = Vec<Pokemon>;
+
+#[expose]
+pub fn init_program() -> i64 {
```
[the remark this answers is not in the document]
samliok: yea that's a good idea. This would mean we need extra logic to handle initializing a program in the host though. Gonna make this an issue.

x/programs/rust/examples/lottery/src/lib.rs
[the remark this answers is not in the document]
richardpringle: What happens if my code looks like this:
```rust
#[expose]
fn init_program() -> i64 {
    -1
}
```
The program will never work? You may have already replied to my other comment, but it seems like calling the Go function before calling the wasm function would make more sense. You could then return a `bool` or something where `false` causes a revert or whatever.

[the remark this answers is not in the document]
richardpringle: you answered me here lol
https://github.com/ava-labs/hypersdk/pull/388#discussion_r1304898447

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -1,33 +1,26 @@
...
 use wasmlanche_sdk::types::Address;
 
 /// Initializes the program. This program maps addresses with a count.
 #[expose]
 fn init_program() -> i64 {
     let mut counter_program = Program::new();
-    counter_program.add_field(String::from("counter"), 0.into());
-    counter_program.add_field(String::from("counts"), ProgramValue::MapObject);
-    counter_program.publish().unwrap().into()
+    let _ = counter_program.add_field(String::from("counter"), 0);
```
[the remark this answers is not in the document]
samliok: gonna change this with #417 and use the return values there.

x/programs/rust/wasmlanche_sdk/src/types/mod.rs
```diff
@@ -1,73 +1,113 @@
...
     }
 }
 
-impl From<&str> for ProgramValue {
-    fn from(value: &str) -> Self {
-        ProgramValue::StringObject(String::from(value))
+/// A struct representing a fixed length of 32 bytes.
+/// This can be used for passing strings to the host. It caps the string at 32 bytes,
+#[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Debug)]
+pub struct Bytes32([u8; Self::LEN]);
```
richardpringle, line 29: Not for this PR, but `Bytes32` and `Address` are the exact same thing; why have separate types?

x/programs/rust/wasmlanche_sdk/src/types/mod.rs
```diff
@@ -1,73 +1,113 @@
...
-    fn from(value: i64) -> Self {
-        ProgramValue::IntObject(value)
+/// Implement the Display trait for Bytes32 so that we can print it.
+/// Enables to_string() on Bytes32.
+impl std::fmt::Display for Bytes32 {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        // Find the first null byte and only print up to that point.
+        let null_pos = self.0.iter().position(|&b| b == b'\0').unwrap_or(Self::LEN);
+        String::from_utf8_lossy(&self.0[..null_pos])
+            .to_string()
```
richardpringle, line 47: I don't think you need to allocate here. 
```suggestion
```

x/programs/rust/wasmlanche_sdk/src/types/mod.rs
```diff
@@ -1,73 +1,113 @@
...
-        Self { bytes }
+        Self(bytes)
+    }
+}
+
+pub trait Argument {
+    fn as_bytes(&self) -> Cow<'_, [u8]>;
+    fn is_primitive(&self) -> bool {
+        false
+    }
```
richardpringle, line 76: 👍

x/programs/rust/wasmlanche_sdk/src/types/mod.rs
samliok: Was thinking we would use bytes32 for Strings that we want to pass across go/rust

richardpringle (approved):
LGTM

x/programs/rust/examples/lottery/src/lib.rs
```diff
@@ -10,65 +9,52 @@ static TOKEN_PROGRAM_NAME: &str = "token_contract";
 /// Initializes the program.
 #[expose]
 fn init_program() -> i64 {
```
patrick-ogrady, line 11: Can we change this to `init`?

patrick-ogrady (approved):

PR #389, opened 2023-08-22
[x/programs] simplify metering

hexfusion:
This PR simplifies metering by removing unnecessary channel to support Run method. The meter is intended to be shared between runtimes (program to program) so starting a go routine that only closes on error was unnecessary. runtime.Call will return error if the balance is not sufficient rendering the runtime unable to make progress. Shutdown of the runtime should be handled independently.

This PR also adds a logger to the meter to assist with debugging.


samliok (commented):
lgtm

samliok (approved):

samliok (approved):

PR #395, opened 2023-08-22
[x/programs/rust] Use clippy::pedantic in the SDK

richardpringle:
`Clippy::pendantic` saves you from yourself. I recommend this for any open-source Rust (then ignore rules where it's overkill).

### DIFF
--- a/x/programs/rust/wasmlanche_sdk/src/host/mod.rs
+++ b/x/programs/rust/wasmlanche_sdk/src/host/mod.rs
@@ -32,13 +32,14 @@ extern "C" {
     ) -> i64;
 }
 
-/// Stores the bytes at value_ptr to the bytes at key ptr on the host.
+/// Stores the bytes at `value_ptr` to the bytes at key ptr on the host.
 ///
 /// # Safety
-/// The caller must ensure that key_ptr + key_len and
-/// value_ptr + value_len point to valid memory locations.
+/// The caller must ensure that `key_ptr` + `key_len` and
+/// `value_ptr` + `value_len` point to valid memory locations.
+#[must_use]
 pub unsafe fn store_bytes(
-    ctx: &Context,
+    ctx: Context,
     key_ptr: *const u8,
     key_len: usize,
     value_ptr: *const u8,
@@ -50,21 +51,24 @@ pub unsafe fn store_bytes(
 /// Gets the length of the bytes associated with the key from the host.
 ///
 /// # Safety
-/// The caller must ensure that key_ptr + key_len points to valid memory locations.
-pub unsafe fn get_bytes_len(ctx: &Context, key_ptr: *const u8, key_len: usize) -> i32 {
+/// The caller must ensure that `key_ptr` + `key_len` points to valid memory locations.
+#[must_use]
+pub unsafe fn get_bytes_len(ctx: Context, key_ptr: *const u8, key_len: usize) -> i32 {
     unsafe { _get_bytes_len(ctx.program_id, key_ptr, key_len) }
 }
 
 /// Gets the bytes associated with the key from the host.
 ///
 /// # Safety
-/// The caller must ensure that key_ptr + key_len points to valid memory locations.
-pub unsafe fn get_bytes(ctx: &Context, key_ptr: *const u8, key_len: usize, val_len: i32) -> i32 {
+/// The caller must ensure that `key_ptr` + `key_len` points to valid memory locations.
+#[must_use]
+pub unsafe fn get_bytes(ctx: Context, key_ptr: *const u8, key_len: usize, val_len: i32) -> i32 {
     unsafe { _get_bytes(ctx.program_id, key_ptr, key_len, val_len) }
 }
 
 /// Invokes another program and returns the result.
-pub fn host_program_invoke(call_ctx: &Context, method_name: &str, args: &[u8]) -> i64 {
+#[must_use]
+pub fn program_invoke(call_ctx: Context, method_name: &str, args: &[u8]) -> i64 {
     let method_name_bytes = method_name.as_bytes();
     unsafe {
         _invoke_program(
@@ -83,7 +87,7 @@ pub fn host_program_invoke(call_ctx: &Context, method_name: &str, args: &[u8]) -
 /// Allocate memory into the module's linear memory
 /// and return the offset to the start of the block.
 #[no_mangle]
-pub fn alloc(len: usize) -> *mut u8 {
+pub extern "C" fn alloc(len: usize) -> *mut u8 {
     // create a new mutable buffer with capacity `len`
     let mut buf = Vec::with_capacity(len);
     // take a mutable pointer to the buffer
@@ -103,7 +107,7 @@ pub fn alloc(len: usize) -> *mut u8 {
 ///
 /// deallocates the memory block at `ptr` with a given `capacity`.
 #[no_mangle]
-pub unsafe fn dealloc(ptr: *mut u8, capacity: usize) {
+pub unsafe extern "C" fn dealloc(ptr: *mut u8, capacity: usize) {
     // always deallocate the full capacity, initialize vs uninitialized memory is irrelevant here
     let data = Vec::from_raw_parts(ptr, capacity, capacity);
     std::mem::drop(data);
--- a/x/programs/rust/wasmlanche_sdk/src/store/mod.rs
+++ b/x/programs/rust/wasmlanche_sdk/src/store/mod.rs
@@ -64,7 +64,7 @@ impl Context {
         let key_bytes = key.as_bytes();
 
         match store_key_value(self, key_bytes, value) {
-            Ok(_) => StoreResult::Ok(self),
+            Ok(()) => StoreResult::Ok(self),
             Err(e) => StoreResult::Err(e),
         }
     }
@@ -80,7 +80,7 @@ impl Context {
         };
 
         match store_key_value(self, &key_bytes, value) {
-            Ok(_) => StoreResult::Ok(self),
+            Ok(()) => StoreResult::Ok(self),
             Err(e) => StoreResult::Err(e),
         }
     }
--- a/x/programs/rust/wasmlanche_sdk/src/store/mod.rs
+++ b/x/programs/rust/wasmlanche_sdk/src/store/mod.rs
@@ -87,7 +87,7 @@ impl Context {
 
     /// Returns the value of the field `name` from the host.
     /// # Errors
-    /// Returns a `StorageError` if the field does not exist or if the bytes are invalid.
+    /// Returns a `StorageError` if there was an error retrieving from the host or the retrieved bytes are invalid.
     pub fn get_value<T>(self, name: &str) -> Result<T, StorageError>
     where
         T: DeserializeOwned,
--- a/x/programs/rust/wasmlanche_sdk/src/store/mod.rs
+++ b/x/programs/rust/wasmlanche_sdk/src/store/mod.rs
@@ -167,7 +167,7 @@ fn get_field_as_bytes(ctx: Context, name: &[u8]) -> Result<Vec<u8>, StorageError
     Ok(bytes)
 }
 
-/// Gets the field [name] from the host and returns it as a `ProgramValue`.
+/// Gets the field `name` from the host and returns it as a `ProgramValue`.
 fn get_field<T>(ctx: Context, name: &str) -> Result<T, StorageError>
 where
     T: DeserializeOwned,
@@ -176,7 +176,7 @@ where
     from_slice(&bytes).map_err(|_| StorageError::InvalidBytes)
 }
 
-/// Gets the correct key to in the host storage for a [`map_name`] and [key] within that map  
+/// Gets the correct key to in the host storage for a `map_name` and `key` within that map
 fn get_map_key<T>(map_name: &str, key: &T) -> Result<Vec<u8>, StorageError>
 where
     T: Serialize,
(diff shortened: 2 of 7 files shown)


x/programs/rust/wasmlanche_sdk/src/host/mod.rs
```diff
@@ -97,35 +102,3 @@ pub fn host_program_invoke(
...
-/// # Safety
-/// `ptr` must be a pointer to a block of memory.
-///
-/// deallocates the memory block at `ptr` with a given `capacity`.
-#[no_mangle]
-pub unsafe fn dealloc(ptr: *mut u8, capacity: usize) {
-    // always deallocate the full capacity, initialize vs uninitialized memory is irrelevant here
-    let data = Vec::from_raw_parts(ptr, capacity, capacity);
-    std::mem::drop(data);
-}
```
richardpringle, line 131: Clippy didn't like the usage of `#[no_mangle]` here. It doesn't make sense to have the `#[no_mangle]` attribute if the function isn't being exported with `extern "C" fn // ...`. 

So instead of exposing a function that isn't being called from the wasm-host, I figured it's probably better to delete these. Let me know if you (the reviewer) thinks otherwise.

richardpringle:
If @hexfusion or @samliok have any Rust work-in-progress, that should definitely be merged first. It would be hard to redo this stuff. 

Please pay close attention to the `# Errors` sections that were missing on functions. The docs are my best guess as describing error scenarios and could probably use better language. Please make `suggestion`s!

x/programs/rust/wasmlanche_sdk/src/host/mod.rs
samliok: i think you need #[no_mange] when dealing with wazero and the rust compiler. Otherwise, the compiler can you change the function signature which causes an error in the host.

samliok (commented):
Definitely should wait on this PR until #388 . I've changed a lot of the stuff edited here. The `Error` changes are helpful tho.

x/programs/rust/wasmlanche_sdk/src/host/mod.rs
richardpringle: you should also have `extern "C"` in that case (at least that's what clippy says)

x/programs/rust/wasmlanche_sdk/src/host/mod.rs
```diff
@@ -50,21 +51,24 @@ pub unsafe fn store_bytes(
 /// Gets the length of the bytes associated with the key from the host.
 ///
 /// # Safety
-/// The caller must ensure that key_ptr + key_len points to valid memory locations.
-pub unsafe fn get_bytes_len(ctx: &Context, key_ptr: *const u8, key_len: usize) -> i32 {
+/// The caller must ensure that `key_ptr` + `key_len` points to valid memory locations.
+#[must_use]
+pub unsafe fn get_bytes_len(ctx: Context, key_ptr: *const u8, key_len: usize) -> i32 {
```
richardpringle, line 56: `Context` is `Copy` which means that on 64-bit architecture, copying directly from the stack is going to be faster than pointer operations. I'm not quite sure about 32-bit architecture since, technically, a pointer is smaller than the data, but the pointer indirection could cause a slowdown that's greater than copying the extra 32-bits. 

Happy to change this back if you (the reviewer) thinks I should.

x/programs/rust/wasmlanche_sdk/src/host/mod.rs
```diff
@@ -83,7 +87,7 @@ pub fn host_program_invoke(call_ctx: &Context, method_name: &str, args: &[u8]) -
 /// Allocate memory into the module's linear memory
 /// and return the offset to the start of the block.
 #[no_mangle]
-pub fn alloc(len: usize) -> *mut u8 {
+pub extern "C" fn alloc(len: usize) -> *mut u8 {
```
richardpringle, line 90: Rust does not have a stable ABI, and while it's unlikely to change for function signatures, there are no guarantees. What that means is external calls could break when calling the same "binary" (WASM in this case) compiled with different versions of Rust. `extern "C"` uses the C ABI, which is stable.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -1,40 +1,48 @@
...
 use crate::host::{get_bytes, get_bytes_len, store_bytes};
 use crate::types::Argument;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use serde_bare::{from_slice, to_vec};
 use std::str;
 
-pub enum StoreResult<'a> {
-    Ok(&'a Context),
+#[allow(clippy::module_name_repetitions)]
```
richardpringle, line 10: This is another stutter, but I think it's fine to ignore here since causing a collision with `std::result::Result` would be a pain.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -1,40 +1,48 @@
...
+pub enum StoreResult {
+    Ok(Context),
```
richardpringle, line 12: Changed this since the `Context` type is `Copy`... same reasoning as explained above.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -1,40 +1,48 @@
...
+impl StoreResult {
+    #[must_use]
     pub fn store_value<T>(self, key: &str, value: &T) -> Self
     where
         T: Serialize + ?Sized,
     {
         match self {
             Self::Ok(ctx) => ctx.store_value(key, value),
-            err => err,
+            err @ Self::Err(_) => err,
```
richardpringle, line 24: Now if we are to add new variants to `StoreResult`, we have to handle the behaviour explicitly.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -69,18 +78,27 @@ impl Context {
...
+        match store_key_value(self, &key_bytes, value) {
             Ok(_) => StoreResult::Ok(self),
             Err(e) => StoreResult::Err(e),
         }
     }
-    pub fn get_value<T>(&self, name: &str) -> Result<T, StorageError>
+
+    /// Returns the value of the field `name` from the host.
+    /// # Errors
+    /// Returns a `StorageError` if the field does not exist or if the bytes are invalid.
```
richardpringle, line 90: @samliok, are these comments accurate?

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -69,18 +78,27 @@ impl Context {
...
     where
         T: DeserializeOwned,
     {
         get_field(self, name)
     }
-    pub fn get_map_value<T, U>(&self, map_name: &str, key: &T) -> Result<U, StorageError>
+
+    /// Returns the value of the field `key` from the map `map_name` from the host.
+    /// # Errors
+    /// Returns a `StorageError` if the field does not exist or if the bytes are invalid.
```
richardpringle, line 100: @samliok, are these comments accurate?

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -102,7 +120,7 @@ impl From<i64> for Context {
     }
 }
 
-fn store_key_value<T>(ctx: &Context, key_bytes: Vec<u8>, value: &T) -> Result<(), StorageError>
+fn store_key_value<T>(ctx: Context, key_bytes: &[u8], value: &T) -> Result<(), StorageError>
```
richardpringle, line 123: This was an interesting one... passing in a `Vec<u8>` here assures that the memory is deallocated after this function exits, which *might* be desired. However, I did find one place where we were calling `to_vec` to pass in a value to `store_key_value`, so taking in a slice requires one less allocation. 

*I think* this is better, but I'm open to an argument for the contrary.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -139,20 +157,26 @@ fn get_field_as_bytes(ctx: &Context, name: &[u8]) -> Result<Vec<u8>, StorageErro
     let bytes_ptr = bytes_ptr as *mut u8;
 
     // Take ownership of those bytes grabbed from the host. We want Rust to manage the memory.
-    let bytes = unsafe { Vec::from_raw_parts(bytes_ptr, bytes_len as usize, bytes_len as usize) };
+    let bytes = unsafe {
+        Vec::from_raw_parts(
+            bytes_ptr,
+            bytes_len.try_into().unwrap(),
+            bytes_len.try_into().unwrap(),
```
richardpringle, line 164: I know there's a check above assuring that the cast will succeed, but generally it's better to avoid using `as` casts with numbers unless you're doing something that's guaranteed to succeed (like `usize as u64`). 

I think there are some other improvements that can be made to this function (like calling `into_boxed_slice` so that you can't expand the `Vec`) and possible changing the return type for `get_bytes_len`, but that's for another PR.

x/programs/rust/wasmlanche_sdk/src/host/mod.rs
samliok: i think passing `Context` is better

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -1,40 +1,48 @@
...
-    Ok(&'a Context),
+#[allow(clippy::module_name_repetitions)]
+pub enum StoreResult {
+    Ok(Context),
     Err(StorageError),
 }
 
-impl StoreResult<'_> {
+impl StoreResult {
+    #[must_use]
```
samliok, line 17: why `must_use`?

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
samliok, line 100:
```suggestion
```

yea I think these comments are old

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
samliok, line 90:
```suggestion
```

yep these are old

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
samliok: passed in a Vec<u8> since we don't need the memory after, but if its faster I'm all for it.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -139,20 +157,26 @@ fn get_field_as_bytes(ctx: &Context, name: &[u8]) -> Result<Vec<u8>, StorageErro
...
+fn get_field<T>(ctx: Context, name: &str) -> Result<T, StorageError>
 where
     T: DeserializeOwned,
 {
     let bytes = get_field_as_bytes(ctx, name.as_bytes())?;
     from_slice(&bytes).map_err(|_| StorageError::InvalidBytes)
 }
 
-/// Gets the correct key to in the host storage for a [map_name] and [key] within that map  
+/// Gets the correct key to in the host storage for a [`map_name`] and [key] within that map
```
samliok, line 179: why ` `` ` and `[]`?

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
richardpringle: 🤷  `clippy --fix`. 

Means you have to use the return value. I think it makes sense.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
richardpringle: One of the `clippy::pedantic` lints is making sure that there's an `# Errors` section for any function that errors, so we need this. I just don't know how accurate the comment is.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
richardpringle: `clippy --fix`... but good call out, these docs are actually not formatted properly.

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -69,18 +78,27 @@ impl Context {
...
+        match store_key_value(self, &key_bytes, value) {
+            Ok(()) => StoreResult::Ok(self),
             Err(e) => StoreResult::Err(e),
         }
     }
-    pub fn get_value<T>(&self, name: &str) -> Result<T, StorageError>
+
+    /// Returns the value of the field `name` from the host.
+    /// # Errors
+    /// Returns a `StorageError` if the field does not exist or if the bytes are invalid.
```
samliok, line 90:
```suggestion
    /// Returns a `StorageError` if there was an error retrieving from the host or the retrieved bytes are invalid.
```

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -69,18 +78,27 @@ impl Context {
...
     where
         T: DeserializeOwned,
     {
         get_field(self, name)
     }
-    pub fn get_map_value<T, U>(&self, map_name: &str, key: &T) -> Result<U, StorageError>
+
+    /// Returns the value of the field `key` from the map `map_name` from the host.
+    /// # Errors
+    /// Returns a `StorageError` if the field does not exist or if the bytes are invalid.
```
samliok, line 100:
```suggestion
    /// Returns a `StorageError` if there was an error retrieving from the host or the retrieved bytes are invalid.
```

samliok (approved):

patrick-ogrady (approved):

PR #403, opened 2023-08-23
[x/programs] add metering benchmarks

hexfusion:
This PR adds a benchmark with metering disabled to understand current perf for the token contract. ~8% CPU time overhead currently.

```sh
goos: linux
goarch: amd64
BenchmarkTokenProgram/benchmark_token_program-12         	      72	  15810518 ns/op	 5038119 B/op	   43264 allocs/op
BenchmarkTokenProgram/benchmark_token_program_no_metering-12         	      92	  14548991 ns/op	 5036648 B/op	   43261 allocs/op
```

### DIFF
--- a/x/programs/examples/token.go
+++ b/x/programs/examples/token.go
@@ -15,25 +15,19 @@ import (
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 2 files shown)


hexfusion:
these are now stale will update with wasmtime.

PR #413, opened 2023-08-24
[xprograms] add vm simulator

hexfusion:
This PR is meant to be a testing tool for folks to interact with their compiled WASM programs. 

## Changes
- adds a cli to interact with WASM programs
- program state is now written to disk

#### build

```sh
go build simulator.go
```

### generate new keys

```sh
./simulator key generate

created new private key with public address: sim_key_dc2
```

### create new program tx

This tx publishes and initializes the compiled binary. The compiled wasm must include an `init` function.

```sh
./simulator program create --caller sim_key_dc2 ../examples/testdata/token.wasm

create program action successful program id: 1
```

### invoke program tx

Reference the program id returned from the `create` tx, to invoke functions on your program!
Remember to reference the correct program id, as each published program has its own storage space on disk.

```sh
./simulator program invoke --id 1 --caller sim_key_dc2 --function mint_to --params sim_key_dc2,100

response: [1]

./simulator program invoke --id 1 --caller sim_key_dc2 --function get_balance --params sim_key_dc2

response: [100]
```


x/programs/examples/cmd/simulator/cmd/program.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+
+		params = append(params, caller)
+		resp, err := runtime.Call(ctx, functionName, params...)
+		if err != nil {
+			return err
+		}
+
+		// spoof txID
+		txID := ids.GenerateTestID()
```
hexfusion, line 144: not random :)

x/programs/runtime/dependencies.go
```diff
@@ -25,6 +27,6 @@ type Runtime interface {
 }
 
 type Storage interface {
-	Get(context.Context, uint32) ([]byte, bool, error)
+	Get(context.Context, uint32) (bool, ed25519.PublicKey, []byte, error)
```
samliok, line 30: @hexfusion whats the reason for returning a pubkey?

hexfusion: In the context of program to program calls the owner of the program might be useful.

hexfusion (commented):
quick pass few questions

x/programs/examples/simulator/cmd/program.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+func initalizeProgram(programBytes []byte, caller ed25519.PublicKey) (uint64, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	runtime := runtime.New(log, runtime.NewMeter(log, maxFee, costMap), db, caller)
+	defer runtime.Stop(ctx)
+
+	programID, err := runtime.Create(ctx, programBytes)
```
hexfusion, line 83: the ID here should be a successful transaction as the result of something like create_program action. Flow.

- create_program action -> programID.
- all invoke calls to the program reference that id. The ID will not be created by wasm in the VM so we should not set that pattern in the examples IMO.

x/programs/runtime/runtime.go
```diff
@@ -23,11 +26,12 @@ const (
 	deallocFnName = "dealloc"
 )
 
-func New(log logging.Logger, meter Meter, storage Storage) *runtime {
+func New(log logging.Logger, meter Meter, db database.Database, callerAddress ed25519.PublicKey) *runtime {
```
hexfusion, line 29: having runtime take a database interface here I think is pretty limiting. The storage interface was meant as a way to abstract the backend. It is certainly needs work but I still think runtime should not care about what kind of storage it uses.

x/programs/runtime/runtime.go
```diff
@@ -36,12 +40,36 @@ type runtime struct {
...
+}
+
+func (r *runtime) initProgramStorage(programBytes []byte) (int64, error) {
+	count, err := GetProgramCount(r.db)
+	if err != nil {
+		r.log.Error("failed to get program counter", zap.Error(err))
+		return 0, err
+	}
+	// increment
+	err = IncrementProgramCount(r.db)
```
hexfusion, line 61: These implementation details should not be part of the runtime IMO. This type of logic would be part of an action.

x/programs/runtime/dependencies.go
hexfusion: But yeah in the context of the interface this isn't ideal, ideally the return object could be generic or maybe a string, this was just wip and need to think about it more.

x/programs/runtime/storage.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+const (
+	HRP                = "simulator_key_"
+	programPrefix      = 0x0
+	programCountPrefix = 0x1
+	storagePrefix      = 0x2
+)
+
+func ProgramKey(asset uint64) (k []byte) {
```
hexfusion, line 19: I think the VM will probably want to define their own storage logic. Which is why I didn't go this route to begin with and was poking around the storage interface. Otherwise each VM will be implementing the runtime interface which I don't think we want. They **could** but I would imagine most to use the implementation in hypersdk when it becomes stable. So this needs more thought.

x/programs/utils/utils.go
```diff
@@ -72,3 +74,13 @@ func GetBuffer(mod api.Module, ptr uint32, length uint32) ([]byte, bool) {
 func GetProgramBytes(filePath string) ([]byte, error) {
 	return os.ReadFile(filePath)
 }
+
+func NewLoggerWithLogLevel(level logging.Level) logging.Logger {
```
hexfusion, line 78: I added this but there must be something that already exists :)

x/programs/runtime/runtime.go
```diff
@@ -36,12 +40,36 @@ type runtime struct {
...
+	// store program bytes
+	err = SetProgram(r.db, count, r.callerAddress, programBytes)
+	if err != nil {
+		r.log.Error("failed to set program", zap.Error(err))
+		return 0, err
+	}
+	return int64(count), nil
 }
 
 func (r *runtime) Create(ctx context.Context, programBytes []byte) (uint64, error) {
```
hexfusion, line 75: as noted in the other review I think we should drop the `Create` method from the Runtime interface.

x/programs/runtime/dependencies.go
hexfusion: But yeah none of that should be part of the interface the signature should remain `([]byte, bool, error)`

x/programs/runtime/dependencies.go
```diff
@@ -25,6 +27,6 @@ type Runtime interface {
 }
 
 type Storage interface {
```
hexfusion, line 29: maybe Storage name is too generic been back and forth on that but we should probably make it less vauge.

x/programs/examples/simulator/server/server.go
```diff
@@ -0,0 +1,236 @@
+package main
```
samliok, line 1: don't review this file yet, probably wont have time to review it before friday's demo. Much of the stuff is specifically for the demo, many things will likely change before then.

x/programs/runtime/invoke_program.go
```diff
@@ -25,19 +28,21 @@ const (
...
+func NewInvokeModule(log logging.Logger, mu state.Mutable, meter Meter, db database.Database, callerAddress ed25519.PublicKey) *InvokeModule {
 	return &InvokeModule{
-		mu:      mu,
-		meter:   meter,
-		storage: storage,
-		log:     log,
+		mu:            mu,
+		meter:         meter,
+		log:           log,
+		db:            db,
```
samliok, line 44: while using db might not be the most efficient way now, especially since before we were storing data in memory. I will create and link an issue, but I think this first iteration is a step in the right direction. I don't think we will reach a perfect solution in this PR.

x/programs/runtime/invoke_program.go
```diff
@@ -25,19 +28,21 @@ const (
...
+		callerAddress: callerAddress,
```
samliok, line 45: `caller address` isn't being used for anything now. I see the vision for this, but for now I say we remove, and come back to this change to a separate pr. Will create an issue.

x/programs/runtime/dependencies.go
```diff
@@ -22,9 +24,14 @@ type Runtime interface {
 	WriteGuestBuffer(context.Context, []byte) (uint64, error)
 	// Stop performs a shutdown of the engine.
 	Stop(context.Context) error
+	// Gets the user defined functions
+	GetUserData(ctx context.Context) (map[string]int, error)
+
+   GetCurrentGas(ctx context.Context) uint64
```
samliok, line 30: these methods are used for the server in order grab info about the runtime. I don't necessarily think we should require these in the interface tho...

also need to fmt this

x/programs/runtime/invoke_program.go
```diff
@@ -25,19 +28,21 @@ const (
...
 type InvokeModule struct {
-	mu      state.Mutable
-	meter   Meter
-	storage Storage
-	log     logging.Logger
+	mu            state.Mutable
+	meter         Meter
+	log           logging.Logger
+	db            database.Database
+	callerAddress ed25519.PublicKey // the callerAddress of the new runtime during invoke.
```
samliok, line 35: removing the calleradress requirement from the runtime would also clean this up

samliok (commented):
Just a few notes

x/programs/examples/examples.go
```diff
@@ -10,6 +10,7 @@ import (
```
samliok, line 10: i don't think we ever use programStorage

x/programs/examples/simulator/cmd/program.go
```diff
@@ -0,0 +1,158 @@
...
+		var callParams []uint64
+		if params != "" {
+			for _, param := range strings.Split(params, ",") {
+				switch p := strings.ToLower(param); {
+				case p == "true":
+					callParams = append(callParams, 1)
+				case p == "false":
+					callParams = append(callParams, 0)
+				case strings.HasPrefix(p, HRP):
+					// address
```
hexfusion, line 125:
```suggestion
					// public key
```

x/programs/examples/simulator/cmd/program.go
```diff
@@ -0,0 +1,158 @@
...
+					ptr, err := runtime.WriteGuestBuffer(ctx, pk[:])
+					if err != nil {
+						return err
+					}
+					callParams = append(callParams, ptr)
+				default:
+					// treat like a number
+					var num uint64
+					num, err := strconv.ParseUint(p, 10, 64)
+
```
hexfusion, line 139: minor: drop newline

x/programs/examples/simulator/server/server.go
hexfusion: "io/ioutil" is deprecated :)

x/programs/examples/token.go
```diff
@@ -36,10 +36,15 @@ type Token struct {
 
 func (t *Token) Run(ctx context.Context) error {
 	meter := runtime.NewMeter(t.log, t.maxFee, t.costMap)
-	db := utils.NewTestDB()
-	store := newProgramStorage(db)
 
-	runtime := runtime.New(t.log, meter, store)
+	db, _, err := pebble.New("test.db", pebble.NewDefaultConfig())
```
hexfusion: nit: ideally names like this are consts.

hexfusion (commented):
first pass more soon

x/programs/runtime/storage.go
```diff
@@ -0,0 +1,121 @@
...
+	if errors.Is(err, database.ErrNotFound) {
+		return false, nil, nil
+	}
+	if err != nil {
+		return false, nil, err
+	}
+	return true, v, nil
+}
+
+// [owner]
```
hexfusion, line 94: we dont encode owner anymore

x/programs/runtime/storage.go
hexfusion: per conversation offline we can change this after Friday.

x/programs/runtime/storage.go
```diff
@@ -0,0 +1,121 @@
...
+func ProgramCountKey() []byte {
+	return []byte{programCountPrefix}
+}
+
+// IncrementProgramCount increments the program count by 1
+func IncrementProgramCount(db database.Database) error {
+	countKey := ProgramCountKey()
+	bytes, err := db.Get(countKey)
+	var count uint64
+	if err != nil {
```
hexfusion, line 46: this should be checking for not found else return error or just use `GetProgramCount`

x/programs/runtime/map.go
```diff
@@ -118,28 +87,19 @@ func (m *MapModule) getBytesFn(ctx context.Context, mod api.Module, id int64, ke
...
+	val, err := GetValue(m.db, uint64(id), buf)
 	if err != nil {
-		m.log.Error("failed to allocate memory for value: %v", zap.Error(err))
 		return mapErr
 	}
-	ptr := result[0]
-	// write to memory
-	if !mod.Memory().Write(uint32(ptr), val) {
+
+	ptr, err := utils.WriteBuffer(ctx, mod, val)
```
hexfusion, line 101: we need to dealloc here right?

x/programs/runtime/mock_storage.go
```diff
@@ -1,68 +0,0 @@
-// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
```
hexfusion, line 1: also remove it from mock_gen script

x/programs/runtime/runtime.go
```diff
@@ -24,11 +27,11 @@ const (
 	deallocFnName = "dealloc"
 )
 
-func New(log logging.Logger, meter Meter, storage Storage) *runtime {
+func New(log logging.Logger, meter Meter, db database.Database) *runtime {
```
hexfusion, line 30: just a TODO for now but we don't want  db database.Database on New the runtime should be capable of using a vm defined storage.

x/programs/runtime/runtime.go
hexfusion: this is future TODO

x/programs/runtime/runtime.go
hexfusion: TODO: future conversation

x/programs/runtime/runtime.go
```diff
@@ -128,7 +157,7 @@ func (r *runtime) Call(ctx context.Context, name string, params ...uint64) ([]ui
 	if api == nil {
 		return nil, fmt.Errorf("%w: %s", ErrMissingExportedFunction, name)
 	}
-
+	// api.Definition()
```
hexfusion, line 160: nit

x/programs/runtime/dependencies.go
```diff
@@ -22,9 +22,13 @@ type Runtime interface {
 	WriteGuestBuffer(context.Context, []byte) (uint64, error)
 	// Stop performs a shutdown of the engine.
 	Stop(context.Context) error
+	// Gets the user defined functions
+	GetUserData(ctx context.Context) (map[string]int, error)
+
+	GetCurrentGas(ctx context.Context) uint64
```
hexfusion, line 28: TODO: both of these should be removed.

x/programs/examples/simulator/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### lottery example
+
+```sh
+./simulator key generate <---- alice\'s key
+./simulator key generate <---- bob\'s key
+alice=sim_key_xxx
+bob=sim_key_xxx
+./simulator program create --caller $alice ../examples/testdata/token.wasm
+./simulator program create --caller $alice ../examples/testdata/lottery.wasm
```
hexfusion, line 55: nit: update paths

```suggestion
./simulator program create --caller $alice ../testdata/token.wasm
./simulator program create --caller $alice ../testdata/lottery.wasm
```

x/programs/examples/simulator/cmd/program.go
```diff
@@ -0,0 +1,158 @@
...
+var programCmd = &cobra.Command{
+	Use: "program",
+	RunE: func(*cobra.Command, []string) error {
+		return ErrMissingSubcommand
+	},
+}
+
+var programCreateCmd = &cobra.Command{
+	Use:   "create [path to wasm program]",
+	Short: "Creates a program from a wasm file, calls init_program and returns the program ID",
```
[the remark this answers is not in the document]
samliok: nope, just `init`. changing

x/programs/examples/simulator/cmd/program.go
```diff
@@ -0,0 +1,158 @@
...
+	Short: "Invokes a wasm program stored on disk",
+	PreRunE: func(cmd *cobra.Command, args []string) (err error) {
+		if programID == 0 {
+			return fmt.Errorf("program --id cannot be empty")
+		}
+		if callerAddress == "" {
+			return ErrMissingAddress
+		}
+		if functionName == "" {
+			return fmt.Errorf("function --name cannot be empty")
```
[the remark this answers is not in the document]
samliok: added to this

x/programs/runtime/map.go
samliok: no, the guest takes control of the memory so rust will free it.

x/programs/runtime/invoke_program.go
samliok: created #454

hexfusion:
I can't tag my own PR :)  but this looks good for the next iteration nice job!

samliok (dismissed):

samliok (approved):

go.mod
```diff
@@ -27,6 +29,26 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require (
```
patrick-ogrady, line 32: Why aren't these indirect requires in the other require section?

x/programs/examples/pokemon.go
```diff
@@ -35,10 +35,15 @@ type Pokemon struct {
 
 func (t *Pokemon) Run(ctx context.Context) error {
 	meter := runtime.NewMeter(t.log, t.maxFee, t.costMap)
-	db := utils.NewTestDB()
-	store := newProgramStorage(db)
+	db, _, err := pebble.New(DBPath, pebble.NewDefaultConfig())
```
patrick-ogrady, line 38: is it ok that this is shared across examples? is prefixing handled somehow?

x/programs/runtime/invoke_program.go
```diff
@@ -50,6 +51,8 @@ func (m *InvokeModule) Instantiate(ctx context.Context, r wazero.Runtime) error
 }
 
 // invokeProgramFn makes a call to an entry function of a program in the context of another program's ID.
+// TODO: need to ensure the invokeProgramID is passed in correctly on rust side.
```
patrick-ogrady, line 54: the programID should just be the `txID` when the contract is deployed

x/programs/runtime/storage.go
```diff
@@ -0,0 +1,123 @@
...
+	k[0] = programPrefix
+	return
+}
+
+// StorageKey returns the key used to store a value in the program's storage
+// [storagePrefix] + [programID] + [key]
+func StorageKey(programID uint64, key []byte) (k []byte) {
+	k = make([]byte, 1+consts.Int64Len+len(key))
+	k[0] = storagePrefix
+	binary.BigEndian.PutUint64(k[1:], programID)
```
patrick-ogrady, line 38: leave a TODO here => "should we just hash this value"?

x/programs/examples/consts.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+var (
+	DBPath = "examples.db"
+	// TODO: add more robust default
+	// example cost map
+	CostMap = map[string]uint64{
+		"ConstI32 0x0": 1,
+		"ConstI64 0x0": 2,
+	}
```
patrick-ogrady, line 19: if an instruction isn't in here, what happens?

x/programs/examples/pokemon.go
samliok: yea, there is a program counter stored in the db. when a program is "published" we increment and return the counter as the id.

x/programs/examples/consts.go
samliok: the gas value is 0

hexfusion:
replaced by #537

PR #416, opened 2023-08-25
[x/programs] add runtime client

hexfusion:
This PR adds a client abstraction around the wazero api.Module for managing exported functions of the guest. This PR also adds memory management for any allocations made during the client life cycle.

### DIFF
--- a/x/programs/examples/token.go
+++ b/x/programs/examples/token.go
@@ -40,8 +40,6 @@ func (t *Token) Run(ctx context.Context) error {
 		"mint_to",
 		"get_balance",
 		"transfer",
-		"alloc",
-		"dealloc",
 		"init_program",
 	}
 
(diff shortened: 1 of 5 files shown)


x/programs/examples/token.go
```diff
@@ -112,22 +112,6 @@ func (t *Token) Run(ctx context.Context) error {
 		zap.Int64("alice", int64(result[0])),
 	)
 
-	// deallocate bytes
```
hexfusion, line 115: this is handled automatically by the client

x/programs/examples/token.go
```diff
@@ -40,8 +40,6 @@ func (t *Token) Run(ctx context.Context) error {
 		"mint_to",
 		"get_balance",
 		"transfer",
-		"alloc",
-		"dealloc",
```
hexfusion, line 44: I think we can assume these will always be supported so the user should not need to define them. please conform @samliok

samliok: yea 👍

richardpringle (dismissed):
Not that my approval means much 😉

PR #432, opened 2023-08-29
[x/programs] refactor init

samliok:
Removes `initialize_program` host function and the `Program` struct in Rust. 

Instead we generate and initialize the programs storage and id in `go` and inject it into the `init` function. 

Changes
- Removes the `Program` folder
- Removes the host function `initialize_function`
- Adds a `Create` method to the go runtime that sets up global storage and calls `init` on the program
   -  `Create` "publishes" the program, while `Initialize` grabs the published program.
- Updated the expose macro to ensure the first argument is always of type `Context`
- updates examples & tests


x/programs/runtime/runtime.go
richardpringle, line 1: 👍

x/programs/rust/examples/counter/src/lib.rs
richardpringle, line 1: 🚀

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,22 +1,17 @@
...
 use expose_macro::expose;
 
 /// Initializes the program with a name, symbol, and total supply.
 #[expose]
-pub fn init_program() -> i64 {
-    let mut token_program = Program::new();
-    token_program
-        .add_field(String::from("name"), Bytes32::from("WasmCoin".to_string()))
+pub fn init(ctx: Context) -> bool {
+    ctx.store_value("total_supply", &123456789_i64)
```
richardpringle, line 9: For another PR, but could use some type of builder pattern here to clean this up such that the code looks like this:
```rust
ctx
    .store_value("total_supply", &123456789_i64)
    .store_value("name", "WasmCoin")
    .store_value("symbol", "WACK")
    .is_ok()
```
where 
```rust
impl Context {
    // not sure what the actual signature is, but I think you can get the gist from my example
    fn store_value<K:Serialize, V: Serialize>(&self, key: K, value: V) -> StoreResult<'_> {
        match store_value(key, value) {
            Ok(_) => StoreResult::Ok(&ctx),
            Err(e) => StoreResult::Err(e),
        }
    }
}

enum StoreResult<'a> {
    Ok(&'a ctx),
    Err(Error),
}

impl StoreResult<'_> {
    fn store_value(self, key: K, value: V) -> Self {
        match self {
            Self::Ok(ctx) => ctx.store_value(k, v),
            err => err,
        }
    }
}

File: x/programs/rust/examples/token/src/lib.rs [hunk 2]
```diff
@@ -1,22 +1,17 @@
...
         .and_then(|_| {
-            token_program.add_field(String::from("symbol"), Bytes32::from("WACK".to_string()))
+            ctx.store_value("name", &String::from("WasmCoin"))
+                .and_then(|_| ctx.store_value("symbol", &String::from("WACK")))
```
richardpringle, line 12: Change `store_value<T>` and `store_key_value<T>` to have `T: Serialize + ?Sized` and this will work.

```suggestion
            ctx.store_value("name", "WasmCoin")
                .and_then(|_| ctx.store_value("symbol", "WACK"))
```

x/programs/rust/expose_macro/src/lib.rs
```diff
@@ -4,8 +4,8 @@ use core::panic;
 
 use proc_macro::TokenStream;
 use quote::{quote, ToTokens};
+use syn::__private::Span;
```
richardpringle, line 7: You definitely should not be using `__private` haha. They probably leak what should be private types which is why that module exists and actually isn't private. [The example](https://docs.rs/syn/latest/syn/struct.Ident.html#examples) shows that you should be able to use `use proc_macro2::{Ident, Span};`.

x/programs/rust/expose_macro/src/lib.rs
```diff
@@ -17,10 +17,15 @@ pub fn expose(_: TokenStream, item: TokenStream) -> TokenStream {
...
     let input_args = &input.sig.inputs;
     let new_name = Ident::new(&format!("{}_guest", name), name.span()); // Create a new name for the generated function(name that will be called by the host)
-
-    let full_params = input_args.iter().map(|fn_arg| {
+    let empty_param = Ident::new("ctx", Span::call_site()); // Create an empty parameter for the generated function
+    let full_params = input_args.iter().enumerate().map(|(index, fn_arg)| {
         // A typed argument is a parameter. An untyped(reciever) argument is a self parameter.
         if let FnArg::Typed(PatType { pat, ty, .. }) = fn_arg {
+            // ensure first parameter is Context
+            if index == 0 && !is_context(ty) {
```
richardpringle, line 25: You should additionally check if the name is `init`. Otherwise, there's nothing stopping someone from adding a `ctx` param to their other functions when it doesn't actually get injected which would probably cause UB.

samliok: i think we want to force every function to take a `ctx` as the first param. The host will always inject this as the first parameter.

x/programs/rust/expose_macro/src/lib.rs
richardpringle: (required change)

x/programs/rust/expose_macro/src/lib.rs
richardpringle: (required change)

richardpringle: whoops, my bad haha

x/programs/rust/examples/token/src/lib.rs
samliok: added this to this pr, makes it much cleaner. however we have to implement the Result<> functions like `is_ok` and `is_err`.

x/programs/rust/examples/lottery/src/lib.rs
```diff
@@ -8,19 +7,18 @@ static TOKEN_PROGRAM_NAME: &str = "token_contract";
...
 /// Sets the token contract address and the lotto address. This needs to be set
 /// before play can be called, otherwise there is no reference contract and address.
 #[expose]
-fn set(ctx: Context, counter_ctx: Context, lot_address: Address) {
+fn set(ctx: Context, counter_ctx: Context, lot_address: Address) -> bool {
     ctx.store_value(TOKEN_PROGRAM_NAME, &counter_ctx)
-        .expect("Failed to store token contract address");
-    ctx.store_value("address", &lot_address)
-        .expect("Failed to store address");
+        .store_value("address", &lot_address)
```
richardpringle, line 20: 🥇

x/programs/rust/examples/token/src/lib.rs
richardpringle: Can just have a `into_result` method instead. But save that for another PR if you want to do it. Feel free to resolve this comment.

x/programs/rust/expose_macro/src/lib.rs
```diff
@@ -3,9 +3,9 @@ extern crate proc_macro;
 use core::panic;
 
 use proc_macro::TokenStream;
+use proc_macro2::Span;
```
richardpringle, line 6: 👍

x/programs/rust/wasmlanche_sdk/src/store/mod.rs
```diff
@@ -6,6 +6,40 @@ use serde::de::DeserializeOwned;
...
+            Self::Ok(ctx) => ctx.store_map_value(map_name, key, value),
+            err => err,
+        }
+    }
+    pub fn is_ok(self) -> bool {
+        matches!(self, Self::Ok(_))
+    }
+    pub fn is_err(self) -> bool {
+        matches!(self, Self::Err(_))
+    }
```
richardpringle, line 40: Yeah, I would convert back to a `Result` when you're ready to do other things like check `is_ok`. For another PR though!

richardpringle (approved):

patrick-ogrady (approved):

x/programs/runtime/dependencies.go
```diff
@@ -8,6 +8,8 @@ import (
 )
 
 type Runtime interface {
+	// Create initializes and calls the init function.
+	Create(context.Context, []byte) (uint64, error)
```
hexfusion, line 12: I don't believe this is actually a necessary part of the runtime interface. Once the runtime is initialized then the user can make whatever call they want. My understanding is that a programID would be the result of a transaction generated by the VM.

x/programs/runtime/runtime.go
```diff
@@ -44,6 +44,27 @@ type runtime struct {
 	log    logging.Logger
 }
 
+func (r *runtime) Create(ctx context.Context, programBytes []byte) (uint64, error) {
+	err := r.Initialize(ctx, programBytes)
```
hexfusion, line 48: why isn't this part of `Initialize` Runtime?

x/programs/runtime/runtime.go
```diff
@@ -44,6 +44,27 @@ type runtime struct {
...
+	if err != nil {
+		return 0, err
+	}
+	// get programId
+	programID := initProgramStorage()
+
+	// call initialize
+	result, err := r.Call(ctx, "init", uint64(programID))
```
hexfusion, line 56: init in this context is perhaps required by the examples as they are now but as stated above the programID is going to be the result something like a `create_program` action. I would like us moving towards that direction in each iteration if possible.

x/programs/rust/examples/lottery/src/lib.rs
```diff
@@ -8,19 +7,18 @@ static TOKEN_PROGRAM_NAME: &str = "token_contract";
 
 /// Initializes the program.
 #[expose]
-fn init_program() -> i64 {
+fn init(_: Context) -> bool {
     // Initialize the program with no fields
-    Program::new().into()
+    true
 }
```
hexfusion, line 13: I think the fact that we end up with a hard coded bool here speaks to the fact that init should not be a requirement.

PR #446, opened 2023-09-05
[x/programs] rewording for better dev ui

samliok:
Changes
- `Context` -> `State`
   - methods implemented by `Context` are just changes to `State`
- `ProgramInvoke` -> `InvokeProgram`
   - Since we have `get_map`, `set_map` we should use `invoke_program` to follow the same phrasing style
- `Expose` -> `Public`
  - analogous to public functions in solidity
- Removes the requirement for an `init` function if one is not needed.


x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -1,23 +1,23 @@
-use expose_macro::expose;
-use wasmlanche_sdk::store::Context;
+use expose_macro::public;
```
richardpringle: Should rename the macro-crate while you're at it!

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 # extend if new rust crates are added so rust analyzer isnt confused
 [workspace]
 members = [
-    "x/programs/rust/expose_macro",
+    "x/programs/rust/public_macro",
```
richardpringle, line 4: This should be `sdk_macros` since you can create multiple macros in here. Best leave it open-ended so it doesn't need to be renamed again in the future.

x/programs/rust/examples/counter/src/lib.rs
richardpringle, line 1: 👍

x/programs/rust/examples/token/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-expose_macro = { version = "0.1.0", path = "../../expose_macro" }
+public_macro = { version = "0.1.0", path = "../../public_macro" }
```
richardpringle, line 9: You should also expose this from the SDK instead of making all the examples depend on it directly

x/programs/rust/wasmlanche_sdk/src/host/mod.rs
```diff
@@ -39,7 +39,7 @@ extern "C" {
 /// `value_ptr` + `value_len` point to valid memory locations.
 #[must_use]
 pub unsafe fn store_bytes(
-    ctx: Context,
+    ctx: State,
```
richardpringle, line 42:
```suggestion
    state: State,
```

richardpringle: There are other places for this too. Should be easy to do a global search and replace

richardpringle (changes requested):
Couple of small comments here

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,50 +1,53 @@
...
         .store_value("name", "WasmCoin")
         .store_value("symbol", "WACK")
         .is_ok()
 }
 
 /// Gets total supply or -1 on error.
-#[expose]
-pub fn get_total_supply(ctx: Context) -> i64 {
-    ctx.get_value("total_supply").unwrap()
+#[public]
```
patrick-ogrady, line 16: ❤️

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 # extend if new rust crates are added so rust analyzer isnt confused
 [workspace]
 members = [
-    "x/programs/rust/expose_macro",
+    "x/programs/rust/sdk_macros",
```
richardpringle, line 4: 🚀

x/programs/rust/examples/pokemon/src/lib.rs
```diff
@@ -1,6 +1,6 @@
-use expose_macro::expose;
+use wasmlanche_sdk::public;
```
richardpringle, line 1: I'm guessing this is what `cargo fmt` will fix

x/programs/rust/wasmlanche_sdk/src/lib.rs
```diff
@@ -4,3 +4,4 @@ pub mod errors;
 pub mod host;
 pub mod store;
 pub mod types;
+pub use sdk_macros::public;
```
richardpringle, line 7: 🥇

richardpringle (approved):
Lookin' pretty slick!

x/programs/runtime/runtime.go
```diff
@@ -123,7 +126,7 @@ func (r *runtime) Call(ctx context.Context, name string, params ...uint64) ([]ui
 	}
 
 	if api == nil {
```
hexfusion, line 128: as pre comment above this should never happen if we properly register.

x/programs/runtime/runtime.go
```diff
@@ -52,15 +53,17 @@ func (r *runtime) Create(ctx context.Context, programBytes []byte) (uint64, erro
 	// get programId
 	programID := initProgramStorage()
 
-	// call initialize
+	// call initialize if it exists
 	result, err := r.Call(ctx, "init", uint64(programID))
```
hexfusion, line 57: I think we should remove this. if the user of the runtime wants to call init they should do so. I think we simplify things a lot if we remove create.

flow
- new runtime
- initialize runtime
- use runtime

create would be logic inside of an action right?

samliok: yea, most of the logic in `Create` would be inside of an action that calls `initProgramStorage`.  I say for now, we leave this in because 
- we don't have actions
- we need logic for storing the program in state and returning the programID
  - this logic must be called once per runtime. 
  - However, Initialize can be called multiple times, say to initialize program bytes without creating a new `programID`

x/programs/runtime/runtime.go
samliok: so previously we passed in a list of function names, each that mapped to an `api.function`. I think it simplifies the flow if we just grab the `api_function` when we need it(this would be in `call`).  That way we don't have to store the map, users don't need to pass in function names, and call becomes much more modular.

x/programs/runtime/runtime.go
samliok: also without a function like `Create` i found myself having to duplicate this logic. Would need to add similar code to all of the `examples` &  this helps with the `simulator` pr as well. I'm  using `Create` in the simulator to simulate the action.

x/programs/runtime/runtime.go
hexfusion: Yeah I guess once the mod is initialized the work here is already done.

x/programs/runtime/runtime.go
```diff
@@ -52,15 +53,17 @@ func (r *runtime) Create(ctx context.Context, programBytes []byte) (uint64, erro
...
 	if err != nil {
-		return 0, err
-	}
-
-	// check boolean result from init
-	if result[0] == 0 {
-		return 0, fmt.Errorf("failed to initialize program")
+		if !errors.Is(err, ErrMissingExportedFunction) {
```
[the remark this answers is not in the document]
hexfusion: getting the module in this context is pretty cheap.

x/programs/runtime/runtime.go
hexfusion: fine to leave it for now as its already merged.

hexfusion (approved):

patrick-ogrady (approved):

PR #464, opened 2023-09-12
[x/programs] [poc] validate wasmtime-go

hexfusion:
POC hack PR please do not review. The goal is to evaluate performance of wasmtime-go with fuel metering.

wazero in the slower interpreter mode appears faster than wasmtime-go with only a partial token example.

>goos: linux
goarch: amd64
BenchmarkTokenWazeroProgram/benchmark_token_program_wazero-12         	      78	  15678089 ns/op	 4653056 B/op	   42360 allocs/op

>cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkTokenWasmtimeProgram/benchmark_token_program-12         	      55	  18746760 ns/op	  104953 B/op	     394 allocs/op


hexfusion:
as per #466 we are making the change closing this and will open a new PR

PR #477, opened 2023-09-18
[x/programs] migrate to wasm runtime to wasmtime and refactor for VM integration

hexfusion:
This PR replaces wazero with wasmtime for the runtime engine.

TODO
- [x] Replace and extend meter test
- [x] Add support for precompiling wasm and bench
- [x] Update removed examples to work with new runtime
- [x] Remove Wasi preview support
- [x] Improve UX to only require unsigned ints similar to wazero


x/programs/examples/imports/balance/balance.go
```diff
@@ -0,0 +1,205 @@
...
+	assetID, err := runtime.IDFromOffset(caller, assetPtr)
+	if err != nil {
+		i.log.Error("failed to read assetID from memory",
+			zap.Error(err),
+		)
+		return -1
+	}
+
+	// TODO: add cost for state access?
+	// i.meter.Spend(BalanceCost)
```
patrick-ogrady, line 104: This will be handled by the HyperSDK outside of this context. All we need is for this to capture compute cost.

You should leave a comment indicating as much.

x/programs/examples/storage/storage.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+)
+
+//
+// Program
+//
+
+func ProgramKey(id ids.ID) (k []byte) {
+	k = make([]byte, 1+consts.IDLen)
+	// convert uint64 to bytes
```
patrick-ogrady, line 44: nit: remove this

x/programs/examples/storage/storage.go
```diff
@@ -0,0 +1,222 @@
...
+func GetProgram(
+	db state.Immutable,
+	programID ids.ID,
+) (
+	[]byte, // program bytes
+	bool, // exists
+	error,
+) {
+	k := ProgramKey(programID)
+	ctx, cancel := context.WithCancel(context.Background())
```
patrick-ogrady, line 60: why not make this an argument of `GetProgram`? Nonetheless, the current cancel usage doesn't make sense to me.

x/programs/examples/storage/storage.go
```diff
@@ -0,0 +1,222 @@
...
+
+// SetProgram stores [program] at [programID]
+func SetProgram(
+	db database.KeyValueWriter,
+	programID ids.ID,
+	program []byte,
+) error {
+	k := ProgramKey(programID)
+	v := make([]byte, len(program))
+	copy(v, program)
```
patrick-ogrady, line 80: Do we need to copy this?

x/programs/examples/token.go
```diff
@@ -5,52 +5,79 @@ package examples
...
+func (t *TokenWasmtime) Run(ctx context.Context) error {
+	rt := runtime.New(t.log, t.cfg, t.imports)
+	err := rt.Initialize(ctx, t.programBytes)
 	if err != nil {
 		return err
 	}
 
-	t.log.Debug("initial cost",
-		zap.Int("gas", 0),
+	t.log.Debug("unit",
```
patrick-ogrady, line 65: nit: more descriptive name than `unit`

x/programs/examples/imports/balance/balance.go
patrick-ogrady: We should still charge for any "context switch" operation but it shouldn't be based on state. I think we could start with a single `ContextSwitchCost` in `cfg` and use that for all of these?

x/programs/examples/imports/hashmap/hashmap.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"go.uber.org/zap"
+
+	"github.com/ava-labs/avalanchego/utils/logging"
+
+	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/x/programs/runtime"
+)
+
+// GlobalStorage is a global variable that holds program state.
```
patrick-ogrady, line 17: Can you add a TODO that this will be replaced with `storage` usage?

x/programs/examples/imports/program/program.go
```diff
@@ -0,0 +1,191 @@
...
+		i.log.Error("program does not exist")
+	}
+	if err != nil {
+		i.log.Error("failed to get program from storage",
+			zap.Error(err),
+		)
+		return -1
+	}
+
+	// spend the maximum number of units allowed for this call
```
patrick-ogrady, line 104: nit: we should call this the `ContractInvocationCost`? We should only charge fees as we use them in the other program, no?

patrick-ogrady, line 104: nit: can you elaborate more in this comment on how you do this (spend max, run, refund difference)?

patrick-ogrady: And that this is required because we have different meters?

x/programs/examples/storage/storage.go
```diff
@@ -0,0 +1,222 @@
...
+	copy(k[1+ed25519.PublicKeyLen:], asset[:])
+	binary.BigEndian.PutUint16(k[1+ed25519.PublicKeyLen+consts.IDLen:], BalanceChunks)
+	return
+}
+
+// If locked is 0, then account does not exist
+func GetBalance(
+	ctx context.Context,
+	im state.Immutable,
+	pk ed25519.PublicKey,
```
patrick-ogrady, line 102: We will eventually replace this with a much more generic address scheme but this is ok for now.

x/programs/examples/token.go
```diff
@@ -124,18 +135,18 @@ func (t *Token) Run(ctx context.Context) error {
 	)
 
 	// transfer 50 from alice to bob
-	transferToBob := uint64(50)
-	_, err = runtime.Call(ctx, "transfer", contractId, alicePtr, bobPtr, transferToBob)
+	transferToBob := int64(50)
```
patrick-ogrady, line 138: Is there a reason why this is `int64` now instead of `uint64`?

x/programs/examples/token_test.go
```diff
@@ -12,47 +12,94 @@ import (
...
+	"github.com/ava-labs/hypersdk/x/programs/runtime"
+	"github.com/ava-labs/hypersdk/x/programs/utils"
 )
 
 var (
 	//go:embed testdata/token.wasm
 	tokenProgramBytes []byte
 
-	// example cost map
-	costMap = map[string]uint64{
```
patrick-ogrady, line 22: ❤️

x/programs/examples/token_test.go
```diff
@@ -12,47 +12,94 @@ import (
...
 		logging.NewWrappedCore(
-			logging.Debug,
+			logging.Info,
 			os.Stderr,
 			logging.Plain.ConsoleEncoder(),
 		))
 )
 
-// go test -v -timeout 30s -run ^TestTokenProgram$ github.com/ava-labs/hypersdk/x/programs/examples
+// go test -v -timeout 30s -run ^TestTokenWazeroProgram$ github.com/ava-labs/hypersdk/x/programs/examples
```
patrick-ogrady, line 34: Should this be `Wasmtime`?

x/programs/examples/token_test.go
```diff
@@ -12,47 +12,94 @@ import (
...
+		Build()
+	require.NoError(err)
+	program := NewToken(log, tokenProgramBytes, cfg, imports)
+	err = program.Run(context.Background())
 	require.NoError(err)
 }
 
 // go test -v -benchmem -run=^$ -bench ^BenchmarkTokenProgram$ github.com/ava-labs/hypersdk/x/programs/examples -memprofile benchvset.mem -cpuprofile benchvset.cpu
-func BenchmarkTokenProgram(b *testing.B) {
+func BenchmarkTokenWazeroProgram(b *testing.B) {
```
patrick-ogrady, line 55: Should this be `Wasmtime`?

x/programs/runtime/config.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+// TODO: review Cranelift fine tune knobs when exposed
+// https://github.com/bytecodealliance/wasmtime-go/pull/188
+
+func NewConfigBuilder(meterMaxUnits uint64) *builder {
+	return &builder{meterMaxUnits: meterMaxUnits}
+}
+
+type builder struct {
```
patrick-ogrady, line 25: We should also be limiting max depth in the invoke call^

x/programs/runtime/config.go
```diff
@@ -0,0 +1,178 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package runtime
+
+import "github.com/bytecodealliance/wasmtime-go/v12"
+
+const (
```
patrick-ogrady, line 8: should there be an SIMD default?

x/programs/runtime/dependencies.go
```diff
@@ -5,26 +5,81 @@ package runtime
...
 
-type Storage interface {
-	Get(context.Context, uint32) ([]byte, bool, error)
-	Set(context.Context, uint32, uint32, []byte) error
+type Meter interface {
+	// GetBalance returns the balance of the meters units remaining.
+	GetBalance() uint64
+	// Spend attempts to spend the given amount of units. If the meter has
+	Spend(uint64) (uint64, error)
+	// AddUnits add units back to the meters balance.
```
patrick-ogrady, line 83: nit: `meter's`

x/programs/runtime/runtime.go
```diff
@@ -5,178 +5,220 @@ package runtime
...
+
+	switch r.cfg.compileStrategy {
+	case PrecompiledWasm:
+		// Note: that to deserialize successfully the bytes provided must have been
+		// produced with an `Engine` that has the same compilation options as the
+		// provided engine, and from the same version of this library.
+		//
+		// A precompile is not something we would store on chain.
+		// Instead we would prefetch programs and precompile them.
+		r.mod, err = wasmtime.NewModuleDeserialize(r.store.Engine, programBytes)
```
patrick-ogrady, line 76: Can we create a simple precompile + cache function that can be called?

x/programs/runtime/runtime.go
```diff
@@ -5,178 +5,220 @@ package runtime
...
+		r.mod, err = wasmtime.NewModule(r.store.Engine, programBytes)
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("unsupported compile strategy: %v", r.cfg.compileStrategy)
+	}
 
-	r.engine = wazero.NewRuntimeWithConfig(ctx, wazero.NewRuntimeConfigInterpreter())
+	// TODO: remove wasi host functions.
```
patrick-ogrady, line 89: Can you add this as an explicit TODO on the PR?

hexfusion: done

x/programs/runtime/config.go
hexfusion: default is disabled but it should be explicit I will set a default for every config option and try to document why.

x/programs/runtime/config.go
hexfusion: plus test adding todo

x/programs/examples/token.go
hexfusion: add todo and fixed on next push

x/programs/examples/imports/program/program.go
hexfusion: I extended the meter interface and added a `TransferUnits` method along with docs. TL;DR yes each runtime has an internal meter. Each level we go deeper that accounting must be kept. I think this is fine because all of this logic will be in `programs`.

hexfusion: But we do need a rollback on failure mechanism to ensure unspent units are returned to parent on failure. Will add a TODO but it might need to be a followup.

x/programs/examples/imports/hashmap/hashmap.go
hexfusion: this is now called `storage` and uses a db backend.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -46,3 +58,9 @@ fn get_value(program: Program, of: Address) -> i64 {
         .get(StateKeys::Counter(of).to_vec())
         .expect("failed to get counter")
 }
+
+/// Gets the count at the address for an external program.
+#[public]
+fn get_value_external(program: Program, target: Program, max_units: i64, of: Address) -> i64 {
+    program.call_program(&target, max_units, "get_value", &[Box::new(of)])
+}
```
richardpringle, line 66: I thought of a way that this can be done directly in the `#[public]` macro. The `inc` and `get_value` can be compiled behind a `cfg(lib)` and `cfg(not(lib))` where you can pull another program in as a dependency. 

The limitation here would be that you couldn't design a program that calls another deployment of itself, but I'm not sure that's a bad thing.

x/programs/rust/wasmlanche_sdk/src/host/balance.rs
```diff
@@ -0,0 +1,18 @@
...
+        value_ptr: *const u8,
+        value_len: usize,
+    ) -> i32;
+
+    #[link_name = "get_bytes_len"]
+    fn _get_bytes_len(caller_id: i64, key_ptr: *const u8, key_len: usize) -> i32;
+
+    #[link_name = "get_bytes"]
+    fn _get_bytes(caller_id: i64, key_ptr: *const u8, key_len: usize, val_len: i32) -> i32;
+}
```
richardpringle, line 18: Which IDE are you using? You should make it so the IDE automatically adds a new line at the end of the file. If you're using Vscode, I can find you the stack-overflow post on how to do it.

richardpringle (commented):
Small comment. not sure if you want to do it here or in a follow up. I'm good with either, but I don't think `*_external` should stick around for very long

x/programs/rust/wasmlanche_sdk/src/host/balance.rs
hexfusion: hmm I thought `cargo fmt` did that or I guess why doesn't it? anyways I will look into it.

Cargo.toml
```diff
@@ -9,9 +9,10 @@ members = [
 resolver = "2"
 
 [profile.release]
-opt-level = 3
+opt-level = "s"
```
patrick-ogrady, line 12: Would be great to have a comment about what this is or where you can learn more? (or this is just standard rust and you should disregard)

x/programs/examples/counter_test.go
```diff
@@ -0,0 +1,50 @@
...
+	supported := runtime.NewSupportedImports()
+	supported.Register("state", func() runtime.Import {
+		return pstate.New(log, db)
+	})
+	supported.Register("program", func() runtime.Import {
+		return program.New(log, db)
+	})
+
+	cfg, err := runtime.NewConfigBuilder(maxUnits).
+		WithLimitMaxMemory(18 * runtime.MemoryPageSize). // 18 pages
```
patrick-ogrady, line 38: Is it expected that this will be overridden?

patrick-ogrady: or are we just doing this as an example everywhere?

x/programs/examples/imports/program/program.go
```diff
@@ -0,0 +1,225 @@
...
+	res, err := rt.Call(ctx, function, params...)
+	if err != nil {
+		i.log.Error("failed to call entry function",
+			zap.Error(err),
+		)
+		return -1
+	}
+
+	// stop the runtime to prevent further execution
+	rt.Stop()
```
patrick-ogrady, line 167: ❤️

x/programs/rust/examples/counter/src/lib.rs
hexfusion: sounds interesting please create an issue for that so we can track it.

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -0,0 +1,182 @@
...
+
+	keyBytes, err := memory.Range(uint64(keyPtr), uint64(keyLength))
+	if err != nil {
+		i.log.Error("failed to read key from memory",
+			zap.Error(err),
+		)
+		return -1
+	}
+
+	k := storage.ProgramPrefixKey(programIDBytes, keyBytes)
```
patrick-ogrady, line 153: This will also need to have the max size (in chunks) suffixed at the end of it but we can do that in a follow-up.

x/programs/examples/storage/storage.go
```diff
@@ -0,0 +1,209 @@
...
+	mu state.Mutable,
+	programID ids.ID,
+	program []byte,
+) error {
+	k := ProgramKey(programID)
+	return mu.Insert(ctx, k, program)
+}
+
+//
+// Balance
```
patrick-ogrady, line 89: Typically this storage would not have "balance" access and that would be implemented as a native smart contract. Can you elaborate on how this works (seems to be out-of-scope of this package unless I'm missing something)?

patrick-ogrady: Unless this is a type of primitive variable type we want to expose (could buy that)?

x/programs/examples/util.go
```diff
@@ -0,0 +1,35 @@
...
+package examples
+
+import (
+	"context"
+
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/x/programs/runtime"
+)
+
+func newKeyPtr(ctx context.Context, key ed25519.PublicKey, runtime runtime.Runtime) (uint64, error) {
```
patrick-ogrady, line 13: We'll turn `ed25519.PublicKey` into a new type soon but good enough for now

x/programs/runtime/config.go
```diff
@@ -0,0 +1,187 @@
...
+	defaultMaxWasmStack    = 256 * 1024 * 1024 // 256 MiB
+	defaultWasmThreads     = false
+	defaultFuelMetering    = true
+	defaultWasmMultiMemory = false
+	defaultWasmMemory64    = false
+	defaultLimitMaxMemory  = 16 * 64 * 1024 // 16 pages
+)
+
+// TODO: review Cranelift fine tune knobs when exposed
```
patrick-ogrady, line 17: Can we create an issue to do this (I think its been merged and released)?

x/programs/runtime/config.go
```diff
@@ -0,0 +1,187 @@
...
+		b.limitMaxMemory = defaultLimitMaxMemory
+	}
+
+	return &Config{
+		engine: cfg,
+
+		limitMaxTableElements: 8192,
+		limitMaxMemory:        b.limitMaxMemory,
+		limitMaxTables:        1,
+		limitMaxInstances:     32,
```
patrick-ogrady, line 166: Very curious how this is used but I'll have to do my own research

x/programs/runtime/config.go
```diff
@@ -0,0 +1,187 @@
...
+		compileStrategy:       b.compileStrategy,
+		meterMaxUnits:         b.meterMaxUnits,
+		contextSwitchCost:     b.contextSwitchCost,
+	}, nil
+}
+
+// non-configurable defaults
+func defaultWasmtimeConfig() *wasmtime.Config {
+	cfg := wasmtime.NewConfig()
+	// defaults
```
patrick-ogrady, line 177: Don't we need to set SIMD?

x/programs/runtime/runtime.go
```diff
@@ -5,178 +5,229 @@ package runtime
...
+	})
+}
 
-	// allocate to guest and return offset
-	result, err := r.Call(ctx, allocFnName, uint64(len(buf)))
+// PreCompileWasm returns a precompiled wasm module.
+//
+// Note: these bytes can be deserialized by an `Engine` that has the same version.
+// For that reason precompiled wasm modules should not be stored on chain.
+func PreCompileWasmBytes(programBytes []byte, cfg *Config) ([]byte, error) {
```
patrick-ogrady, line 200: LOVE that this is a separate function. Provides a ton of flexibility on how this can be done (rather than relying on the engine to do all caching)

x/programs/examples/imports/pstate/pstate.go
patrick-ogrady: https://github.com/ava-labs/hypersdk/blob/main/keys/keys.go

x/programs/runtime/config.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	if b.limitMaxMemory == 0 {
+		b.limitMaxMemory = defaultLimitMaxMemory
+	}
+
+	return &Config{
+		engine: cfg,
+
+		limitMaxTableElements: 8192,
```
patrick-ogrady, line 163: I think we can decrease this?

x/programs/runtime/config.go
patrick-ogrady: discussed and we decided to specify ALL options even if default to false

Cargo.toml
hexfusion: good callout and added the link but adding more details here.

So what I found with default vs "s" is a 13% reduction in wasm size and a potential performance gain. What was interesting here is when I run the bench -count 10 it blows up[1].

```
                                                                │ without.txt  │              with_s.txt              │
                                                                │    sec/op    │    sec/op     vs base                │
TokenProgram/benchmark_token_program_compile_and_cache-12         2.595m ± ∞ ¹   2.575m ± ∞ ¹       ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_compile_and_cache_short-12   1.946m ± ∞ ¹   1.817m ± ∞ ¹       ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_precompile-12                1.702m ± ∞ ¹   1.705m ± ∞ ¹       ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_precompile_short-12          1.072m ± ∞ ¹   1.062m ± ∞ ¹       ~ (p=1.000 n=1) ²
geomean                                                           1.742m         1.706m        -2.07%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                                                │  without.txt  │               with_s.txt               │
                                                                │     B/op      │     B/op       vs base                 │
TokenProgram/benchmark_token_program_compile_and_cache-12         80.76Ki ± ∞ ¹   72.70Ki ± ∞ ¹        ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_compile_and_cache_short-12   47.44Ki ± ∞ ¹   39.55Ki ± ∞ ¹        ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_precompile-12                264.4Ki ± ∞ ¹   240.8Ki ± ∞ ¹        ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_precompile_short-12          231.3Ki ± ∞ ¹   207.3Ki ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                                                           123.7Ki         109.5Ki        -11.54%
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                                                                │ without.txt  │              with_s.txt              │
                                                                │  allocs/op   │  allocs/op    vs base                │
TokenProgram/benchmark_token_program_compile_and_cache-12         2.479k ± ∞ ¹   2.474k ± ∞ ¹       ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_compile_and_cache_short-12    395.0 ± ∞ ¹    414.0 ± ∞ ¹       ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_precompile-12                2.471k ± ∞ ¹   2.472k ± ∞ ¹       ~ (p=1.000 n=1) ²
TokenProgram/benchmark_token_program_precompile_short-12           394.0 ± ∞ ¹    394.0 ± ∞ ¹       ~ (p=1.000 n=1) ³
geomean                                                            988.1          999.4        +1.14%
```

[1] 
> thread '<unnamed>' panicked at 'failed to spawn thread: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/thread/mod.rs:686:29
stack backtrace:
   0:          0x168d351 - <unknown>
   1:          0x16f1adf - <unknown>
   2:          0x16814e7 - <unknown>
   3:          0x168d165 - <unknown>
   4:          0x168ff43 - <unknown>
   5:          0x168fcd4 - <unknown>
   6:          0x1690589 - <unknown>
   7:          0x1690487 - <unknown>
   8:          0x168d7b6 - <unknown>
   9:          0x16901d2 - <unknown>
  10:           0x7077b3 - <unknown>
  11:           0x707d43 - <unknown>
  12:          0x13b69f5 - <unknown>
  13:          0x139db4f - <unknown>
  14:          0x139928d - <unknown>
  15:          0x139912b - <unknown>
  16:          0x114657e - <unknown>
  17:           0xde920a - <unknown>
  18:           0xcbeb5f - <unknown>
  19:           0x7a1ce4 - <unknown>
fatal runtime error: failed to initiate panic, error 5
SIGABRT: abort
PC=0x7faad5c969fc m=6 sigcode=18446744073709551610
signal arrived during cgo execution

x/programs/rust/examples/counter/src/lib.rs
richardpringle: #523 for tracking

x/programs/rust/wasmlanche_sdk/src/host/balance.rs
richardpringle: I thought cargo fmt did that too... Weird

richardpringle (approved):

x/programs/examples/counter_test.go
hexfusion: Yeah I am trying to be explicit in the examples to understand how the memory needs change. Also to understand what is reasonable as a default.

x/programs/examples/storage/storage.go
hexfusion: yeah I brought in a bunch of stuff for the example/storage that needs to get cleaned up.

x/programs/runtime/config.go
hexfusion: I did update to the latest and added a default for `enable_nan_canonicalization`. But will add an issue for digging further.

x/programs/runtime/config.go
hexfusion: yeah these defaults are from other production environments but that doesn't mean they are right for us.

x/programs/runtime/config.go
hexfusion: docs claim 
// number of instances of this module can be instantiated in parallel

x/programs/examples/imports/pstate/pstate.go
hexfusion: https://github.com/ava-labs/hypersdk/issues/528

patrick-ogrady (approved):

PR #499, opened 2023-09-25
[x/programs] Refactor SDK storage

hexfusion:
The current host module `map` was a POC example of storing data on the host. In order to integrate programs into a `VM` we need  to refactor `Program` storage. This PR proposes a new host module called `state_keys` and changes to the SDK to provide clear separation between interactions with module `Memory` and the hosts `State`.

- [x] rename `State` to `Program`
- [x] propose a new enum based `Program` state schema which emulates HyperSDK.
- [x] Document how a Program can utilize `Memory` and `State`
- [x] New macro `state_keys` which implements `StateKeys` for the user and helps to standardize the format.
- [x] Handle errors inline with ETH standards. In general all examples will show halt on error usage.
- [x] Remove Wasi dependencies from example programs
- [x] Add tests

cargo expand for new `state_key` macro shows its implementation
```rust
/// The program state keys.
#[repr(u8)]
enum StateKey {
    /// The total supply of the token.
    TotalSupply,
    /// The name of the token.
    Name,
    /// The symbol of the token.
    Symbol,
    /// The balance of the token by address.
    Balance(Address),
}
#[automatically_derived]
impl ::core::marker::Copy for StateKey {}
#[automatically_derived]
impl ::core::clone::Clone for StateKey {
    #[inline]
    fn clone(&self) -> StateKey {
        let _: ::core::clone::AssertParamIsClone<Address>;
        *self
    }
}
impl StateKey {
    pub fn to_vec(self) -> Vec<u8> {
        match self {
            Self::TotalSupply => {
                <[_]>::into_vec(#[rustc_box] ::alloc::boxed::Box::new([0u8]))
            }
            Self::Name => <[_]>::into_vec(#[rustc_box] ::alloc::boxed::Box::new([1u8])),
            Self::Symbol => <[_]>::into_vec(#[rustc_box] ::alloc::boxed::Box::new([2u8])),
            Self::Balance(a) => std::iter::once(3u8).chain(a.into_iter()).collect(),
        }
    }
}
```


x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+supported by WebAssembly Memory is critical for communication between `Program`
+and host.
+
+- `State` is a host module which is imported into the `Program` instance, which
+exposes access to a database store. Unlike `Memory` state can be durable.
+
+### Memory
+
+In this example we share a byte slice between our host and a `Program`. To
+achieve this we can write our byte slice to the shared memory from he host using
```
exdx, line 43:
```suggestion
achieve this we can write our byte slice to the shared memory from the host using
```

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+memory := runtime.Memory()
+
+bytes := []byte("hello world")
+bytesLen := uint64(len(bytes))
+
+// create a memory allocation 
+offset, _ := memory.Alloc(bytesLen)
+
+// copy bytes into memory allocation
+offset, _ = memory.Write(offset, bytes)
```
exdx, line 60: I'm a little confused by calling both variables `offset`. From what I can tell, the first `offset` is actually a pointer to the start of the allocation of memory for the byte array. The second `offset` is actually an offset, a length between the start and finish of the new allocation. If this is correct, should we name the first offset differently?

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
 
 On the Go side, we unmarshal in the same order.
 
+## Storage
+
+A Program has 2 storage types.
+
+- `Memory` in WebAssembly is a linear buffer of unsigned bytes that can read and
+written to by both the guest or host. Because of the very minimal types
```
exdx, line 33: nit: The disambiguation between guest and host is non-obvious in this context. Is the `Program` a guest? Might be worth providing more context.

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+  let from_balance :=  program.state().get(Storage::Balance(from).into().as_ref())?;
+
+  if from_balance.lt(amount) {
+    return -1
+  }
+
+  // update balances to state
+  program.state().
+    put(Storage::Balance(to).into().as_ref(), &(to_balance + amount )).
+    put(Storage::Balance(from).into().as_ref(), &(from_balance - amount ))?
```
exdx, line 116: Is this atomic? What if the first `put` succeeds and the second `put` fails, will the state revert back?

exdx (approved):
LGTM just some thoughts

x/programs/rust/README.md
hexfusion: they are the same write should probably not return the offset as it is passed directly.

exdx:
In terms of actual design feedback, I'm not caught up enough to provide a clear opinion. But the proposed design seems reasonable, and importantly, simple to understand. So that's good.

x/programs/rust/README.md
hexfusion: They are the same `Write` should not return the offset as it is passed directly. Will update.

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
 On the Go side, we unmarshal in the same order.
 
+## Storage
+
+A Program has 2 storage types.
+
+- `Memory` in WebAssembly is a linear buffer of unsigned bytes that can read and
+written to by both the guest or host. Because of the very minimal types
+supported by WebAssembly Memory is critical for communication between `Program`
+and host.
```
richardpringle, line 35: We don't need this last sentence. But if we do, it definitely needs a `,` after `WebAssembly`, then `Memory` should be surrounded by backticks

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+
+- `State` is a host module which is imported into the `Program` instance, which
+exposes access to a database store. Unlike `Memory` state can be durable.
```
richardpringle, line 38:
```suggestion
- `State` is a host module which is imported into the `Program` instance. It
exposes access to a database store. Unlike `Memory` state can be durable.
```

x/programs/rust/README.md
richardpringle: Also, "To achieve this," (missing comma)

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+
+### Memory
+
+In this example we share a byte slice between our host and a `Program`. To
+achieve this we can write our byte slice to the shared memory from he host using
+an exported functions called `alloc` and `memory` To do this we would first
```
richardpringle, line 44:
```suggestion
an exported functions called `alloc` and `memory`. To do this, we would first
```

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+allocate  a chunk of continuous memory of the length desired.  `Alloc` will
```
richardpringle, line 45:
```suggestion
allocate a chunk of continuous memory of the length desired.  `Alloc` will
```

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+The below shows an exported `Program` function `some_function` which accepts a
+message of arbitrary length. Unlike fixed size length objects such as `Address`
+the length can be variable. So the `length` of the memory along with the
+`offset` are both required.
+
+```rust
+#[public]
+pub fn some_function(program: Program, message: i64, message_length: i64) -> i64 {
+  // read message bytes from memory
+  let msg := program.memory().read(message, message_length)?;
```
richardpringle, line 83:
```suggestion
  let msg = program.memory().read(message, message_length)?;
```

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+          
+Each  Memory is tightly bound to the lifecycle of the module. For this reason
+`Memory` is non durable and can not be relied upon. 
+
+### State
+
+- `Storage` is the definition of the schema used to persist Program data.
+
+```rust
+#[reps(u8)]
```
richardpringle, line 96:
```suggestion
#[repr(u8)]
```

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+```rust
+#[public]
+pub fn some_function(program: Program, message: i64, message_length: i64) -> i64 {
+  // read message bytes from memory
+  let msg := program.memory().read(message, message_length)?;
+  // msg = [119 111 114]
+}
+```
+          
+Each  Memory is tightly bound to the lifecycle of the module. For this reason
```
richardpringle, line 88:
```suggestion
Each `Memory` is tightly bound to the lifecycle of the module. For this reason,
```
you need a "format on save" on your markdown!

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+`Memory` is non durable and can not be relied upon. 
+
+### State
+
+- `Storage` is the definition of the schema used to persist Program data.
+
+```rust
+#[reps(u8)]
+#[storage]
```
richardpringle, line 97: 🤔, the `storage` attribute should automatically add a `repr` attribute. I'm guessing that `repr(Rust)` would be invalid here, so the `Storage(u8)` could validate its input and pass it to `repr`

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+enum Storage {
+  /// [program_prefix_byte][program_id_bytes][0x0][address_bytes] = amount
+  Balance(Address) // 0x0
+}
+
+#[public]
+pub fn transfer(program: Program, from: Address, to: Address, amount: i64) -> i64 {
+  // check balances from state
+  let to_balance := program.state().get(Storage::Balance(to).into().as_ref())?;
+  let from_balance :=  program.state().get(Storage::Balance(from).into().as_ref())?;
```
richardpringle, line 107:
```suggestion
  let to_balance = program.state().get(Storage::Balance(to).into().as_ref())?;
  let from_balance =  program.state().get(Storage::Balance(from).into().as_ref())?;
```

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+
+  if from_balance.lt(amount) {
+    return -1
```
richardpringle, line 110:
```suggestion
    return -1;
```

x/programs/rust/README.md
```diff
@@ -25,6 +25,100 @@ Just a couple things to note. Serialization is minimal, yet there are certain as
...
+  }
+
+  // update balances to state
+  program.state().
+    put(Storage::Balance(to).into().as_ref(), &(to_balance + amount )).
+    put(Storage::Balance(from).into().as_ref(), &(from_balance - amount ))?
+  [...]
```
richardpringle, line 117:
```suggestion
  program
    .state()
    .put(Storage::Balance(to).into().as_ref(), &(to_balance + amount ))?
    .put(Storage::Balance(from).into().as_ref(), &(from_balance - amount ))?
```
Just guessing here.

x/programs/rust/README.md
```diff
@@ -34,9 +128,30 @@ This folder provides the necessary tools to build WASM programs using rust.
 - `/host` : Imports necessary functions from the host.
 - `/Program`: Defines the `ProgramValue` and `Progam` types.
 
-### Expose Macro
-
-A rust crate that contains an attribute procedural macro `expose` allowing program functions to be exposed to the host.
+### Exporting Program Functions
+
+A program exposes functions that can be called by the host called exports. To
```
richardpringle, line 133:
```suggestion
A program exposes functions that can be called by the host. To
```
? "Called" twice like that is kind of confusing. Maybe `named` or something like that?

x/programs/rust/README.md
```diff
@@ -34,9 +128,30 @@ This folder provides the necessary tools to build WASM programs using rust.
...
+export a function using the Rust SDK we use the `#[public]` attribute above the signature
```
richardpringle, line 134:
```suggestion
export a function using the Rust SDK, we use the `#[public]` attribute above the signature
```

x/programs/rust/README.md
```diff
@@ -34,9 +128,30 @@ This folder provides the necessary tools to build WASM programs using rust.
...
+of the function we want to export.
+
+The example below shows a basic function `add` which is accessible by the host
```
richardpringle, line 137:
```suggestion
The example below shows a basic function `add`, which is accessible by the host
```

x/programs/rust/README.md
hexfusion: makes sense

exdx:
In Stellar they use `env` as a wrapper around storage. Might be worth considering, the UX is nice. https://soroban.stellar.org/docs/getting-started/storing-data#contract-data-access

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,27 +1,89 @@
...
+}
+
+impl StateKey {
+    pub fn to_vec(self) -> Vec<u8> {
+        match self {
+            StateKey::TotalSupply => vec![self.into()],
+            StateKey::Name => vec![self.into()],
+            StateKey::Symbol => vec![self.into()],
+            StateKey::Balances(a) => {
+                let mut bytes = Vec::with_capacity(1 + 32);
```
hexfusion, line 20: 32 should be a const

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,27 +1,89 @@
...
+/// The internal state keys used by the program.
+#[repr(u8)]
+enum StateKey {
+    TotalSupply,
+    Name,
+    Symbol,
+    Balances(Address),
+}
+
+impl StateKey {
```
hexfusion, line 13: TODO: a proc macro should implement all of below

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,27 +1,89 @@
...
-pub fn init(state: State) -> bool {
-    state
-        .store_value("total_supply", &123456789_i64)
-        .store_value("name", "WasmCoin")
-        .store_value("symbol", "WACK")
-        .is_ok()
+pub fn init(program: Program) -> bool {
+    program
+        .state()
+        .insert(&StateKey::TotalSupply.to_vec(), &123456789_i64)
```
hexfusion, line 51: I think this would ideally be 
```
.insert(StateKey::TotalSupply.into().as_ref(), &123456789_i64)
```
but my rust is rusty

hexfusion:
> In Stellar they use `env` as a wrapper around storage. Might be worth considering, the UX is nice. https://soroban.stellar.org/docs/getting-started/storing-data#contract-data-access

I think we started there and have changed a few times. I think `Program` makes sense but open to further conversation.

x/programs/rust/examples/token/src/lib.rs
richardpringle: OR... you could use an iterator and let the compiler handle capacities for you, 😉.
```rust
StateKey::Balances(a) => {
    std::iter::once(self as u8).chain(a.as_bytes().copied()).collect()
}
```

hexfusion: cool sounds good

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -105,9 +106,9 @@ impl Argument for i64 {
     }
 }
 
-impl Argument for State {
+impl Argument for Program {
     fn as_bytes(&self) -> Cow<'_, [u8]> {
-        self.program_id.as_bytes()
+        Cow::Owned(self.id().to_be_bytes().to_vec())
```
richardpringle, line 120: This doesn't make sense...

x/programs/rust/wasmlanche_sdk/src/host/state.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn _len(caller_id: i64, key_ptr: *const u8, key_len: usize) -> i32;
+}
+
+/// Persists the bytes at `value_ptr` to the bytes at key ptr on the host storage.
+///
+/// # Safety
+/// The caller must ensure that `key_ptr` + `key_len` and
+/// `value_ptr` + `value_len` point to valid memory locations.
+#[must_use]
+pub unsafe fn put_bytes(
```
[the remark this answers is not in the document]
hexfusion: gotcha yeah that makes sense.

x/programs/rust/wasmlanche_sdk/src/memory.rs
```diff
@@ -0,0 +1,71 @@
...
+    let ptr = buf.as_mut_ptr();
+    // take ownership of the memory block and
+    // ensure that its destructor is not
+    // called when the object goes out of scope
+    // at the end of the function
+    std::mem::forget(buf);
+    // return the pointer so the runtime
+    // can write data at this offset
+    ptr
+}
```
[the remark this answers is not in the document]
hexfusion: basically anything not an int.

x/programs/rust/wasmlanche_sdk/src/program.rs
```diff
@@ -0,0 +1,81 @@
...
+    // Calculate the total size of the combined byte slices
+    let total_size = args.iter().map(|cow| cow.len() + META_SIZE).sum();
+
+    // Create a mutable Vec<u8> to hold the combined bytes
+    let mut bytes = Vec::with_capacity(total_size);
+
+    for arg in args {
+        // if we want to be efficient we dont need to add length of bytes if its an int
+        let len = i64::try_from(arg.len()).expect("Error converting to i64");
+        bytes.extend_from_slice(&len.as_bytes());
```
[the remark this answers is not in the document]
hexfusion: existing code

x/programs/rust/wasmlanche_sdk/src/types.rs
hexfusion: yeah the usage is in the entire file I will dig into it.

x/programs/rust/wasmlanche_sdk/src/program.rs
```diff
@@ -0,0 +1,81 @@
...
+        bytes.extend_from_slice(&[u8::from(arg.is_primitive())]);
+        bytes.extend_from_slice(&arg.as_bytes());
```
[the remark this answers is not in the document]
hexfusion: in this case it does not

```
mismatched types
expected reference `&[u8]`
        found enum `Cow<'_, [u8]>`
```

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,27 +1,89 @@
...
+impl Into<u8> for StateKey {
+    fn into(self) -> u8 {
+        match self {
+            StateKey::TotalSupply => 0,
+            StateKey::Name => 1,
+            StateKey::Symbol => 2,
+            StateKey::Balances(_) => 3,
+        }
+    }
+}
```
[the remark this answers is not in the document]
hexfusion: new macro solves this

x/programs/rust/wasmlanche_sdk/src/state.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    pub fn program_id(&self) -> Program {
+        self.program
+    }
+     /// Insert a key and value to the host storage.
+    ///
+    /// # Errors Returns an `StateError` if the key or value cannot be
+    /// serialized or if the host fails to write the key and value.
+    pub fn insert<V>(&self, key: &[u8], value: &V) -> Result<(), StateError>
```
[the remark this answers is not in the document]
hexfusion: That would depend on the implementation of the underlying storage. In the case of merkledb if execution fails then no state is actually mutated.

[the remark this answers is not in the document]
hexfusion: Updated to store as insert does imply a new record.

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,27 +1,89 @@
-use wasmlanche_sdk::{public, store::State, types::Address};
+use wasmlanche_sdk::{program::Program, public, types::Address};
+
+
+/// The internal state keys used by the program.
+#[repr(u8)]
```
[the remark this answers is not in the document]
hexfusion: automatically handled by the macro

x/programs/rust/examples/token/src/lib.rs
hexfusion: Done

x/programs/rust/README.md
richardpringle: @hexfusion, FYI. I stumbled upon this if you're looking for inspiration
https://docs.rs/enumflags2_derive/latest/src/enumflags2_derive/lib.rs.html#56-79

They force you to add `repr` separately, but I think the validation would work the same way

x/programs/rust/examples/token/src/lib.rs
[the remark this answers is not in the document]
richardpringle: TIL

x/programs/rust/README.md
hexfusion: Check out the state macro I added it explicitly. lmk what you think

https://github.com/ava-labs/hypersdk/pull/499/files#diff-3ca848a15e3cf3a85a35434debabf380e80ce7a22ef64446e861cb2b93c9d1c1R96

x/programs/rust/wasmlanche_sdk/src/memory.rs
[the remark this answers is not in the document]
richardpringle: @hexfusion, I meant from a high level. Addresses maybe?

x/programs/rust/wasmlanche_sdk/src/state.rs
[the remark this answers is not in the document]
richardpringle: >That would depend on the implementation of the underlying storage. In the case of merkledb if execution fails then no state is actually mutated.

There might be a better place to have this conversation, but does merkledb commit on every contract execution? I would have expected many (?) actions per block, some of which could fail while others could succeed. 

But I have no idea how Hyper works, so I may be way off-base.

exdx (dismissed):

x/programs/rust/wasmlanche_sdk/src/memory.rs
```diff
@@ -0,0 +1,92 @@
...
+        Self { ptr }
+    }
+    /// Attempts return owned bytes from a pointer created by the global allocator.
+    ///
+    /// # Safety
+    /// `ptr` must be a pointer to a block of memory created using alloc.
+    /// `length` must be the length of the block of memory.
+    #[must_use]
+    pub unsafe fn range(&self, length: usize) -> Vec<u8> {
+        unsafe { Vec::from_raw_parts(self.ptr.raw, length, length) }
```
hexfusion, line 34: hmm I think we don't want to own these bytes testing ` std::slice::from_raw_parts`

hexfusion: added another variant range_owned with tests

richardpringle:
image
@hexfusion... wait a minute, aren't you supposed to do that first 😝

hexfusion:
> image @hexfusion... wait a minute, aren't you supposed to do that first 😝

Yeah I realized after the fact that I was making some assumptions on in the memory module so added some tests.

x/programs/rust/wasmlanche_sdk/src/memory.rs
```diff
@@ -0,0 +1,175 @@
+//! Temporary storage allocated during the Program runtime.
```
hexfusion, line 1: adding a note on what the standard memory access pattern is.

x/programs/rust/wasmlanche_sdk/src/types.rs
hexfusion: still working through this.

richardpringle (approved):
Looking good! Majority of my comments are more thoughts for the future.

x/programs/rust/wasmlanche_sdk/src/types.rs
richardpringle: Let me know if you need help. We used `Cow` so that you didn't always have to allocate, but I don't think that's relevant anymore.

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -75,6 +78,72 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
 
+/// This macro assists in defining the schema for a program's state.  A user can
+/// simply define an enum with the desired state keys and the macro will
+/// generate the necessary code to convert the enum to a byte vector.
+/// The enum will automatically derive the Copy and Clone traits. As well as the
+/// repr(u8) attribute.
+///
+/// Note: The enum variants with named fields are not supported.
+#[proc_macro_attribute]
+pub fn state(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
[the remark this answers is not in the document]
hexfusion: haha yeah was back and forth

x/programs/rust/README.md
```diff
@@ -1,30 +1,134 @@
...
+message of arbitrary length. Unlike fixed size length objects such as `Address`
+the length can be variable. So the `length` of the memory along with the
+`offset` are both required.
+
+```rust
+#[public]
+pub fn some_function(program: Program, message: i64, message_length: i64) -> i64 {
+    // read message bytes from memory
+    let mem = Memory::new(Pointer::new(message));
+    let msg = unsafe { mem.range(12) };
```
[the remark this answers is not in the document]
hexfusion: Yeah it doesn't even require it now. I just have these available as we experiment so there are some guardrails. I imagine these will be removed or cut back.

x/programs/rust/README.md
```diff
@@ -1,30 +1,134 @@
...
+    // msg = [119 111 114]
+}
+```
+          
+Each `Memory` is tightly bound to the lifecycle of the module. For this reason,
+`Memory` is non durable and can not be relied upon.
```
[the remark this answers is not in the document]
hexfusion: Yeah I will add a note that a memory is bound to the runtime execution so it can not be relied upon outside of that. Which yes to your point could be a single exported function call. But it could be longer in program to program for example.

x/programs/rust/README.md
```diff
@@ -1,30 +1,134 @@
...
+enum StateKeys {
+    Proposal(ProposalId) // 0x0 + [u8; 8]
+}
+
+#[public]
+pub fn vote(program: Program, proposal: i64, weight: i64) -> i64 {
+  // check current votes
+  let votes: i64 = program
+        .state()
+        .get(StateKey::Proposal(proposal.to_be_bytes()).to_vec())
```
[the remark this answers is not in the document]
hexfusion: nice makes sense

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -75,6 +78,72 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
+    let mut item_enum = parse_macro_input!(item as ItemEnum);
+    // add default attributes
+    item_enum.attrs.push(syn::parse_quote! {
+         #[derive(Clone, Copy, Debug)]
```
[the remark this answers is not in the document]
hexfusion: For now fixed length types are supported. 

```
type ProposalId = [u8; 8];

#[state_keys]
enum StateKeys {
    Proposal(ProposalId) // 0x0 + [u8; 8]
}
```

x/programs/rust/examples/pokemon/src/lib.rs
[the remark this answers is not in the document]
hexfusion: I will return it @samliok I promise :)

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -1,21 +1,48 @@
...
+    {
+        panic!("counter already initialized for address")
+    }
+
+    program
+        .state()
+        .store(StateKeys::Counter(address).to_vec(), &0_i64)
+        .expect("failed to store counter");
+
+    true
```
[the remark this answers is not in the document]
hexfusion: Yeah I think that can be part of the event PR?

richardpringle (approved):
🚀

x/programs/rust/wasmlanche_sdk/src/types.rs
hexfusion: As we talked about this is as expected with possible change in future followup.

x/programs/rust/README.md
```diff
@@ -1,56 +1,179 @@
...
+memory := runtime.Memory()
+
+bytes := []byte("hello world")
+bytesLen := uint64(len(bytes))
+
+// create a memory allocation 
+offset, _ := memory.Alloc(bytesLen)
+
+// copy bytes into memory allocation
+_ = memory.Write(offset, bytes)
```
patrick-ogrady, line 70: Do we need to ensure that the amount written is equal to `bytesLen`?

x/programs/rust/README.md
```diff
@@ -1,56 +1,179 @@
...
+
+```go
+runtime := // create new runtime and initialize
+memory := runtime.Memory()
+
+bytes := []byte("hello world")
+bytesLen := uint64(len(bytes))
+
+// create a memory allocation 
+offset, _ := memory.Alloc(bytesLen)
```
patrick-ogrady, line 67: Is there some error we should be checking here? (maybe out of scope for readme)

x/programs/rust/README.md
```diff
@@ -1,56 +1,179 @@
...
+pub fn some_function(program: Program, message: i64, message_length: i64) -> i64 {
+    // read message bytes from memory
+    let mem = Memory::new(Pointer::new(message));
+    let msg = unsafe { mem.range(12) };
+    // msg = [119 111 114]
+}
+```
+          
+Each `Memory` is tightly bound to the lifecycle of the module. For this reason,
+`Memory` is non durable and can not be relied upon outside the scope of a `#[public]` function.
```
patrick-ogrady, line 100: nit: `non-durable`

x/programs/rust/README.md
hexfusion: yeah just example usage error removed for brevity.

x/programs/rust/examples/token/Cargo.toml
```diff
@@ -11,3 +11,12 @@ wasmlanche_sdk = { version = "0.1.0", path = "../../wasmlanche_sdk" }
...
 crate-type = ["cdylib"] # set the crate(needed for cargo build to work properly)
+
+# when creating a new project, you can uncomment the following lines to enable optimizations
+# [profile.release]
+# opt-level = 3
+# lto = true
+# codegen-units = 1
+# overflow-checks = true
+# # https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html#unwinding-the-stack-or-aborting-in-response-to-a-panic
+# panic = 'abort'
```
patrick-ogrady, line 22: is this considered an optimization or required?

hexfusion: panic=abort is the default for the `wasm32-unknown-unknown` target which is now our default target. So this is just visibility to the default in case we wanted to changed it. I do have a question of how well the memory is cleaned up in this case by the OS.

hexfusion: The rust book describes this as well as says "dont panic" :)
```
Ultimately, panics translate into aborts in wasm32-unknown-unknown anyways, so this gives you the same behavior but without the code bloat.
```
ref. https://rustwasm.github.io/docs/book/reference/code-size.html?highlight=panic#avoid-panicking

patrick-ogrady (approved):

PR #500, opened 2023-09-26
[x/programs] Update token contract

exdx:
Improves the overall utility of the existing token contract.

- token: Add comments to program functions
- token: Use assert macro
- token: Use checked math to prevent overflows
- programs: add declarative macro utility fns

### DIFF
--- a/x/programs/rust/examples/token/src/lib.rs
+++ b/x/programs/rust/examples/token/src/lib.rs
@@ -1,12 +1,12 @@
 use wasmlanche_sdk::{public, store::State, types::Address};
 
-/// @title A basic ERC-20 contract 
-/// @author Ava Labs 
+/// @title A basic ERC-20 compatible contract
+/// @author Ava Labs
 /// @notice Serves as a fungible token
-/// @dev Limits the token functionality to the bare minimum
+/// @dev Limits the token functionality to the bare minimum required
 
 /// @notice Initializes the program with a name, symbol, and total supply
-/// @param state The initial program state 
+/// @param state The initial program state
 /// @return Whether the initialization succeeded
 #[public]
 pub fn init(state: State) -> bool {
@@ -54,9 +54,14 @@ pub fn transfer(state: State, sender: Address, recipient: Address, amount: i64) 
     assert!(amount >= 0 && sender_balance >= amount);
 
     let recipient_balance: i64 = state.get_map_value("balances", &recipient).unwrap_or(0);
+
+    let sender_delta = sender_balance.checked_sub(amount);
+    let recipient_delta = recipient_balance.checked_add(amount);
+    assert!(sender_delta.is_some() && recipient_delta.is_some());
+
     state
-        .store_map_value("balances", &sender, &(sender_balance - amount))
-        .store_map_value("balances", &recipient, &(recipient_balance + amount))
+        .store_map_value("balances", &sender, &sender_delta.unwrap())
+        .store_map_value("balances", &recipient, &recipient_delta.unwrap())
         .is_ok()
 }
 
--- a/x/programs/rust/examples/token/src/lib.rs
+++ b/x/programs/rust/examples/token/src/lib.rs
@@ -47,14 +47,12 @@ pub fn mint_to(state: State, recipient: Address, amount: i64) -> bool {
 #[public]
 pub fn transfer(state: State, sender: Address, recipient: Address, amount: i64) -> bool {
     // require sender != recipient
-    if sender == recipient {
-        return false;
-    }
+    assert_ne!(sender, recipient);
+
     // ensure the sender has adequate balance
     let sender_balance: i64 = state.get_map_value("balances", &sender).unwrap_or(0);
-    if amount < 0 || sender_balance < amount {
-        return false;
-    }
+    assert!(amount >= 0 && sender_balance >= amount);
+
     let recipient_balance: i64 = state.get_map_value("balances", &recipient).unwrap_or(0);
     state
         .store_map_value("balances", &sender, &(sender_balance - amount))
@@ -64,9 +62,9 @@ pub fn transfer(state: State, sender: Address, recipient: Address, amount: i64) 
 
 /// @notice Gets the balance of the recipient
 /// @param state The current program state
-/// @param recipient The address whose balance will be returned
+/// @param address The address whose balance will be returned
 /// @return The balance of the recipient
 #[public]
-pub fn get_balance(state: State, recipient: Address) -> i64 {
-    state.get_map_value("balances", &recipient).unwrap_or(0)
+pub fn get_balance(state: State, address: Address) -> i64 {
+    state.get_map_value("balances", &address).unwrap_or(0)
 }
--- a/x/programs/rust/examples/token/src/lib.rs
+++ b/x/programs/rust/examples/token/src/lib.rs
@@ -74,10 +74,14 @@ pub fn transfer(state: State, sender: Address, recipient: Address, amount: i64) 
         return false;
     }
 
-    state
-        .store_map_value("balances", &sender, &sender_delta.unwrap())
-        .store_map_value("balances", &recipient, &recipient_delta.unwrap())
-        .is_ok()
+    if let (Some(sender_delta), Some(recipient_delta)) = (sender_delta, recipient_delta) {
+        state
+            .store_map_value("balances", &sender, &sender_delta)
+            .store_map_value("balances", &recipient, &recipient_delta)
+            .is_ok()
+    } else {
+        false
+    }
 }
 
 /// @notice Gets the balance of the recipient
--- a/x/programs/rust/examples/token/src/lib.rs
+++ b/x/programs/rust/examples/token/src/lib.rs
@@ -1,3 +1,4 @@
+use sdk_dec_macros::{assert_eq, assert_gt, require};
 use wasmlanche_sdk::{public, store::State, types::Address};
 
 /// @title A basic ERC-20 compatible contract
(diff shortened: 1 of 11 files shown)


x/programs/rust/examples/token/src/lib.rs
```diff
@@ -25,27 +38,38 @@ pub fn mint_to(state: State, recipient: Address, amount: i64) -> bool {
...
-        return false;
-    }
+    assert_ne!(sender, recipient);
+
     // ensure the sender has adequate balance
     let sender_balance: i64 = state.get_map_value("balances", &sender).unwrap_or(0);
-    if amount < 0 || sender_balance < amount {
-        return false;
-    }
+    assert!(amount >= 0 && sender_balance >= amount);
```
hexfusion, line 54: before this change invalid balance returned false now it panics? just making sure I understand the intent here.

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -25,27 +38,38 @@ pub fn mint_to(state: State, recipient: Address, amount: i64) -> bool {
...
+
     let recipient_balance: i64 = state.get_map_value("balances", &recipient).unwrap_or(0);
+
+    let sender_delta = sender_balance.checked_sub(amount);
+    let recipient_delta = recipient_balance.checked_add(amount);
+    assert!(sender_delta.is_some() && recipient_delta.is_some());
```
hexfusion, line 60: same here I think we want to consider better error handling vs panic right?

x/programs/rust/examples/token/src/lib.rs
hexfusion: can you compile and test this case I am curious how the engine handles this.

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -25,27 +38,38 @@ pub fn mint_to(state: State, recipient: Address, amount: i64) -> bool {
...
+/// @param recipient The address receiving tokens
+/// @param amount The amount of tokens to transfer
+/// @return Whether the transfer succeeded
 #[public]
 pub fn transfer(state: State, sender: Address, recipient: Address, amount: i64) -> bool {
     // require sender != recipient
-    if sender == recipient {
-        return false;
-    }
+    assert_ne!(sender, recipient);
```
patrick-ogrady, line 50: ❤️

x/programs/rust/examples/token/src/lib.rs
patrick-ogrady: This is closer to how the EVM handles these sorts of things fwiw (reverting out as soon as an assert is hit with no return).

hexfusion: OK gothca, looking around I see Soroban does something similar WRT to `panic!`. But also adds context to the panic condition. Perhaps this could be a middle ground? Ideally we have something to return to the caller?

https://github.com/stellar/soroban-examples/blob/7a7cc6268ada55113ce0b82a3ae4405f7ec8b8f0/token/src/balance.rs#L29-L34

hexfusion: Also important is that they explicitly define panic = abort to release which appears to have some interesting properties.

```
[profile.release]
panic = 'abort'
```

[1] https://github.com/stellar/soroban-examples/blob/7a7cc6268ada55113ce0b82a3ae4405f7ec8b8f0/token/Cargo.toml#L23
[2] https://doc.rust-lang.org/book/ch09-01-unrecoverable-errors-with-panic.html#unwinding-the-stack-or-aborting-in-response-to-a-panic

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,6 +1,13 @@
...
 
-/// Initializes the program with a name, symbol, and total supply.
+/// @title A basic ERC-20 compatible contract
+/// @author Ava Labs
+/// @notice Serves as a fungible token
+/// @dev Limits the token functionality to the bare minimum required
+
+/// @notice Initializes the program with a name, symbol, and total supply
+/// @param state The initial program state
+/// @return Whether the initialization succeeded
```
richardpringle, line 10: This does not look good when generating docs
image

x/programs/rust/examples/token/src/lib.rs
richardpringle: Should probably add a custom `panic` handler:
https://doc.rust-lang.org/nomicon/panic-handler.html

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -10,13 +17,19 @@ pub fn init(state: State) -> bool {
         .is_ok()
 }
 
-/// Gets total supply or -1 on error.
+/// @notice Gets total token supply or -1 on error
+/// @param state The current program state
```
richardpringle, line 21: I don't think it's useful to continuously repeat the exact same definition of `State` when the struct itself should be clearly documented. The `public` macro should also be documented and linked. 

We don't need `@param` and `@returns` like you would in JavaScript.

x/programs/rust/examples/token/src/lib.rs
richardpringle: This isn't idiomatic. Why waste computation if `sender_delta` is `None`? Prefer using `expect` if you want to panic.

richardpringle: Alernatively, you could do something like this:
```rust
let (Some(sender_delta), Some(recipient_delta)) = (sender_delta, recipient_delta) else {
    panic!("transfer failed");
};
```
but it's still a wasteful computation if the first fails.

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -25,27 +38,38 @@ pub fn mint_to(state: State, recipient: Address, amount: i64) -> bool {
...
+
+    let sender_delta = sender_balance.checked_sub(amount);
+    let recipient_delta = recipient_balance.checked_add(amount);
+    assert!(sender_delta.is_some() && recipient_delta.is_some());
+
     state
-        .store_map_value("balances", &sender, &(sender_balance - amount))
-        .store_map_value("balances", &recipient, &(recipient_balance + amount))
+        .store_map_value("balances", &sender, &sender_delta.unwrap())
+        .store_map_value("balances", &recipient, &recipient_delta.unwrap())
```
richardpringle, line 64: Using `Option::unwrap` is generally considered a bad practice.

exdx:
I updated to use custom dec macros. The nice thing is with custom macros we have full control over our design. And they compose easily. 

Re program comments, the Solidity based approach was the first option I chose since I'm familiar with it. I think we can certainly iterate and come up with a custom comment design over time.

x/programs/rust/examples/token/src/lib.rs
exdx: It's true but we assert that they are both `Some(T)` above. I don't see an elegant way to get to the inner value otherwise.

x/programs/rust/examples/token/src/lib.rs
exdx: Good point

x/programs/rust/examples/token/src/lib.rs
exdx: I will come up with a comment spec, this is just a start

x/programs/rust/examples/token/src/lib.rs
exdx: Ok nvm found a way

x/programs/rust/examples/token/src/lib.rs
richardpringle: You don't really need a spec. If all your types are well documented, the built-in doc links pretty much handle everything you would want. The function signature itself partially serves as documentation.

x/programs/rust/sdk_dec_macros/src/macros.rs
richardpringle, line 1: I don't understand the point of these and why they aren't just function calls. You can annotate functions with `#[inline(always)]` if you want the code to be inlined. 

But `PartialOrd` and `PartialEq` already exist, why re-invent the wheel? 

`assert` also implies a panic as it's even in `core::prelude`. I don't think these macros are improving the developer experience.

x/programs/rust/sdk_dec_macros/Cargo.toml
richardpringle, line 1: Declarative macros shouldn't be in their own crate.

x/programs/rust/sdk_dec_macros/src/macros.rs
exdx: So how would you propose replicating the `assert()` and `require()` statements used in Solidity? If we don't want to panic we can't use the built-in macros. The alternative is to provide nothing and have every developer manually check every precondition themselves using if statements like it's written now.

exdx: You're saying just use generic function calls instead? With what type parameters, `?Sized` or something catch-all? I don't think the experience is as nice but it's possible.

richardpringle:
```rust
#[inline(always)]
pub fn require<T: PartialEq<U>, U>(t: T, u: U) -> Result<(), Box<dyn Error>> {
    if t != u {
        Ok(())
    } else {
        Err("bad".into())
    }
}
```

So that it can be called as 
```rust
#[public]
pub fn my_contract_method(ctx: Ctx) -> Result<(), Error> {
    require(ctx.sender, ctx.owner)?; // pseudocode, I realize this stuff doesn't exist yet
    
    // continue executing the contract
}
```


In any case, my argument is that you aren't providing more than what's already built in to `core`. 
```rust
if a != b {
    return false;
}

// vs

if !assert_eq!(a, b) {
    return false;
}
```

The former is definitely going to be more familiar to developers, especially Rust devs (I realize we actually aren't just catering toward Rust devs) who use `assert_eq` in unit tests and expect that macro to panic. While we shouldn't pander to Rust devs, we also shouldn't have an SDK actively goes against Rust idioms.

Issue #475, opened 2023-09-17
[chain] Only store last X blocks in memory


patrick-ogrady:
To reduce memory pressure, we should only store the latest X accepted blocks in memory (ideally much less than the "backup range" we are keeping for recovery).

Issue #553, opened 2023-10-14
[chain] store blocks in flat files instead of using pebble


patrick-ogrady:

patrick-ogrady:
May make sense to not use fixed-size chunks and instead store file metadata/offsets in dedicated pebble db.

PR #558, opened 2023-10-14
Limit Accepted Block Cache + Store BlockID Index On-Disk

patrick-ogrady:
Resolves: #475 

We purposely store blocks separately from block metadata to make block shuffling on-disk less onerous (should be properly fixed with #553).

This should make implementing archival-required features much easier.

### DIFF
--- a/vm/storage.go
+++ b/vm/storage.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ava-labs/avalanchego/cache"
 	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/choices"
 	"github.com/ava-labs/avalanchego/utils/crypto/bls"
 	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
 	"go.uber.org/zap"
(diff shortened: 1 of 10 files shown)


patrick-ogrady:
Memory usage:
profile001

patrick-ogrady:
alloc space:
profile002

patrick-ogrady:
This reduced memory usage by ~50% (compared to what was seen on devnet):
image

vm/storage.go
```diff
@@ -96,16 +113,35 @@ func (vm *VM) shouldComapct(expiryHeight uint64) bool {
...
-	if err := batch.Put(lastAccepted, binary.BigEndian.AppendUint64(nil, blk.Height())); err != nil {
+	bigEndianHeight := binary.BigEndian.AppendUint64(nil, blk.Height())
+	if err := batch.Put(lastAccepted, bigEndianHeight); err != nil {
+		return err
+	}
+	if err := batch.Put(PrefixBlockKey(blk.Height()), blk.Bytes()); err != nil {
 		return err
 	}
-	if err := batch.Put(PrefixBlockHeightKey(blk.Height()), blk.Bytes()); err != nil {
+	if err := batch.Put(PrefixBlockIDHeightKey(blk.ID()), bigEndianHeight); err != nil {
```
patrick-ogrady, line 123: We need to store this on-disk now so that we can service historical `Get` requests. We make this tradeoff so that we can avoid db compaction (from storing blocks all over disk).

vm/storage.go
```diff
@@ -96,16 +113,35 @@ func (vm *VM) shouldComapct(expiryHeight uint64) bool {
...
+		return err
+	}
+	blkID := blk.ID()
+	if err := batch.Put(PrefixBlockHeightIDKey(blk.Height()), blkID[:]); err != nil {
```
patrick-ogrady, line 127: We store a height lookup to avoid pulling entire blocks from disk to service historical queries.

PR #567, opened 2023-10-17
[x/programs] Add simulator feature to RustSDK

hexfusion:
This PR adds the types to describe VM simulation `Plans` in Rust. This is the API that the Simulator will use, I adding this first will allow us to come to consensus on the API and unblock writing tests for program examples. 

The idea behind this feature is to provide a simple API to communicate with the simulator using Rust types that serialize to `JSON` instead of `JSON/YAML` directly. This will allow for Rust developers to write tests for their programs completely in Rust or directly with `YAML/JSON`.

### DIFF
--- a/x/programs/rust/wasmlanche_sdk/Cargo.toml
+++ b/x/programs/rust/wasmlanche_sdk/Cargo.toml
@@ -9,6 +9,11 @@ edition = "2021"
 thiserror = "1.0.46"
 serde = { version = "1.0.185", features = ["derive"] }
 serde_bare = "0.5.0"
+serde_yaml = {version = "0.9.25", optional = true}
+serde_json = { version = "1.0.107", optional = true }
 sdk_macros = { version = "0.1.0", path = "../sdk_macros" }
+
+[features]
+simulator = ["serde_yaml", "serde_json"]
 
 [lib]
--- a/x/programs/rust/wasmlanche_sdk/Cargo.toml
+++ b/x/programs/rust/wasmlanche_sdk/Cargo.toml
@@ -9,11 +9,14 @@ edition = "2021"
 thiserror = "1.0.46"
 serde = { version = "1.0.185", features = ["derive"] }
 serde_bare = "0.5.0"
-serde_yaml = {version = "0.9.25", optional = true}
-serde_json = { version = "1.0.107", optional = true }
 sdk_macros = { version = "0.1.0", path = "../sdk_macros" }
 
+# simulator
+serde_yaml = {version = "0.9.25", optional = true}
+serde_json = { version = "1.0.107", optional = true }
+
 [features]
+default = []
 simulator = ["serde_yaml", "serde_json"]
 
 [lib]
--- a/x/programs/rust/wasmlanche_sdk/src/simulator.rs
+++ b/x/programs/rust/wasmlanche_sdk/src/simulator.rs
@@ -1,0 +1,353 @@
+//! A client and types for the VM simulator.
+
+use std::{
+    error::Error,
+    io::Write,
+    process::{Command, Stdio},
+};
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum Endpoint {
+    /// Perform an operation against the key api.
+    Key,
+    /// Make a read-only call to a program function and return the result.
+    View,
+    /// Create a transaction from a possible state changing program function
+    /// call. A program's function can internally optionally call other
+    /// functions including program to program.
+    Execute,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct Step<'a> {
+    /// A description of the step.
+    description: &'a str,
+    /// The API endpoint to call.
+    endpoint: Endpoint,
+    /// The method to call on the endpoint.
+    method: &'a str,
+    /// The parameters to pass to the method.
+    params: Vec<Param<'a>>,
+    #[serde(default)]
+    require: Option<Require>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum ParamType {
+    U64,
+    String,
+    Key(Key),
+    ID,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum Key {
+    Ed25519,
+    Secp256r1,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct Param<'a> {
+    /// The optional name of the parameter. This is used for readability.
+    name: &'a str,
+    #[serde(rename = "type")]
+    /// The type of the parameter.
+    param_type: ParamType,
+    /// The value of the parameter.
+    value: &'a str,
+}
+
+impl Serialize for ParamType {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            ParamType::U64 => serializer.serialize_str("u64"),
+            ParamType::String => serializer.serialize_str("string"),
+            ParamType::Key(Key::Ed25519) => serializer.serialize_str("ed25519"),
+            ParamType::Key(Key::Secp256r1) => serializer.serialize_str("secp256r1"),
+            ParamType::ID => serializer.serialize_str("id"),
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for ParamType {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        use serde::de::Error;
+        let s = String::deserialize(deserializer)?;
+        match s.as_str() {
+            "u64" => Ok(ParamType::U64),
+            "string" => Ok(ParamType::String),
+            "ed25519" => Ok(ParamType::Key(Key::Ed25519)),
+            "secp256r1" => Ok(ParamType::Key(Key::Secp256r1)),
+            "id" => Ok(ParamType::ID),
+            _ => Err(D::Error::custom(format!("unknown param type: {}", s))),
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct Require {
+    result: ResultCondition,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub enum Operator {
+    #[serde(rename = "==")]
+    NumericEq,
+    #[serde(rename = "!=")]
+    NumericNe,
+    #[serde(rename = ">")]
+    NumericGt,
+    #[serde(rename = "<")]
+    NumericLt,
+    #[serde(rename = ">=")]
+    NumericGe,
+    #[serde(rename = "<=")]
+    NumericLe,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct ResultCondition {
+    operator: Operator,
+    operand: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct Simulation<'a> {
+    name: &'a str,
+    description: &'a str,
+    caller_key: &'a str,
+    steps: Vec<Step<'a>>,
+}
+
+pub struct Client {
+    /// Path to the simulator binary
+    path: String,
+}
+
+impl Client {
+    pub fn new(path: String) -> Self {
+        Self { path }
+    }
+
+    /// Runs a simulation against the simulator and returns the result.
+    pub fn run<T>(&self, simulation: &Simulation) -> Result<T, Box<dyn Error>>
+    where
+        T: serde::de::DeserializeOwned + serde::Serialize,
+    {
+        call_run_stdin(&self.path, simulation)
+    }
+
+    /// Performs a view step against the simulator and returns the result.
+    pub fn view<T>(&self, data: Step, key: &str) -> Result<T, Box<dyn Error>>
+    where
+        T: serde::de::DeserializeOwned + serde::Serialize,
+    {
+        let simulation = &Simulation {
+            name: "view",
+            description: "single view request",
+            caller_key: key,
+            steps: vec![data],
+        };
+
+        call_run_stdin(&self.path, simulation)
+    }
+
+    /// Performs a single execution step against the simulator and returns the result.
+    pub fn execute<T>(&self, data: Step, key: &str) -> Result<T, Box<dyn Error>>
+    where
+        T: serde::de::DeserializeOwned + serde::Serialize,
+    {
+        let simulation = &Simulation {
+            name: "execute",
+            description: "single execution request",
+            caller_key: key,
+            steps: vec![data],
+        };
+
+        call_run_stdin(&self.path, simulation)
+    }
+}
+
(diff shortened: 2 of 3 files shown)


exdx (approved):

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+    fn parse_yaml<'a>(yaml_content: &'a str) -> Result<Simulation<'a>, Box<dyn std::error::Error>> {
+        let sim: Simulation<'a> = serde_yaml::from_str(yaml_content)?;
+        Ok(sim)
+    }
+
+    #[test]
+    fn test_parse_key_yaml() {
+        let yaml_content = "
+name: token program
+description: Deploy and execute token program
```
exdx, line 231: should we add versioning?

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+caller_key: alice_key
```
exdx, line 232: This API name is a little awkward. I believe it's "the key of the person invoking the contract" right? Maybe we can call it `invoking_key` to be more precise?

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+    method: create
+    params:
+      - name: key name
+        type: ed25519
+        value: alice_key
+  - description: create bob key
+    endpoint: key
+    method: create
+    params:
+      - name: key name
```
exdx, line 245: nit: both `name` sections are the same for alice and bob. We probably want to have different names? Also I imagine the name shouldn't have spaces in it?

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+#[serde(rename_all = "lowercase")]
+pub enum Endpoint {
+    /// Perform an operation against the key api.
+    Key,
+    /// Make a read-only call to a program function and return the result.
+    View,
+    /// Create a transaction from a possible state changing program function
+    /// call. A program's function can internally optionally call other
+    /// functions including program to program.
+    Execute,
```
exdx, line 21: If this is solely related to transactions maybe we can call it `Transactions` or `Transact`? It feels a bit vague.

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+    /// A description of the step.
+    description: &'a str,
+    /// The API endpoint to call.
+    endpoint: Endpoint,
+    /// The method to call on the endpoint.
+    method: &'a str,
+    /// The parameters to pass to the method.
+    params: Vec<Param<'a>>,
+    #[serde(default)]
+    require: Option<Require>,
```
exdx, line 35: What is this field for? Maybe adding a docstring here helps?

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+    #[serde(rename = ">=")]
+    NumericGe,
+    #[serde(rename = "<=")]
+    NumericLe,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct ResultCondition {
+    operator: Operator,
+    operand: String,
```
exdx, line 121: Not in love with `operand` ... maybe `value` is an alternative?

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: spaces are valid in yaml. name in params is only for readability I will add a note on that. this will help a lot with multiple params.

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: I don't have strong feelings over than it has a logical flow.

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: But I do want separation for non transaction events

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+    path: String,
+}
+
+impl Client {
+    #[must_use]
+    pub fn new(path: String) -> Self {
+        Self { path }
+    }
+
+    /// Runs a simulation against the simulator and returns the result.
```
patrick-ogrady, line 143: I think we should use something clearer than `running a simulation against the simulator`. We should call the entire "simluation" a `plan` IMO. `Plans` have many `steps` (which you already use).

patrick-ogrady: I also think `simulation` can imply that the results aren't persisted (like a "trial run")

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+    require: Option<Require>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum ParamType {
+    U64,
+    String,
+    Key(Key),
+    ID,
+}
```
richardpringle, line 44:
```suggestion
#[derive(Debug, PartialEq, Serialize, Deserialize)]
#[serde(rename_all = "lowercase")]
pub enum ParamType {
    U64,
    String,
    #[serde(untagged)]
    Key(Key),
    Id,
}
```

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+pub enum Key {
+    Ed25519,
+    Secp256r1,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct Param<'a> {
+    /// The optional name of the parameter. This is used for readability.
+    name: &'a str,
+    #[serde(rename = "type")]
```
richardpringle, line 57: 🚀

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+        match s.as_str() {
+            "u64" => Ok(ParamType::U64),
+            "string" => Ok(ParamType::String),
+            "ed25519" => Ok(ParamType::Key(Key::Ed25519)),
+            "secp256r1" => Ok(ParamType::Key(Key::Secp256r1)),
+            "id" => Ok(ParamType::ID),
+            _ => Err(D::Error::custom(format!("unknown param type: {s}"))),
+        }
+    }
+}
```
richardpringle, line 95: See comment above, you should be able to derive these

x/programs/rust/wasmlanche_sdk/src/simulator.rs
richardpringle, line 35: Do we want the `null` for json serialization? If not, `#[serde(skip_serializing_if = "Option::is_none")]`

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+    NumericNe,
+    #[serde(rename = ">")]
+    NumericGt,
+    #[serde(rename = "<")]
+    NumericLt,
+    #[serde(rename = ">=")]
+    NumericGe,
+    #[serde(rename = "<=")]
+    NumericLe,
+}
```
richardpringle, line 116: Going to expand this in the future?
If not
```suggestion
#[derive(Debug, Serialize, Deserialize, PartialEq)]
pub enum NumericOperator {
    #[serde(rename = "==")]
    Eq,
    #[serde(rename = "!=")]
    Ne,
    #[serde(rename = ">")]
    Gt,
    #[serde(rename = "<")]
    Lt,
    #[serde(rename = ">=")]
    Ge,
    #[serde(rename = "<=")]
    Le,
}
```

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+{
+    let mut child = Command::new(path)
+        .arg("run")
+        .stdin(Stdio::piped())
+        .spawn()?;
+
+    // write json to stdin
+    let input =
+        serde_json::to_string(simulation).map_err(|e| format!("failed to serialize json: {e}"))?;
+    if let Some(ref mut stdin) = child.stdin {
```
richardpringle, line 203: nit
```suggestion
    let input =
        serde_json::to_string(simulation).map_err(|e| format!("failed to serialize json: {e}"))?;

    if let Some(ref mut stdin) = child.stdin {
```

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+            .map_err(|e| format!("failed to write to stdin: {e}"))?;
+    }
+
+    let output = child
+        .wait_with_output()
+        .map_err(|e| format!("failed to wait for command to finish: {e}"))?;
+
+    let resp: T = serde_json::from_str(String::from_utf8(output.stdout)?.as_ref())
+        .map_err(|e| format!("failed to parse output to json: {e}"))?;
+    Ok(resp)
```
richardpringle, line 215: nit:
```suggestion
        .map_err(|e| format!("failed to parse output to json: {e}"))?;
        
    Ok(resp)
```

richardpringle (commented):
Fine as is, but I think my *suggestions* would improve things

x/programs/rust/wasmlanche_sdk/Cargo.toml
```diff
@@ -9,6 +9,11 @@ edition = "2021"
 thiserror = "1.0.46"
 serde = { version = "1.0.185", features = ["derive"] }
 serde_bare = "0.5.0"
+serde_yaml = {version = "0.9.25", optional = true}
+serde_json = { version = "1.0.107", optional = true }
```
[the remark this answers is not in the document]
hexfusion: They are required for anyone using the SDK feature so I **think** they are are required gated by the feature?

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: sweeeeet

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: Yeah I was THINKING of strings in the future but I'm going to use this

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: I think we will leave this as it does logically flow to `execute` a program.

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: I like caller key but I did add more details in the docs.

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: going to leave it for now actually.

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,363 @@
...
+                            operand: "1000".into(),
+                        },
+                    }),
+                },
+            ],
+        };
+
+        assert_eq!(parse_yaml(yaml_content).unwrap(), expected);
+    }
+}
```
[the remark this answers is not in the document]
hexfusion: yeah its explicit now only for the sake of coverage while I test go. added TODO to simplify

x/programs/rust/wasmlanche_sdk/src/simulator.rs
exdx: https://github.com/ava-labs/hypersdk/issues/568

x/programs/rust/wasmlanche_sdk/Cargo.toml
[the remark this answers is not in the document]
hexfusion: broke them out in the cargo file to clarify

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,326 @@
...
+    process::{Command, Stdio},
+};
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum Endpoint {
+    /// Perform an operation against the key api.
+    Key,
```
samliok, line 18: what is the key api?

richardpringle (approved):
🚀

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,326 @@
...
+    Secp256r1,
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct Param<'a> {
+    /// The optional name of the parameter. This is only used for readability.
+    name: &'a str,
+    #[serde(rename = "type")]
+    /// The type of the parameter.
+    param_type: ParamType,
```
samliok, line 64: the parameter type must be specified so json knows how to parse?

patrick-ogrady (approved):

x/programs/rust/wasmlanche_sdk/src/simulator.rs
```diff
@@ -0,0 +1,326 @@
...
+        };
+
+        call_run_stdin(&self.path, plan)
+    }
+
+    /// Performs a single `Execute` step against the simulator and returns the result.
+    /// # Errors
+    ///
+    /// Returns an error if the if serialization or plan fails.
+    pub fn execute<T>(&self, data: Step, key: &str) -> Result<T, Box<dyn Error>>
```
samliok, line 155: what is the point of explicitly noting `execute` vs `read_only`?

x/programs/rust/wasmlanche_sdk/src/simulator.rs
samliok: ah nvm, understood after I looked at the examples 👍

samliok (commented):
Just a few questions

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: the expectation is that one returns a tx the other returns the result of a function.

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: yeah I want to use `type` but its reserved in rust and r#type is even more annoying to type :)

x/programs/rust/wasmlanche_sdk/src/simulator.rs
hexfusion: I think we will want the SDK to be able to better describe this. Some functions are read_only thus can be run against a local nodes as a read only call. But something that explicitly could change state would be part of a transaction. That's my thinking at least.

PR #569, opened 2023-10-17
[x/programs] Improve runtime Config defaults 

hexfusion:
This PR adds.

- improved `Config` default handling
- support for serializing the runtime `ConfigBuilder`
- allow multiple instances of the runtime.Config to be passed to additional runtimes. Currently this would fail as the wasmtime.Config can only be used once per engine. This is important so we can ensure program to program calls use the same runtime configuration as the parent.
- bumps defaultLimitMaxMemory from 16 to 18 pages. We found that 18 pages is required for all of our current examples.
- Add testing only mode which allows for non production debug modes by enabling `Wasi` preview for the upcoming simulator work.
- Removed `maxFee` from `ConfigBuilder` and added to Initialize()
- renamed `TransferUnits` -> `TransferUnitsTo`


exdx (dismissed):
lgtm

x/programs/examples/token_test.go
```diff
@@ -25,7 +25,7 @@ var (
 	log = logging.NewLogger(
 		"",
 		logging.NewWrappedCore(
-			logging.Info,
+			logging.Debug,
```
hexfusion, line 28: nit

x/programs/rust/scripts/build.sh
```diff
@@ -1,15 +1,30 @@
...
-  --target wasm32-unknown-unknown \
-  --target-dir "$CARGO_TARGET_DIR" \
+  --target "${target}" \
+  --target-dir "${target_dir}" \
   --release
+
+# Copy the compiled program into the build directory
+cp ./"${target_dir}"/"${target}"/release/*.wasm "./${target_dir}/"
+
+echo "Wasm files are in ./${target_dir}"
```
hexfusion, line 30: nit

x/programs/runtime/config.go
patrick-ogrady, line 1: Much cleaner. Nice.

x/programs/runtime/runtime.go
```diff
@@ -82,13 +87,30 @@ func (r *WasmRuntime) Initialize(ctx context.Context, programBytes []byte) (err
...
 	// setup metering
 	r.meter = NewMeter(r.store)
 	_, err = r.meter.AddUnits(r.cfg.meterMaxUnits)
 	if err != nil {
 		return err
 	}
 
+	// reset the config max units to 0 this ensures that any subsequent usage of
+	// this config will not inherit the meter units. ex (program -> program)
+	r.cfg.meterMaxUnits = 0
```
patrick-ogrady, line 112: This must've been a fun bug.

patrick-ogrady: Should it instead be set to the remaining units? Or is that handled elsewhere?

x/programs/rust/README.md
```diff
@@ -18,6 +18,17 @@ cargo build --target wasm32-unknown-unknown --target-dir $CARGO_TARGET_DIR --rel
 ./scripts/build.sh
 ```
 
+## Debugging
+
+While developing programs you can optionally compile the program with wasi
+support. This provides for example logging from your program. The compile Wasm
```
patrick-ogrady, line 24: If you leave the logs in and don't use `TESTING_ONLY`, what happens? If you have to remove them, I'd make sure to say that here as a reminder.

x/programs/runtime/runtime.go
hexfusion: Good question, the idea is the caller sets the maxfee for the entire execution. once the parent runtime sets the meter I want this cfg value to be 0. This way if we use the config again for program to program the expectation is the meter should start at 0. We then can handle remaining units with `TransferUnits`. You can find an example of this in the program import example[1].

 https://github.com/ava-labs/hypersdk/blob/9c184578d8d1536964abe34055754e15bde644fe/x/programs/runtime/meter.go#L55
 
 [1] https://github.com/ava-labs/hypersdk/blob/9c184578d8d1536964abe34055754e15bde644fe/x/programs/examples/imports/program/program.go#L170

x/programs/rust/README.md
hexfusion: improved the docs on this.

x/programs/rust/README.md
```diff
@@ -20,10 +20,14 @@ cargo build --target wasm32-unknown-unknown --target-dir $CARGO_TARGET_DIR --rel
...
-While developing programs you can optionally compile the program with wasi
-support. This provides for example logging from your program. The compile Wasm
-will be automatically supported by the VM simulator or if you explicitly set
-WithEnableTestingOnlyMode(true) for the runtime `Config`.
+While developing programs you can optionally compile the program with `Wasi`
+support. This allows, for example, logging from your program. The compiled
+`Wasm` will be automatically supported by the VM simulator, or if you explicitly
+set `WithEnableTestingOnlyMode` for the runtime `Config`. 
+
+**NOTE**: Once testing is complete, remove all print statements from your
```
patrick-ogrady, line 28: If you keep them around, will it error or just skip it?

x/programs/runtime/runtime.go
patrick-ogrady: Ah ok. We should probably document this very explicitly (maybe right there?)

x/programs/rust/README.md
hexfusion: there is no error if you recompile, if you don't it will error.

x/programs/runtime/runtime.go
hexfusion: Made a few other changes here. I think its important that there is only one flow for populating the meter with a single config to reduce the scope of leaky metering.

- new cfg populated by genesis sets `maxFee`
- any other inline runtimes must `TransferUnits` meter <-> meter.

hexfusion: I thought this over and the Unit fee really should not be part of the cfg for the reasons I am trying to escape :). I have been back and forth on this but I think it should be a param to Intialize.

x/programs/rust/README.md
hexfusion: Decided to leave the script and clarify the docs

x/programs/examples/counter.go
```diff
@@ -1,197 +0,0 @@
-// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
-// See the file LICENSE for licensing terms.
-
-package examples
```
hexfusion, line 4: these are really just tests

x/programs/runtime/runtime.go
patrick-ogrady: Ok sounds good.

exdx (dismissed):

x/programs/rust/scripts/build.sh
```diff
@@ -1,15 +1,30 @@
...
+# $ TESTING_ONLY=1 ./scripts/build.sh .
+#
+# If TESTING_ONLY=1 is passed we compile the program with wasi support for
+# easier debugging.
 set -x
 
-# Set a default value for CARGO_TARGET_DIR if it's not provided as an argument
-CARGO_TARGET_DIR="${1:-../examples/testdata/target}"
+target="wasm32-unknown-unknown"
+if [ -n "${TESTING_ONLY:-}" ]; then
```
patrick-ogrady, line 10: Can we just call this `DEBUG`?

x/programs/runtime/meter.go
```diff
@@ -52,8 +52,9 @@ func (m *meter) AddUnits(units uint64) (uint64, error) {
 	return m.GetBalance(), nil
 }
 
-func (m *meter) TransferUnits(to Meter, units uint64) (uint64, error) {
-	// TODO: add rollback support
+// TransferUnitsTo moves units from this meter to another meter.
+func (m *meter) TransferUnitsTo(to Meter, units uint64) (uint64, error) {
+	// TODO: add rollback support?
```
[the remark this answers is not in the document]
hexfusion: true only concern was if the transfer itself failed but not much that really can be done. will drop

Issue #267, opened 2023-07-22
[morpheusvm/tokenvm/indexvm] Change the `auth` on one of these to better showcase how to provide your own crypto


patrick-ogrady:
We should use BLS for this...we'll need to generalize `crypto.PublicKey` and `crypto.Signature`, which are hardcoded to 32 and 64 bytes, respectively. BLS has 48 and 96.

## Potential Changes
* We should consider a "SmallBytes" codec which uses uint8 to encode length, rather than int.
* We may want to send to an address/arbitrary bytes instead of sending to a public key (will be needed for WASM anyways). Addresses/Programs should be "indistinguishable".

0xJohnnyGault:
We are hacking on an `Auth` that uses `secp256r1` (P256) which is what WebAuthn uses, and have a POC where the browser can sign transactions using biometrically protected private keys. The P256 curve basically turns every browser into a overpowered crypto wallet. The WebAuthn plumbing even includes scannable QR codes to sign from your mobile while using a desktop browser. Anyway, it seems like the `ed25519.PublicKey` type is pretty deeply baked throughout the HyperSDK code. It would be very useful to see some sketches of how to best incorporate a generic Auth pattern.

patrick-ogrady:
> We are hacking on an `Auth` that uses `secp256r1` (P256) which is what WebAuthn uses, and have a POC where the browser can sign transactions using biometrically protected private keys. The P256 curve basically turns every browser into a overpowered crypto wallet. The WebAuthn plumbing even includes scannable QR codes to sign from your mobile while using a desktop browser. Anyway, it seems like the `ed25519.PublicKey` type is pretty deeply baked throughout the HyperSDK code. It would be very useful to see some sketches of how to best incorporate a generic Auth pattern.

Exactly why I want to fix this 😓. Thanks for sharing your use case!

I am planning to work on this right after the first public Devnet is launched. It'll be a pretty easy integration (Address will basically become a bytes array instead of a fixed-size ed25519 public key).

patrick-ogrady:
Great blog on this topic: https://gogo-webauthn.fly.dev/blog

Issue #378, opened 2023-08-19
[chain] Use TxIDs as address


patrick-ogrady:
This will unblock the ability to treat programs as accounts/allow arbitrary crypto without complicating address field parsing.

PR #571, opened 2023-10-18
[chain] Address Refactor

patrick-ogrady:
Resolves: #267 
Resolves: #603
Blocks: #378 

The HyperSDK is tightly bound to `crypto.ed25519`. This PR explicitly integrates `crypto.secp256r1` to force us to come up with a more generic abstractions (`codec.Address`). This is required for anyone adding custom cryptography and for upcoming HyperSDK Program integration.

## TODO
- [x] migrate `utils` to `address`
- [x] move address creation functions in `auth` to `address`
- [x] remove balance key printing
- [x] create var for address bytes instead of calling address creator each time (integration + e2e)
- [x] move `NewED25519Address` to `address` and make it work for `secp256r1`
- [x] only turn on DEBUG logs for HyperSDK
- [x] set redirect STDOUT for AvalancheGO processes created outside of initial 5: https://github.com/ava-labs/avalanche-network-runner/pull/660
- [x] fix `tokenvm` compilation
- [x] add generic key support to CLI
- [ ] ~use home directory for CLI default~
- [x] ~Change `ShortBytes()` to `Address()` (explain prefix is used to properly extract value from bech32 encoding)~
- [x] Migrate to just using `ids.ID` for all addresses (makes entirely uniform for spending) -> anything can receive balances/pay fees?
  - [x] auth module calls have an ID that is used to regulate access to which addresses? that seems much too lockstep. We need to think deeply about how to support `Auth` defined as WASM here.
  - [x] could put typeID ahead of address payload to scope to some sort of auth mechanism? (this ensures there is no way for auth to collide)
  - [x] Add `Actor` to Auth interface explicitly
- [x] Add generic crypto support to `README`
  - [x] add `Address` description to README (mention bytes are always 32 byte hash, which prevents someone from knowing publicKey until spend)
- [x] Print key type during selection
- [x] Find replacement for "Payer" (maybe "Sponsor")
- [x] `codec.AddressBytes` -> `codec.Address`
- [x] Update `MorpheusVM` README
- [x] Migrate CLI to use WS


examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -413,7 +416,7 @@ var _ = ginkgo.Describe("[Test]", func() {
 
 		other, err := ed25519.GeneratePrivateKey()
 		gomega.Ω(err).Should(gomega.BeNil())
-		aother := utils.Address(other.PublicKey())
+		aother, err := utils.Address(other.PublicKey().ShortBytes())
```
patrick-ogrady, line 419: use new address functions

PR #585, opened 2023-10-26
[x/programs] Add support for Pointer types as public function params 

hexfusion:
This is going to change please don't review.

This PR adds support for using the `Pointer` types directly in the signature of a public function. This helps improve reading non fixed length bytes from the stack.

```rust
#[public]
pub fn init( program: Program, name_ptr: Pointer, name_length: i64) -> bool {
    let name = unsafe { name_ptr.read(name_length as usize) };
```

This works because of the public macro
https://github.com/ava-labs/hypersdk/blob/f6588c52f6982dfa0412e7679dd7e6d3006b56ee/x/programs/rust/sdk_macros/src/lib.rs#L75

### DIFF
--- a/x/programs/rust/wasmlanche_sdk/src/types.rs
+++ b/x/programs/rust/wasmlanche_sdk/src/types.rs
@@ -132,7 +132,7 @@ impl Argument for Program {
 /// use wasmlanche_sdk::types::Pointer;
 /// #[public]
 /// pub fn init( program: Program, name_ptr: Pointer, name_length: i64) -> bool {
-/// let nft_name = name_ptr.read(name_length as usize);
+///     let name = unsafe { name_ptr.read(name_length as usize) };
 /// ```
 ///
 #[derive(Clone, Copy)]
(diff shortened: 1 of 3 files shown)


x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -123,3 +123,56 @@ impl Argument for Program {
...
 }
+
+/// Represents a pointer to a block of memory allocated by the global allocator.
+/// A `Pointer` can be used directly as a param type on a `public` functions.
+///
+/// # Example
+/// ```
+/// use wasmlanche_sdk::types::Pointer;
+/// #[public]
+/// pub fn init( program: Program, name_ptr: Pointer, name_length: i64) -> bool {
```
patrick-ogrady, line 134: Would this allow us to remove `name_length` then? Do you think `Bytes` or `Memory` would be a better name here?

patrick-ogrady: Cool

PR #599, opened 2023-11-02
[x/programs] Set bulk memory support on debug mode

samliok:
Was receiving this error when trying to run my compiled WASM(with IO) using debug mode.
```
 Error:       Received unexpected error:
                  WebAssembly translation error
     
                    Caused by:
                    Invalid input WebAssembly code at offset 40132: bulk memory support is not enabled
```

Added a check for the `debugMode` flag before we pass `engineConfig` into the `wastime` runtime.


x/programs/runtime/runtime.go
```diff
@@ -53,6 +53,12 @@ func (r *WasmRuntime) Initialize(ctx context.Context, programBytes []byte, maxUn
 		return err
 	}
 
+	// enable bulk memory support only in testing/debug mode
+	// this is because some io operations from WASM require bulk memory support
+	if r.cfg.debugMode {
+		engineConfig.SetWasmBulkMemory(true)
+	}
+
```
hexfusion, line 61: Good call, if we want to auto enable bulk memory we should add this directly to the config flag and document it so it is explicit.

https://github.com/ava-labs/hypersdk/blob/f735d9422ef7db029ec64bb26a2bf9af563dac98/x/programs/runtime/config.go#L206-L209

x/programs/runtime/config.go
```diff
@@ -196,15 +196,21 @@ func (c *ConfigBuilder) WithDefaultCache(enabled bool) *ConfigBuilder {
 }
 
 // WithDebugMode enables debug mode which provides access to
-// useful debugging information. This should not be set for a live
-// system as it has both performance and security considerations.
+// useful debugging information. If [enabled] is true we set
+// EnableBulkMemory to support useful IO operations during debugging.
```
hexfusion, line 200: :heart:

x/programs/cmd/simulator/vm/actions/program_execute.go
```diff
@@ -86,7 +86,6 @@ func (t *ProgramExecute) Execute(
 	// TODO: get cfg from genesis
 	cfg, err := runtime.NewConfigBuilder().
 		WithDebugMode(true).
-		WithBulkMemory(true).
 		// only required for Wasi support exposed by testing only.
```
hexfusion, line 89: nit: my fault but could we remove this dangling doc line as it is no longer useful. :pray:

hexfusion (commented):
super minor nit then LGTM

hexfusion (approved):

Issue #429, opened 2023-08-28
[x/programs] new-type wrapper around `u64`


samliok:
We should be able to pass `u64`s and `u32` across the guest/host boundary.

richardpringle:
I edited the issue description... but I think we can just close this as the signatures have changed a little.

PR #600, opened 2023-11-02
[x/programs] Improve overflow checks

samliok:
Any reason to not support `#public` functions from wasm to return nothing?


x/programs/runtime/runtime.go
```diff
@@ -189,6 +189,8 @@ func (r *WasmRuntime) Call(_ context.Context, name string, params ...uint64) ([]
 	case int64:
 		value := uint64(result.(int64))
 		return []uint64{value}, nil
+	case nil:
+		return []uint64{}, nil
```
samliok, line 193: better to return `nil` or empty `[]uint64` here?

hexfusion: This should return nil as it appears to be a valid resposne. Can you also add a note on the interface :pray: ?

since we are here we can also clean up this switch a bit.

```go
switch v := result.(type) {
case int32, int64:
    value := uint64(v)
    return []uint64{value}, nil
case nil:
    return nil, nil
default:
    return nil, fmt.Errorf("invalid result type: %T", result)
}
```
https://github.com/ava-labs/hypersdk/blob/f735d9422ef7db029ec64bb26a2bf9af563dac98/x/programs/runtime/dependencies.go#L32-L33

hexfusion: Also lets add some docs on the nil case Ideally we could reference the actual spec.

samliok:
I think we should refactor how we handle uint64 & int64. I updated the PR to reduce our risk of overflow and allowing greater functionality with signed integers.

In the codebase we were restricting call params to our wasm programs to only `uint64` types. This approach doesn't make sense because
- `wasm` does not have unsigned integer types and will just convert our params into either `i32` or `i64`
-  this artificially limits us from passing negative values into `wasm`. 
-  potential unchecked overflow errors as `max uint64 > max int64`

Because of WASM's nature we should default to using `int64`.

x/programs/runtime/memory.go
```diff
@@ -111,7 +111,7 @@ func (m *memory) Len() (uint64, error) {
 
 // WriteBytes is a helper function that allocates memory and writes the given
 // bytes to the memory returning the offset.
-func WriteBytes(m Memory, buf []byte) (uint64, error) {
+func WriteBytes(m Memory, buf []byte) (int64, error) {
```
samliok, line 114: this is a bit awkward because a memory location _should_ be a uint64. However, it might makes sense to keep the `int64` here to avoid potential overflow errors. We pass these memory pointers  into `wasm` which will convert to `int64`, so if we somehow return a `ptr` here that's greater than `MaxInt64` we will have undefined behavior.

samliok:
In addition to  #429 we need to make sure our wrapper u64 does not exceed `MaxInt64`

x/programs/runtime/dependencies.go
```diff
@@ -30,7 +30,7 @@ type Runtime interface {
 	// module with the given imports.  Initialize should only be called once.
 	Initialize(context.Context, []byte, uint64) error
 	// Call invokes the an exported guest function with the given parameters.
-	Call(context.Context, string, ...uint64) ([]uint64, error)
+	Call(context.Context, string, ...int64) ([]int64, error)
```
hexfusion, line 33: While I see your point technically, I would like to see a strong usecase to make this change. Wazero[1] made the same design choice and I feel it is the right one for the sake of simplicity. Now one could argue that this could be an interface{} but again the amount of complexity we save by assuming unsigned ints is non trivial.

[1] https://github.com/tetratelabs/wazero/blob/d3b83a596cde165d0fe69bc4992f31801e516be3/api/wasm.go#L386

Question: what would the usecase be for an exported function to take a param that has a negative value? Could this be solved in other ways with a tradeoff being simplicity?

x/programs/runtime/memory.go
hexfusion: I guess I would need to dig into this further but my general understand is that while technically possible for a pointer to exceed that limit. A 64bit system would not utilize even close to that entire range and a Program which has a capped memory limit via the runtime store/engine is a much smaller subnet of that address space.

I guess what I am saying is there is a trade off between extreme edge case safety and usability/complexity. Its a great topic though and we should try to write some tests to exercise this.

x/programs/examples/util.go
```diff
@@ -13,7 +13,11 @@ import (
 	"github.com/ava-labs/hypersdk/x/programs/runtime"
 )
 
-func newKeyPtr(ctx context.Context, key ed25519.PublicKey, runtime runtime.Runtime) (uint64, error) {
+const MaxInt64 = runtime.MaxInt64
+
+var ErrIntegerConversionOverflow = runtime.ErrIntegerConversionOverflow
```
hexfusion, line 18: nit: Use `runtime` directly vs declaring

x/programs/runtime/dependencies.go
samliok: Hmm, why do you say assuming unsigned ints decreases complexity? The wasmtime `call` [function](https://pkg.go.dev/github.com/bytecodealliance/wasmtime-go#Func.Call) converts any `uint` to `int` which can potentially overflow. Therefore we get no benefit from using `uint64`(since anything above `Int64Max` will cause overflow) _AND_ we limit our programs to only unsigned numbers. 

So we could use an interface{} and accept `int32`, `int64` and `uint32` but as you said this would increase complexity. And I don't think `interface{}` is worth it here.

There are plenty of use cases for signed ints. The `rust sdk` is a tool for building programs, so any use case for signed ints in traditional programs are use cases for signed ints in wasm programs. Of course any function that would like to accept a negative number could do something like `reduceBalanceBy(uint balance)` & `increaseBalance(uint balance)` but I feel like this is an artificial workaround that increases complexity on the program dev.

x/programs/runtime/memory.go
samliok: I can write this up as an issue for a deeper dive later, or should I address/look into this now? I find it a bit unsettling that the host can return pointers our wasm programs wont be able to reference.

samliok: linked #605

x/programs/runtime/dependencies.go
hexfusion: Yeah since we really won't be supporting floats I think interface{} is also overkill. While I think that programs although arbitrary by definition will be very much tied to transactions. So in the context of a VM it seems unsigned ints are really a much more native type and IMO easier to deal with.

But I do see that it is used in for example Solidity to represent debts as you outlined above. So I don't want to make the interface overly opinionated. Lets think about this a little bit more but I am leaning towards the change. Thanks for the feedback.

x/programs/runtime/runtime.go
```diff
@@ -239,14 +242,14 @@ func PreCompileWasmBytes(programBytes []byte, cfg *Config) ([]byte, error) {
 }
 
 // mapFunctionParams maps call input to the expected wasm function params.
-func mapFunctionParams(input []uint64, values []*wasmtime.ValType) ([]interface{}, error) {
+func mapFunctionParams(input []int64, values []*wasmtime.ValType) ([]interface{}, error) {
 	params := make([]interface{}, len(values))
 	for i, v := range values {
 		switch v.Kind() {
 		case wasmtime.KindI32:
```
hexfusion, line 249: we should also ensure this conversion is safe within max,min int32 range

x/programs/runtime/dependencies.go
hexfusion: Yeah I think this makes sense for a lot of reasons. In the future we will probably be passing byte slices to int64 types exclusively. But I think this change makes sense for now.

hexfusion (commented):
This makes sense just a few last nits :pray: 

- ensure safe conversion from int64 to int32

x/programs/runtime/runtime.go
samliok: tried to do this in go, but we need a type assertion in both cases
```go
value := int64(v.(int32))
```
or in the int64 case
```go
value := v.(int64)
```
Is there a workaround for this?

hexfusion: actually there isn't, so we can skip this.

x/programs/runtime/runtime.go
```diff
@@ -239,14 +242,18 @@ func PreCompileWasmBytes(programBytes []byte, cfg *Config) ([]byte, error) {
...
+			// ensure this value is within the range of an int32
+			if input[i] > int64(MaxInt64) || input[i] < int64(MinInt64) {
```
hexfusion, line 251: I think we should make this a helper maybe something like.

```
func ensureInt64ToInt32(v int64) bool {
	return v >= math.MinInt32 && v <= math.MaxInt32
}
```

x/programs/runtime/dependencies.go
```diff
@@ -29,8 +29,10 @@ type Runtime interface {
 	// units. The engine will handle the compile strategy and instantiate the
 	// module with the given imports.  Initialize should only be called once.
 	Initialize(context.Context, []byte, uint64) error
-	// Call invokes the an exported guest function with the given parameters.
-	Call(context.Context, string, ...uint64) ([]uint64, error)
+	// Call invokes an exported guest function with the given parameters.
+	// Returns the results of the call or an error if the call failed.
+	// If there are 0 results this value is set to nil.
```
hexfusion, line 34:
```suggestion
	// If the function called does not return a result this value is set to nil.
```

x/programs/examples/util.go
```diff
@@ -13,19 +13,22 @@ import (
...
 
 	// write programID to memory which we will later pass to the program
-	err = runtime.Memory().Write(ptr, key[:])
+	err = rt.Memory().Write(ptr, key[:])
 	if err != nil {
 		return 0, err
 	}
 
-	return ptr, err
+	if ptr > runtime.MaxInt64 {
```
hexfusion, line 28: same here Alloc should ensure this

hexfusion (commented):
very close!

samliok:
@hexfusion should I also check the return value for overflow in this function?
```rust
pub extern "C" fn alloc(len: usize) -> *mut u8 {
```

hexfusion:
> @hexfusion should I also check the return value for overflow in this function?
> 
> ```rust
> pub extern "C" fn alloc(len: usize) -> *mut u8 {
> ```

Well we should ensure that all the conversions are safe for Alloc. I think we still need to check u64 -> i32 here?

https://github.com/ava-labs/hypersdk/blob/8a7191159adde3480c8fd3965f652e739ff4bbb6/x/programs/runtime/memory.go#L77-L82

the export to alloc takes an i32 as param and usize can't be negative so I am not sure if something strange could happen there I suppose you could check against `i32::MAX` but it might not be nessisary.

x/programs/rust/wasmlanche_sdk/src/memory.rs
```diff
@@ -94,6 +94,10 @@ pub extern "C" fn alloc(len: usize) -> *mut u8 {
     let mut buf = Vec::with_capacity(len);
     // take a mutable pointer to the buffer
     let ptr = buf.as_mut_ptr();
+    // ensure memory pointer is fits in an i64
+    // to avoid potential issues 
+    // when passing across wasm boundary
+    assert!(ptr as u64 <= i64::MAX as u64);
```
samliok, line 100: wrote an example to see what happens if we return > `i64::MAX` and wasm will just round to `i64::MAX`. Still an important check IMO since this ensures the ptr's we get back are correct.

x/programs/rust/wasmlanche_sdk/src/memory.rs
```diff
@@ -84,16 +84,20 @@ pub unsafe fn deallocate(ptr: *mut u8, capacity: usize) {
...
 #[no_mangle]
 pub extern "C" fn alloc(len: usize) -> *mut u8 {
     // create a new mutable buffer with capacity `len`
     let mut buf = Vec::with_capacity(len);
     // take a mutable pointer to the buffer
     let ptr = buf.as_mut_ptr();
+    // ensure memory pointer is fits in an i64
+    // to avoid potential issues when passing
+    // across wasm boundary
+    assert!(i64::try_from(ptr as u64).is_ok());
```
richardpringle, line 100: 🚀

hexfusion (approved):

PR #610, opened 2023-11-07
[x/programs] add support for import fn callbacks

hexfusion:
This PR
- Adds support for import function callbacks. 
- Adds `ContextSwitchUnits` to the runtime config which allows for a user define fee in units for context switching and is set with an arbitrary default of 1000 units.

The callback are currently for internal use only.


token.wat
richardpringle, line 1: Did you mean to commit this?

hexfusion: no its WIP, will cleanup on the next push

x/programs/runtime/import.go
```diff
@@ -57,3 +63,84 @@ func (f *Factory) Register(name string) *Factory {
...
+
+// RegisterFn registers a host function exposed to the guest (import).
+func (l *Link) RegisterFn(module, name string, f interface{}) error {
+	val := reflect.ValueOf(f)
+	funcType := val.Type()
+	if funcType.NumOut() == 0 {
+		return fmt.Errorf("%w: host functions must return a value", ErrInvalidFunction)
+	}
+
+	wrapper := func(args []reflect.Value) []reflect.Value {
```
hexfusion, line 99: ftr not going to use reflection this was just for testing.

PR #623, opened 2023-11-20
[x/programs] Improve State-keys & Argument Passing

samliok:
This PR improves argument passing from `rust` programs to `host functions`. It also removes the need to call the `.to_vec()` function on `StateKeys`, and generalizes external function calls by defaulting to borsh serialization. 

Changes
- All argument types are now serialized using borsh when being passed to/from host.
- All arguments passed in/from the host are `SmartPtr`' which are `int64` with the format `[length][ptr]`. `Length` and `Ptr` are 4 bytes each.
   - improves function signatures 
	   - ex. `fn _put(caller_id: i64, key: *const u8, value: *const u8) -> i32;`
	   - instead of `fn _put(caller_id: i64, key_ptr: *const u8, key_len: usize, value_ptr: *const u8, value_len: usize) -> i32;`
    - this removes the need for the `len` host function call. 
    - `get(key)` no longer requires calling the `len` host function call 
 - add an `Into<Key>` implementation for the `#[state_keys]` proc macro
 - update `call_program` function to require `BorshSerialize` instead of `Argument` trait
    - improves usability for `call_program` & external function calls
    - the `params!` declarative macro serializes arguments required for external function calls
 - Update go to use `SmartPtr` and consolidate associated helper functions into `runtime/utils.go`
   

also a small ux improvement from macro change

before 
```rust
    program
        .state()
        .store(StateKey::TotalSupply.to_vec(), &123456789_i64)
        .expect("failed to store total supply");
```
now
```rust
    program
        .state()
        .store(StateKey::TotalSupply, &123456789_i64)
        .expect("failed to store total supply");
```


x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -47,20 +47,20 @@ fn inc_external(
     of: Address,
     amount: i64,
 ) -> i64 {
-    program.call_program(&target, max_units, "inc", &[Box::new(of), Box::new(amount)])
+    program.call_program(&target, max_units, "inc", serialize_params!(of, amount))
```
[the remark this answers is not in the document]
samliok: imo I think serializing _every_ parameter passed to/from WASM would add nice simplicity to the code.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -47,20 +47,20 @@ fn inc_external(
...
-        .get(StateKeys::Counter(of).to_vec())
+        .get(StateKeys::Counter(of))
         .expect("failed to get counter")
 }
 
 /// Gets the count at the address for an external program.
 #[public]
 fn get_value_external(program: Program, target: Program, max_units: i64, of: Address) -> i64 {
-    program.call_program(&target, max_units, "get_value", &[Box::new(of)])
+    program.call_program(&target, max_units, "get_value", serialize_params!(of))
```
samliok, line 65: this works as intended because no `primitve` types

x/programs/rust/wasmlanche_sdk/src/program.rs
```diff
@@ -56,24 +59,23 @@ impl From<i64> for Program {
     }
 }
 
-/// Marshals arguments into byte slice which can be unpacked by the host.
-fn marshal_args(args: &[Box<dyn Argument>]) -> Vec<u8> {
-    use std::mem::size_of;
-    // Size of meta data for each argument
-    const META_SIZE: usize = size_of::<i64>() + 1;
+#[macro_export]
+macro_rules! serialize_params {
```
samliok, line 63: not sure how else to serialize an unspecified # of arguments with different types in rust. Can `macro_rules` be used here @richardpringle ?

x/programs/rust/examples/counter/src/lib.rs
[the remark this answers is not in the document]
richardpringle: Agreed. I think we talked about this offline, but just want to capture here for posterity. I actually think your declarative macro is nice. I would rename it `params` because it's both more terse, and I don't think the declarative macro should perform the serialization. 

Another thing we could do is pass in a tuple and generate a bunch of trait implementations for tuples of various sizes, that might even be better for DX (program-developer experience). Libraries like `Warp` do this [here](https://docs.rs/warp/latest/src/warp/generic.rs.html). It's a little crazy, but definitely doable. 

As for the actual serialization, it should be possible to implement a trait such that the primitives serialize to themselves and everything else serializes to a byte-slice? 

Anyway, not adding anything new here, I think you already have a good idea how to implement this.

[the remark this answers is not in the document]
samliok: To recap what we talked about offline. 
- We are going to stay with the `params` declarative macro for now.
- All types are now being serialized. 
- In the future, to reduce the amount of serialization/deserialization, we will serialize all params once inside of a struct. Then we will pass the struct to `WASM` and deserialize it inside the `#[public]` proc macro

x/programs/rust/wasmlanche_sdk/src/state.rs
```diff
@@ -110,3 +88,42 @@ pub unsafe fn bytes_and_length(ptr: i64) -> (Vec<u8>, usize) {
...
+        .to_vec();
+    len_bytes.extend(bytes);
+    len_bytes
+}
+
+/// Key is a wrapper around a Vec<u8> that represents a key in the host storage.
+#[derive(Debug, Default, Clone)]
+pub struct Key(Vec<u8>);
+
+impl Key {
```
richardpringle, line 109: Just implement `Deref<Target = [u8]>` instead. `Deref` guarantees you can't mutate the Vector.

x/programs/rust/wasmlanche_sdk/src/errors.rs
```diff
@@ -22,4 +22,7 @@ pub enum StateError {
 
     #[error("failed to serialize bytes")]
     Serialization,
+
+    #[error("failed to deserialize bytes")]
+    Deserialization,
```
richardpringle, line 27: 🥇

richardpringle (changes requested):
Newline and Deref trait are all that need to change right now.

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -55,15 +56,13 @@ func (i *Import) Register(link runtime.Link, meter runtime.Meter, _ runtime.Supp
...
-	if err := link.FuncWrap(Name, "len", i.getLenFn); err != nil {
-		return err
-	}
 
 	return nil
 }
 
-func (i *Import) putFn(caller *wasmtime.Caller, idPtr int64, keyPtr int32, keyLength int32, valuePtr int32, valueLength int32) int32 {
-	memory := runtime.NewMemory(runtime.NewExportClient(caller))
+func (i *Import) putFn(caller *wasmtime.Caller, idPtr int32, keyPtr int32, valuePtr int32) int32 {
```
hexfusion, line 63: i think now would be a good time to just move to int64 args which support [len][ptr]. We are creating a fair amount of rework doing this is phases.

I would like to see this signature.
```go
func (i *Import) putFn(caller *wasmtime.Caller, id int64, key int64, value int64) int64 {
   memory := runtime.NewMemory(client)
   idBytes, err := memory.RangeFromInt64(id)
   ...
```

Then extend memory with a method like `memory.RangeFromInt64()`[1] or a helper function.

This ensures we have the full int32 range for the ptr

[1] https://github.com/ava-labs/hypersdk/pull/631/files#diff-46bbbce39f0b86c15aea460cf21fa799bd6fabad826e7ea351c7577e13997ac9R48-R50

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -47,20 +47,20 @@ fn inc_external(
...
+    program.call_program(&target, max_units, "inc", params!(of, amount))
 }
 
 /// Gets the count at the address.
 #[public]
 fn get_value(program: Program, of: Address) -> i64 {
     program
         .state()
-        .get(StateKeys::Counter(of).to_vec())
+        .get(StateKeys::Counter(of))
```
hexfusion, line 58: I just love this ;)

hexfusion (commented):
Looks great I noted I think its time to move to int64 params I think this should be done before this merges because it will save another round of rework and is already supported in my PR.

Perf 

### Before
````goos: linux
goarch: amd64
BenchmarkTokenProgram/benchmark_token_program_compile_and_cache-12         	     528	   2860427 ns/op	   96511 B/op	    3430 allocs/op
BenchmarkTokenProgram/benchmark_token_program_compile_and_cache_short-12   	     608	   1985926 ns/op	   48759 B/op	     413 allocs/op
BenchmarkTokenProgram/benchmark_token_program_precompile-12                	     630	   1979494 ns/op	  301309 B/op	    3452 allocs/op
BenchmarkTokenProgram/benchmark_token_program_precompile_short-12          	    1124	   1084352 ns/op	  253480 B/op	     411 allocs/op
```

### After
```
goos: linux
goarch: amd64
BenchmarkTokenProgram/benchmark_token_program_compile_and_cache-12         	     488	   2685640 ns/op	   90604 B/op	    3422 allocs/op
BenchmarkTokenProgram/benchmark_token_program_compile_and_cache_short-12   	     703	   1822811 ns/op	   49032 B/op	     515 allocs/op
BenchmarkTokenProgram/benchmark_token_program_precompile-12                	     771	   1618551 ns/op	  279075 B/op	    3416 allocs/op
BenchmarkTokenProgram/benchmark_token_program_precompile_short-12          	    1413	    865186 ns/op	  237559 B/op	     502 allocs/op
```

x/programs/examples/imports/pstate/pstate.go
hexfusion: As I noted I am working on adding direct support so it can look like this.

```
func (i *Import) putFn(caller *program.Caller, keyBytes []byte, valueBytes []byte) int64 {
   id, err :=caller.ID() 
   ...

File: x/programs/rust/wasmlanche_sdk/src/program.rs [hunk 8]
```diff
@@ -32,48 +41,37 @@ impl Program {
...
+                params.push(wasmlanche_sdk::program::serialize_params(&$param).unwrap());
+            )*
 
-impl From<i64> for Program {
-    fn from(value: i64) -> Self {
-        Self { id: value }
-    }
+            params
+        }
+    };
```
richardpringle, line 65: This should be simplified (and publicly documented)
```rust
/// serialize every parameter into a byte vector and return a vector of byte vectors
#[macro_export]
macro_rules! params {
    ($($param:expr),*) => {
        vec![$(wasmlanche_sdk::program::serialize_params($param).unwrap(),)*]
    };
}
```
Also, notice I removed the `&`. It's repetitive for the caller to have to add the character, but then it makes `params!` have something more akin to function calling semantics, where it's clear that you can still use whatever was passed to the macro.

x/programs/examples/imports/program/program.go
```diff
@@ -170,31 +162,34 @@ func (i *Import) callProgramFn(
...
+		lenBytes := buffer[i : i+consts.Uint32Len]
+		length := binary.BigEndian.Uint32(lenBytes) + consts.Uint32Len
+
+		// we include the length in the value bytes because the program
+		// still needs to know how many bytes to read
+		valueBytes := buffer[i : i+int(length)]
+		i += int(length)
+
+		// every argument is a pointer
+		ptr, err := runtime.WriteBytes(memory, valueBytes)
```
patrick-ogrady, line 186: Would it be more efficient to write once and then set offsets into the pointer?

patrick-ogrady: (or is that not really a good idea in this context)

x/programs/examples/imports/util.go
patrick-ogrady, line 1: usually this would be `utils.go`

x/programs/examples/testdata/counter.wasm
patrick-ogrady, line 1: Would it make sense to compile these on-the-fly rather than committing executables to git?

x/programs/examples/counter_test.go
```diff
@@ -105,14 +107,17 @@ func TestCounterProgram(t *testing.T) {
 	require.Equal(int64(1), result[0])
 
 	// increment alice's counter on program 2 by 10
-	result, err = rt2.Call(ctx, "inc", programID2Ptr, alicePtr2, 10)
+	incAmount := int64(10)
+	incAmountPtr, err := newParameterPtr(ctx, incAmount, rt2)
```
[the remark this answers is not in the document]
samliok: no, even native types are passed in as a ptr now.

x/programs/examples/imports/pstate/pstate.go
samliok: actually, what you mentioned will still be necessary because we are still passing in `ptrs`. I'll update the pr

x/programs/examples/imports/program/program.go
samliok: yea that could work, i think what you are thinking of is similar to this issue I just created #638

x/programs/examples/imports/program/program.go
```diff
@@ -60,8 +62,8 @@ func (i *Import) Register(link runtime.Link, meter runtime.Meter, imports runtim
 // callProgramFn makes a call to an entry function of a program in the context of another program's ID.
 func (i *Import) callProgramFn(
 	caller *wasmtime.Caller,
-	callerIDPtr int64,
-	programIDPtr int64,
+	callerIDPtr int32,
```
samliok, line 65: change to i64

x/programs/examples/testdata/counter.wasm
samliok: as in compile them using github actions? It only makes sense to recompile them if the rust code changes. Even then we need the updated binaries for unit tests.

x/programs/rust/wasmlanche_sdk/src/memory.rs
```diff
@@ -68,6 +71,73 @@ impl Memory {
     }
 }
 
+/// `SmartPtr` is an i64 where the first 4 bytes represent the length of the bytes
+/// and the following 4 bytes represent a pointer to WASM memeory where the bytes are stored.
+pub type SmartPtr = i64;
```
richardpringle, line 76: Do you want to create an issue to rename this? Smart-pointers are already thing in Rust [^1]. An `Arc` is an example of a smart-pointer. While there are tons of terms that are overloaded, I wouldn't want a new program dev to land in this code for whatever reason and get confused. 

[^1]: https://doc.rust-lang.org/book/ch15-00-smart-pointers.html

x/programs/rust/wasmlanche_sdk/src/memory.rs
```diff
@@ -68,6 +71,73 @@ impl Memory {
...
+/// Panics if the bytes cannot be deserialized.
+/// # Safety
+/// This function is unsafe because it dereferences raw pointers.
+/// # Errors
+/// Returns an `StateError` if the bytes cannot be deserialized.
+pub unsafe fn from_smart_ptr<V>(ptr: SmartPtr) -> Result<V, StateError>
+where
+    V: BorshDeserialize,
+{
+    let (bytes, _) = bytes_and_length(ptr);
```
richardpringle, line 125: 🤔, what else is this `bytes_and_length` function used for? It looks to me like the function is actually safe. 

Not a change suggestion for this PR, I just noticed this.

x/programs/runtime/utils.go
```diff
@@ -0,0 +1,104 @@
...
+// ToSmartPtr converts [ptr] and [len] to a SmartPtr.
+func ToSmartPtr(ptr uint32, len uint32) (SmartPtr, error) {
+	// ensure length of bytes is not greater than int32 to prevent overflow
+	if !EnsureUint32ToInt32(len) {
+		return 0, fmt.Errorf("length of bytes is greater than int32")
+	}
+
+	// convert to int64 with length of bytes in the upper 32 bits and ptr in the lower 32 bits
+	smartPtr := int64(len) << 32
+	smartPtr |= int64(ptr)
```
hexfusion, line 52: optional. I think we could break this up a little to make it easier to understand the bitwise op.
```go
lenUpperBits := int64(len) << 32
ptrLowerBits := int64(ptr)

smartPtr := lenUpperBits | ptrLowerBits
```

richardpringle (dismissed):

x/programs/runtime/utils.go
```diff
@@ -0,0 +1,104 @@
...
+	return v >= math.MinInt32 && v <= math.MaxInt32
+}
+
+func EnsureUint32ToInt32(v uint32) bool {
+	return v <= math.MaxInt32
+}
+
+// SmartPtr is an int64 where the first 4 bytes represent the length of the bytes
+// and the following 4 bytes represent a pointer to WASM memeory where the bytes are stored.
+type SmartPtr int64
```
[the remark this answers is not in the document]
samliok: created these helper functions, not sure what you meant by `Verify()` tho

x/programs/rust/wasmlanche_sdk/src/memory.rs
samliok: yea, overloading`SmartPtr` isn't the best but every other name felt not as good. Maybe `argPtr` or `memoryPtr`? will create an issue.

x/programs/runtime/memory.go
```diff
@@ -164,3 +168,46 @@ func WriteParams(m Memory, p []CallParam) ([]int64, error) {
...
+	// read the range of PtrOffset + length from memory
+	bytes, err := memory.Range(uint64(s.PtrOffset()), uint64(s.Len()))
+	if err != nil {
+		return nil, err
+	}
+
+	return bytes, nil
+}
+
+func BytesToSmartPtr(bytes []byte, memory Memory) (SmartPtr, error) {
```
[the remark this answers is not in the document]
hexfusion: stale comments

x/programs/runtime/memory.go
```diff
@@ -164,3 +168,46 @@ func WriteParams(m Memory, p []CallParam) ([]int64, error) {
 
 	return params, nil
 }
+
+func (s SmartPtr) Get() int64 {
```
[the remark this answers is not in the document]
hexfusion: stale comments

hexfusion (approved):

x/programs/rust/wasmlanche_sdk/src/memory.rs
richardpringle: What about... wait for it... `Ptr` or `HostPtr`? I also like `ArgPtr` and `MemPtr`. The only reason I don't like `SmartPtr` is because it immediately makes me think that the type is something it isn't, whereas all the other names are equally nebulous and I therefore have to look at the implementation. 

🤷 , I saw you created the issue. We can deal with this later

richardpringle (approved):
🚀

Issue #576, opened 2023-10-22
[x/programs/rust] Add support for non fixed length program function args. 


hexfusion:
We should be able to simplify the signature of a HyperSDK program that accepts arbitrary length strings. Perhaps add functionality to `Argument` or make a new type that takes the 2 params  `ptr` and  `len` and handles the memory access in the background. Today this is possible but requires 2 params on the signature for each. 

Another alternative might be using the first byte for length prefix.

### today
```rust
pub fn init(program: Program, name_ptr: i32, name_len:i32) -> bool {
```
### desired
```rust
pub fn init(program: Program, name: Argument) -> bool {
    program
        .state()
        .store(StateKey::Name, name.to_bytes())
        .expect("failed to store coin name");
}
```

exdx:
I agree. Passing in multiple params gets pretty unwieldy. More than 8 params is a clippy lint violation I believe.

hexfusion:
Ideally we are dealing with objects via Borsh I suspect.

richardpringle:
In your example, `Name` should be a string and the `public` macro should provide all of the logic to fetch the pointer or deserialize, or whatever needs to happen. We should not restrict the types that can be used in the function signature beyond requiring a specific trait implementation (marker-trait or otherwise).

Thoughts?

richardpringle:
> It would be non trivial to update the macro to support `String` of arbitrary length. Because from the go side we would need to pass in two arguments anyway(ptr + length) and then our macro would need to: squash the function signature & complete the conversion. Would it make sense to have a variable say `String32` that is a null terminated string with a fixed length like `32` or `64`? This is kinda what I already do with the `Bytes32` type
> 
> Also: the macro calls `.into()` to convert from an`int` to the user specified type so we would need extra handling for two arguments coming from go. Maybe we can serialize the length as a prefix of the string and deserialize in our macro?

Makes sense. Seems like serialization should solve this? Looks like borsh encodes strings lengths, so that should work?

samliok:
> Makes sense. Seems like serialization should solve this? Looks like borsh encodes strings lengths, so that should work?

I think a good solution would be to update the public macro so that it automatically deserializes into the function arguments. 

So at a high level a user would specify 

```rust
pub fn do_something(program: Program, name: String, object: CustomUserObject) {
 ```

and the `public` macro would expose this signature to the host.
```rust
[...]

PR #624, opened 2023-11-21
[x/Programs] Dynamic Arguments

samliok:
## Dynamic Arguments

This PR add support for non fixed length program function args. #576 . 

Changes:
- Arguments passed into #[public] functions must be prefixed with a uint32 length.
  - This does not include supported primitives and the `Program` type
  - after the length prefix, parameters for `#[public]` functions are serialized using `borsh`.
- The `public` proc_macro was updated to call `from_raw_pointer` on ptrs.
   - This function unpacks the 4 byte length prefix, and converts the subsequent bytes into the specified argument.
 

example usage

```rust 
#[public]
pub fn print_name(_: Program, name: String) {
    println!("name: {}", name);
}
```
```rust
/// Mints tokens to multiple recipients.
#[public]
pub fn mint_to_many(program: Program, recipients: Vec<Address>, amounts: Vec<i32>) -> bool {
    assert_eq!(recipients.len(), amounts.len(), "invalid input");

    for (recipient, amount) in recipients.iter().zip(amounts.iter()) {
        mint_to(program, *recipient, *amount as i64);
    }

    true
}
```


x/programs/examples/token.go
```diff
@@ -179,10 +223,23 @@ func (t *Token) Run(ctx context.Context) error {
...
 	t.log.Debug("balance", zap.Int64("bob", result[0]))
 
+
+	// call example string function
+	name := []byte("Alice")
+	namePtr, err := newPtr(ctx, name, rt, true)
+	if err != nil {
+		return err
+	}
+	_, err = rt.Call(ctx, "print_name", programIDPtr, namePtr)
```
samliok, line 233: this is just temporary, was making a trivial example to show passing `Strings`.

x/programs/examples/token_test.go
```diff
@@ -24,7 +24,7 @@ var (
 	log = logging.NewLogger(
 		"",
 		logging.NewWrappedCore(
-			logging.Info,
+			logging.Debug,
```
samliok, line 27: also temp

x/programs/examples/util.go
```diff
@@ -5,29 +5,61 @@ package examples
...
 	"github.com/ava-labs/hypersdk/state"
 	"github.com/ava-labs/hypersdk/x/programs/runtime"
 )
 
-func newKeyPtr(ctx context.Context, key ed25519.PublicKey, rt runtime.Runtime) (int64, error) {
-	ptr, err := rt.Memory().Alloc(ed25519.PublicKeyLen)
+// newPtr allocates memory and writes [bytes] to it.
+// If [prependLength] is true, it prepends the length of [bytes] as a uint32 to [bytes].
+// It returns the pointer to the allocated memory.
+func newPtr(ctx context.Context, bytes []byte, rt runtime.Runtime, prependLength bool) (int64, error) {
```
samliok, line 21: made the function more generic

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-use wasmlanche_sdk::{program::Program, public, state_keys, types::Address};
+use wasmlanche_sdk::{program::Program, public, state_keys, types::Address, state::from_raw_ptr};
```
samliok: the macro requires this function, is there a way to use this in `sdk_macros` without a dependency cycle?

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -100,6 +100,23 @@ pub fn transfer(program: Program, sender: Address, recipient: Address, amount: i
...
+
+    for (recipient, amount) in recipients.iter().zip(amounts.iter()) {
+        mint_to(program, *recipient, *amount as i64);
+    }
+
+    true
+}
+
+#[public]
+pub fn print_name(_: Program, name: String) {
```
samliok, line 116: temp

x/programs/rust/wasmlanche_sdk/src/state.rs
```diff
@@ -77,3 +78,40 @@ impl State {
...
+/// Converts a raw pointer to a deserialized value.
+/// Expects the first 4 bytes of the pointer to represent the [length] of the serialized value,
+/// with the subsequent [length] bytes comprising the serialized data.
+// pub fn from_raw_ptr<V>(ptr: i64) -> V
+// where
+//     V: serde::de::DeserializeOwned,
+// {
+//     let (bytes, _) = bytes_and_length(ptr);
+//     from_slice(&bytes).expect("failed to deserialize")
+// }
```
samliok, line 91: might warrant a discussion about using a serialization package. Would make the code cleaner and have more out of the box functionality

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -31,24 +37,33 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
 
             if let Pat::Ident(ref pat_ident) = **pat {
                 let param_name = &pat_ident.ident;
-                // We only set the type to i64 if it is not a supported WASM primitive.
+                let param_descriptor = if is_supported_primitive(ty) {
```
richardpringle, line 40: Would be nicer to have 
```rust
impl From<&Type> for ParamKind {
    // ...
}
```

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -31,24 +37,33 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
+                };
                 let param_type = if is_supported_primitive(ty) {
                     ty.to_token_stream()
                 } else {
                     parse_str::<Type>("i64")
                         .expect("valid i64 type")
                         .to_token_stream()
                 };
-                return (param_name, param_type);
+                return (param_name, (param_type, param_descriptor));
```
richardpringle, line 54: Can we flatten out this tuple?

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -58,8 +73,35 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
         panic!("Unsupported function parameter format.");
     });
 
+    // Converts the parameters that are pointers to their original type.
+    let converted_params = full_params.clone().map(|param| {
```
richardpringle, line 77: You shouldn't need to `clone` here.

x/programs/rust/sdk_macros/src/lib.rs
richardpringle, line 1: I think there's a bit better way to do this with the types... something like
```rust
// I don't know the actual type of `ty`, but that's what I'm calling `Type`. 
// you can also store the name here if you need it
enum Param {
    SupportedPrimitive(Type),
    Program(Type),
    Pointer(Type),
}
```
Then you go from `TokenStream` -> `Type` -> `Param` -> `TokenStream` without exception. 

Maybe open a new issue from this comment? The output code will be the same either way so we don't need to change it now.

x/programs/rust/wasmlanche_sdk/src/state.rs
richardpringle: I think we agreed on Borsh for now. @hexfusion?

x/programs/rust/wasmlanche_sdk/src/state.rs
```diff
@@ -77,3 +78,40 @@ impl State {
...
+    let (bytes, _) = bytes_and_length(ptr);
+    V::from_bytes(&bytes)
+}
+
+// TODO: move this logic to return a Memory struct that conatins ptr + length
+/// Returns a tuple of the bytes and length of the argument.
+/// # Panics
+/// Panics if the value cannot be converted from i32 to usize.
+/// # Safety
+/// This function is unsafe because it dereferences raw pointers.
```
richardpringle, line 110: 🤔  haha. If the function is `unsafe`, it should be marked as such. 

We can figure out how to fix it later, but it should be glaringly obvious where our unsafe calls are. 

Also, if you look at the [Safety parameters for `from_raw_parts`](https://doc.rust-lang.org/stable/std/slice/fn.from_raw_parts.html#safety), it looks like we have a bunch of these covered because we're using `u8` as `T` (don't have to worry about alignment, etc. 

Would it make sense for the host to call the `program` to do the allocation, guaranteeing the memory exists? Then the guest could place the bytes into the memory location? That's the only thing we're missing for safety (I think).

x/programs/rust/wasmlanche_sdk/src/state.rs
```diff
@@ -77,3 +78,40 @@ impl State {
...
+#[must_use]
+pub fn bytes_and_length(ptr: i64) -> (Vec<u8>, i32) {
+    let len = unsafe { std::slice::from_raw_parts(ptr as *const u8, 4) };
+    let len = u32::from_be_bytes(len.try_into().unwrap()) as usize;
```
richardpringle, line 114:
```suggestion
    type LenType = u32;
    
    let len = unsafe { std::slice::from_raw_parts(ptr as *const u8, 4) };

    assert_eq!(len.len(), std::mem::size_of::<LenType>());    
    let len = LenType::from_be_bytes(len.try_into().unwrap()) as usize;
```

x/programs/rust/wasmlanche_sdk/src/state.rs
```diff
@@ -77,3 +78,40 @@ impl State {
...
+}
+
+// TODO: move this logic to return a Memory struct that conatins ptr + length
+/// Returns a tuple of the bytes and length of the argument.
+/// # Panics
+/// Panics if the value cannot be converted from i32 to usize.
+/// # Safety
+/// This function is unsafe because it dereferences raw pointers.
+#[must_use]
+pub fn bytes_and_length(ptr: i64) -> (Vec<u8>, i32) {
```
richardpringle, line 112: Why return an `i32` for the length here?

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -81,8 +81,12 @@ impl From<i64> for Bytes32 {
     }
 }
 
+/// A trait that represents an argument that can be passed to & from the host.
 pub trait Argument {
     fn as_bytes(&self) -> Cow<'_, [u8]>;
+    fn from_bytes(bytes: &[u8]) -> Self
```
richardpringle, line 87: This needs to return `Result<Self, _>`. The error type can be changed later, but you know this will always be fallible so it's best to do the work to change it now. 

Calling `unwrap` is great until you've had to go refactor a ton of code that calls `unwrap` everywhere and you think to yourself, "it would have been so much easier to just return a `Result` in the first place". 

I'll leave a suggestion below for how to not get bogged down in dealing with failure modes while you're prototyping.

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -98,12 +102,24 @@ impl Argument for Bytes32 {
     fn as_bytes(&self) -> Cow<'_, [u8]> {
         Cow::Borrowed(&self.0)
     }
+    fn from_bytes(bytes: &[u8]) -> Self
+    where
+        Self: Sized,
+    {
+        let bytes = bytes[..ADDRESS_LEN].try_into().unwrap();
+        Self::new(bytes)
+    }
```
richardpringle, line 111: You can do this everywhere and connect the proper error types later. 

```suggestion
    fn from_bytes(bytes: &[u8]) -> Result<Self, Box<dyn std::error::Error>>
    where
        Self: Sized,
    {
        let bytes = bytes[..ADDRESS_LEN].try_into()?;
        Self::new(bytes)
    }
```

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -81,8 +81,12 @@ impl From<i64> for Bytes32 {
     }
 }
 
+/// A trait that represents an argument that can be passed to & from the host.
 pub trait Argument {
     fn as_bytes(&self) -> Cow<'_, [u8]>;
+    fn from_bytes(bytes: &[u8]) -> Self
+    where
+        Self: Sized;
```
richardpringle, line 89: I think you might as well put the bound on the trait, 
```rust
pub trait Argument: Sized {
```

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -113,6 +129,23 @@ impl Argument for i64 {
...
+
+impl Argument for i32 {
+    fn as_bytes(&self) -> Cow<'_, [u8]> {
+        Cow::Owned(self.to_be_bytes().to_vec())
+    }
+    fn is_primitive(&self) -> bool {
+        true
+    }
+    fn from_bytes(bytes: &[u8]) -> Self {
+        let bytes = bytes[..4].try_into().unwrap();
```
richardpringle, line 146: Should `assert` on the length here

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -122,4 +155,56 @@ impl Argument for Program {
...
+    }
+}
+
+impl Argument for String {
+    fn as_bytes(&self) -> Cow<'_, [u8]> {
+        Cow::Owned(self.as_bytes().to_vec())
+    }
+    fn from_bytes(bytes: &[u8]) -> Self {
+        String::from_utf8_lossy(bytes).to_string()
+    }
```
richardpringle, line 170: I think this is fine for now, but there has to be a better way to do this...

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -122,4 +155,56 @@ impl Argument for Program {
...
+        result_vec
+    }
+
+    fn as_bytes(&self) -> Cow<'_, [u8]> {
+        if self.is_empty() {
+            return Cow::Owned(Vec::new());
+        }
+
+        // avoid reallocation
+        let mut bytes = Vec::with_capacity(self.len() * self[0].len());
```
richardpringle, line 204: 🤔, can the `len` of `T` vary?

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -122,4 +155,56 @@ impl Argument for Program {
...
+        for elem in self {
+            bytes.extend_from_slice(&elem.as_bytes());
+        }
+        Cow::Owned(bytes)
```
richardpringle, line 208: Don't worry about re-allocating for now. We should either handle the case where the length of `T` varies, or we should force `T` to have a constant length with another trait-bound. In either case, the compiler should take care of minimizing allocations. 

```suggestion
        let bytes = self.iter().flat_map(|val| val.as_bytes()).collect();
        Cow::Owned(bytes)
```

richardpringle (changes requested):
Minor change requests

x/programs/rust/wasmlanche_sdk/src/state.rs
patrick-ogrady: Yes 👍

x/programs/rust/wasmlanche_sdk/src/types.rs
samliok: deleted `from_bytes` for argument so gonna resolve a lot of these comments

x/programs/rust/wasmlanche_sdk/src/state.rs
samliok: yep should be a u32

x/programs/rust/wasmlanche_sdk/src/state.rs
samliok: I'm not sure what you mean by 
> The guest could place the bytes into the memory location? 

the host allocated and writes the data for these pointers when it creates parameters.

x/programs/rust/wasmlanche_sdk/src/state.rs
samliok: using borsh now 🙌

x/programs/rust/wasmlanche_sdk/src/state.rs
samliok: also if I mark this function as unsafe, then all public functions need to be marked unsafe(since the macro calls this function).

x/programs/rust/examples/token/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2021"
 wasmlanche_sdk = { version = "0.1.0", path = "../../wasmlanche_sdk" }
 serde_bytes = "0.11.5"
 serde ={ version = "1.0.130", features = ["derive"] }
+borsh = { version = "1.2.0", features = ["derive"] }
```
richardpringle, line 12: 🎉

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -31,56 +26,11 @@ impl IntoIterator for Address {
...
-                .try_into()
-                .unwrap()
-        };
-        Self(bytes)
+        IntoIterator::into_iter(self.0)
     }
 }
 
+/// A trait that represents an argument that can be passed to & from the host.
 pub trait Argument {
```
samliok, line 34: only place we are using `Argument` now is when we are calling an external program. I think we can remove this trait in the future, going to mark as an issue and hit up in another PR.

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -100,6 +100,18 @@ pub fn transfer(program: Program, sender: Address, recipient: Address, amount: i
     true
 }
 
+/// Mints tokens to multiple recipients.
+#[public]
+pub fn mint_to_many(program: Program, recipients: Vec<Address>, amounts: Vec<i32>) -> bool {
```
richardpringle, line 105: Can we change this so that you can pass in `Vec<(Address, Amount)>`? And can you have that done by EOD, please? Haha

Jokes aside, do you think this would be possible? If yes, I'll create an issue!

x/programs/rust/sdk_macros/Cargo.toml
```diff
@@ -10,3 +10,4 @@ proc-macro = true
 proc-macro2 = "1.0.67"
 quote = "1.0.33"
 syn = { version = "2.0.37", features = ["full"] }
+unzip-n = "0.1.2"
```
richardpringle, line 13: Are you doing this by hand instead of using `cargo add`?

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -8,6 +8,51 @@ use quote::{quote, ToTokens};
 use syn::{
     parse_macro_input, parse_str, Fields, FnArg, Ident, ItemEnum, ItemFn, Pat, PatType, Type,
 };
+use unzip_n::unzip_n;
+unzip_n!(3);
```
richardpringle, line 12: 😳

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -58,8 +104,15 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
-    let (param_names, param_types): (Vec<_>, Vec<_>) = full_params.unzip();
+    let (param_names, param_types, converted_params) = full_params
+        .map(|(param_name, param_type, param_kind)| {
+            (
+                param_name,
+                param_type,
+                param_kind.converted_param_tokenstream(param_name),
+            )
+        })
+        .unzip_n_vec();
```
richardpringle, line 115: I see why you did that now...

x/programs/rust/wasmlanche_sdk/src/state.rs
richardpringle: > I'm not sure what you mean by
> 
> > The guest could place the bytes into the memory location?
> 
> the host allocated and writes the data for these pointers when it creates parameters.

I need to learn more about the different ways arguments can be passed back and forth. You can ignore my question.

richardpringle: > also if I mark this function as unsafe, then all public functions need to be marked unsafe(since the macro calls this function).

You can put the function call in an `unsafe` block inside the macro. It's okay though, don't worry about it for now.

x/programs/rust/examples/token/src/lib.rs
richardpringle, line 1: This example is really nice.

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -123,3 +76,9 @@ impl Argument for Program {
         true
     }
 }
+
+impl Argument for String {
+    fn as_bytes(&self) -> Cow<'_, [u8]> {
+        Cow::Owned(self.as_bytes().to_vec())
```
richardpringle, line 82:
```suggestion
        Cow::Borrowed(self.as_bytes())
```

richardpringle (changes requested):
Only the last comment is a change request

x/programs/rust/examples/token/src/lib.rs
samliok: possible, on it!

samliok: changed

x/programs/rust/sdk_macros/Cargo.toml
samliok: yea, should prob use `cargo add` tho

richardpringle (dismissed):

richardpringle (approved):
Re-approving!

x/programs/examples/token.go
```diff
@@ -28,6 +28,11 @@ func NewToken(log logging.Logger, programBytes []byte, db state.Mutable, cfg *ru
 	}
 }
 
+type mintingRecipient struct {
+	To     [32]byte
```
hexfusion, line 32: In the rust struct you use `recipient` as a field vs `To` shouldn't they be the same? I think it might read more logically as

```
type minter struct {
  To     [32]byte
  Amount int32
}

File: x/programs/examples/token.go [hunk 28]
```diff
@@ -183,6 +191,49 @@ func (t *Token) Run(ctx context.Context) error {
 		zap.Uint64("unit", rt.Meter().GetBalance()),
 	)
 
+	// combine alice and bobs addresses
+	payments := []mintingRecipient{
```
hexfusion, line 195:
```suggestion
	minters := []minter{
```

x/programs/examples/util.go
```diff
@@ -5,29 +5,62 @@ package examples
...
+
+	amountToAllocate := uint64(len(bytes))
+	writeBytes := bytes
+
+	if prependLength {
+		amountToAllocate += consts.Uint32Len
+		writeBytes = marshalArg(bytes)
+	}
+
+	ptr, err := rt.Memory().Alloc(amountToAllocate)
```
hexfusion, line 36: I feel like these should be separate steps see below.

https://github.com/ava-labs/hypersdk/pull/631/files#diff-46bbbce39f0b86c15aea460cf21fa799bd6fabad826e7ea351c7577e13997ac9R206-R228

hexfusion (commented):
Overall I think this a great step forward I have a few cosmetic things to review and in general would like to see the go tests moving to simulator but that can be a followup.

x/programs/examples/util.go
```diff
@@ -37,6 +70,12 @@ func newKey() (ed25519.PrivateKey, ed25519.PublicKey, error) {
 	return priv, priv.PublicKey(), nil
 }
 
+func fixedByteKey(key ed25519.PublicKey) [32]byte {
```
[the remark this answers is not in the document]
samliok: ahh, i actually wrote a test and it turned out this wasn't necessary because `key` is fixed length. I think i was running into this issue earlier because I was converting to a dynamic array somewhere in earlier code. Anyways, this isn't needed anymore but still good to note that borsh serializes fixed/dynamic array differently.

hexfusion (approved):

richardpringle (approved):
🚀

x/programs/examples/token.go
```diff
@@ -28,6 +28,11 @@ func NewToken(log logging.Logger, programBytes []byte, db state.Mutable, cfg *ru
 	}
 }
 
+type minter struct {
+	To     [32]byte
```
patrick-ogrady, line 32: Node that Address is 33 bytes. I recommend using `Address` directly here, if possible: https://github.com/ava-labs/hypersdk/blob/283846bac51de111d18c225c8b0a7b92b6cf6ec8/codec/address.go

x/programs/examples/token.go
```diff
@@ -28,6 +28,11 @@ func NewToken(log logging.Logger, programBytes []byte, db state.Mutable, cfg *ru
...
+	Amount int32
```
patrick-ogrady, line 33: Recommend adding a comment that a production contract would use uint64 here.

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -1,28 +1,23 @@
...
 pub const ADDRESS_LEN: usize = 32;
 /// A struct that enforces a fixed length of 32 bytes which represents an address.
-#[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Debug)]
-pub struct Address(Bytes32);
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug, BorshSerialize, BorshDeserialize)]
+pub struct Address([u8; Self::LEN]);
 
 impl Address {
+    pub const LEN: usize = 32;
```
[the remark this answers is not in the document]
patrick-ogrady: What is the difference between the `LEN` const and `ADDRESS_LEN` const?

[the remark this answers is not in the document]
samliok: `ADDRESS_LEN` should be removed

x/programs/examples/token.go
samliok: will create an issue to move to `HyperSDK.address` later. Currently use ed25519 public keys instead of addresses which have 32 bytes and not 33.

patrick-ogrady (approved):

PR #651, opened 2023-12-11
[X/Programs] Update hypersdk programs to use codec.Address

samliok:


x/programs/runtime/auth.go
patrick-ogrady, line 1: This is something that would normally be configured in the VM? I'd advocate for not hardcoding IDs/creating hardcoded IDs here.

x/programs/runtime/consts.go
```diff
@@ -14,4 +12,5 @@ const (
 	guestSuffix         = "_guest"
 	wasiPreview1ModName = "wasi_snapshot_preview1"
 	MemoryPageSize      = 64 * units.KiB
-)
+	ED25519ID uint8 = 0
```
patrick-ogrady, line 15: left a comment in another file about considering not defining this const here.

x/programs/rust/wasmlanche_sdk/src/types.rs
```diff
@@ -1,15 +1,12 @@
-use crate::program::Program;
 use borsh::{BorshDeserialize, BorshSerialize};
-use std::borrow::Cow;
 
 /// A struct that enforces a fixed length of 32 bytes which represents an address.
```
patrick-ogrady, line 3: nit: 33 bytes

x/programs/examples/imports/program/program.go
```diff
@@ -170,44 +159,54 @@ func (i *Import) callProgramFn(
...
-func getProgramWasmBytes(log logging.Logger, db state.Immutable, idBytes []byte) ([]byte, error) {
-	id, err := ids.ToID(idBytes)
-	if err != nil {
-		return nil, err
+func getProgramWasmBytes(log logging.Logger, db state.Immutable, addressBytes []byte) ([]byte, error) {
+	if len(addressBytes) != codec.AddressLen {
+		return nil, fmt.Errorf("invalid address length: %d", len(addressBytes))
 	}
 
+	address := codec.Address(addressBytes)
```
patrick-ogrady, line 204: You are probably going to want to assert that the first byte is the right typeID (in the VM setting).

x/programs/examples/storage/storage.go
```diff
@@ -29,23 +29,23 @@ func ProgramPrefixKey(id []byte, key []byte) (k []byte) {
 // Program
 //
 
-func ProgramKey(id ids.ID) (k []byte) {
-	k = make([]byte, 1+consts.IDLen)
-	copy(k[1:], id[:])
+func ProgramKey(address codec.Address) (k []byte) {
+	k = make([]byte, 1+codec.AddressLen)
+	copy(k[1:], address[:])
```
patrick-ogrady, line 34: Should probably leave a comment why you don't populate k[0]

x/programs/examples/token.go
```diff
@@ -58,23 +57,23 @@ func (t *Token) Run(ctx context.Context) error {
...
 
-	programIDPtr, err := runtime.WriteBytes(rt.Memory(), programID[:])
+	programAddressPtr, err := argumentToSmartPtr(programAddress, rt.Memory())
 	if err != nil {
 		return err
 	}
 
 	t.log.Debug("new token program created",
-		zap.String("id", programID.String()),
+		zap.String("id", codec.MustAddressBech32("token_", programAddress)),
```
patrick-ogrady, line 72: usually, there is no trailing dash here...there is already a "separator" character in bech32.

x/programs/examples/utils.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/ava-labs/avalanchego/database/memdb"
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/x/programs/runtime"
+	"github.com/near/borsh-go"
+)
+
+func newAddress() (ed25519.PrivateKey, codec.Address, error) {
```
patrick-ogrady, line 18: It was super annoying to make this generic in the hypersdk, recommend not assuming everything is ed25519 here.

x/programs/examples/storage/storage.go
samliok: will leave a comment, i think we should explicitly set `k[0]` to `programPrefix` here to make it clear as well.

x/programs/examples/imports/program/program.go
samliok: to clarify, this should be done in the VM? Not sure how the engine(at least in our current implementation) can ensure the right typeID's here. 
More generally, the SDK needs to handle address and `Program` types in a more tamper-proof way, but I'm still looking into that.

samliok: i guess, is this comment suggesting I should address(no pun intended) something now or more of a note for when we start building out the VMs later?

x/programs/examples/counter_test.go
```diff
@@ -48,70 +49,74 @@ func TestCounterProgram(t *testing.T) {
 	require.Equal(maxUnits, rt.Meter().GetBalance())
 
 	// simulate create program transaction
-	programID := ids.GenerateTestID()
-	err = storage.SetProgram(ctx, db, programID, counterProgramBytes)
+	programAddress := codec.CreateAddress(ED25519ID, ids.GenerateTestID())
```
patrick-ogrady, line 52: I don't think a program will have an ed25519ID address?

x/programs/examples/imports/program/program.go
patrick-ogrady: > to clarify, this should be done in the VM? Not sure how the engine(at least in our current implementation) can ensure the right typeID's here. More generally, the SDK needs to handle address and `Program` types in a more tamper-proof way, but I'm still looking into that.

You are right. This should be handled by the wrapping VM. Can you leave a comment/invariant stating that?

x/programs/examples/imports/program/program.go
```diff
@@ -170,44 +159,54 @@ func (i *Import) callProgramFn(
...
+
 	// get the program bytes from storage
-	bytes, exists, err := storage.GetProgram(context.Background(), db, id)
+	bytes, exists, err := storage.GetProgram(context.Background(), db, address)
 	if !exists {
-		log.Debug("key does not exist", zap.String("id", id.String()))
+		log.Debug("key does not exist", zap.String("address", codec.MustAddressBech32("addr", address)))
```
patrick-ogrady, line 209: I would just use `prog` as the prefix here?

x/programs/examples/utils.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/ava-labs/avalanchego/database/memdb"
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/utils"
+	"github.com/ava-labs/hypersdk/x/programs/runtime"
+	"github.com/near/borsh-go"
+)
+
+var ED25519ID uint8 = 0
```
patrick-ogrady, line 19: Does this need to be exported if only used within `examples`?

x/programs/rust/wasmlanche_sdk/src/program.rs
```diff
@@ -8,8 +8,8 @@ use crate::{errors::StateError, host::call_program, state::State};
 pub struct Program([u8; Self::LEN]);
 
 impl Program {
-    /// The length of ids.ID
-    pub const LEN: usize = 32;
```
hexfusion: isn't 32 bytes still needed here the id() method will return an ID that should be 32 bytes right?

x/programs/runtime/auth.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/utils"
+)
+
+func NewED25519Address(pk ed25519.PublicKey) codec.Address {
+	return codec.CreateAddress(ED25519ID, utils.ToID(pk[:]))
+}
```
hexfusion: nit

x/programs/rust/wasmlanche_sdk/src/program.rs
samliok: I was under the impression that programs id's would also be saved as `Address` with a different typeID

hexfusion: I guess we should verify how that will look. HyperShift still used a 32bit `ID`[1]. So if we are returning a program.id() that should be 32bits. I think we can achieve what your describing by dropping the first bit. 


[1] https://github.com/ava-labs/hypersdk/blob/ad3ac9427c178c85cc55e5dbc9b044b04152d89e/examples/morpheusvm/storage/storage.go#L74

https://github.com/ava-labs/hypersdk/blob/b3da5dec37df3f6bb7a27069e6d878f485aa4b34/x/programs/rust/wasmlanche_sdk/src/program.rs#L14C1-L18C6

samliok: ah ok, guess we'll keep it as id then. Can see this also becoming generic tho

x/programs/examples/storage/storage.go
```diff
@@ -25,17 +27,15 @@ func ProgramPrefixKey(id []byte, key []byte) (k []byte) {
...
 
-//
-// Program
-//
-
+// ProgramKey returns the key used to store the program bytes at [id].
 func ProgramKey(id ids.ID) (k []byte) {
 	k = make([]byte, 1+consts.IDLen)
 	copy(k[1:], id[:])
+	k[0] = programPrefix
```
patrick-ogrady, line 34: set this before the copy line (set byte 0 before byte 1)

x/programs/examples/storage/storage.go
```diff
@@ -17,6 +17,8 @@ const (
 	programPrefix = 0x0
 )
 
+// ProgramPrefixKey returns a properly formatted key
+// for storing a value at [id][key].
 func ProgramPrefixKey(id []byte, key []byte) (k []byte) {
```
patrick-ogrady, line 22: This function is wrong. You are setting byte 0 then overriding it with an id copy.

x/programs/examples/storage/storage.go
```diff
@@ -17,6 +17,8 @@ const (
...
 	k = make([]byte, consts.IDLen+1+len(key))
```
patrick-ogrady, line 23: nit: put the 1+ first (reflects what is actually in the key)

x/programs/examples/token.go
```diff
@@ -31,7 +31,7 @@ func NewToken(log logging.Logger, programBytes []byte, db state.Mutable, cfg *ru
 
 type minter struct {
 	// TODO: use a HyperSDK.Address instead
```
patrick-ogrady, line 33: remove todo

x/programs/examples/utils.go
```diff
@@ -8,19 +8,27 @@ import (
...
 	"github.com/ava-labs/hypersdk/crypto/ed25519"
 	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/utils"
 	"github.com/ava-labs/hypersdk/x/programs/runtime"
 	"github.com/near/borsh-go"
 )
 
-func newKey() (ed25519.PrivateKey, ed25519.PublicKey, error) {
+var (
+	ed25519ID     uint8 = 0
```
patrick-ogrady, line 20: nit: add a comment for what this is/why it is needed

hexfusion (approved):

x/programs/examples/storage/storage.go
```diff
@@ -17,25 +17,25 @@ const (
...
 
+// ProgramPrefixKey returns a properly formatted key
+// for storing a value at [id][key].
 func ProgramPrefixKey(id []byte, key []byte) (k []byte) {
-	k = make([]byte, consts.IDLen+1+len(key))
+	k = make([]byte, 1+consts.IDLen+len(key))
 	k[0] = programPrefix
-	copy(k, id[:])
+	copy(k[1:consts.IDLen], id[:])
 	copy(k[consts.IDLen:], (key[:]))
```
patrick-ogrady, line 26: This should be 1+consts.IDLen

patrick-ogrady: Also unneeded parenthesis on second param

patrick-ogrady (changes requested):
Key encoding is almost right

x/programs/examples/storage/storage.go
```diff
@@ -17,25 +17,25 @@ const (
 	programPrefix = 0x0
 )
 
+// ProgramPrefixKey returns a properly formatted key
+// for storing a value at [id][key].
 func ProgramPrefixKey(id []byte, key []byte) (k []byte) {
```
dboehm-avalabs, line 22: Are ids always a fixed length?  If so we should probably make it a fixed size []byte.  If they aren't, then we probably shouldn't use a const length for the offsets into the result.

x/programs/examples/storage/storage.go
```diff
@@ -17,25 +17,25 @@ const (
...
-	k = make([]byte, consts.IDLen+1+len(key))
+	k = make([]byte, 1+consts.IDLen+len(key))
 	k[0] = programPrefix
-	copy(k, id[:])
-	copy(k[consts.IDLen:], (key[:]))
+	copy(k[1:1+consts.IDLen], id[:])
```
dboehm-avalabs, line 25: Don't need the 1+1+consts.IDLen for the copy, since it will stop at len(id), but it doesn't really change the functionality to have it vs not so 🤷

dboehm-avalabs (dismissed):

x/programs/examples/storage/storage.go
samliok: ids are fixed length, but we grab the bytes from the runtime memory as a slice. I guess it may make sense to ensure the slice is `const.IDLen` but this could be redundant because the `Program` type in the rust sdk is already set to `const.IDLen`.

PR #658, opened 2023-12-15
[x/programs] Add host abstractions

hexfusion:
This breaks out host related `Import` types into its own package.

Depends on #657


x/programs/host/imports.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/ava-labs/hypersdk/x/programs/engine"
+)
+
+const (
+	wasiPreview1ModName = "wasi_snapshot_preview1"
+)
+
+var NoSupportedImports = make(SupportedImports)
+
+type Imports map[string]Import
```
samliok, line 18: don't see where we use this type

x/programs/host/imports.go
```diff
@@ -0,0 +1,69 @@
...
+
+type ImportFnCallback struct {
```
samliok, line 20: 🚀

samliok (dismissed):

x/programs/host/imports.go
hexfusion: yeah I dropped it as it was already explicitly defined below

dboehm-avalabs (approved):

samliok (approved):

PR #659, opened 2023-12-22
[x/programs] Improve external program function call DX

samliok:
I think this is a slight improvement to the current way we are calling external programs. We never use the calling `programID` anyways. 

I do see us needing to distinguish between a `caller` vs `target` for reentrancy purposes later on, but imo the calling vs target programID should be managed by the engine/runtime. Thus passing the program ID of the `caller` should be handled outside of rust. If needed, wouldn't be a hard change to revert.


richardpringle (dismissed):
🥇

richardpringle (approved):

PR #660, opened 2023-12-22
[x/programs] add program abstraction

hexfusion:
Depends on #657 and #658


x/programs/host/link.go
```diff
@@ -0,0 +1,154 @@
...
+				}
+			}()
+		}
+		return f
+	}
+	return l.wasmLink.FuncWrap(module, name, wrapper())
+}
+
+// RegisterImportWrapFn registers a wrapped host function exposed to the guest (import).
+func (l *Link) RegisterImportWrapFn(module, name string, paramCount int, f func(caller *program.Caller, args ...wasmtime.Val) (*types.Val, error)) error {
```
samliok, line 108: I think it would be useful to add a more detailed function comment here. Specifically citing the difference between registering a Wrapped vs non wrapped import fn.

x/programs/program/memory.go
```diff
@@ -0,0 +1,243 @@
...
+			params = append(params, uint64(ptr))
+		case int:
+			if v < 0 {
+				return nil, fmt.Errorf("failed to write param: %w", ErrUnderflow)
+			}
+			params = append(params, uint64(v))
+		case uint64:
+			params = append(params, v)
+		default:
+			return nil, fmt.Errorf("invalid param: supported types int, uint64 and string")
```
samliok, line 185: should we support SmartPtrs?

x/programs/program/utils.go
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package program
+
+import "math"
+
+// EnsureInt64ToInt32 returns true if a [v] (int64) can be safely converted to an int32.
+func EnsureInt64ToInt32(v int64) bool {
```
samliok, line 9: we can remove this func since we don't use it

x/programs/examples/imports/wrap/wrap.go
```diff
@@ -0,0 +1,130 @@
...
+	importFn := ImportFn[F]{fn: src}
+	fn := func(c *program.Caller, wargs ...wasmtime.Val) (*types.Val, error) {
+		args := make([]int64, 0, len(wargs))
+		for _, arg := range wargs {
+			args = append(args, arg.I64())
+		}
+		return importFn.Invoke(c, args...)
+	}
+	return fn
+}
```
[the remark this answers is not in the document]
hexfusion: Yes but this is an example of how to not require that approach and have named params.

x/programs/program/memory.go
hexfusion: this should all get cleaned up when you update the simulator

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -92,29 +90,35 @@ func (i *Import) putFn(caller *wasmtime.Caller, id int64, key int64, value int64
 		i.log.Error("failed to insert into storage",
 			zap.Error(err),
 		)
-		return -1
+		return nil, err
 	}
 
-	return 0
+	return types.ValI64(0), nil
```
samliok, line 96: bug: need to change the rust sdk to expect a `i32`

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -124,31 +128,31 @@ func (i *Import) getFn(caller *wasmtime.Caller, id int64, key int64) int64 {
 				zap.Error(err),
 			)
 		}
-		return -1
+		return nil, err
```
samliok, line 131: bug: need to update the rust sdk to handle for errors now

samliok (changes requested):

x/programs/examples/imports/pstate/pstate.go
samliok: created an issue to hit up after this is merged. Temporarily returning -1 here
https://github.com/ava-labs/hypersdk/issues/666

samliok (approved):

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -44,12 +44,34 @@ func (i *Import) Name() string {
...
 	wrap := wrap.New(link)
-	if err := wrap.RegisterThreeParamInt64Fn(Name, "put", i.putFn); err != nil {
+	if err := wrap.RegisterAnyParamFn(Name, "put", 3, i.putFnVariadic); err != nil {
 		return err
 	}
-	return wrap.RegisterTwoParamInt64Fn(Name, "get", i.getFn)
+	return wrap.RegisterAnyParamFn(Name, "get", 2, i.getFnVariadic)
 }
 
+func (i *Import) putFnVariadic(caller *program.Caller, args ...int64) (*types.Val, error) {
```
samliok, line 53: we could change our host functions to be variadic functions and throw an error if an incorrect # of params are passed in to avoid this. Would make the host functions less readable.

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -44,12 +44,34 @@ func (i *Import) Name() string {
...
+}
+
+func (i *Import) getFnVariadic(caller *program.Caller, args ...int64) (*types.Val, error) {
+	if len(args) != 2 {
+		return nil, errors.New("expected 2 arguments")
+	}
+	return i.getFn(caller, args[0], args[1])
+}
+
+type MsgLevel int
```
samliok: remove from pr, was playing around for #666

x/programs/runtime/instance.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	mem, err := exp.Memory()
+	if err != nil {
+		return nil, err
+	}
+
+	exp, err = i.GetExport(program.AllocFnName)
```
samliok, line 64: we require the instance to export an `alloc` function here, might make sense to make this optional

samliok (approved):

patrick-ogrady (approved):

PR #672, opened 2024-01-10
[fee manager] add unit tests

### LINKED ISSUE #674: [chain] feeManager add unit tests

wlawt:
Closes #674 

Unit tests for all FeeManager functions, currently at ~80% total coverage for the `fee_manager.go`

Some notes: I encountered into this issue `Caught SIGILL in blst_cgo_init, consult <blst>/bindings/go/README.md` when running `go test` within `chain` package. To fix it, I ran:

```sh
export CGO_CFLAGS="-O -D__BLST_PORTABLE__"
export CGO_CFLAGS_ALLOW="-O -D__BLST_PORTABLE__"
```

### DIFF
--- a/chain/fee_manager_test.go
+++ b/chain/fee_manager_test.go
@@ -1,0 +1,85 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package chain
+
+import (
+	//"context"
+	//"fmt"
+	"time"
+	"testing"
+
+	"github.com/ava-labs/avalanchego/ids"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSetUnitPrice(t *testing.T) {
+	require := require.New(t)
+
+	var d Dimensions
+	feeManager := NewFeeManager(nil)
+	for i := Dimension(0); i < FeeDimensions; i++ {
+		// Set unit prices for different dimensions
+		feeManager.SetUnitPrice(i, uint64(i))
+		d[i] = uint64(i)
+	}
+
+	unitPrices := feeManager.UnitPrices()
+	for i := Dimension(0); i < FeeDimensions; i++ {
+		require.Equal(unitPrices[i], feeManager.UnitPrice(i))
+	}
+
+	_, err := feeManager.MaxFee(d)
+	require.NoError(err)
+}
+
+func TestSetLastConsumed(t *testing.T) {
+	require := require.New(t)
+	feeManager := NewFeeManager(nil)
+	for i := Dimension(0); i < FeeDimensions; i++ {
+		feeManager.SetLastConsumed(i, uint64(i))
+	}
+
+	unitsConsumed := feeManager.UnitsConsumed()
+	for i := Dimension(0); i < FeeDimensions; i++ {
+		require.Equal(unitsConsumed[i], feeManager.LastConsumed(i))
+	}
+}
+
+func TestConsume(t *testing.T) {
+	require := require.New(t)
+	feeManager := NewFeeManager(nil)
+
+	consumed := Dimensions{100, 100, 100, 100, 100}
+	maxUnits := Dimensions{1_800_000, 2_000, 2_000, 2_000, 2_000}
+
+	ok, dimension := feeManager.Consume(consumed, maxUnits)
+	require.True(ok)
+	require.Equal(dimension, Dimension(0))
+}
+
+func TestComputeNext(t *testing.T) {
+	/*require := require.New(t)
+	//var vm VM
+	var r Rules
+	parent := &StatelessBlock{
+		StatefulBlock: &StatefulBlock{
+			Prnt: ids.GenerateTestID(),
+			Tmstmp: time.Date(2024, time.January, 1, 0, 0, 0, 0, time.UTC).UnixMilli(),
+			Hght: 10000,
+		},
+	}
+
+	parentTime := time.Date(2024, time.January, 1, 0, 0, 0, 0, time.UTC).UnixMilli()
+	nextTime := time.Now().UnixMilli()
+	//parentID := parent.ID()
+	//r := parent.vm.Rules(nextTime)
+
+	parentFeeManager := NewFeeManager(nil)
+	_, err := parentFeeManager.ComputeNext(parentTime, nextTime, r)
+	require.NoError(err)*/
+
+	parent := NewGenesisBlock(ids.GenerateTestID())
+	
+}
--- a/chain/fee_manager_test.go
+++ b/chain/fee_manager_test.go
@@ -4,7 +4,6 @@
 package chain
 
 import (
-	//"fmt"
 	"time"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+	require := require.New(t)
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	// Mock VM
+	rules := NewMockRules(ctrl)
+	rules.EXPECT().GetMinUnitPrice().Return(Dimensions{100, 100, 100, 100, 100})
+	rules.EXPECT().GetMaxBlockUnits().Return(Dimensions{1_800_000, 2_000, 2_000, 2_000, 2_000})
+
+	// Similar to vm.go L326
```
patrick-ogrady, line 27: If you want to link to code, you should use a permalink here.

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+func TestUnitPrice(t *testing.T) {
+	require := require.New(t)
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	// Mock VM
+	rules := NewMockRules(ctrl)
+	rules.EXPECT().GetMinUnitPrice().Return(Dimensions{100, 100, 100, 100, 100})
```
patrick-ogrady, line 24: Nice usage of mock 👍

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+		d[i] = uint64(i)
+	}
+
+	// Check unit prices
+	unitPrices := feeManager.UnitPrices()
+	for i := Dimension(0); i < FeeDimensions; i++ {
+		require.Equal(unitPrices[i], feeManager.UnitPrice(i))
+	}
+
+	// Similar to spam.go
```
patrick-ogrady, line 44: Would recommend explaining what you are testing here rather than just saying it is similar to other usage.

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+	maxUnits := rules.GetMaxBlockUnits()
+	_, err := MulSum(unitPrices, maxUnits)
```
patrick-ogrady, line 46: I recommend checking the output of this function, not just that it didn't error.

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+	rules.EXPECT().GetStorageValueWriteUnits().Return(uint64(3))
+
+	// Get usage for bandwith, compute, read, allocate, writes
+	bandwidthOp := math.NewUint64Operator(200)
+	bandwidthUnits, err := bandwidthOp.Value()
+	require.NoError(err)
+
+	computeUnitsOp := math.NewUint64Operator(rules.GetBaseComputeUnits())
+	computeUnitsOp.Add(50)
+	computeUnitsOp.Add(100)
```
patrick-ogrady, line 78: Unless we are also attempting to test the math package here, I recommend just doing a single addition.

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+	writesOp.MulAdd(uint64(chunks), rules.GetStorageValueWriteUnits())
+	writeUnits, err := writesOp.Value()
+	require.NoError(err)
+
+	consumed := Dimensions{bandwidthUnits, computeUnits, readUnits, allocateUnits, writeUnits}
+
+	// Similar to processor.go
+	ok, dimension := feeManager.Consume(consumed, rules.GetMaxBlockUnits())
+	require.True(ok)
+	require.Equal(dimension, Dimension(0))
```
patrick-ogrady, line 105: Recommend also adding a test to ensure the Dimension returned in the error case is correct.

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// The fee required 
+	_, err = feeManager.MaxFee(consumed)
```
patrick-ogrady, line 108: Recommend checking the value of this MaxFee

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+	require.NoError(err)
+
+	// Update last consumed
+	for i := Dimension(0); i < FeeDimensions; i++ {
+		feeManager.SetLastConsumed(i, consumed[i])
```
patrick-ogrady, line 113: This is already done as part of `Consume`: https://github.com/ava-labs/hypersdk/blob/797d38ae84c8d7dca027a6b1d3e875e49218082e/chain/fee_manager.go#L138

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// Similar to genesis config
+	rules := NewMockRules(ctrl)
+	rules.EXPECT().GetWindowTargetUnits().Return(Dimensions{20_000_000, 1_000, 1_000, 1_000, 1_000})
+	rules.EXPECT().GetUnitPriceChangeDenominator().Return(Dimensions{48, 48, 48, 48, 48})
+	rules.EXPECT().GetMinUnitPrice().Return(Dimensions{100, 100, 100, 100, 100})
+
+	// Create our parent "block"
+	parentTime := time.Date(2024, time.January, 1, 0, 0, 0, 0, time.UTC).UnixMilli()
+	nextTime := time.Now().UnixMilli()
```
patrick-ogrady, line 135: Recommend only using hardcoded timestamps in unit tests (otherwise can get flaky)

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	parentFeeManager := NewFeeManager(nil)
+	_, err := parentFeeManager.ComputeNext(parentTime, nextTime, rules)
```
patrick-ogrady, line 138: I think a core part of this test will be checking the output in different scenarios.

Of all the functions to put time into writing a ton of tests for, this is probably the one.

You might find it useful to use table-driven tests for this guy.

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+func TestAdd(t *testing.T) {
+	require := require.New(t)
+	// Similar to tokenvm/cmd/token-wallet/backend/backend.go
+	consumed := Dimensions{1,2,3,4,5}
+	nconsumed := Dimensions{}
+	tmpConsumed, err := Add(nconsumed, consumed)
+	require.NoError(err)
+	nconsumed = tmpConsumed
+	for i := 0; i < FeeDimensions; i++ {
+		require.Equal(nconsumed[i], consumed[i])
```
patrick-ogrady, line 151: I think we should add an `Equals` function like we have for `Add`

chain/fee_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+func TestDimensionGreater(t *testing.T) {
+	require := require.New(t)
+	d := Dimensions{2,3,4,5,6}
+	o := Dimensions{1,2,3,4,5}
+	result := d.Greater(o)
+	require.True(result)
+}
+
+func TestInvalidDimensionGreater(t *testing.T) {
```
patrick-ogrady, line 163: For this Valid/Invalid tests of Greater, you should use table-driven testing (here is a basic example): https://github.com/ava-labs/avalanchego/blob/310b77e221b8d0c5fcee3254900079b1987466f5/network/p2p/gossip/bloom_test.go#L18

patrick-ogrady (changes requested):

chain/fee_manager_test.go
wlawt: Moved to its separate unit test

chain/fee_manager_test.go
```diff
@@ -0,0 +1,346 @@
...
+	// We set unit prices by the min unit price assigned at Genesis
+	// This is similar to a usage described in:
+	// https://github.com/ava-labs/hypersdk/blob/main/vm/vm.go#L326 
+	minUnitPrice := rules.GetMinUnitPrice()
+
+	var d Dimensions
+	feeManager := NewFeeManager(nil)
+	for i := Dimension(0); i < FeeDimensions; i++ {
+		// Set unit prices for different dimensions 
+		feeManager.SetUnitPrice(i, minUnitPrice[i])
```
dboehm-avalabs, line 34: Might want to make these all different and set a different value for last consumed.  I doubt it is the case that these are getting scrambled, but useful to confirm it.

wlawt:
Now that we have a `fees` package, I'm going to be revisiting this and placing the unit test in that package

Issue #613, opened 2023-11-09
[auth/morpheusvm] Add `BLS` signing


patrick-ogrady:

wlawt:
@patrick-ogrady i think this one can be closed too?

PR #675, opened 2024-01-11
[crypto]: add bls

### LINKED ISSUE #676: [crypto] support bls option

wlawt:
Closes #676 and #677

Add support for BLS by using the BLS library that exists in AvalancheGo. This would let developers integrate BLS, if they choose to, into their VM.

Overview:
- Introduces BLS into the `crypto` package
- The BLS package is a wrapper of a selection of functions from the AvalancheGo library that aligns with the existing function signatures that exist for ed25519 and secp256r1
- Example of integration into the MorpheusVM

Test Plan:
- Unit tests for the crypto bls package
   - Run `go test` within `bls` package
- Integration test in MorpheusVM sending between ed25519 and bls
   - Run `./scripts/tests.integration.sh` within `examples/morpheusvm`
- Manual testing with CLI interacting with MorpheusVM
   - Run `./scripts/build.sh`
   - Generate a BLS key
   - Send funds to the BLS address 
   - Send funds from the BLS address to any other wallet 

Future TODOs:
- Possibly replace usage of `avalanchego/utils/crypto` throughout the repo with this internal package
- Add developer docs to show how to integrate own crypto into their own HyperVM


wlawt:
Related: #613

crypto/bls/private.go
```diff
@@ -0,0 +1,52 @@
...
+// See the file LICENSE for licensing terms.
+
+package bls
+
+import (
+	"errors"
+
+	blst "github.com/supranational/blst/bindings/go"
+
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
```
patrick-ogrady, line 11: ❤️

crypto/bls/private.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+const PrivateKeyLen = blst.BLST_SCALAR_BYTES
```
patrick-ogrady, line 14: recommend changing this to `const PrivateKeyLen = bls.SecretKeyLen` so we don't need to keep our reference to `blst` (I think there are a few places you do something like this)

examples/morpheusvm/auth/bls.go
```diff
@@ -0,0 +1,165 @@
...
+func (d *BLS) Marshal(p *codec.Packer) {
+	p.PackFixedBytes(bls.SerializePublicKey(&d.Signer))
+	p.PackFixedBytes(bls.SignatureToBytes(&d.Signature))
+}
+
+func UnmarshalBLS(p *codec.Packer, _ *warp.Message) (chain.Auth, error) {
+	var d BLS
+
+	signer := bls.SerializePublicKey(&d.Signer)
+	p.UnpackFixedBytes(bls.PublicKeyLen*2, &signer)
```
patrick-ogrady, line 98: why is this *2?

examples/morpheusvm/auth/bls.go
```diff
@@ -0,0 +1,165 @@
...
+
+var _ chain.Auth = (*BLS)(nil)
+
+const (
+	BLSComputeUnits = 10
+	BLSSize         = bls.PublicKeyLen + bls.SignatureLen
+)
+
+type BLS struct {
+	Signer    bls.PublicKey `json:"signer"`
```
patrick-ogrady, line 28: If the AvalancheGo library operates over pointers, I would recommend using them here instead of dereferencing things.

examples/morpheusvm/auth/bls.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (d *BLS) Marshal(p *codec.Packer) {
+	p.PackFixedBytes(bls.SerializePublicKey(&d.Signer))
+	p.PackFixedBytes(bls.SignatureToBytes(&d.Signature))
+}
+
+func UnmarshalBLS(p *codec.Packer, _ *warp.Message) (chain.Auth, error) {
+	var d BLS
+
+	signer := bls.SerializePublicKey(&d.Signer)
```
patrick-ogrady, line 97: You can just create a byte array of the size you are expecting and pass that into `Deserialize*` instead of doing this (which is going to be much less efficient)

examples/morpheusvm/cmd/morpheus-cli/cmd/key.go
```diff
@@ -22,10 +23,11 @@ import (
 const (
 	ed25519Key   = "ed25519"
 	secp256r1Key = "secp256r1"
+	blsKey       = "bls"
 )
 
 func checkKeyType(k string) error {
-	if k != ed25519Key && k != secp256r1Key {
+	if k != ed25519Key && k != secp256r1Key && k != blsKey {
```
patrick-ogrady, line 30: I think we could convert this to a `switch`

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -194,7 +194,7 @@ var _ = ginkgo.BeforeSuite(func() {
 	app := &appSender{}
 	for i := range instances {
 		nodeID := ids.GenerateTestNodeID()
-		sk, err := bls.NewSecretKey()
```
patrick-ogrady, line 197: You can actually leave this as the avalanchego import (this is for Warp, which uses that lib)

examples/morpheusvm/auth/bls.go
wlawt: `bls.PublicKeyLen` is defined to be `BLST_P1_COMPRESS_BYTES` which is 48 bytes. When we serialize the public key this has a length of `BLST_P1_SERIALIZE_BYTES` which is twice of `PublicKeyLen`. When just unpacking 48 bytes the rest are all 0-filled which led to my `invalid signature`. I'll also make a note of this in the code

crypto/bls/private.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"errors"
+
+	"github.com/ava-labs/avalanchego/utils/crypto/bls"
+)
+
+const PrivateKeyLen = bls.SecretKeyLen
+
+var (
```
patrick-ogrady, line 14: nit: you should be able to do `var errFailedPrivateKeyDeserialize = errors.New("couldn't deserialize secret key")`

examples/morpheusvm/auth/bls.go
patrick-ogrady: I think we should make a new const of serialize bytes...we need to align `BLSSize` with the actual marshal size or else it won't be big enough?

examples/morpheusvm/auth/bls.go
```diff
@@ -21,7 +21,8 @@ var _ chain.Auth = (*BLS)(nil)
 
 const (
 	BLSComputeUnits = 10
-	BLSSize         = bls.PublicKeyLen + bls.SignatureLen
+	BLSCompressSize	= bls.PublicKeyLen * 2
```
patrick-ogrady, line 24: We should just use the byte length of the serialized key here?

What use is the non-serialized key length?

patrick-ogrady: I'm not sure `PublicKeyLen` isn't this already: https://github.com/ava-labs/avalanchego/blob/7623ffd4be915a5185c9ed5e11fa9be15a6e1f00/utils/crypto/bls/public.go#L12C41-L12C41

examples/morpheusvm/auth/bls.go
```diff
@@ -148,22 +148,22 @@ func (d *BLS) Refund(
 var _ chain.AuthFactory = (*BLSFactory)(nil)
 
 type BLSFactory struct {
-	priv bls.PrivateKey
+	priv *bls.PrivateKey `json:"priv,omitempty"`
```
patrick-ogrady, line 151: private fields aren't marshaled in `json`, so I think this is unnecessary?

examples/morpheusvm/auth/bls.go
```diff
@@ -88,22 +87,27 @@ func (*BLS) Size() int {
...
 	// https://github.com/supranational/blst/blob/master/bindings/go/blst.go#L165C41-L165C41
-	p.UnpackFixedBytes(BLSCompressSize, &signer)
+	p.UnpackFixedBytes(bls.PublicKeyLen, &signer)
 	signature := make([]byte, bls.SignatureLen)
 	p.UnpackFixedBytes(bls.SignatureLen, &signature)
 
-	d.Signer = bls.DeserializePublicKey(signer)
+	pk, err := bls.PublicKeyFromBytes(signer)
+	if err != nil {
+		return &BLS{}, nil
```
patrick-ogrady, line 107: this should be `return nil, err`

examples/morpheusvm/auth/bls.go
```diff
@@ -88,22 +87,27 @@ func (*BLS) Size() int {
...
+	}
+	d.Signer = pk
+	
 	sig, err := bls.SignatureFromBytes(signature)
 	if err != nil {
 		return &BLS{}, nil
```
patrick-ogrady, line 113: should be `nil, err`

examples/morpheusvm/auth/bls.go
```diff
@@ -0,0 +1,173 @@
...
+func (d *BLS) Marshal(p *codec.Packer) {
+	p.PackFixedBytes(bls.PublicKeyToBytes(d.Signer))
+	p.PackFixedBytes(bls.SignatureToBytes(d.Signature))
+}
+
+func UnmarshalBLS(p *codec.Packer, _ *warp.Message) (chain.Auth, error) {
+	var d BLS
+
+	signer := make([]byte, bls.PublicKeyLen)
+	// Public key length is defined to be 48 bytes.
```
patrick-ogrady, line 98: Can remove this comment which is no longer necessary

examples/morpheusvm/auth/bls.go
```diff
@@ -0,0 +1,173 @@
...
+type BLSFactory struct {
+	priv *bls.PrivateKey
+}
+
+func NewBLSFactory(priv *bls.PrivateKey) *BLSFactory {
+	return &BLSFactory{priv}
+}
+
+func (d *BLSFactory) Sign(msg []byte, _ chain.Action) (chain.Auth, error) {
+	sig := bls.Sign(msg, d.priv)
```
patrick-ogrady, line 163: You can do this inline below (since it returns no error)

examples/morpheusvm/auth/bls.go
```diff
@@ -0,0 +1,173 @@
...
+	r chain.Rules,
+	_ state.Immutable,
+	_ chain.Action,
+) (uint64, error) {
+	// We don't do anything during verify (there is no additional state to check
+	// to authorize the signer other than verifying the signature)
+	return d.MaxComputeUnits(r), nil
+}
+
+func (d *BLS) Actor() codec.Address {
```
patrick-ogrady, line 77: nit: I would use `b *BLS`

patrick-ogrady (approved):
Nice work!

PR #686, opened 2024-01-18
[x/programs] Limit Reentrancy at the Program Method Level

samliok:
This PR allows a dev to specify whether `public` functions allow reenterancy or not. Reentrancy is not allowed by default. 

To allow reentrancy, pass in the `reentrant` attribute into the public macro.

`#[public(reentrant)]`

## How it works

The host holds a new struct `ReentrancyGuard` which keeps track of visited functions during a program's execution. 

We also define a new host funciton call `enter_program` which sets the calling function to visited in `ReentrancyGuard` and returns whether or not the function has already been visited. 

Inside the `#public` macro, we call `enter_program`, and if the function is not reentrant and has already been visited we panic.


x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -46,6 +46,16 @@ fn inc_external(_: Program, target: Program, max_units: i64, of: Address, amount
         .unwrap()
 }
 
+// the macro would input a boolean into the function and some logic, checking if reentrancy is allowed. 
+// A default value would be false, but if the runtime provides true it must check the the with_reentrancy flag
+
+#[public]
+fn reentrance_example(_: Program, target_a: Program, target_b: Program, max_units: i64) -> i64 {
```
samliok, line 50: need to make a better example and potential in a separate crate

x/programs/examples/counter_test.go
```diff
@@ -79,71 +81,53 @@ func TestCounterProgram(t *testing.T) {
 
 	// create counter for alice on program 1
 	result, err := rt.Call(ctx, "initialize_address", programIDPtr, alicePtr)
+	reentrancyGaurd.Reset()
```
samliok, line 84: temporarily calling reset manually. Need to think about a better way

x/programs/examples/counter_test.go
```diff
@@ -79,71 +81,53 @@ func TestCounterProgram(t *testing.T) {
...
 	// validate counter at 0
 	result, err = rt.Call(ctx, "get_value", programIDPtr, alicePtr)
+	reentrancyGaurd.Reset()
 	require.NoError(err)
 	require.Equal(int64(0), result[0])
 
-	// initialize second runtime to create second counter program with an empty
-	// meter.
-	rt2 := runtime.New(log, eng, imports, cfg)
-	err = rt2.Initialize(ctx, wasmBytes, engine.NoUnits)
```
samliok, line 93: instantiating 2 runtimes is not needed

x/programs/examples/counter_test.go
```diff
@@ -160,22 +144,29 @@ func TestCounterProgram(t *testing.T) {
...
-	balance, err = rt.Meter().GetBalance()
+	// balance, err = rt.Meter().GetBalance()
+	// require.NoError(err)
+	// require.Greater(balance, uint64(0))
+
+	_, err = rt.Call(ctx, "reentrance_example", caller, target, caller, maxUnitsProgramToProgramPtr)
+	reentrancyGaurd.Reset()
+
 	require.NoError(err)
-	require.Greater(balance, uint64(0))
```
samliok, line 180: re-add this

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -59,6 +65,9 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
         #input
         #[no_mangle]
         pub extern "C" fn #new_name(#(#param_names: #param_types), *) #return_type {
+            if !#is_reentrant && !wasmlanche_sdk::host::enter_program(#program_param, #program_name).expect("error calling enter_program host function") {
+                panic!("Reentrancy not allowed for this function");
```
samliok, line 69: another option is for the engine to panic during the host function call

richardpringle (commented):
Bunch of comments. I think you'll agree with a couple of little suggested changes, but then some of the other comments are up for discussion. Once discussed, you can re-request a review, and I'll *likely* approve.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -46,6 +46,20 @@ fn inc_external(_: Program, target: Program, max_units: i64, of: Address, amount
         .unwrap()
 }
 
+/// multiply is an example of a reentrant function. It will call inc with value `x`, `y` times.
+#[public(reentrant)]
```
richardpringle, line 50: this is nice, 🥇

x/programs/examples/imports/program/program.go
```diff
@@ -215,6 +238,45 @@ func getCallArgs(ctx context.Context, memory *program.Memory, buffer []byte, pro
 	return args, nil
 }
 
+// enterProgram tries to enter [function] of [programID].
+// Returns 0 if [function] has already been entered, 1 otherwise.
```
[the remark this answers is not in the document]
samliok: updated the function to return an error if we are reentering. This way we can error directly inside the host, and not in wasm.

x/programs/engine/reentrancy.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+type reentrancyMap map[ids.ID]map[string]bool
+
+// ReentrancyGuard keeps track of which methods have been during execution by the runtime.
+type ReentrancyGuard struct {
+	// In the future, every call context might need its own reentrancyMap
+	// with fine grain locking across call contexts if we want to parellellize program calls.
```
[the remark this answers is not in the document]
samliok: yea, the context should have its own guard. I'm not sure how to store the reentrancy guard in the call-context tho. Going to look into it further

x/programs/engine/reentrancy.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package engine
+
+import (
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+type reentrancyMap map[ids.ID]map[string]bool
```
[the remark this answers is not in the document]
samliok: right

x/programs/rust/sdk_macros/src/lib.rs
samliok: yea agree, updated to abort in the host.

x/programs/program/reentrancy.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+var ErrReentrancy = errors.New("reentrant call not allowed")
+
+type reentrancyMap map[ids.ID]map[string]bool
```
richardpringle, line 15: 🤔, on a re-entrant call, do instantiate a new call context? 

If we have program A that calls program B that calls back into A, do we instantiate two As? 

If so, we probably shouldn't do that? 

I think this might help with the whole context thing. We shouldn't really need a map of maps, we should only need a map attached to the call context and/or program or whatever. 

Do I know how to do that? Absolutely not haha. It's definitely possible though. 

Also curious about @dboehm-avalabs's thoughts here.

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -46,19 +46,41 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
         }
         panic!("Unsupported function parameter format.");
     });
-
     let (param_names, param_types): (Vec<_>, Vec<_>) = full_params.unzip();
+
+    let program_param = &param_names[0];
+    // convert to a string
+    let program_name = name.to_string();
```
richardpringle, line 53: nit: this is the function name isn't it?

x/programs/program/reentrancy.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Note, you can bypass this check by calling a different method for the same program which call the intended no reenter method.
+// ex. Program A has method no_reenter and reenter. Reenter calls method no_reenter.
+// If an external call wants to call no_reenter, it can call reenter instead and will bypass the reentrancy check.
+// The second call wont be an external program call, but it will be a reentrant call and is not handled.
+
+// Enter returns if we have never entered this function before, errors otherwise.
+// If the function has not been entered before, we set the visited flag.
+// We can return an error because only non-reentrant functions will call this method.
+func (r *ReentrancyGuard) Enter(id ids.ID, fn string) error {
```
dboehm-avalabs, line 43: Could we make this a little more abstract?  It might be nice if there were a reentrancyGuard interface with the param of Enter either being generic or just an interface so we could have different functionality without having to add a ton of plumbing.

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -46,19 +46,41 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
+        ),
+    };
+    let rentrancy_check = if !is_reentrant {
+        quote! {
+            // enter_program will error in the host if the program is already entered
+            wasmlanche_sdk::host::enter_program(#program_param, #program_name).expect("error calling enter_program host function");
+        }
+    } else {
+        quote! {}
+    };
```
richardpringle, line 76: I think this is fine the way it is now, but this whole thing might be a lot simpler if we do the locking in the guest and only track the active-call count in the host. We can even pass that in directly as an argument instead of needing to allocate in the guest. I think that would really simplify the host code.

richardpringle (approved):
I think we need to iterate on this still, but that doesn't mean we can't land this now. Checkout my comments anyway, though.

x/programs/rust/sdk_macros/src/lib.rs
samliok: yep

richardpringle:
```suggestion
    let function_name = name.to_string();
```
?

x/programs/program/reentrancy.go
samliok: yea, I agree it should be attached on a per context level. However, I'm not sure how to attach it... Wasmtime requires this function signature for imported functions
`fn := func(caller *wasmtime.Caller, args []wasmtime.Val) ([]wasmtime.Val, *wasmtime.Trap) {`

ideally we could pass along our reentrancy data into `caller`

richardpringle: I case it's not clear from the fact that I approved this, I'm not suggesting you make any of the changes from this approval now. I left another comment somewhere saying that we could actually do ALL of the locking, etc, in the guest, and there would be fewer host calls. 

Of course, that's contingent on us not spinning multiple instances of the same program...

dboehm-avalabs (approved):

richardpringle (approved):

PR #656, opened 2023-12-13
[storage] Add serializeData helper function

martineckardt:
### Problem

I've found the Set functions in the storage.go somewhat hard to understand:

```go
func SetAsset(
	ctx context.Context,
	mu state.Mutable,
	asset ids.ID,
	symbol []byte,
	decimals uint8,
	metadata []byte,
	supply uint64,
	owner codec.Address,
	warp bool,
) error {
	k := AssetKey(asset)
	symbolLen := len(symbol)
	metadataLen := len(metadata)
	v := make([]byte, consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len+metadataLen+consts.Uint64Len+codec.AddressLen+1)
	binary.BigEndian.PutUint16(v, uint16(symbolLen))
	copy(v[consts.Uint16Len:], symbol)
	v[consts.Uint16Len+symbolLen] = decimals
	binary.BigEndian.PutUint16(v[consts.Uint16Len+symbolLen+consts.Uint8Len:], uint16(metadataLen))
	copy(v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len:], metadata)
	binary.BigEndian.PutUint64(v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len+metadataLen:], supply)
	copy(v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len+metadataLen+consts.Uint64Len:], owner[:])
	b := byte(0x0)
	if warp {
		b = 0x1
	}
	v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len+metadataLen+consts.Uint64Len+codec.AddressLen] = b
	return mu.Insert(ctx, k, v)
}
```

Therefore I was thinking about if there is a way to make them more readable and easier to maintain.

### Possible solution

Introduce a new helper function:

```go
func serializeData(data ...interface{}) ([]byte, error) {
	totalLen := 0
	for _, d := range data {
		switch value := d.(type) {
		case byte: // includes uint8
			totalLen += consts.ByteLen
		case uint16:
			totalLen += consts.Uint16Len
		case uint64:
			totalLen += consts.Uint64Len
		case []byte:
			totalLen += len(value)
		default:
			return nil, fmt.Errorf("unsupported type: %T", value)
		}
	}

	v := make([]byte, totalLen)
	offset := 0
	for _, d := range data {
		switch value := d.(type) {
		case byte: // includes uint8
[...]

PR #687, opened 2024-01-18
[storage] Utilize codec packer

martineckardt:
Another approach for making the storage functions more readable in the reference implementations as outlined in #656

This approach uses codec.packer instead of the inefficient helper function:

```go
func SetAsset(
	ctx context.Context,
	mu state.Mutable,
	asset ids.ID,
	symbol []byte,
	decimals uint8,
	metadata []byte,
	supply uint64,
	owner codec.Address,
	warp bool,
) error {
	k := AssetKey(asset)

	p := codec.NewWriter(0, consts.MaxInt)

	p.PackBytes(symbol)
	p.PackByte(decimals)
	p.PackBytes(metadata)
	p.PackUint64(supply)
	p.PackAddress(owner)
	p.PackBool(warp)

	if err := p.Err(); err != nil {
		return err
	}

	return mu.Insert(ctx, k, p.Bytes())
}
```

The unpacking is equally elegant:

```go
func innerGetAsset(
	v []byte,
	err error,
) (bool, []byte, uint8, []byte, uint64, codec.Address, bool, error) {
	if errors.Is(err, database.ErrNotFound) {
		return false, nil, 0, nil, 0, codec.EmptyAddress, false, nil
	}
	if err != nil {
		return false, nil, 0, nil, 0, codec.EmptyAddress, false, err
	}
	p := codec.NewReader(v, consts.MaxInt)

	var symbol []byte
	var metadata []byte
	var addr codec.Address

	p.UnpackBytes(consts.MaxInt, false, &symbol)
	decimals := p.UnpackByte()
	p.UnpackBytes(consts.MaxInt, false, &metadata)
	supply := p.UnpackUint64(false)
	p.UnpackAddress(true, &addr)
	warp := p.UnpackBool()

	if err := p.Err(); err != nil {
		return false, nil, 0, nil, 0, codec.EmptyAddress, false, err
	}

	return true, symbol, decimals, metadata, supply, addr, warp, nil
}

```

For the unpacking I had to adjust the packer.UnpackAddress method. The previous implementation would throw an error if the EmptyAddress was unpacked. However, the integration test had a test case with an empty owner. Therefore, I introduced a new parameter canBeEmpty.

```go
func (p *Packer) UnpackAddress(canBeEmpty bool, dest *Address) {
	copy((*dest)[:], p.p.UnpackFixedBytes(AddressLen))
	if !canBeEmpty && *dest == EmptyAddress {
		p.addErr(fmt.Errorf("%w: Address field is not populated", ErrFieldNotPopulated))
	}
}
```


martineckardt:
### Benchmarks

#### Bottom line
The previous implementation produces the following result:
```
BenchmarkSerializeOriginal-12           48373767                22.17 ns/op           64 B/op          1 alloc
```

#### New Implementation
```
BenchmarkSerializeWithPacker-12         11368502               101.9 ns/op           120 B/op          4 alloc
```
We see that the new implementation performs significantly worse than the original implementation and only slightly better than the helper function proposed before.

#### Optimizations
One thing I tried is initializing the writer with an estimate of the length of the byte array:
```go
p := codec.NewWriter(0, consts.MaxInt)
```
=>
```go
symbolLen := len(symbol)
metadataLen := len(metadata)
p := codec.NewWriter(consts.Uint32Len+symbolLen+consts.Uint8Len+consts.Uint32Len+metadataLen+consts.Uint64Len+codec.AddressLen+consts.ByteLen, consts.MaxInt)
```
We can cut the ns/op by 60%
```
BenchmarkSerializeWithPacker-12         29186575                39.35 ns/op           64 B/op          1 alloc
```
#### Conclusion
Is this acceptable or are there any other ideas for optimizations?

patrick-ogrady:
image

Nice! I think allocating maxInt memory would definitely take a good bit of time lol.

PR #688, opened 2024-01-18
[WIP] Update Simulator to begin running again

dboehm-avalabs:
TODO: Add CI to prevent this from happing again

Update the simulator code to the latest interfaces and fix associated param writing code.


x/programs/cmd/simulator/cmd/config.go
richardpringle, line 1: Have we tried using `uint64` instead of `int64`? From my understanding, any 32-bit or 64-bit parameter is allowed. You can definitely do this with the Rust version of wasmtime, not sure if the cgo glue prevents this.

x/programs/cmd/simulator/cmd/plan.go
richardpringle, line 1: Haha what a fun job trying to find the renames like this. 🏆 for your efforts!

x/programs/program/errors.go
```diff
@@ -63,20 +64,18 @@ func HandleTrapError(err error) error {
...
+	case 8:
 		trapErr = ErrTrapBadConversionToInteger
-	case 10:
+	case 9:
 		trapErr = ErrTrapUnreachableCodeReached
-	case 11:
+	case 10:
 		trapErr = ErrTrapInterrupt
-	case 12:
+	case 11:
```
richardpringle, line 78: 😥

richardpringle (commented):
Looks good so far

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -282,49 +281,49 @@ func (c *runCmd) createCallParams(ctx context.Context, db state.Immutable, param
...
 			val, ok := param.Value.(string)
 			if !ok {
 				return nil, fmt.Errorf("%w: %s", ErrFailedParamTypeCast, param.Type)
 			}
 
-			var key string
+			key := val
 			// get named public key from db
 			pk, ok, err := storage.GetPublicKey(ctx, db, val)
 			if !ok {
```
samliok, line 300: since we are setting `key := val` above we can change the if-else statement to just
```go
if ok {			
  key = string(pk[:])
}
```

x/programs/program/memory.go
```diff
@@ -164,31 +167,52 @@ type CallParam struct {
...
+			smartPtr, err := BytesToSmartPtr(v[:], m)
 			if err != nil {
 				return nil, err
 			}
-			params = append(params, uint64(ptr))
-		case int:
-			if v < 0 {
-				return nil, fmt.Errorf("failed to write param: %w", ErrUnderflow)
+			params = append(params, smartPtr)
+		case string:
```
samliok, line 180: why handle strings and ids.ID differently here? Can they just be passed in as SmartPtrs?

x/programs/program/memory.go
```diff
@@ -164,31 +167,52 @@ type CallParam struct {
...
+			params = append(params, ptr)
 		}
 	}
 
 	return params, nil
 }
 
+// SerializeParameter serializes [obj] using Borsh
+func serializeParameter(obj interface{}) ([]byte, error) {
+	bytes, err := borsh.Serialize(obj)
```
samliok, line 202: I think the serialization should be done outside of `memory.go`. That way we can abstract away the serializer(borsh) and just require all params to just be bytes/SmartPtrs.

dboehm-avalabs: I'll make this same change when I redo the write params.  I'd like to move all of that logic out, but I'd like to do it in the followup.

dboehm-avalabs: for now, moving this out of the program module and into simulator, since it isn't used anywhere else.

x/programs/program/memory.go
dboehm-avalabs: for now, moving this out of the program module and into simulator, since it isn't used anywhere else.

.github/workflows/simulator-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+          components: rustfmt, clippy
+          override: true
+      - name: Run token simulator test
+        working-directory: ./x/programs/rust/examples/token
+        shell: bash
+        run: scripts/tests.simulator.sh
+        env:
+          MODE: "full-test"
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
richardpringle, line 40: cool

richardpringle (dismissed):

.github/workflows/simulator-tests.yml
patrick-ogrady: that took forever to get right 😉

.github/workflows/simulator-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  simulator-tests:
+    runs-on:
+      labels: ubuntu-20.04-32
```
patrick-ogrady, line 15: If we are always running this, I think we should use `ubuntu-latest` or we should put this behind `run all ci`: https://github.com/ava-labs/hypersdk/blob/db1a98e691fd47c4db39650adbc72ff694470517/.github/workflows/tokenvm-release.yml#L18

x/programs/cmd/simulator/simulator
patrick-ogrady, line 1: Did you mean to commit this?

patrick-ogrady: I think you'll want to add this to `.gitignore`

x/programs/cmd/simulator/vm/config/config.go
```diff
@@ -87,17 +86,14 @@ func New(nodeID ids.NodeID, b []byte) (*Config, error) {
 func (c *Config) setDefault() {
 	c.LogLevel = c.Config.GetLogLevel()
 	c.MempoolSize = c.Config.GetMempoolSize()
-	c.MempoolPayerSize = c.Config.GetMempoolPayerSize()
 	c.StateSyncServerDelay = c.Config.GetStateSyncServerDelay()
 	c.StreamingBacklogSize = c.Config.GetStreamingBacklogSize()
-	c.VerifySignatures = c.Config.GetVerifySignatures()
```
patrick-ogrady, line 93: Why remove this one?

dboehm-avalabs: it wasn't being used as far as I could tell.

dboehm-avalabs: more specifically https://github.com/ava-labs/hypersdk/blob/main/vm/dependencies.go#L28 doesn't seem to have GetVerifySignatures() anymore.

x/programs/cmd/simulator/simulator
dboehm-avalabs: you are correct, snuck in

x/programs/cmd/simulator/vm/config/config.go
patrick-ogrady: It is now `GetVerifyAuth`: https://github.com/ava-labs/hypersdk/blob/0bd179ff03e4cc7672faaa69502d06a73d1d9fb5/vm/dependencies.go#L32

patrick-ogrady: I'll update this rq

Issue #503, opened 2023-10-02
[chain] Add tx root to each block header


patrick-ogrady:
Add tx root of successful txs to each block so that inclusion can be verified just by processing headers (see Bitcoin SPV).

Alternatively, nodes could produce signatures over accepted txs periodically (may require extra latency) so anyone can verify inclusion with just p-chain syncing.

patrick-ogrady:
Could accomplish this by exporting a "proof" warp message so people that don't need this don't need to pay more.

patrick-ogrady:
To avoid interference with warp (where we want to prove + warp), we could add a Boolean to txs that makes it "prove-able" and saves txID to state.

patrick-ogrady:
May also want to make exported warp messages per block a root too (so any network can verify arbitrary amounts of data with a single multi-signature).

patrick-ogrady:
Related (maybe better approach): https://github.com/ava-labs/hypersdk/issues/504

patrick-ogrady:
related: #950

Issue #412, opened 2023-08-24
[chain] Hash Block Based on Hash Trie of Transactions + State Root + Results Root + Warp Outputs Root


patrick-ogrady:
Otherwise, we'd need to provide all transactions to prove anything about a block.

bianyuanop:
Hey Patrick, 

I'm Chan from NodeKit, I'm taking this with #503. 

I'm curious about that does  **_Hash Block Based on Hash Trie of Transactions + State Root + Results Root + Warp Outputs Root_** mean concatenation of Root of transaction trie with all other roots or it means using a single trie and putting transactions and all other roots in the trie to calculate the root as hash of the block

patrick-ogrady:
Related: #950

PR #689, opened 2024-01-19
Bloch hash generation by merkle trie root of transactions, results and state root

bianyuanop:
### Context
This PR covers #412 and #503, the goal is to generate block hash _Based on Hash Trie of Transactions + State Root + Results Root + Warp Outputs Root_ . Several things aren't clear to me 
1. Warp Outputs are not found under the `builder.go` that seems the outputs from warp messages are results of transactions if `warpMessage` field is not nil of a transaction
2. How to allow users to verify transactions/results, two ways i can think of depending on the current implementation, we either make modification on the existing merkletree implementation that let it return proofs in order of given `TrieItems` or we hash the transactions/results and letting the `TrieItem.x` to be the hash of each item then double hashing won't be needed. 
### Testing
e2e tests under morpheusvm and tokenvm are passed 
### Dependencies 
The merkletrie implementation borrowed from [cbergoon/merkletree](https://github.com/cbergoon/merkletree), which is a pure implementation so no more dependencies are introduced except itself


chain/block.go
```diff
@@ -284,12 +312,32 @@ func (b *StatelessBlock) initializeBuilt(
...
 		if tx.WarpMessage != nil {
 			b.containsWarp = true
 		}
 	}
+	// transaction results
+	for _, result := range b.results {
+		items2prove = append(items2prove, TrieItem{x: result.Output})
+	}
+
+	trie, err := merkletree.NewTree(items2prove)
```
patrick-ogrady, line 333: We should use a stateless instance of the `MerkleDB` to build this trie, rather than this external package.

chain/block.go
```diff
@@ -35,12 +39,36 @@ var (
...
+}
+
 type StatefulBlock struct {
 	Prnt   ids.ID `json:"parent"`
 	Tmstmp int64  `json:"timestamp"`
 	Hght   uint64 `json:"height"`
 
-	Txs []*Transaction `json:"txs"`
+	Txs       []*Transaction `json:"txs"`
+	BlockHash []byte         `json:"txsRoot"`
```
patrick-ogrady, line 71: nit: this should be `TxsRoot` not `BlockHash`. The hash of any block is returned by `ID()`.

chain/block.go
bianyuanop: Could a `MerkleDB` backed by a `memdb` necessarily be a stateless `MerkleDB`?

bianyuanop:
New root generation with stateless merkledb was updated. One thing I'm not sure that will the `tx.Bytes()` contains info about actor and transaction content itself to distinguish it from other transactions so the keys generated will be different. 

[block.go#L313](https://github.com/bianyuanop/hypersdk/blob/main/chain/block.go#L313)

chain/block.go
patrick-ogrady: Yeah, I'm looping in @dboehm-avalabs to assist here 👍

dboehm-avalabs: Looks like it has already been updated, but let me know if you have any additional questions.

chain/block.go
```diff
@@ -290,6 +294,46 @@ func (b *StatelessBlock) initializeBuilt(
 			b.containsWarp = true
 		}
 	}
+
+	// transaction hash generation
+	db, err := merkledb.New(ctx, memdb.New(), merkledb.Config{
```
patrick-ogrady, line 299: It would be great to create a helper for this inside of `HyperSDK` so this root generation technique could be used more easily in other places.

patrick-ogrady: But this is the right direction 👍

bianyuanop:
The root generation has been made into a function, I put it under `utils/utils.go` since it seems unmatch if i put it under `crypto`. e2e tests under two hypervms were passed

patrick-ogrady:
> The root generation has been made into a function, I put it under `utils/utils.go` since it seems unmatch if i put it under `crypto`. e2e tests under two hypervms were passed

could you move it to its own package called `merkle`? We'll eventually want to write some tests/benchmarks around its performance (say if we added 10k transactions)

utils/utils.go
```diff
@@ -116,3 +122,43 @@ func LoadBytes(filename string, expectedSize int) ([]byte, error) {
 	}
 	return bytes, nil
 }
+
+// Generate merkle root for a set of items
+func GenerateMerkleRoot(ctx context.Context, tracer trace.Tracer, merkleItems [][]byte, consumeBytes bool) ([]byte, merkledb.MerkleDB, error) {
```
patrick-ogrady, line 127: ❤️

patrick-ogrady: @dboehm-avalabs it would be great if you could confirm correctness here.

chain/block.go
```diff
@@ -290,6 +291,23 @@ func (b *StatelessBlock) initializeBuilt(
 			b.containsWarp = true
 		}
 	}
+
+	// transaction hash generation
+	merkleItems := make([][]byte, 0, len(b.Txs)+len(b.results))
+	for _, tx := range b.Txs {
+		merkleItems = append(merkleItems, tx.Bytes())
```
patrick-ogrady, line 298: Would it make sense to use `txID` as key and then put something useful in the value?

chain/block.go
```diff
@@ -290,6 +291,23 @@ func (b *StatelessBlock) initializeBuilt(
...
+	}
+	for _, result := range b.results {
+		merkleItems = append(merkleItems, result.Output)
```
patrick-ogrady, line 301: Not sure what just storing hashes of outputs of results will provide? It seems like it should at least be associated with a tx...

hypersdk.code-workspace
patrick-ogrady, line 1: Please add `*.code-workspace` to `.gitignore` and remove this from the PR

chain/block.go
bianyuanop: Yea, one way to do that is to concanate transaction with corresponding output then feed into merkle tree.

hypersdk.code-workspace
bianyuanop: Sure

chain/block.go
bianyuanop: will the type of transactions have distinct id for different actors? nor there will be collision

bianyuanop:
Root generation function has been moved to `merkle` package. Now the `TxsRoot` are generated by root of merkle trie with items in format [txID + result]

iFrostizz:
Having the transaction hashes concatenated with the transaction results doesn't look very idiomatic to me, and at the same time if we provide results in a flat data structure then it will do a collision. Do we really need to be able to prove that the result of a transaction given the current state is correct using the merkle tree ? If it's something that people really want to do, then they could execute the transactions in the block sequentially to find that their execution results matches the one stored in the block.

Issue #562, opened 2023-10-16
[cli] Add argument in `chain watch` to start stream at an earlier block height


patrick-ogrady:
Blocker: #555 
Blocker: #556

Issue #556, opened 2023-10-14
[rpc/storage] Allow for optionally storing `block.Results` on-disk for stream replay after restart


patrick-ogrady:
To offer a production-ready solution, we need to guarantee there won't be a gap in block history, even if there is an unclean shutdown. This means we need to persist the block during `block.Accept` (before updating the last accepted block on-disk and/or committing state).

patrick-ogrady:
We need to decide if we want to utilize the otherwise stored blocks on-disk or to just store the entire message on-disk (in case the block on-disk representation changes).

patrick-ogrady:
> We need to decide if we want to utilize the otherwise stored blocks on-disk or to just store the entire message on-disk (in case the block on-disk representation changes).

We should not assume anything about how the block is stored elsewhere on the node and should store the entire message on-disk separately.

manojkgorle:
Hey, I am working on this issue. I have a few questions.

1) Should `txIDs` be stored along with `block.Results`?

> We need to decide if we want to utilize the otherwise stored blocks on-disk or to just store the entire message on-disk (in case the block on-disk representation changes).

2) what does the *message* here refer to? 

3) should the number of `block.Results` stored be limited? or make it tunable?

I have something in mind: 

- optional storage can be tuned by config params passed during `vm.Initialize`.  

- a new DB instance to store `block.Results`, similar to how blocks are stored.

manojkgorle:
> Hey @patrick-ogrady, should the number of `block.Results` stored be limited to `acceptedBlockWindow`? or make it tunable?

it won't make sense if `block.Results` are stored with out `block` to the disk.

patrick-ogrady:
Related: #961

Issue #555, opened 2023-10-14
[rpc] Allow caller to specify a start block for streaming in WSS


patrick-ogrady:
Blocked by: #556 

This would allow an indexer to start where they left off in the stream (assuming the node has the block in its history).

PR #692, opened 2024-01-20
[rpc] optionally storing `block.Results` on disk, start stream from a specified block

manojkgorle:
This PR covers #556, #555, #562

Currently `block.Bytes()`  are stored to disk. now allowing to optionally store `block.Results`  to disk.

`block.Results` & `block.FeeManager` are now stored in `VM.vmDB`.

- [x] optionally store `block.Results` & `block.FeeManager` during `block.Accept` before committing to disk or updating the last accepted block.
- [x] limit `block.Results` stored on disk to same number of `blocks` stored on disk 
- [x] remove `block.Results` for expired blocks.
- [x] getter for fetching `block.Results` & `block.FeeManager` from disk.
- [x] Allow caller to specify a start block to stream if the block is stored on disk by node
- [x] [cli] Added arguments in `chain watch` to start stream at an earlier block height
- [x] update `scripts/run.sh` in `morpheusvm`  and `tokenvm`.

### DIFF
--- a/chain/block.go
+++ b/chain/block.go
@@ -730,6 +730,13 @@ func (b *StatelessBlock) Accept(ctx context.Context) error {
 		}
 	}
 
+	// optionally store [block.Results] to disk
+	if b.vm.GetStoreBlockResultsOnDisk() {
+		if err := b.vm.StoreBlockResultsOnDisk(b); err != nil {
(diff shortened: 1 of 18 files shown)


bianyuanop (commented):
Should delete `expiryHeight` here, not `blk.Height()`

Issue #690, opened 2024-01-19
[executor] improving parallelism in the HyperSDK


wlawt:
In an effort to improve/modify how parallelism is handled in the HyperSDK, this issue will track a series of smaller issues/PRs that would need to be completed.

- [x] #628 
- [x] #701
- [x] #712

wlawt:
Related: #561

wlawt:
Related: #561

wlawt:
Related: #709

Issue #561, opened 2023-10-16
[chain] Add ability to have Read-Only State Keys


patrick-ogrady:
This will allow multiple HyperSDK Programs to read the contract code concurrently (which would otherwise be a conflict key and require serial execution -> It should only require serial execution if the code is being modified).

This will also lead to lower "max fees" and is good for the HyperSDK regardless to provide.

wlawt:
@patrick-ogrady we can probably close this issue now right?

PR #693, opened 2024-01-20
[chain] allow StateKeys to be specified with a Permission (Read/Allocate/Write)

### LINKED ISSUE #628: [chain] Separate `[Allocate/Write/Read]StateKeys`
This would allow users to have a lower balance to issue transactions (otherwise charged as if every key is an allocation) and would allow us to parallelize transactions that only read from the same state (not write).

wlawt:
Closes #628 

### Overview

This PR introduces the ability for a `Transaction` to specify a key type (Permission) for StateKeys. Currently, the permissions include Read, Write, and RWrite, but with the ability to easily add additional permissions in the future (i.e., Allocate). By default, no permissions are set. To add StateKeys to the `state.Keys` map, use `AddKeyPermission` to prevent new insertions overriding the original permission value.

At this point in time, it should be noted that the Execution Engine still handles these keys the same way, regardless of the key type specified. This refactor is the first step in changing the way parallelism is done in the HyperSDK, which is tracked by #690.  

### Test Plan

- Integrated into TokenVM
- Integrated into MorpheusVM
- Added tests for setting and checking permissions in `tstate_test.go`
- Added tests for GetValue/Insert/Delete with different key permissions in `tstate_test.go`
- Ensure union of permission bits for insertion of a key that already exists. Tests are in `tstate_test.go`

To test, run `./scripts/tests.integration.sh` within each VM example and `go test -race` for the unit tests.

### Review Guide

- `chain/`: where most of the refactoring happened
- `examples/`: where the testing happened 

### TODO

- [x] Refactor Mocks
- [x] Update README
- [x] Update tstate and executor tests


wlawt:
Related: #561

chain/transaction.go
```diff
@@ -130,39 +132,109 @@ func (t *Transaction) Expiry() int64 { return t.Base.Timestamp }
...
 	}
 
 	// Cache keys if called again
-	t.stateKeys = stateKeys
-	return stateKeys, nil
+	t.readKeys = readKeys
+	return readKeys, nil
+}
+
+func (t *Transaction) WriteKeys(stateMapping StateManager) (set.Set[string], error) {
```
wlawt, line 170: `ReadKeys`, `WriteKeys`, and `RWKeys` all have similar functionality, is there a way to prevent duplicate code among these functions?

chain/processor.go
```diff
@@ -52,11 +53,26 @@ func (b *StatelessBlock) Execute(
...
+		writeKeys, err := tx.WriteKeys(sm)
+		if err != nil {
+			e.Stop()
+			return nil, nil, err
+		}
+		rwKeys, err := tx.RWKeys(sm)
+		if err != nil {
+			e.Stop()
+			return nil, nil, err
+		}
```
wlawt, line 70: Not sure if there's a better way to refactor this

examples/tokenvm/actions/import_asset.go
```diff
@@ -38,7 +38,15 @@ func (*ImportAsset) GetTypeID() uint8 {
...
-func (i *ImportAsset) StateKeys(auth chain.Auth, _ ids.ID) []string {
+func (i *ImportAsset) ReadKeys(auth chain.Auth, _ ids.ID) []string {
+	return []string{}
+}
+
+func (i *ImportAsset) WriteKeys(auth chain.Auth, _ ids.ID) []string {
+	return []string{}
+}
+
+func (i *ImportAsset) RWKeys(auth chain.Auth, _ ids.ID) []string {
```
wlawt, line 49: Not sure if all these keys are `RWKeys`

examples/tokenvm/actions/export_asset.go
```diff
@@ -37,7 +37,15 @@ func (*ExportAsset) GetTypeID() uint8 {
...
-func (e *ExportAsset) StateKeys(auth chain.Auth, _ ids.ID) []string {
+func (e *ExportAsset) ReadKeys(auth chain.Auth, _ ids.ID) []string {
+	return []string{}
+}
+
+func (e *ExportAsset) WriteKeys(auth chain.Auth, _ ids.ID) []string {
+	return []string{}
+}
+
+func (e *ExportAsset) RWKeys(auth chain.Auth, _ ids.ID) []string {
```
wlawt, line 48: Not sure if all these keys are `RWKeys`

chain/builder.go
```diff
@@ -181,14 +181,26 @@ func BuildBlock(
 				continue
 			}
 
-			stateKeys, err := tx.StateKeys(sm)
+			readKeys, err := tx.ReadKeys(sm)
```
patrick-ogrady, line 184: I think it may be better to return a map of `StateKeys` where the value in the map is a "mode" (Read/Write/RWrite/Allocate/etc). This provides future flexibility as well to keep adding new modes that the execution engine can use.

chain/transaction.go
patrick-ogrady: See my comment here: https://github.com/ava-labs/hypersdk/pull/693/files#r1460969811

examples/tokenvm/actions/export_asset.go
```diff
@@ -37,18 +37,18 @@ func (*ExportAsset) GetTypeID() uint8 {
...
-func (e *ExportAsset) StateKeys(actor codec.Address, _ ids.ID) []string {
-	if e.Return {
-		return []string{
-			string(storage.AssetKey(e.Asset)),
-			string(storage.BalanceKey(actor, e.Asset)),
-		}
+func (e *ExportAsset) StateKeys(actor codec.Address, _ ids.ID) map[string]chain.Mode {
+	result := map[string]chain.Mode{
+		string(storage.AssetKey(e.Asset)):          chain.RWrite,
+		string(storage.BalanceKey(actor, e.Asset)): chain.RWrite,
```
wlawt, line 43: Not sure these are all `RWrite`

examples/tokenvm/actions/import_asset.go
```diff
@@ -38,38 +38,38 @@ func (*ImportAsset) GetTypeID() uint8 {
 	return importAssetID
 }
 
-func (i *ImportAsset) StateKeys(actor codec.Address, _ ids.ID) []string {
+func (i *ImportAsset) StateKeys(actor codec.Address, _ ids.ID) map[string]chain.Mode {
```
wlawt, line 41: Not sure if all these keys are RWrite here as well

chain/dependencies.go
```diff
@@ -25,6 +25,7 @@ import (
 type (
 	ActionRegistry *codec.TypeParser[Action, *warp.Message, bool]
 	AuthRegistry   *codec.TypeParser[Auth, *warp.Message, bool]
+	Mode           int
```
patrick-ogrady, line 28: nit: should be unsigned imo

patrick-ogrady: probably a uint8 is more than sufficient

chain/dependencies.go
```diff
@@ -176,7 +177,9 @@ type FeeHandler interface {
 	//
 	// All keys specified must be suffixed with the number of chunks that could ever be read from that
 	// key (formatted as a big-endian uint16). This is used to automatically calculate storage usage.
-	SponsorStateKeys(addr codec.Address) []string
+	//
+	// All keys must be specified with a Mode (i.e., Read, Write, RWrite).
+	SponsorStateKeys(addr codec.Address) map[string]Mode
```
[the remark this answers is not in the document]
joshua-kim: We can also just return a slice here to make the PR diff smaller, i don't think this PR requires us to return a map

wlawt:
Instead of returning a map from `tx.StateKeys(..)`, after some thought, I think keeping it as a set with a struct that holds onto the key info (name and mode) would align more with the current function signatures that rely on `StateKeys` (i.e., tstate.NewView, executor.Run)

types/types.go
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package types
```
wlawt, line 4: Not sure if this could go somewhere else, but it's to avoid circular dependency when using it in `executor`

tstate/tstate_view.go
```diff
@@ -49,14 +52,30 @@ type TStateView struct {
 	writes      map[string]uint16
 }
 
-func (ts *TState) NewView(scope set.Set[string], storage map[string][]byte) *TStateView {
+func (ts *TState) NewView(scope set.Set[types.Key], storage map[string][]byte) *TStateView {
+	// TODO: set length to be the number of keys for each type
+	readScope := set.NewSet[string](len(scope))
+	writeScope := set.NewSet[string](len(scope))
+	rwScope := set.NewSet[string](len(scope))
```
wlawt, line 59: Not sure if it's better to have the scopes already separated and then passed into NewView or done within NewView (or if it makes a difference)

chain/builder.go
```diff
@@ -181,14 +181,18 @@ func BuildBlock(
...
-			stateKeys, err := tx.StateKeys(sm)
+			stateKeysWithMode, err := tx.StateKeys(sm)
 			if err != nil {
 				// Drop bad transaction and continue
 				//
 				// This should not happen because we check this before
 				// adding a transaction to the mempool.
 				continue
 			}
+			stateKeys := set.NewSet[string](len(stateKeysWithMode))
```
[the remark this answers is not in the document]
wlawt: I ended up removing this and instead passed a `set[Key]` into Executor and TState so we can preserve the key mode info. this would allow the Executor to use these modes and also the TState can prefetch the keys according to the mode

types/types.go
wlawt: Moved into `state` package

chain/block.go
```diff
@@ -609,7 +609,12 @@ func (b *StatelessBlock) innerVerify(ctx context.Context, vctx VerifyContext) er
 	heightKeyStr := string(heightKey)
 	timestampKeyStr := string(timestampKey)
 	feeKeyStr := string(feeKey)
-	tsv := ts.NewView(set.Of(heightKeyStr, timestampKeyStr, feeKeyStr), map[string][]byte{
+
+	heightKeyKey := state.NewKey(heightKeyStr, state.Read, state.Write)
+	timestampKeyKey := state.NewKey(timestampKeyStr, state.Read, state.Write)
+	feeKeyKey := state.NewKey(feeKeyStr, state.Read, state.Write)
```
joshua-kim, line 615: The naming for these variables is weird and it's only referenced in the construction of `NewView` so i would just inline them all like 

```
tsv := ts.NewView(
    set.Of(
	state.NewKey(heightKeyStr, state.Read, state.Write),
	state.NewKey(timestampKeyStr, state.Read, state.Write),
	state.NewKey(feeKeyStr, state.Read, state.Write),
    )
...
)
```

chain/builder.go
```diff
@@ -228,7 +229,8 @@ func BuildBlock(
 					toLookup = make([]string, 0, len(stateKeys))
 				)
 				cacheLock.RLock()
-				for k := range stateKeys {
+				for sk := range stateKeys {
+					k := sk.Name
```
joshua-kim, line 233: Don't think we need this var

state/state.go
```diff
@@ -27,3 +27,41 @@ type View interface {
...
+type Key struct {
+	Name string
+	Mode [ModeLen]byte
+}
+
+const (
+	ModeLen = 8
+	Read    = 0
+	Write   = 1
+)
```
joshua-kim, line 40: nit: consts are usually declared at the top of the file

state/state.go
```diff
@@ -27,3 +27,41 @@ type View interface {
...
 	GetMerkleRoot(ctx context.Context) (ids.ID, error)
 }
+
+type Key struct {
+	Name string
+	Mode [ModeLen]byte
+}
+
+const (
+	ModeLen = 8
```
joshua-kim, line 37: We are probably fine with getting away with a hard-coded value/a single byte but should consider a dynamically sized []byte as an alternative

state/state.go
```diff
@@ -27,3 +27,41 @@ type View interface {
 	NewView(ctx context.Context, changes merkledb.ViewChanges) (merkledb.View, error)
 	GetMerkleRoot(ctx context.Context) (ids.ID, error)
 }
+
+type Key struct {
+	Name string
+	Mode [ModeLen]byte
```
joshua-kim, line 33: nit: name `Permission/Perm`?

state/state.go
```diff
@@ -27,3 +27,41 @@ type View interface {
...
+func NewKey(name string, bits ...int) Key {
+	var key Key
+	key.Name = name
+
+	for _, bit := range bits {
+		byteIdx := bit / 8
+		bitIdx := uint(bit % 8)
+
+		if byteIdx < ModeLen {
+			key.Mode[byteIdx] |= (1 << bitIdx)
```
joshua-kim, line 51: This will panic if we use a bit outside of the bound, we should add a bounds check + tests for these new functions

state/state.go
```diff
@@ -27,3 +27,41 @@ type View interface {
...
+		}
+	}
+
+	return key
+}
+
+func (k *Key) HasMode(mode int) bool {
```
joshua-kim, line 58: We should just make `NewKey#bits` and `HasMode#mode` have the same variable name (either bit or index or mode or perm or something like that...) since they all mean the same thing of the bit to check for consistency

state/state.go
```diff
@@ -27,3 +27,41 @@ type View interface {
...
+	byteIdx := mode / 8
+	bitIdx := uint(mode % 8)
+
+	if byteIdx >= ModeLen {
+		return false
+	}
+
+	return (k.Mode[byteIdx] & (1 << bitIdx)) != 0
```
joshua-kim, line 66: nit: reads better if we do `== 1` instead of `!= 0` imo

tstate/tstate_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/units"
```
joshua-kim, line 15: Should add unit tests for the new code in `state.go` here

tstate/tstate_view.go
```diff
@@ -39,8 +40,10 @@ type TStateView struct {
 	// operations allows for reverting state to a certain point-in-time.
 	ops []*op
 
-	// We don't differentiate between read and write scope.
-	scope        set.Set[string] // stores a list of managed keys in the TState struct
+	// stores a list of managed keys in the TState struct
+	readScope    set.Set[string]
+	writeScope   set.Set[string]
+	rwScope      set.Set[string]
```
joshua-kim, line 46: I would just have a single `map[string]Mode` of our state keys, and just look up the key's bit value in its `Mode` and get modify `checkScope` to take in a parameter of a key + the perm to look up instead of maintaining separate maps for every permutation of scopes.

tstate/tstate_view.go
```diff
@@ -49,14 +52,31 @@ type TStateView struct {
...
+	rwScope := set.NewSet[string](len(scope))
+	for k := range scope {
+		switch {
+		case k.HasMode(state.Read):
+			readScope.Add(k.Name)
+		case k.HasMode(state.Write):
+			writeScope.Add(k.Name)
+		case k.HasMode(state.Read) && k.HasMode(state.Write):
+			rwScope.Add(k.Name)
+		}
```
joshua-kim, line 68: See earlier comment, we should be able to get rid of this switch

joshua-kim (changes requested):

chain/builder.go
```diff
@@ -444,7 +447,12 @@ func BuildBlock(
 	timestampKey := TimestampKey(b.vm.StateManager().TimestampKey())
 	timestampKeyStr := string(timestampKey)
 	feeKeyStr := string(feeKey)
-	tsv := ts.NewView(set.Of(heightKeyStr, timestampKeyStr, feeKeyStr), map[string][]byte{
+
+	heightKeyKey := state.NewKey(heightKeyStr, state.Read, state.Write)
```
joshua-kim, line 451: TODO make sure these scopes are all correct

chain/processor.go
```diff
@@ -57,6 +57,7 @@ func (b *StatelessBlock) Execute(
 			e.Stop()
 			return nil, nil, err
 		}
+
```
joshua-kim, line 60: nit: Clean diff

tstate/tstate_view.go
```diff
@@ -49,14 +52,31 @@ type TStateView struct {
...
+
 	return &TStateView{
 		ts:                 ts,
 		pendingChangedKeys: make(map[string]maybe.Maybe[[]byte], len(scope)),
 
 		ops: make([]*op, 0, defaultOps),
 
-		scope:        scope,
+		readScope:    readScope,
+		writeScope:   writeScope,
```
joshua-kim, line 78: I believe we should be checking for this scope in `Insert` call, I think the current code is still checking against the read + readwrite scope map when it should actually be checking against write + readwrite. We should add a test for this to make sure scopes are honored

chain/builder.go
wlawt: When I had this convo with Patrick, my understanding is that it would be read at the beginning of the block and then written later

tstate/tstate_view.go
wlawt: I added additional tests in `tstate_test.go` where I tested it on GetValue/Insert/Remove using different key permissions (i.e., No perms/Read/Write/RW), and the scopes look like they are honored (that is, trying to Insert a key that was specified to only be read or had no permissions would error, etc.)

state/state.go
wlawt: Will look into this

state/state.go
```diff
@@ -27,3 +33,37 @@ type View interface {
...
+
+type Key struct {
+	Name string
+	// TODO: consider a dynamically sized []byte
+	Permission Permission
+}
+
+type Permission [PermissionLen]byte
+
+func NewKey(name string, permissions ...int) Key {
```
samliok, line 45: even though the methods are easy to follow, a basic comment would be nice.

state/state.go
```diff
@@ -27,3 +33,37 @@ type View interface {
...
+	var key Key
+	key.Name = name
+
+	for _, perm := range permissions {
+		// TODO: handle out of bounds
+		byteIdx := perm % 8
+
+		if byteIdx < PermissionLen {
```
samliok, line 53: don't think this if does anything? 
 `byteIdx := perm % 8` so byteIdx will always be less than `PermissionLen`.

state/state.go
```diff
@@ -27,3 +33,37 @@ type View interface {
...
+
+type Permission [PermissionLen]byte
+
+func NewKey(name string, permissions ...int) Key {
+	var key Key
+	key.Name = name
+
+	for _, perm := range permissions {
+		// TODO: handle out of bounds
+		byteIdx := perm % 8
```
samliok, line 51: is supposed to be the same as `PermissionLen`? if so, we should just reference `PermissionLen`

state/state.go
```diff
@@ -27,3 +33,37 @@ type View interface {
...
+}
+
+func (p *Permission) HasPermission(permission int) bool {
+	byteIdx := permission % 8
+
+	if byteIdx >= PermissionLen {
+		return false
+	}
+
+	return p[byteIdx] == 1
```
samliok, line 68: same comments as above

state/state.go
```diff
@@ -27,3 +33,38 @@ type View interface {
...
+// StateKey holds the name of the key and its permission (Read/Write)
+// Specifying RWrite is setting both the Read and Write bit
+type Key struct {
+	Name string
+	// TODO: consider a dynamically sized []byte
+	Permission Permission
+}
+
+// TODO: Support Alloc bit
+type Permission [PermissionLen]byte
```
patrick-ogrady, line 46: It should be sufficient to use a single byte for this, not 8 (unless you think we'll have more than 255 permissions 😉 ).

state/state.go
```diff
@@ -27,3 +33,38 @@ type View interface {
...
+
+// Returns a StateKey with the appropriate permission bits set
+// By default, no permission bits are set
+func NewKey(name string, permissions ...int) Key {
+	var key Key
+	key.Name = name
+	for _, perm := range permissions {
+		// Only set bit if it's within the byte slice
```
patrick-ogrady, line 54: You are setting bytes here, not bits.

patrick-ogrady: I agree, we should set bits.

patrick-ogrady: We have some utilities for this: https://github.com/ava-labs/avalanchego/blob/master/utils/set/bits.go

state/state.go
```diff
@@ -27,3 +33,38 @@ type View interface {
...
+	// TODO: consider a dynamically sized []byte
+	Permission Permission
+}
+
+// TODO: Support Alloc bit
+type Permission [PermissionLen]byte
+
+// Returns a StateKey with the appropriate permission bits set
+// By default, no permission bits are set
+func NewKey(name string, permissions ...int) Key {
```
patrick-ogrady, line 50: I think this is a solid use of a variadic function 👍

state/state.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/ava-labs/avalanchego/x/merkledb"
 )
 
+const (
+	Read          = 0
```
patrick-ogrady, line 14: recommend using a smaller type than `int`...would at least recommend unsigned (maybe `uint8`)

chain/transaction.go
```diff
@@ -122,18 +122,18 @@ func (t *Transaction) Expiry() int64 { return t.Base.Timestamp }
...
-	stateKeys := set.NewSet[string](len(actionKeys) + len(sponsorKeys))
-	for _, arr := range [][]string{actionKeys, sponsorKeys} {
+	stateKeys := set.NewSet[state.Key](len(actionKeys) + len(sponsorKeys))
+	for _, arr := range [][]state.Key{actionKeys, sponsorKeys} {
 		for _, k := range arr {
-			if !keys.Valid(k) {
+			if !keys.Valid(k.Name) {
 				return nil, ErrInvalidKeyValue
 			}
 			stateKeys.Add(k)
```
patrick-ogrady, line 139: How do you handle cases of conflicting keys? What happens if you insert a R/W key then later add a R key?

chain/transaction.go
```diff
@@ -144,12 +144,12 @@ func (t *Transaction) StateKeys(sm StateManager) (set.Set[string], error) {
 	if t.WarpMessage != nil {
 		p := sm.IncomingWarpKeyPrefix(t.WarpMessage.SourceChainID, t.warpID)
 		k := keys.EncodeChunks(p, MaxIncomingWarpChunks)
-		stateKeys.Add(string(k))
+		stateKeys.Add(state.NewKey(string(k), state.Read, state.Write))
```
patrick-ogrady, line 147: Should just be write

chain/transaction.go
```diff
@@ -144,12 +144,12 @@ func (t *Transaction) StateKeys(sm StateManager) (set.Set[string], error) {
...
 	}
 	if t.Action.OutputsWarpMessage() {
 		p := sm.OutgoingWarpKeyPrefix(t.id)
 		k := keys.EncodeChunks(p, MaxOutgoingWarpChunks)
-		stateKeys.Add(string(k))
+		stateKeys.Add(state.NewKey(string(k), state.Read, state.Write))
```
patrick-ogrady, line 152: Should just be write

state/state.go
```diff
@@ -27,3 +33,38 @@ type View interface {
 	NewView(ctx context.Context, changes merkledb.ViewChanges) (merkledb.View, error)
 	GetMerkleRoot(ctx context.Context) (ids.ID, error)
 }
+
+// StateKey holds the name of the key and its permission (Read/Write)
+// Specifying RWrite is setting both the Read and Write bit
+type Key struct {
+	Name string
+	// TODO: consider a dynamically sized []byte
+	Permission Permission
```
patrick-ogrady, line 42: You may want to comment that we use bits here instead of a set because we expect transactions to include this encoding in transactions.

tstate/tstate_view.go
```diff
@@ -39,8 +40,8 @@ type TStateView struct {
 	// operations allows for reverting state to a certain point-in-time.
 	ops []*op
 
-	// We don't differentiate between read and write scope.
-	scope        set.Set[string] // stores a list of managed keys in the TState struct
+	// stores a map of managed keys and its permissions in the TState struct
+	scope        map[string]state.Permission
```
patrick-ogrady, line 44: See comment about how duplicate keys are handled.

tstate/tstate_view.go
```diff
@@ -49,7 +50,11 @@ type TStateView struct {
 	writes      map[string]uint16
 }
 
-func (ts *TState) NewView(scope set.Set[string], storage map[string][]byte) *TStateView {
+func (ts *TState) NewView(scopeKeys set.Set[state.Key], storage map[string][]byte) *TStateView {
```
patrick-ogrady, line 53: The approach of using `state.Key` as the `key` in `set.Set` means you can now have duplicate (same `Name`) state keys in the same `set.Set`. I recommend using a map for this instead of defining a new struct.

state/state.go
patrick-ogrady: Related: https://github.com/ava-labs/hypersdk/issues/705

chain/transaction.go
wlawt: With the map now, I tried to have it behave like a `set`. That is, a key that already exists in the map (i.e., say with RW) will still have RW. Therefore, adding a R key later will have no change. This is done using the `Add` function that I added in `state.go`. 

I also added tests in `tstate_test.go` to check this behavior too

chain/transaction.go
wlawt: I ran the `panic` and it showed that GetValue was being called here: https://github.com/ava-labs/hypersdk/blob/797d38ae84c8d7dca027a6b1d3e875e49218082e/chain/transaction.go#L376-L379

patrick-ogrady: That's my bad, this should be RW (we use this to check for duplicates).

chain/block.go
```diff
@@ -609,7 +609,12 @@ func (b *StatelessBlock) innerVerify(ctx context.Context, vctx VerifyContext) er
 	heightKeyStr := string(heightKey)
 	timestampKeyStr := string(timestampKey)
 	feeKeyStr := string(feeKey)
-	tsv := ts.NewView(set.Of(heightKeyStr, timestampKeyStr, feeKeyStr), map[string][]byte{
+
+	keys := make(state.Keys)
+	keys.AddKeyAndPermission(heightKeyStr, state.NewPermission(state.Write))
```
patrick-ogrady, line 614: nit: why not just call this `Add`? Seems like we always add `KeyAndPermission`

chain/block.go
```diff
@@ -609,7 +609,12 @@ func (b *StatelessBlock) innerVerify(ctx context.Context, vctx VerifyContext) er
...
+	keys.AddKeyAndPermission(timestampKeyStr, state.NewPermission(state.Write))
+	keys.AddKeyAndPermission(feeKeyStr, state.NewPermission(state.Write))
```
patrick-ogrady, line 616: nit: why not call this `state.Permission`? I don't think of us as creating a "new" permission, rather referencing existing ones.

executor/executor.go
```diff
@@ -105,7 +106,7 @@ func (e *Executor) createWorker() {
 
 // Run executes [f] after all previously enqueued [f] with
 // overlapping [conflicts] are executed.
-func (e *Executor) Run(conflicts set.Set[string], f func() error) {
+func (e *Executor) Run(conflicts state.Keys, f func() error) {
```
patrick-ogrady, line 109: Can you add a "TODO" with a link to an issue to handle read-only/write-only keys (everything rn is just treated as RW)?

state/state.go
```diff
@@ -27,3 +34,44 @@ type View interface {
...
+// Permission uses bits since transactions are expected to include this encoding
+type Keys map[string]Permission
+
+// The access permission (Read/Write) for each StateKey
+// Specifying RWrite is setting both the Read and Write bit
+type Permission set.Bits
+
+// By default, no permission bits are set
+func NewPermission(permissions ...uint8) Permission {
+	withinBoundsPermission := []int{} // NewBits requires int type
```
patrick-ogrady, line 50: Why not create a new `set.Bit` and just `Add` during the iteration?

patrick-ogrady: Seems like that will avoid this unnecessary arr

state/state.go
```diff
@@ -27,3 +34,44 @@ type View interface {
...
+// insertions from overriding the original permissions, use the Add function below.
+// Permission uses bits since transactions are expected to include this encoding
+type Keys map[string]Permission
+
+// The access permission (Read/Write) for each StateKey
+// Specifying RWrite is setting both the Read and Write bit
+type Permission set.Bits
+
+// By default, no permission bits are set
+func NewPermission(permissions ...uint8) Permission {
```
patrick-ogrady, line 49: I recommend creating a type `PermissionBit` and use that here instead of `uint8`

state/state.go
```diff
@@ -27,3 +34,44 @@ type View interface {
...
+	withinBoundsPermission := []int{} // NewBits requires int type
+	for _, v := range permissions {
+		// Only set bit to 1 if we're within 8 bits
```
patrick-ogrady, line 52: This seems like a fatal error that should not happen if the type system is respected. Nonetheless, I think the skip approach is correct.

state/state.go
```diff
@@ -7,9 +7,16 @@ import (
 	"context"
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/ava-labs/avalanchego/x/merkledb"
 )
 
+const (
+	Read          uint8 = 0
```
patrick-ogrady, line 15: I didn't realize `bit.Set` used `int`...for that reason, it is probably ok to keep that and then just ensure it is a valid enum below.

state/state.go
```diff
@@ -27,3 +34,44 @@ type View interface {
...
+// Specifying RWrite is setting both the Read and Write bit
+type Permission set.Bits
+
+// By default, no permission bits are set
+func NewPermission(permissions ...uint8) Permission {
+	withinBoundsPermission := []int{} // NewBits requires int type
+	for _, v := range permissions {
+		// Only set bit to 1 if we're within 8 bits
+		if v < PermissionLen {
+			withinBoundsPermission = append(withinBoundsPermission, int(v))
```
patrick-ogrady, line 54: I recommend using a `switch` over the `v` and just skip if `default`

state/state.go
```diff
@@ -27,3 +34,44 @@ type View interface {
...
+
+// Transactions are expected to use this to prevent
+// overriding of key permissions
+func (k Keys) AddKeyAndPermission(name string, permission Permission) {
+	_, exists := k[name]
+	if !exists {
+		k[name] = permission
+	} else {
+		// Transaction's permissions are the union of all
+		// Actions and Auths it contains
```
patrick-ogrady, line 74: nit: union of all state keys from both Actions and Auth

tstate/tstate_view.go
```diff
@@ -39,8 +39,7 @@ type TStateView struct {
 	// operations allows for reverting state to a certain point-in-time.
 	ops []*op
 
-	// We don't differentiate between read and write scope.
-	scope        set.Set[string] // stores a list of managed keys in the TState struct
+	scope        state.Keys // stores a map of managed keys and its permissions in the TState struct
```
patrick-ogrady, line 42: TODO: connect TODO to treat them difference (we won't prefetch a write key)

tstate/tstate_view.go
```diff
@@ -157,17 +156,22 @@ func (ts *TStateView) KeyOperations() (map[string]uint16, map[string]uint16) {
 	return ts.allocates, ts.writes
 }
 
-// checkScope returns whether [k] is in ts.readScope.
-func (ts *TStateView) checkScope(_ context.Context, k []byte) bool {
-	return ts.scope.Contains(string(k))
+// checkScope returns whether [k] is in scope and has appropriate permissions.
+func (ts *TStateView) checkScope(_ context.Context, k []byte, perm uint8) bool {
```
patrick-ogrady, line 160: again, recommend replacing `uint8` with `PermissionBit`

tstate/tstate_view.go
```diff
@@ -157,17 +156,22 @@ func (ts *TStateView) KeyOperations() (map[string]uint16, map[string]uint16) {
...
 
 // GetValue returns the value associated from tempStorage with the
-// associated [key]. If [key] does not exist in readScope or if it is not found
+// associated [key]. If [key] does not exist in scope, or is not read/rw, or if it is not found
 // in storage an error is returned.
 func (ts *TStateView) GetValue(ctx context.Context, key []byte) ([]byte, error) {
-	if !ts.checkScope(ctx, key) {
-		return nil, ErrKeyNotSpecified
+	// Getting a value requires a Read permission, so we pass state.Read
+	if !ts.checkScope(ctx, key, state.Read) {
```
patrick-ogrady, line 173: I love this function signature...very nice.

state/state.go
```diff
@@ -7,9 +7,16 @@ import (
 	"context"
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/ava-labs/avalanchego/x/merkledb"
 )
 
+const (
```
patrick-ogrady, line 14: Could you add one more value here for `Allocate` (fine to leave as a `TODO`)? -> This won't help with parallelization but will help with fee calculation (if you don't allow `Allocate`, can charge less fee).

state/state.go
```diff
@@ -7,9 +7,18 @@ import (
...
+	Read  PermissionBit = 0
+	Write PermissionBit = 1
+	// TODO: Handle Allocate permission
+	Allocate      PermissionBit = 2
```
patrick-ogrady, line 18: nit: let's put this above `Write` (make it =1)

patrick-ogrady: Id did the same with dimension definition: https://github.com/ava-labs/hypersdk/blob/4bd6d720a5bc8ab9c6361e05f3c00d9c4780720d/chain/fee_manager.go#L20-L22

state/state.go
```diff
@@ -27,3 +36,49 @@ type View interface {
...
 	GetMerkleRoot(ctx context.Context) (ids.ID, error)
 }
+
+// StateKey holds the name of the key and its permission (Read/Write). By default,
+// initialization of Keys with duplicate key will not work. And to prevent duplicate
+// insertions from overriding the original permissions, use the Add function below.
+// Permissions uses bits since transactions are expected to include this encoding
+type Keys map[string]Permissions
+
+// The access permission (Read/Write) for each StateKey
```
patrick-ogrady, line 46: nit: `Read/Allocate/Write`

state/state.go
```diff
@@ -27,3 +36,49 @@ type View interface {
...
+// By default, no permission bits are set
+func Permission(permissions ...PermissionBit) Permissions {
+	perms := set.NewBits()
+	for _, v := range permissions {
+		switch {
+		case v < PermissionLen:
+			// Only set bit to 1 if we're within 8 bits
+			perms.Add(int(v))
+		default:
+			continue
```
patrick-ogrady, line 61: You can remove `continue` here as that is implied.

state/state.go
```diff
@@ -27,3 +36,49 @@ type View interface {
...
+type Keys map[string]Permissions
+
+// The access permission (Read/Write) for each StateKey
+// Specifying RWrite is setting both the Read and Write bit
+type Permissions set.Bits
+
+type PermissionBit int
+
+// By default, no permission bits are set
+func Permission(permissions ...PermissionBit) Permissions {
```
patrick-ogrady, line 53: nit: note that any undefined permissions are dropped.

state/state.go
```diff
@@ -27,3 +36,49 @@ type View interface {
...
+// Transactions are expected to use this to prevent
+// overriding of key permissions
+func (k Keys) Add(name string, permission Permissions) {
+	_, exists := k[name]
+	if !exists {
+		k[name] = permission
+	} else {
+		// Transaction's permissions are the union of all
+		// state keys from both Actions and Auth
+		set.Bits(k[name]).Union(set.Bits(permission))
```
patrick-ogrady, line 82: nit: I think you need to do `k[name] = set.Bits(k[name]).Union(set.Bits(permission))`?

Maybe this works because it is a pointer

patrick-ogrady, line 82: Can we write a simple test that ensures this actually gets set? There is some potentially confusing indirection going on here (because `set.Bits` is modified inplace).

state/state.go
```diff
@@ -27,3 +36,49 @@ type View interface {
 	NewView(ctx context.Context, changes merkledb.ViewChanges) (merkledb.View, error)
 	GetMerkleRoot(ctx context.Context) (ids.ID, error)
 }
+
+// StateKey holds the name of the key and its permission (Read/Write). By default,
+// initialization of Keys with duplicate key will not work. And to prevent duplicate
+// insertions from overriding the original permissions, use the Add function below.
+// Permissions uses bits since transactions are expected to include this encoding
+type Keys map[string]Permissions
```
patrick-ogrady, line 44: IMO this should all be in a new file called `state/keys.go`?

state/state.go
wlawt: I wrote some tests here earlier: https://github.com/ava-labs/hypersdk/pull/693/files#diff-31074509c62c06d7c8b41444cf3f7ac6ee9aa9a1079d2f6d699ae0b063427a69R858 

But I will also make it more clear for that line of code (with the earlier comment you suggested)

state/keys.go
patrick-ogrady, line 1: 🙏

tstate/tstate_test.go
```diff
@@ -613,3 +613,328 @@ func TestCreateView(t *testing.T) {
 		require.ErrorIs(err, database.ErrNotFound, "value not removed from db")
 	}
 }
+
+func TestGetValuePermissions(t *testing.T) {
```
patrick-ogrady, line 617: nit: Feels like these tests that just check `state.Permission` should be in `state/keys_test`?

state/keys.go
```diff
@@ -0,0 +1,58 @@
...
+
+// StateKey holds the name of the key and its permission (Read/Allocate/Write). By default,
+// initialization of Keys with duplicate key will not work. And to prevent duplicate
+// insertions from overriding the original permissions, use the Add function below.
+// Permissions uses bits since transactions are expected to include this encoding
+type Keys map[string]Permissions
+
+// The access permission (Read/Allocate/Write) for each StateKey
+// Specifying RWrite is setting both the Read and Write bit
+type Permissions set.Bits
```
patrick-ogrady, line 24: I wonder if it is more efficient to just use a struct here like `{Allocate:bool, Write: bool, Read: bool}` instead of using this bitset.

This is not sent over the wire, so using a bitset seems a bit unnecessary. When it eventually is sent over the wire (when access pattern can't be inferred by the transaction), it makes sense.

state/keys.go
```diff
@@ -0,0 +1,87 @@
...
+		// state keys from both Actions and Auth
+		if permission.Read {
+			k[name].Read = true
+		}
+		if permission.Allocate {
+			k[name].Allocate = true
+		}
+		if permission.Write {
+			k[name].Write = true
+		}
```
wlawt, line 50: nit: is there a cleaner way to do this? (`switch` cases don't follow through)

patrick-ogrady (approved):

Issue #531, opened 2023-10-09
[archiver/tokenvm] Create an S3 Archiver + RPC 


patrick-ogrady:
Would be very nice to have an out-of-the-box archiver that could run next to any hypersdk and store blocks in S3. Bonus points if blocks can be easily served from it (with results/warp messages) -> this could be used to make an explorer (nodes should only need to serve reads of current state).

PR #697, opened 2024-01-21
S3 Archiver in TokenVM + RPC

bianyuanop:
### Context

This PR covers #531, including

- an `archiver` package with s3 archiver and noop archiver implementations
- RPC methods and a controller resolution function to handle block related JSON-RPC requests
- configuration of archiver from file that passed by a env variable `$ARCHIVER_CONFIG_PATH` used by `./scripts/run.sh`
- e2e tests to verify if archiver functions well 

### Testing
- archiver functionalities has been tested with a s3 archiver, see below
- The JSONRPC client cannot parse `StatefulBlock` directly since it relys on `json.Unmarshal`, and it doesn't do reflection automatically. To be specific, `Block.txs.action` cannot be unmarshalled since it's an interface rather than a struct. This is handled by [requester.go#L146](https://github.com/ava-labs/hypersdk/blob/main/requester/requester.go#L146), possible solution to this need to be discussed. 
image

### How to test
Create a json file `s3conf.json` that contains s3 archiver config, such as 
```json
{"archiverType":"aws","enabled":true,"tokenID":"s3 access key","secretKey":"secretkey","region":"us-east-2","bucketName":"tokenvm"}
```
or `disk_config.json`
```json
{
    "archiverType": "disk",
    "enabled": true,
    "location": "/tmp/tokenvm"
}
```
Then run
```shell
MODE="test" ARCHIVER_CONFIG_PATH="./s3conf.json" ./scripts/run.sh
```
or 
```shell
MODE="test" ARCHIVER_CONFIG_PATH="./disk_config.json" ./scripts/run.sh
```
BTW, `jq` need to be installed since it provides a way to generate a escaped string of a json


aaronbuchwald:
Closing this PR for now because it's become stale.

If you want to continue this work, we're likely going to solidify the interface under the idea of allowing an `AcceptedSubscriber`.

It's not 100% clear yet what should be inside of the HyperSDK vs. external. If you're interested to continue this work, could you open a GitHub discussion on this topic?

Once https://github.com/ava-labs/hypersdk/pull/1143 is merged, we can spec out a POC to export block data to an arbitrary consumer that ensures indexes are not dropped as per https://github.com/ava-labs/hypersdk/issues/1145.

Another possible direction is to define a plugin interface for an external indexer, so these can be more easily maintained externally.

Issue #335, opened 2023-08-08
[chain/tx] Make transaction.Action an array (Support Batch Transactions)


patrick-ogrady:
Allow a single transaction to carry multiple actions (this will better amortize the cost per signature verification, which will be cheaper for end users)

PR #698, opened 2024-01-21
[chain/transaction] Introduce Multi-Action Support

patrick-ogrady:
Resolves: #335 

### TODO
- [ ] Add `Rule` to limit number of `Actions` per `Transaction`
- [ ] Introduce `ActionID` (33 bytes) for unique interactions rather than `TxID` (may need more than 1 unique key per tx)

### Future Work (post-merge)
- [ ] Add support for outputting multiple warp messages
- [ ] Consider attaching `Warp.Message` to an `Action` instead of a `Transaction` (then consider each `Action` an `InnerTx`)


chain/transaction.go
```diff
@@ -371,63 +393,65 @@ func (t *Transaction) Execute(
...
 			// the key was not properly specified.
-			return &Result{false, utils.ErrBytes(err), maxUnits, maxFee, nil}, nil
+			return []*Result{{false, utils.ErrBytes(err), maxUnits, maxFee, nil}}, nil
 		}
 	}
 
 	// We create a temp state checkpoint to ensure we don't commit failed actions to state.
 	actionStart := ts.OpIndex()
-	handleRevert := func(rerr error) (*Result, error) {
+	handleRevert := func(rerr error) ([]*Result, error) {
```
patrick-ogrady, line 402: If we revert, return all results up to the failure (with success true)

chain/processor.go
```diff
@@ -41,7 +41,7 @@ func (b *StatelessBlock) Execute(
 		f       = fetcher.New(im, numTxs, b.vm.GetStateFetchConcurrency())
 		e       = executor.New(numTxs, b.vm.GetTransactionExecutionCores(), MaxKeyDependencies, b.vm.GetExecutorVerifyRecorder())
 		ts      = tstate.New(numTxs * 2) // TODO: tune this heuristic
-		results = make([]*Result, numTxs)
+		results = make([]*Result, 0)
```
wlawt, line 44: Do we want a 2d array with a length of `numTxs` instead? or flatten all the results into one array? not sure of the consequences of doing the flatten approach right now

wlawt:
Going to use https://github.com/ava-labs/hypersdk/pull/858 going forward (had too many issues with merge conflicts and made it difficult to reason). Was easier to start fresh

PR #700, opened 2024-01-23
[fees] Introduce Fees package

### LINKED ISSUE #671: [chain] Move `chain/fee_manager.go` to `fees/manager.go`

wlawt:
Closes #671 

### Overview

Moves `fee_manager.go` out of the `chain` package and into its own `fees` package. 

### Changes

Create `Rules` dependency in `fees` to avoid circular dependency with `chain`. This PR makes `chain` the parent and `fees` the leaf of the dependency tree.

### Tests

Integration and build tests pass for TokenVM and MorpheusVM


builder/dependencies.go
```diff
@@ -17,5 +18,5 @@ type VM interface {
 	PreferredBlock(context.Context) (*chain.StatelessBlock, error)
 	Logger() logging.Logger
 	Mempool() chain.Mempool
-	Rules(int64) chain.Rules
+	Rules(int64) fees.Rules
```
patrick-ogrady, line 21: You should not move `Rules` out of `Chain`.

chain/block.go
```diff
@@ -117,7 +118,7 @@ type StatelessBlock struct {
 	vdrState     validators.State
 
 	results    []*Result
-	feeManager *FeeManager
+	feeManager *fees.FeeManager
```
patrick-ogrady, line 121: When we move this to `fees`, we should rename `FeeManager` -> `Manager` to avoid a stutter.

chain/dependencies.go
```diff
@@ -113,51 +114,6 @@ type Mempool interface {
 	FinishStreaming(context.Context, []*Transaction) int
 }
 
-type Rules interface {
```
patrick-ogrady, line 116: As mentioned above, don't move this out of `chain`.

patrick-ogrady: We could embed a fee-focused interface into `Rules` but the top-level struct should stay here.

chain/dependencies.go
```diff
@@ -123,10 +124,7 @@ type Rules interface {
 	GetMinEmptyBlockGap() int64 // in milliseconds
 	GetValidityWindow() int64   // in milliseconds
 
-	GetMinUnitPrice() Dimensions
-	GetUnitPriceChangeDenominator() Dimensions
-	GetWindowTargetUnits() Dimensions
-	GetMaxBlockUnits() Dimensions
+	Fees() fees.Metadata
```
wlawt, line 127: nit: could possibly rename?

examples/morpheusvm/genesis/rules.go
```diff
@@ -18,6 +22,10 @@ type Rules struct {
 	chainID   ids.ID
 }
 
+type FeeMetadata struct {
+	*Rules
+}
```
wlawt, line 27: Not sure if there's a better way to do this implementation or if it's necessarily the correct approach

examples/tokenvm/genesis/rules.go
```diff
@@ -18,6 +22,10 @@ type Rules struct {
 	chainID   ids.ID
 }
 
+type FeeMetadata struct {
+	*Rules
+}
```
wlawt, line 27: Same comment as above

fees/dependencies.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package fees
+
+type Metadata interface {
```
wlawt, line 6: Not sure here if we want a more useful name (or if it suffices)

chain/dependencies.go
wlawt: I made a fee-focused interface that takes out the functions that `fee_manager` uses from `Rules` and placed it into `fees.Metadata` so we can embed it into `Rules`: 

```
type Metadata interface {
	GetMinUnitPrice() Dimensions
	GetUnitPriceChangeDenominator() Dimensions
	GetWindowTargetUnits() Dimensions
	GetMaxBlockUnits() Dimensions
}
```

examples/morpheusvm/genesis/rules.go
patrick-ogrady: I would not recommend doing it this way.

chain/dependencies.go
```diff
@@ -123,10 +124,7 @@ type Rules interface {
...
 
 	GetBaseComputeUnits() uint64
```
patrick-ogrady, line 129: These are all `Fees` related IMO?

fees/dependencies.go
patrick-ogrady, line 6: This isn't a dependency of `fees`, it is what `fees` implements. 

You should define this in `chain`.

patrick-ogrady: I don't think we should use `Metadata` for this...not sure what makes sense yet.

I would call this interface `Fees` and keep it in `chain/dependencies`?

wlawt: The only problem that I'm facing is that `computeNext` takes in `Rules` as a parameter, which could lead to an import cycle. I'm not really sure how to work around that

fees/dependencies.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package fees
+
```
patrick-ogrady, line 5: I would recommend changing this to `Rules` and then not change `chain.Rules`.

patrick-ogrady (approved):

Issue #336, opened 2023-08-08
[chain] Add support for Chain/Validator-Driven Transactions


patrick-ogrady:
At the end of each block, we should make it possible for a validator to add "unauthenticated" actions. The rules for which actions can be issued (like Oracle updates) can be governed by `genesis.Rules`

patrick-ogrady:
There is an open discussion about this topic: https://github.com/ava-labs/hypersdk/discussions/933

PR #704, opened 2024-01-24
Validator driven transactions

bianyuanop:
### Context
This PR covers: #336. This is a guess work of how the _Validator Driven Transactions_ should work. The idea is to use rules from genesis to check if the transaction is validator driven then use a flag to bypass some operations that's only needed by authenticated transactions. 

### Tests
e2e tests were passes but haven't been updated since I need to know if it's the right direction

### DIFF
--- a/chain/builder.go
+++ b/chain/builder.go
@@ -8,6 +8,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


aaronbuchwald:
Going through and closing stale PRs. Sorry this was not reviewed earlier, if you want to pick this back up, feel free to re-open and I'll add a more detailed spec to the GitHub issue, so the target for this work is more clear.

PR #710, opened 2024-02-02
[WIP] Add Context To Functions 

dboehm-avalabs:


x/programs/cmd/simulator/vm/actions/program_execute.go
```diff
@@ -107,8 +106,12 @@ func (t *ProgramExecute) Execute(
 	importsBuilder.Register("state", func() host.Import {
 		return pstate.New(logging.NoLog{}, mu)
 	})
+	callContext := program.Context{
+		ProgramID:        programID,
+		Actor:            [32]byte(actor[1:]),
+		OriginatingActor: [32]byte(actor[1:])}
```
patrick-ogrady, line 112: You know what might be neat is passing the stack of actors rather than just parent actor and original actor 🤔

x/programs/cmd/simulator/vm/actions/program_execute.go
```diff
@@ -107,8 +106,12 @@ func (t *ProgramExecute) Execute(
 	importsBuilder.Register("state", func() host.Import {
 		return pstate.New(logging.NoLog{}, mu)
 	})
+	callContext := program.Context{
+		ProgramID:        programID,
+		Actor:            [32]byte(actor[1:]),
```
patrick-ogrady, line 111: I recommend passing all 33 bytes here as there could be conflicting addresses within the trailing 32

x/programs/cmd/simulator/vm/actions/program_execute.go
iFrostizz: Do you have use case ideas ?
It seems that if a contract is calling itself then this array would be redundant and linearly growing, making the execution and simulation more expensive. Also, it *may* encourage some bad patterns such as looking for every element in the array instead of the last one and the very first one.

x/programs/examples/counter_test.go
```diff
@@ -95,41 +96,41 @@ func TestCounterProgram(t *testing.T) {
...
+	programID2Ptr, err := argumentToSmartPtr(programID2, mem2)
 	require.NoError(err)
 
 	// write alice's key to stack and get pointer
 	alicePtr2, err := argumentToSmartPtr(aliceKey, mem2)
 	require.NoError(err)
 
 	// initialize counter for alice on runtime 2
-	result, err = rt2.Call(ctx, "initialize_address", programID2Ptr, alicePtr2)
+	result, err = rt2.Call(ctx, "initialize_address", program.Context{ProgramID: programID2}, alicePtr2)
```
iFrostizz, line 121: So if I understand correctly, the `program.Context` is passed here, pre-populated with the `ProgramID` only.
Isn't it a bit redundant though ? According to the "old" way, you would pass the `programID` as the function parameter, and as of this feature, the `program.Context` field would all be populated with this `programID` in the `Call` function.

x/programs/cmd/simulator/vm/actions/program_execute.go
patrick-ogrady: I have always found that the more data available during contract execution the better. I think contracts could enforce call depth or only certain calling "paths" (might be nice for security or something).

x/programs/examples/counter_test.go
richardpringle: @iFrostizz, it's easier to keep the function signatures the same and extend the context rather than having to change every signature every time we want to add to the context. This PR simply gives us a place where we can expand things without breaking old programs.

iFrostizz: Thanks, that makes sense !

x/programs/examples/utils.go
```diff
@@ -31,7 +31,7 @@ func serializeParameter(obj interface{}) ([]byte, error) {
 
 // Serialize the parameter and create a smart ptr
 func argumentToSmartPtr(obj interface{}, memory *program.Memory) (program.SmartPtr, error) {
-	bytes, err := serializeParameter(obj)
+	bytes, err := borsh.Serialize(obj)
```
richardpringle, line 34: 👍

x/programs/program/context.go
```diff
@@ -0,0 +1,11 @@
+package program
+
+import (
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+type Context struct {
+	ProgramID        ids.ID   `json:"program"`
+	Actor            [32]byte `json:"actor"`
+	OriginatingActor [32]byte `json:"originating_actor"`
```
richardpringle, line 10: I don't think we need to change this now, but I like @patrick-ogrady's idea of a stack of actors... I can already see myself responding to issues saying, "I think you meant to use `actor`, not `originating_actor`"

x/programs/rust/wasmlanche_sdk/src/context.rs
richardpringle, line 1: 👍  (except for the missing newline)

richardpringle:
LGTM so far

iFrostizz:
Substrate sdk does it using origin checks with helpers such as `ensure_origin` which might be an inspiration. See https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.EnsureOrigin.html

richardpringle:
@dboehm-avalabs, I'm closing this. I don't think it's relevant anymore

PR #715, opened 2024-02-07
[executor] handling read/allocate/write-keys in the execution engine

### LINKED ISSUE #701: [executor] parallelizing reads in the execution engine
See below

Prereq: #693 
Related: #709

wlawt:
Closes #701, #709

### Overview

Please see the 2nd comment in #701 for the overall approach. Some modifications were made from the original idea. That being, in order to ensure reads can't happen before writing, and writing can't happen before reading, the executor keeps track of the last blocked task in `edges` known as the `Parent` within `KeyData`. This allows us to execute the tasks in the order in which they were enqueued, in the case of high contention.

### Test Plan

- Added unit tests in `executor_test.go`, coverage is ~97.4%
- Integration tests in MorpheusVM and TokenVM pass
 
### TODOs

- [x] Add more unit tests
- [ ] Handle GetValue caching

### DIFF
--- a/chain/processor.go
+++ b/chain/processor.go
@@ -81,18 +81,8 @@ func (b *StatelessBlock) Execute(
 			// Fetch keys from disk
 			var toCache map[string]*fetchData
 			if len(toLookup) > 0 {
-				cacheLock.RLock()
 				toCache = make(map[string]*fetchData, len(toLookup))
 				for _, k := range toLookup {
-					// Ensure we don't read the same value from disk more than once
-					// We can use the cache for subsequent reads
-					if v, ok := cache[k]; ok {
-						reads[k] = v.chunks
-						if v.exists {
-							storage[k] = v.v
-						}
-						continue
-					}
 					v, err := im.GetValue(ctx, []byte(k))
 					if errors.Is(err, database.ErrNotFound) {
 						reads[k] = 0
@@ -111,7 +101,6 @@ func (b *StatelessBlock) Execute(
 					toCache[k] = &fetchData{v, true, numChunks}
 					storage[k] = v
 				}
-				cacheLock.RUnlock()
 			}
 
 			// Execute transaction
--- a/chain/processor.go
+++ b/chain/processor.go
@@ -81,8 +81,18 @@ func (b *StatelessBlock) Execute(
 			// Fetch keys from disk
 			var toCache map[string]*fetchData
 			if len(toLookup) > 0 {
+				cacheLock.RLock()
 				toCache = make(map[string]*fetchData, len(toLookup))
 				for _, k := range toLookup {
+					// Ensure we don't read the same value from disk more than once
+					// We can use the cache for subsequent reads
+					if v, ok := cache[k]; ok {
+						reads[k] = v.chunks
+						if v.exists {
+							storage[k] = v.v
+						}
+						continue
+					}
 					v, err := im.GetValue(ctx, []byte(k))
 					if errors.Is(err, database.ErrNotFound) {
 						reads[k] = 0
@@ -101,6 +111,7 @@ func (b *StatelessBlock) Execute(
 					toCache[k] = &fetchData{v, true, numChunks}
 					storage[k] = v
 				}
+				cacheLock.RUnlock()
 			}
 
 			// Execute transaction
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -168,7 +168,7 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 	if t.dependencies != nil && t.dependencies.Len() > 0 {
 		// Ensure that we don't schedule a transaction that is
 		// blocked by another transaction currently running
-		for _, bID := range t.dependencies {
+		for bID := range t.dependencies {
 			// don't execute if any dependency task is in blocking
 			if e.blocking.Contains(bID) {
 				if e.metrics != nil {
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -5,7 +5,6 @@ package executor
 
 import (
 	"sync"
-	//"fmt"
 
 	"github.com/ava-labs/avalanchego/utils/set"
 
@@ -118,9 +117,6 @@ func (e *Executor) createWorker() {
 
 // Run executes [f] after all previously enqueued [f] with
 // overlapping [conflicts] are executed.
-// TODO: Handle read-only/write-only keys (currently the executor
-// treats everything still as ReadWrite, see issue below)
-// https://github.com/ava-labs/hypersdk/issues/709
 func (e *Executor) Run(conflicts state.Keys, f func() error) {
 	e.l.Lock()
 	defer e.l.Unlock()
@@ -146,11 +142,10 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 					lt.blocking = set.NewSet[int](defaultSetSize)
 				}
 				// key has ONLY a Read permission
-				// TODO: clean this up
-				if v.Has(state.Read) && !v.Has(state.Write) && !v.Has(state.Allocate) {
-					// lt contains a Write permission. If lt had only a Read
+				if v == state.Read {
+					// lt contains either Allocate or Write permission. If lt had only a Read
 					// permission, no dependency or blocking would be added
-					if latest.Permissions.Has(state.Write) {
+					if latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
 						t.dependencies.Add(lt.id) // t depends on lt to execute
 						lt.blocking.Add(id)       // lt is blocking this current task
 					}
@@ -158,8 +153,8 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 
 				// key contains a Write permission
 				if v.Has(state.Write) {
-					// lt contains a Read or Write permission
-					if latest.Permissions.Has(state.Read | state.Write) {
+					// lt contains either Read, Allocate, or Write
+					if latest.Permissions.Has(state.Read) || latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
 						t.dependencies.Add(lt.id)
 						lt.blocking.Add(id)
 					}
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -192,17 +192,7 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 					// With a bunch of reads before our write,
 					// we need to update that we're blocked by
 					// all of these reads
-					for b := range key.concurrentReads {
-						bt := e.tasks[b]
-						// In the case one of the concurrent reads have
-						// executed and we're at this stage of the code,
-						// we only want to record that we're blocked on
-						// the concurrent reads that are not yet executed.
-						if !bt.executed {
-							bt.blocking.Add(id)
-							t.dependencies.Add(b)
-						}
-					}
+					e.updateConcurrentReads(key, t)
 					// Any subsequent reads will now be blocked by this
 					// task, if it hasn't executed yet.
 					key.concurrentReads.Clear()
@@ -214,13 +204,7 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 			// consider any outstanding reads that still need
 			// to be executed once its parent write ran.
 			if v.Has(state.Allocate) || v.Has(state.Write) {
-				for b := range key.concurrentReads {
-					bt := e.tasks[b]
-					if !bt.executed {
-						bt.blocking.Add(id)
-						t.dependencies.Add(b)
-					}
-				}
+				e.updateConcurrentReads(key, t)
 			}
 			key.concurrentReads.Clear()
 			if v == state.Read {
@@ -243,6 +227,20 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 	}
 	if e.metrics != nil {
 		e.metrics.RecordBlocked()
+	}
+}
+
+// Update concurrentReads when we encounter a Allocate/Write key
+// or when the last blocked transaction has been executed
+func (e *Executor) updateConcurrentReads(key *keyData, t *task) {
+	for b := range key.concurrentReads {
+		bt := e.tasks[b]
+		// Record that we're blocked on the concurrent reads
+		// that are not yet executed.
+		if !bt.executed {
+			bt.blocking.Add(t.id)
+			t.dependencies.Add(b)
+		}
 	}
 }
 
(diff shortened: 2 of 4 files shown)


executor/executor.go
```diff
@@ -31,7 +31,15 @@ type Executor struct {
...
 	completed int
 	tasks     map[int]*task
-	edges     map[string]int
+	edges     map[string]*KeyData
+	blocking  set.Set[int]
+}
+
+// KeyData keeps track of the key permission and the
+// TaskID using the state key name.
+type KeyData struct {
```
wlawt, line 40: nit: is there a clearer name

chain/processor.go
```diff
@@ -80,8 +80,18 @@ func (b *StatelessBlock) Execute(
...
+					// Ensure we don't read the same value from disk more than once
+					// We can use the cache for subsequent reads
+					if v, ok := cache[k]; ok {
+						reads[k] = v.chunks
+						if v.exists {
+							storage[k] = v.v
+						}
+						continue
+					}
 					v, err := im.GetValue(ctx, []byte(k))
```
patrick-ogrady, line 95: We should avoid holding the lock while fetching keys from the database imo

chain/processor.go
```diff
@@ -100,6 +110,7 @@ func (b *StatelessBlock) Execute(
 					toCache[k] = &fetchData{v, true, numChunks}
```
patrick-ogrady, line 110: We also update the cache with an `RLock` (which probably is unexpected behavior)

chain/processor.go
```diff
@@ -100,6 +110,7 @@ func (b *StatelessBlock) Execute(
...
 					storage[k] = v
 				}
+				cacheLock.RUnlock()
```
patrick-ogrady, line 113: Even if we did want to hold the lock during the entirety of this disk fetch, we wouldn't prevent multiple fetches of the same key because we haven't updated `cache` by the time this lock is released.

executor/executor.go
```diff
@@ -122,35 +129,64 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+						lt.blocking.Add(id)       // lt is blocking this current task
+						// If lt is blocking any other txn, we need to record this
+						// to prevent subsequent txns that depend on lt from executing
+						// at the same time
+						e.blocking.Add(lt.id)
+					}
+				}
+
+				// key contains a Write permission
+				if v.Has(state.Write) {
```
patrick-ogrady, line 158: I think this should be `Write` or `Allocate`?

executor/executor.go
```diff
@@ -122,35 +129,64 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+					// lt contains either Read, Allocate, or Write
+					if latest.Permissions.Has(state.Read) || latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
```
patrick-ogrady, line 160: I don't think it matters what permissions the dependencies have, we need to wait?

Or is the idea that we are "future-looking" and being explicit with our checks?

executor/executor.go
```diff
@@ -122,35 +129,64 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
 	}
 
-	// Start execution if there are no blocking dependencies
-	if t.dependencies == nil || t.dependencies.Len() == 0 {
-		t.dependencies = nil // free memory
-		e.executable <- t
-		if e.metrics != nil {
-			e.metrics.RecordExecutable()
+	if t.dependencies != nil && t.dependencies.Len() > 0 {
+		// Ensure that we don't schedule a transaction that is
```
patrick-ogrady, line 173: In the previous code, we didn't run a transaction if there were dependencies. I'm not sure what `e.blocking` adds?

executor/executor.go
```diff
@@ -122,35 +129,64 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+					// permission, no dependency or blocking would be added
+					if latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
+						t.dependencies.Add(lt.id) // t depends on lt to execute
+						lt.blocking.Add(id)       // lt is blocking this current task
+						// If lt is blocking any other txn, we need to record this
+						// to prevent subsequent txns that depend on lt from executing
+						// at the same time
+						e.blocking.Add(lt.id)
+					}
+				}
```
patrick-ogrady, line 155: I think you could `continue` inside of this `if`?

executor/executor.go
```diff
@@ -122,35 +129,64 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
 					lt.blocking = set.NewSet[int](defaultSetSize)
 				}
-				lt.blocking.Add(id)
+				// key has ONLY a Read permission
+				if v == state.Read {
+					// lt contains either Allocate or Write permission. If lt had only a Read
+					// permission, no dependency or blocking would be added
+					if latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
+						t.dependencies.Add(lt.id) // t depends on lt to execute
+						lt.blocking.Add(id)       // lt is blocking this current task
```
patrick-ogrady, line 149: You need to register the type of blocker this is `state.Read` so that when complete, `lt` can potentially unblock multiple ids.

patrick-ogrady, line 149: If this is a read dependency, we don't necessarily just want to register on the preceding tx...we would want to register on the last blocked transaction (otherwise non-blocking txs will be unnecessarily serialized).

patrick-ogrady: This may also cause a bug if the `lt` is blocked on a write but we don't add a dependency because `lt`'s usage of a key is a read.

executor/executor.go
```diff
@@ -122,35 +129,64 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+						// If lt is blocking any other txn, we need to record this
+						// to prevent subsequent txns that depend on lt from executing
+						// at the same time
+						e.blocking.Add(lt.id)
```
patrick-ogrady, line 153: As mentioned below, I don't know what `e.blocking` adds and think it should be removed.

Blocking is implied via `lt.blocking`. No transaction will be executed until all blockers are cleared by its dependendencies.

executor/executor.go
patrick-ogrady: This could be fixed by adding `KeyData.Parent` and we iterate backwards until we get to executed or a write/alloc I think.

chain/processor.go
```diff
@@ -80,8 +80,18 @@ func (b *StatelessBlock) Execute(
 			// Fetch keys from disk
 			var toCache map[string]*fetchData
 			if len(toLookup) > 0 {
+				cacheLock.RLock()
```
patrick-ogrady, line 83: I'm not sure a `RLock` helps prevent concurrent reads here as any number of transactions would see the value is missing in the cache and try to read it?

executor/executor.go
```diff
@@ -122,22 +143,92 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+		parent, exists := e.edges[k]
+		if !exists {
+			concurrentReads := set.NewSet[int](defaultSetSize)
+			if v == state.Read {
+				// Record any reads on the same key to help
+				// with updating dependencies properly
+				concurrentReads.Add(id)
+			}
+			// Key doesn't exist, so we add it to our edge map
+			e.edges[k] = &KeyData{Parent: id, Permissions: v, ConcurrentReads: concurrentReads}
```
patrick-ogrady, line 157: Haven't reviewed the logic, but you should do a `continue` after this line instead of doing a monster `if/else`

wlawt: oops forgot, updated now

executor/executor.go
```diff
@@ -122,22 +143,92 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
```
patrick-ogrady, line 157: There is no `Parent` here?

executor/executor.go
```diff
@@ -122,22 +143,92 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+					parent.ConcurrentReads.Clear()
+				}
+			}
+		} else {
+			// We don't need to record the case of W->W->W->...,
+			// since that parent write already executed. We
+			// consider any outstanding reads that still need
+			// to be executed once its parent write ran.
+			if v.Has(state.Allocate) || v.Has(state.Write) {
+				for b := range parent.ConcurrentReads {
```
patrick-ogrady, line 216: With this approach, you'll want to remove `ids` from `ConcurrentReads` when they finish execution? Seems like a way to avoid a lot of extra iteration here (and complexity).

executor/executor.go
```diff
@@ -122,22 +143,92 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+					bt := e.tasks[b]
+					if !bt.executed {
+						bt.blocking.Add(id)
+						t.dependencies.Add(b)
+					}
 				}
-				lt.blocking.Add(id)
+			}
+			parent.ConcurrentReads.Clear()
```
patrick-ogrady, line 224: Why would you want to clear `ConcurrentReads` here? Seems like it should be included in the `if` above?

executor/executor.go
```diff
@@ -122,22 +143,92 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
-					t.dependencies = set.NewSet[int](defaultSetSize)
+	for k, v := range conflicts {
+		// Get last blocked transaction
+		parent, exists := e.edges[k]
+		if !exists {
+			concurrentReads := set.NewSet[int](defaultSetSize)
+			if v == state.Read {
+				// Record any reads on the same key to help
+				// with updating dependencies properly
+				concurrentReads.Add(id)
```
patrick-ogrady, line 154: This is a tricky condition to consider. I agree it helps with tracking dependencies.

executor/executor.go
```diff
@@ -122,22 +143,92 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
-	for k := range conflicts {
-		latest, ok := e.edges[k]
-		if ok {
-			lt := e.tasks[latest]
-			if !lt.executed {
-				if t.dependencies == nil {
-					t.dependencies = set.NewSet[int](defaultSetSize)
+	for k, v := range conflicts {
+		// Get last blocked transaction
+		parent, exists := e.edges[k]
```
patrick-ogrady, line 148: I don't love `parent` here given this `keyData` can be either the `parent` or some value back in the ancestry of keys.

executor/executor.go
```diff
@@ -122,22 +143,92 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+			if v == state.Read {
+				// Record any reads on the same key to help
+				// with updating dependencies properly
+				concurrentReads.Add(id)
+			}
+			// Key doesn't exist, so we add it to our edge map
+			e.edges[k] = &KeyData{Parent: id, Permissions: v, ConcurrentReads: concurrentReads}
+			continue
+		}
+		pt := e.tasks[parent.Parent]
```
patrick-ogrady, line 160: I think `e.tasks[key.ID]` is much easier to reason about.

executor/executor.go
```diff
@@ -31,7 +31,31 @@ type Executor struct {
...
+// read if we're trying to write. This is to adhere to the
+// properties of the Executor.
+//
+// Consider the case where four transactions, touching the same state
+// key, with the following permissions: T1_W -> T2_R -> T3_R -> T4_W.
+// Assume that T1 hasn't executed, so this blocks T2 and T3. But,
+// we also note down that T2 and T3 are concurrent reads here. So,
+// when we get to T4, we observe that T2 and T3 are both blocking T4,
+// and we can record the appropriate dependencies with ConcurrentReads.
+type KeyData struct {
```
patrick-ogrady, line 55: should be private struct `keyData`

executor/executor.go
patrick-ogrady, line 157: Don't love `Parent` here. Not sure what to replace it with but it is really `LastWriteAllocateOrRead` lol

executor/executor.go
patrick-ogrady: Maybe just `key`?

executor/executor.go
patrick-ogrady: Maybe just `ID`?

executor/executor.go
```diff
@@ -122,22 +143,92 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
...
+		if !pt.executed {
+			if t.dependencies == nil {
+				t.dependencies = set.NewSet[int](defaultSetSize)
+			}
+			if pt.blocking == nil {
+				pt.blocking = set.NewSet[int](defaultSetSize)
+			}
+			// key has ONLY a Read permission
+			if v == state.Read {
+				parent.ConcurrentReads.Add(id)
```
patrick-ogrady, line 170: Could this lead to the same `id` being added to `pt.blocking` twice?

If `parent.Permissions == state.Write` and 1 tx comes in with `v==state.Read`, it will add here. If another tx comes in with `v==state.Write`, then it will iterate through the concurrent reads below and add the item to `pt.blocking`.

executor/executor.go
patrick-ogrady: I would think `if executed`, you should assume there is nothing to do here besides update `ID` and `Permissions`?

patrick-ogrady (changes requested):

executor/executor.go
wlawt: reposting for visibility:

if i'm understanding this comment correctly, suppose we have the following 3 transactions:
T1 W
T2 R
T3 W

here, in T1 W, it's "parent" is T1. suppose it doesn't yet execute, for some reason

then in T2 R, it gets blocked by its "parent" which is T1, so we have the following:
T1.blocking = (T2)
T2.dep = (T1)
concurrentReads = (T2)

then when we run T3 W, we will go through concurrentReads and update the following:
T2.blocking = (T3)
T3.dep = (T2)

then we clear concurrentReads, and then finally, we update the "parent" to be T3

so in this walkthrough we don't add to pt.blocking twice. if a key has Allocate/Write, it will always be updated to itself

executor/executor.go
wlawt: reposting for visibility:

we want to clear concurrentReads here before updating because if we had the case of only reads like: R->R->R->R, and the first R already executed, concurrentReads was populated with that first R, but was already done by the time we reached this part of the code, so we need to reset it to avoid adding unwanted concurrentReads

wlawt:
Please see #779 for the continuation of the PR (the branch is no longer a fork of the repo)

PR #725, opened 2024-02-09
[vryx] add epochs and partitions

patrick-ogrady:
- [x] ~[?] Avoid construction around epoch boundary (will this lead to gaps in slot building)~
- [x] remove last block check for too far into the future (use current time + validity window)
- [x] ensure there is a p-chain height for an epoch before building a chunk
- [x] partition on gossip
  - [x] allow validators to gossip txs for partitions they aren't responsible for
  - [ ] ~allow RPC to send to specific validators? -> this isn't a realistic scenario~
- [x] cache validator set as soon as set in state during execute (then never need to do a lookback)
  - [x] ~use an `emap` here and clear sets when epoch is finished~ use LRU
- [x] Remove TODO for handling warp messages because pulling from external sources
  - [x] store most recent block ctx to verify warp messages to avoid look backs

### Open Questions
- [x] Including chunks prior to the epoch associated with their slot? Otherwise, might need to wait (or set expiry of slot very close to end of epoch)
  - [x] inclusion based on expiry time, partition verification/cert verification/warp verification based on epoch (because we have n and n+1 epoch height known by start of n. Still limit outstanding by epoch for txs to collect bond.
- [x] How to partition accounts? Sample with epoch+account


examples/morpheusvm/genesis/genesis.go
```diff
@@ -77,7 +75,7 @@ func Default() *Genesis {
 		MaxBlockUnits:              chain.Dimensions{1_800_000, 2_000, 2_000, 2_000, 2_000},
 
 		// Tx Parameters
-		ValidityWindow: 60 * hconsts.MillisecondsPerSecond, // ms
+		ValidityWindow: 10 * hconsts.MillisecondsPerSecond, // ms
```
patrick-ogrady, line 78: Must reduce because reduced epoch size (for faster epoch init)

PR #731, opened 2024-02-13
Rename "smart-pointer" on the Rust side

richardpringle:
While this is a little silly, "smart pointers" are an explicit concept in Rust [^1].

[^1]: https://doc.rust-lang.org/book/ch15-00-smart-pointers.html


wlawt (approved):
looks good to me!

patrick-ogrady (approved):

PR #732, opened 2024-02-14
Fix unit fees dynamics

abi87:
I believe there is an issue with the way we update unit fees here.
I understand unit fees should change in steps of `changeDenom` from a given unit fee.
This is not what happen in the code right now.
Probably the easier way to understand that there is an issue is to note that current code sums a dimensioned quantity (unitfee, which is Avax or <Token> denominated) with a dimensionless quantity (baseDelta).
I propose a change to fix it up.

### DIFF
--- a/fees/manager.go
+++ b/fees/manager.go
@@ -213,10 +213,10 @@ func (f *Manager) UnitsConsumed() Dimensions {
 func computeNextPriceWindow(
 	previous window.Window,
 	previousConsumed uint64,
-	previousPrice uint64,
-	target uint64, /* per window */
+	previousUnitFee uint64,
+	target uint64, /* per window, must be non-zero */
 	changeDenom uint64,
-	minPrice uint64,
+	minUnitFee uint64,
 	since int, /* seconds */
 ) (uint64, window.Window, error) {
 	newRollupWindow, err := window.Roll(previous, since)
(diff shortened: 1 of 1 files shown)


fees/manager.go
```diff
@@ -231,53 +231,45 @@ func computeNextPriceWindow(
...
-		delta := total - target
-		x := previousPrice * delta
-		y := x / target
-		baseDelta := y / changeDenom
-		if baseDelta < 1 {
-			baseDelta = 1
-		}
-		n, over := math.Add64(nextPrice, baseDelta)
+		rawDelta := previousUnitFee * (totalUnitsConsumed - target) / target
+		delta := max(rawDelta/changeDenom, 1) * changeDenom // price must change in increments on changeDenom
```
abi87, line 245: This is the main point: `delta` here is ensured to be a multiple of `changeDenom`. Moverover, from a dimensional analysis standpoint, delta is a "token amont" quantity, which can be added to unit fees.
In the previous version `baseDelta` is a dimensionless quantity (ration of two fees) which cannot be added to a dimensioned quantity like next price

wlawt (approved):
This makes sense to me

fees/manager.go
abi87: This is actually wrong: this code comes from coreth (see https://github.com/ava-labs/coreth/blob/58729bccbbda5fc2ffe4edc1d635d81bb2437421/consensus/dummy/dynamic_fees.go#L40) which in turns comes from EIP1559 proposal (see for instance https://github.com/ethereum/EIPs/issues/1559)

The goal of changeDenom is **not** to make sure that delta happens in multiples of the changeDenom.
Instead the idea is to cap the unit fees changes to ±1/changeDenom (1/8 = 12.5% in eth proposal).

abi87:
I misunderstood the meaning of `changeDenom`, so this PR is wrong.

Issue #666, opened 2024-01-06
[x/programs] pass error messages from host functions


samliok:
Currently if our host function throws an error we propagate it up the runtime and stop executing the program.

Defining a common `hostResponse` in the `host` and `guest` would provide a more descriptive error handling framework. A potential struct like this would give us more programability and allow us to better handle different responses in the guest. 

```go
type  MsgLevel  int
const (
	None MsgLevel  =  iota
	Info
	Error
	Panic
)
type hostResponse struct {
	value int64
	level MsgLevel
	message []byte
}
```
the host function would convert a `hostResponse` to a `smartPtr` and then return as `*types.Val` in order to match the required [function signature](https://github.com/ava-labs/hypersdk/pull/660/files#diff-9619b5dfe09284cc8c7bd4374ad3404b91e577f8c3335cb8791cf14fff0ef99bR53). Or it can stop program execution by throwing an `error`.

 For example, the `getFn` host function returns -1 if it [cannot find a key in its db](https://github.com/ava-labs/hypersdk/blob/main/x/programs/examples/imports/pstate/pstate.go#L128). It also returns a -1 if it [cannot grab bytes from memory](https://github.com/ava-labs/hypersdk/blob/main/x/programs/examples/imports/pstate/pstate.go#L110)(a more serious error). These errors are not descriptive. Even with the new refactor of #660, some errors should still be passed to the host while others can stop the runtime.

richardpringle:
Seems like this is linked with #595?

richardpringle:
See specific comment from the above PR here
https://github.com/ava-labs/hypersdk/pull/737#discussion_r1499098482

richardpringle:
Closing this in favour of #595. Host functions (with the exception of external calls) should not return `Result`.

PR #737, opened 2024-02-19
[x/programs] Support key deletion

### LINKED ISSUE #590: [x/programs] Support key deletion
Currently in wasm program state.rs only (get, store) operations are supported. To remove key/value pairs from storage, a `delete` function can be added. 

Rationale:
* For an NFT, when the owner of an NFT burns their NFT, their balance in the program state should be deleted entirely, if their new balance reaches 0.

najeal:
This PR does the following:
It enhances `state` import module and usage adding `delete` function to delete keys from the State.
It updates the Token example to use the `delete` function throught a new public program function `burn_from`.

Closes #590


x/programs/rust/wasmlanche_sdk/src/host/state.rs
```diff
@@ -38,3 +41,13 @@ pub(crate) unsafe fn get_bytes(caller: &Program, key: &Key) -> Result<i64, State
...
     Ok(unsafe { _get(caller, key) })
 }
+
+/// Deletes the bytes at key ptr from the host storage
+pub(crate) unsafe fn delete_bytes(caller: &Program, key: &Key) -> Result<(), StateError> {
+    let caller = to_host_ptr(caller.id())?;
+    let key = to_host_ptr(key)?;
+    match unsafe { _delete(caller, key) } {
+        0 => Ok(()),
+        _ => Err(StateError::Delete),
```
richardpringle, line 51: 🤔, is this the only thing that can go wrong here? What if the key doesn't exist vs a read-error etc?

x/programs/examples/imports/pstate/pstate.go
richardpringle, line 1: 👍

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -63,6 +63,16 @@ pub fn mint_to(program: Program, recipient: Address, amount: i64) -> bool {
...
 
+/// Burn the token from the recipient.
+#[public]
+pub fn burn_from(program: Program, recipient: Address) -> bool {
+    program
+        .state()
+        .delete(StateKey::Balance(recipient))
+        .expect("failed to burn recipient tokens");
+    true
+}
```
richardpringle, line 74: Can you add a test here? This is kind of an odd example IMO, usually you would burn a specific amount of tokens, not the entire balance. 

A simulator test makes the intended behaviour crystal clear

x/programs/examples/token_test.go
```diff
@@ -8,18 +8,79 @@ import (
...
+		require.Equal(int64(1000), result[0])
+
+		// burn alice tokens
+		_, err = rt.Call(ctx, "burn_from", programIDPtr, alicePtr)
+		require.NoError(err)
+
+		// check balance of alice
+		result, err = rt.Call(ctx, "get_balance", programIDPtr, alicePtr)
+		require.NoError(err)
+		require.Equal(int64(0), result[0])
```
richardpringle, line 81: 🤔, this test doesn't guarantee that the behaviour is as intended. We should check the state and ensure the key is no longer there.

x/programs/rust/wasmlanche_sdk/src/state.rs
richardpringle, line 1: 👍

richardpringle (changes requested):
Hey @najeal, thanks for the PR! 

We're starting to get _a little_ more strict with reviews here, so I apologize if you followed an existing pattern and I asked you to change something. 

The change requests should be simple enough, though!

Feel free to let me know if you have any questions.

x/programs/rust/wasmlanche_sdk/src/host/state.rs
najeal: Following errors can occur golang side:
1)Fail to access caller memory
2)Fail to access/read program & key pointer
3)Fail to Remove from the State

Actually, any returned error by the extern golang function will result in a panic.
The only error I could handle is the third scenario returning `-1`, just like the `get_bytes`. But I cannot add more details than `StateError::Delete`
-> Going to handle this scenario.
There is the same problem with the `put_bytes` method returning `StateError::Write`. It will never happens.

https://github.com/ava-labs/hypersdk/issues/666 looks like this issue is talking about error enhancement.

x/programs/rust/examples/token/src/lib.rs
najeal: I can rename it as `burn_all` or `burn_total` if you prefer.
It is mostly to test on golang side the usage of the `delete` function.

richardpringle: Don't worry about the rename, thanks for adding the test!

richardpringle (approved):

richardpringle:
@najeal, looks like you need to run `cargo fmt`

richardpringle:
@najeal, can you mention me in a comment if I forget about this? Everything looks good, we just need CI to pass

najeal:
@richardpringle CI passed 🙂

PR #743, opened 2024-02-23
simulator tests

richardpringle:
- **Rename wasmlanche to use - instead of _**
- **Specify the default target as wasm for  examples**
- Make `cargo test` run simulator test too

PR is a work in progress still. What we needed is CI that will test all the examples on every run instead of having to specify a separate job. This was a bit of a pain to get working so the simulator binary is written in Go and we need to make sure we compile it every time. It also requires build scripts for every example which sucks, but all the logic has been extracted to the SDK.

This really isn't a scalable form of testing. In an ideal world, all of the tests would actually also be compiled in wasm, sent to the simulator to run, then the results would be returned. `wasm_bindgen` does something along these lines, but relies less on external parameters such as "units" to calculate the cost of compute.

In any case, this unlocks us to be able to add examples and have the wasm binaries run in the simulator by simple writing tests in `lib.rs` and including the exact same `build.rs` script as in the `token` example.


x/programs/rust/examples/token/src/lib.rs
```diff
@@ -120,24 +121,44 @@ pub fn get_balance(program: Program, recipient: Address) -> i64 {
         .unwrap_or_default()
 }
 
+#[cfg(simulator_test)]
+mod tests {
+    #[test]
+    fn hello() {
+        dbg!("hello world!");
+    }
+}
```
richardpringle, line 130:
```suggestion
```

.github/workflows/rust-ci.yml
```diff
@@ -50,30 +50,40 @@ jobs:
...
-          cache: true
-      - name: Install Rust
-        uses: dtolnay/rust-toolchain@stable
-      - name: Run token simulator test
-        working-directory: ./x/programs/rust/examples/token
-        shell: bash
-        run: scripts/tests.simulator.sh
-        env:
-          MODE: 'full-test'
+      # TODO: deleteme
```
richardpringle, line 72: Don't forget about me!

x/programs/cmd/simulator/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "simulator"
+version = "0.1.0"
+edition = "2021"
+include = ["**/*.go"]
+
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
richardpringle, line 8: delete me

x/programs/cmd/simulator/build.rs
```diff
@@ -0,0 +1,50 @@
+use std::process::Command;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=/build");
+    println!("cargo:rerun-if-changed=**/*.go");
```
richardpringle, line 6: Not 100% sure this works as expected

x/programs/cmd/simulator/build.rs
```diff
@@ -0,0 +1,50 @@
...
+    println!("cargo:warning={status}");
+
+    println!("cargo:warning=building simulator...");
+
+    let current_dir = std::env::current_dir().unwrap();
+
+    let simulator_path = "bin/simulator";
+    let simulator_src = "simulator.go";
+
+    // resolve absolute path for simulator_path and create dir if it doesn't exist
```
richardpringle, line 21: delete me

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 use borsh::{BorshDeserialize, BorshSerialize};
 use wasmlanche_sdk::{program::Program, public, state_keys, types::Address};
+// new things
```
richardpringle, line 3: delete me

richardpringle:
@patrick-ogrady, can you remove the simulator-tests from the "Required" checks list and add the `Rust CI / fmt` check, please?

x/programs/cmd/simulator/cmd/root.go
```diff
@@ -89,7 +91,7 @@ func NewRootCmd() *cobra.Command {
 func (s *simulator) Init() error {
 	homeDir, err := os.UserHomeDir()
 	if err != nil {
-		return nil
+		return err
```
patrick-ogrady, line 94: Interesting that we didn't have a linter catch this 🤔

richardpringle: This was all @darioush when he was helping me debug

x/programs/cmd/simulator/cmd/root.go
```diff
@@ -114,23 +116,24 @@ func (s *simulator) Init() error {
...
 	sk, err := bls.NewSecretKey()
 	if err != nil {
-		return nil
+		return err
 	}
 
 	// setup pebble and db manager
-	pdb, _, err := pebble.New(dbPath, pebble.NewDefaultConfig())
+	// pdb, _, err := pebble.New(dbPath, pebble.NewDefaultConfig())
+	pdb := memdb.New()
```
wlawt, line 129: just curious, should we update the comment and remove the pebble instantiation?

richardpringle: Yes, will do!

x/programs/cmd/simulator/cmd/root.go
```diff
@@ -148,13 +151,15 @@ func (s *simulator) Init() error {
 
 	// initialize the simulator VM
 	vm := controller.New()
+	configBytes := []byte("{\"testMode\": true}")
```
richardpringle, line 154: I don't know if this is doing anything, I can delete

x/programs/cmd/simulator/cmd/root.go
richardpringle: Might actually be able to just use `nil`, not sure if this is actually used, will test.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -3,17 +3,15 @@
 //! Alternatively the `Plan` can be written in JSON and passed to the
 //! Simulator binary directly.
 
+use serde::{Deserialize, Serialize};
```
richardpringle, line 6: Write comments `lib.rs` (here) describing that this file is used to wrap the Go cli-tool

wlawt (approved):
Looks good to me, ran the `token` example - I think updating the `token` docs on how to test and troubleshoot may be a good idea

x/programs/cmd/simulator/cmd/root.go
```diff
@@ -124,9 +122,6 @@ func (s *simulator) Init() error {
 		return err
 	}
 
-	// setup pebble and db manager
-	// pdb, _, err := pebble.New(dbPath, pebble.NewDefaultConfig())
-	pdb := memdb.New()
 	if err != nil {
 		return err
 	}
```
darioush, line 127:
```suggestion
```
not needed

darioush (dismissed):

richardpringle:
> Looks good to me, ran the `token` example - I think updating the `token` docs on how to test and troubleshoot may be a good idea

I actually ended up updating the docs to the simulator since I changed how it works. I agree that we need better documentation, but I think I'll do that when I get the example to a real usable state.

x/programs/cmd/simulator/cmd/root.go
richardpringle: Created an issue
https://github.com/ava-labs/hypersdk/issues/752

x/programs/cmd/simulator/README.md
```diff
@@ -1,6 +1,14 @@
-# Program VM simulator
+# Program VM simulator (work in progress)
 
-## Introduction
+The `Simulator` is currently a work-in-progres. It is a cli-tool built in Go, but meant to be used from the `wasmlanche-sdk` as a way to run automated tests on your HyperSDK wasm programs. For relatively up-to-date documentation, run `cargo doc --no-deps -p simulator --open`. Create a `Client`, create a `Plan`, then call `client.execute(plan)`, to execute the plan.
```
patrick-ogrady, line 3: nit: progress

x/programs/cmd/simulator/cmd/root.go
```diff
@@ -148,10 +142,11 @@ func (s *simulator) Init() error {
 
 	// initialize the simulator VM
 	vm := controller.New()
+
 	err = vm.Initialize(
```
patrick-ogrady, line 146: I think there is a future where we rip this out and just stub the host function to the WASM engine?

richardpringle: 100% agree. I don't think this is worth doing right at this moment though. I'll create an issue.

richardpringle: https://github.com/ava-labs/hypersdk/issues/756

wlawt (approved):

PR #744, opened 2024-02-24
[chain/processor] fetch key values on-the-fly

### LINKED ISSUE #749: [chain/processor] address cache stampede
we should only go to disk once on concurrent reads and store it in cache

wlawt:
Closes #749 

### Overview
This PR addresses the cache stampede problem that occurs with concurrent reads to a key. The key may not be in the cache, resulting in us to go to disk. But, we only want to go to disk once and update the cache so that subsequent reads will fetch from cache. 

### Changes
We introduce the `Fetcher` package that will fetch values and store them into the cache on-the-fly. However, if the key is already in cache, then we can say we already fetched this value and reduce the count of the number of keys a given transaction is waiting on.

This also reduces the logic contained within `chain/processor.go` and is replaced by a simple `fetcher.Lookup(..)` and `fetcher.Wait(..)` call.

### Review Guide
- `chain/processor.go`: new, reduced logic to lookup keys
- `fetcher/fetcher.go`: core logic 
   - `Lookup(..)`: enqueues keys for the workers to fetch
   - `Wait(..)`: blocks a txn until all keys are fetched then return reads and storage from cache
- `fetcher/fetcher_tests.go`: unit tests

### Tests
- Added unit tests to `Fetcher` with ~90% coverage
- Integration tests with MorpheusVM and TokenVM pass
- CI

### TODO
- [x] More unit tests
- [x] Add `Stop()`
- [x] Handle bubbling up errors from workers


fetcher/fetcher.go
```diff
@@ -0,0 +1,155 @@
...
+		totalTxns:   numTxs,
+	}
+	for i := 0; i < concurrency; i++ {
+		f.createWorker()
+	}
+	return f
+}
+
+func (f *Fetcher) createWorker() {
+	go func() {
```
patrick-ogrady, line 71: nit: recommend calling `go f.createWorker` above so that you don't need to put all the logic so nested.

fetcher/fetcher.go
```diff
@@ -0,0 +1,155 @@
...
+		stop:        make(chan struct{}), // TODO: implement Stop()
+		totalTxns:   numTxs,
+	}
+	for i := 0; i < concurrency; i++ {
+		f.createWorker()
+	}
+	return f
+}
+
+func (f *Fetcher) createWorker() {
```
patrick-ogrady, line 70: nit: call this `runWorker` (it isn't just creating)

chain/processor.go
```diff
@@ -58,50 +56,27 @@ func (b *StatelessBlock) Execute(
 			e.Stop()
 			return nil, nil, err
 		}
+		// Fetch keys from disk or check if it's in cache
+		wg := f.Lookup(ctx, tx.ID(), stateKeys)
 		e.Run(stateKeys, func() error {
+			// Block until worker finishes fetching keys
+			wg.Wait()
```
patrick-ogrady, line 63: I think this would be a lot cleaner if `wg.Wait()` returned `reads` and `storage`

chain/processor.go
```diff
@@ -40,8 +38,8 @@ func (b *StatelessBlock) Execute(
 		sm        = b.vm.StateManager()
 		numTxs    = len(b.Txs)
 		t         = b.GetTimestamp()
-		cacheLock sync.RWMutex
-		cache     = make(map[string]*fetchData, numTxs)
+		fetchLock sync.RWMutex
```
patrick-ogrady, line 41: I'm not sure what this lock does? the fetcher could still be concurrently writing when you are reading from `f.Cache`?

chain/processor.go
```diff
@@ -58,50 +56,27 @@ func (b *StatelessBlock) Execute(
...
-				toCache = make(map[string]*fetchData, len(toLookup))
-				for _, k := range toLookup {
-					v, err := im.GetValue(ctx, []byte(k))
-					if errors.Is(err, database.ErrNotFound) {
-						reads[k] = 0
-						toCache[k] = &fetchData{nil, false, 0}
-						continue
-					} else if err != nil {
-						return err
+				if v, ok := f.Cache[k]; ok {
```
patrick-ogrady, line 72: `f.Cache` should be private

fetcher/fetcher.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+// Fetcher retrieves values on-the-fly and
+// ensures that a value is only fetched from
+// disk once. Subsequent requests can be retrieved
+// from cache.
+type Fetcher struct {
+	im        state.Immutable
+	cacheLock sync.RWMutex
+	Cache     map[string]*FetchData
```
patrick-ogrady, line 25: this should be a private variable.

fetcher/fetcher.go
```diff
@@ -0,0 +1,155 @@
...
+					if _, ok := f.Cache[k]; ok {
+						f.TxnsToFetch[t.id].Done()
+						f.cacheLock.RUnlock()
+						continue
+					}
+					f.cacheLock.RUnlock()
+
+					// Fetch from disk that aren't already in cache
+					// We only ever fetch from disk once
+					v, err := f.im.GetValue(t.ctx, []byte(k))
```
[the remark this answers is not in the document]
patrick-ogrady: The other map would be something like `map[string][]ids.ID` (where `ids.ID` are all the txs to call `Done` on when it is fetched).

chain/processor.go
patrick-ogrady: As mentioned below, `f.Cache` should be private (and access must be synchronized within `fetcher` to prevent panic)

fetcher/fetcher.go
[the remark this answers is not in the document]
patrick-ogrady: consider the case where 2 transactions are trying to fetch the same key (not in the cache) at the same time?

fetcher/fetcher.go
```diff
@@ -0,0 +1,155 @@
...
+// Fetcher retrieves values on-the-fly and
+// ensures that a value is only fetched from
+// disk once. Subsequent requests can be retrieved
+// from cache.
+type Fetcher struct {
+	im        state.Immutable
+	cacheLock sync.RWMutex
+	Cache     map[string]*FetchData
+
+	TxnsToFetch map[ids.ID]*sync.WaitGroup // Number of keys a txn is waiting on
```
patrick-ogrady, line 27: This should also be private

fetcher/fetcher.go
```diff
@@ -0,0 +1,155 @@
...
+			case <-f.stop:
+				return
+			}
+		}
+	}()
+}
+
+// Lookup enqueues a set of stateKey values that we need to lookup, and
+// returns a WaitGroup for a given transaction.
+func (f *Fetcher) Lookup(ctx context.Context, txID ids.ID, stateKeys state.Keys) *sync.WaitGroup {
```
patrick-ogrady, line 138: Note: this is not safe to call concurrently

fetcher/fetcher.go
```diff
@@ -0,0 +1,155 @@
...
+			select {
+			case t, ok := <-f.fetchable:
+				if !ok {
+					return
+				}
+				for _, k := range t.toLookup {
+					// Allow concurrent reads to cache
+					f.cacheLock.RLock()
+					if _, ok := f.Cache[k]; ok {
+						f.TxnsToFetch[t.id].Done()
```
patrick-ogrady, line 82: You need to synchronize access to `TxnsToFetch` otherwise it could panic (you are reading a value here while another thread may add an id to it)

fetcher/fetcher.go
```diff
@@ -0,0 +1,215 @@
...
+		t := &task{
+			ctx: ctx,
+			key: k,
+		}
+		// Release the lock to avoid deadlock when
+		// calling updateDependencies
+		f.keysFetchLock.Unlock()
+		f.fetchable <- t
+	}
+	return f.txnsToFetch[txID]
```
wlawt, line 185: Will this cause a panic if I'm not holding the `txnsFetchLock` here?

fetcher/fetcher.go
```diff
@@ -0,0 +1,215 @@
...
+func (f *Fetcher) updateCache(k string, v []byte, exists bool, chunks uint16) {
+	f.cacheLock.Lock()
+	defer f.cacheLock.Unlock()
+	f.cache[k] = &fetchData{v, exists, chunks}
+}
+
+// For a key that was fetched from disk or was already in cache, we decrement
+// the count for the txID that was waiting on that key. This function is
+// only called by the worker.
+func (f *Fetcher) updateDependencies(k string) {
```
wlawt, line 147: nit: is there a better name?

wlawt: maybe `updateQueues`?

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+func (db *testDB) GetValue(_ context.Context, key []byte) (value []byte, err error) {
+	val, ok := db.storage[string(key)]
+	if !ok {
+		return nil, database.ErrNotFound
+	}
+	return val, nil
+}
+
+func TestFetchSameKeysConcurrentBasic(t *testing.T) {
```
wlawt, line 51: Would be a good idea to incorporate the executor and pass `f.Wait()` into `e.Run(..)` for better testing

wlawt: Think this is indirectly blocked by #715 since the executor still doesn't know how to handle r/w keys in `main`

wlawt: For now, the unit tests in `fetcher_test.go` will closely mimic its usage in executor using go routines

fetcher/fetcher.go
```diff
@@ -0,0 +1,242 @@
...
+		case t, ok := <-f.fetchable:
+			if !ok {
+				return
+			}
+
+			// Checks the cache twice. In isFirst, we will wait if it's not
+			// in the cache and not the first tx requesting the key. We have
+			// the second check to ensure when the tx is unblocked, it can
+			// fetch from cache.
+			f.isFirst(t)
```
wlawt, line 89: I'm thinking about it a bit more and I don't think we need to check the cache twice? Either it's not in the cache AND not the first tx requesting the key so we will block. When we get unblocked, we know that we're only getting unblocked because all of our keys were fetched so we can just `continue`.

However, if it's in the cache then we just proceed as usual and decrement the counts using `f.updateDependencies`. This saves an extra call to `f.has(t.key)`

chain/dependencies.go
```diff
@@ -79,6 +79,7 @@ type VM interface {
 	IsRepeat(context.Context, []*Transaction, set.Bits, bool) set.Bits
 	GetTargetBuildDuration() time.Duration
 	GetTransactionExecutionCores() int
+	GetKeyStorageConcurrency() int
```
patrick-ogrady, line 82: nit `GetStateFetchConcurrency`

chain/processor.go
```diff
@@ -138,18 +95,12 @@ func (b *StatelessBlock) Execute(
...
-				cacheLock.Lock()
-				for k := range toCache {
-					cache[k] = toCache[k]
-				}
-				cacheLock.Unlock()
-			}
 			return nil
 		})
 	}
+	if err := f.HandleErrors(); err != nil {
```
patrick-ogrady, line 101: I think you'd want to call this after `e.Wait()` because we aren't guaranteed the fetcher is done before then.

fetcher/fetcher.go
```diff
@@ -0,0 +1,237 @@
...
+		case <-f.stop:
+			return
+		}
+	}
+}
+
+// Checks if a key is in the cache.
+func (f *Fetcher) has(k string) bool {
+	f.cacheLock.RLock()
+	defer f.cacheLock.RUnlock()
```
patrick-ogrady, line 137: nit: leave a `\n` after a defer like this

fetcher/fetcher.go
```diff
@@ -0,0 +1,237 @@
...
+	for _, id := range txIDs {
+		f.txnsToFetch[id].Done() // Notify all other txs
+	}
+	f.txnLock.Unlock()
+	f.keysToFetch[k] = nil
+}
+
+// Lookup enqueues keys for the workers to fetch
+func (f *Fetcher) Lookup(ctx context.Context, txID ids.ID, stateKeys state.Keys) *sync.WaitGroup {
+	f.txnLock.Lock()
```
patrick-ogrady, line 166: nit: you can create this wg before acquiring the lock and adding to the map (more efficient).

You should do:
```
wg := &sync.WaitGroup{}
wg.Add(len(stateKeys))
f.txnLock.Lock()
f.txnsToFetch[txID] = wg
f.txnLock.Unlock()
```

fetcher/fetcher.go
```diff
@@ -0,0 +1,237 @@
...
+	wg := &sync.WaitGroup{}
+	f.txnsToFetch[txID] = wg
+	f.txnsToFetch[txID].Add(len(stateKeys))
+	f.txnLock.Unlock()
+
+	f.keyLock.Lock()
+	tasks := make([]*task, 0, len(stateKeys))
+	for k := range stateKeys {
+		if _, ok := f.keysToFetch[k]; !ok {
+			f.keysToFetch[k] = make([]ids.ID, 0)
```
patrick-ogrady, line 176: While this approach is correct, I think it would be more efficient to avoid requeueing dependencies if the key has already been fetched. Seems odd to track them in that case.

patrick-ogrady: I think this would involve decrementing the `wg` count here or waiting to create the `wg` var until after this loop so we know what we are waiting on.

fetcher/fetcher.go
```diff
@@ -0,0 +1,237 @@
...
+				continue
+			}
+
+			// Doesn't exist in cache, check if we were the first to
+			// request the key.
+			f.keyLock.RLock()
+			firstTx := f.keysToFetch[t.key][0]
+			f.keyLock.RUnlock()
+			if firstTx != t.id {
+				t.wg.Wait()
```
patrick-ogrady, line 97: We should fetch other keys while we are in the work, not wait for this key to be notified if someone else is going to get it.

patrick-ogrady: I think we can just remove the `Wait()` here and just continue.

fetcher/fetcher.go
patrick-ogrady: You could handle this by using a `struct` as the value of `keysToFetch` that can communicate that the key was fetched to avoid more locks.

chain/processor.go
```diff
@@ -55,53 +51,14 @@ func (b *StatelessBlock) Execute(
 
 		stateKeys, err := tx.StateKeys(sm)
 		if err != nil {
+			f.Stop()
 			e.Stop()
 			return nil, nil, err
 		}
+		// Fetch keys from disk or check if it's in cache
+		wg := f.Lookup(ctx, tx.ID(), stateKeys)
```
patrick-ogrady, line 59: We should note here that we fetch all keys from disk (whether or not they are read/write/allocate perms because we use the existing value on-disk for fee calculation)

patrick-ogrady: https://github.com/ava-labs/hypersdk/pull/751#discussion_r1513480075

If we require this, then we don't need to leave this comment.

chain/processor.go
```diff
@@ -55,53 +51,14 @@ func (b *StatelessBlock) Execute(
 
 		stateKeys, err := tx.StateKeys(sm)
 		if err != nil {
+			f.Stop()
 			e.Stop()
 			return nil, nil, err
 		}
+		// Fetch keys from disk or check if it's in cache
```
patrick-ogrady, line 58: I thought about this some more and this will actually be a good bit faster than what we currently do because we'll start the lookup right away rather than waiting for a transaction to be considered executable.

While one could argue this means we are prioritizing disk reads that aren't ready to execute (if a set of keys is blocked serially), it isn't really the case because all but one of those reads will be from a cache.

fetcher/fetcher.go
```diff
@@ -0,0 +1,258 @@
...
+				continue
+			} else if err != nil {
+				f.stopOnce.Do(func() {
+					f.err = err
+					close(f.stop)
+				})
+				return
+			}
+
+			// We verify that the [NumChunks] is already less than the number
```
patrick-ogrady, line 124: this comment doesn't match what we do below (where we check `ok`)?

fetcher/fetcher.go
```diff
@@ -0,0 +1,258 @@
...
+	for {
+		select {
+		case t, ok := <-f.fetchable:
+			if !ok {
+				return
+			}
+
+			// Allows concurrent reads to cache.
+			if exists := f.has(t.key); exists {
+				f.updateDependencies(t.key)
```
patrick-ogrady, line 93: You don't need to updateDependencies here anymore. The queue will already be set to be empty by the time it is fetched.

fetcher/fetcher.go
```diff
@@ -0,0 +1,258 @@
...
+				continue
+			}
+
+			// Doesn't exist in cache, check if we were the first to
+			// request the key.
+			f.keyLock.RLock()
```
patrick-ogrady, line 99: This will be easier to reason about if you merge the `cacheLock`/`keyLock`. Then you can check if the key is in the cache and whether you need to fetch atomically.

fetcher/fetcher.go
```diff
@@ -0,0 +1,258 @@
...
+			f.keyLock.RUnlock()
+			if firstTx != t.id {
+				// Fetch other keys instead
+				continue
+			}
+
+			// Fetch from disk once if key isn't already in cache
+			v, err := f.im.GetValue(t.ctx, []byte(t.key))
+			if errors.Is(err, database.ErrNotFound) {
+				f.updateCache(t.key, nil, false, 0)
```
patrick-ogrady, line 113: You can merge `updateCache` and `updateDependencies` since you never call them apart.

fetcher/fetcher.go
```diff
@@ -0,0 +1,258 @@
...
+			key: k,
+		}
+		tasks = append(tasks, t)
+	}
+	f.keyLock.Unlock()
+
+	// Create wg based on number of keys we need to wait on
+	wg.Add(len(tasks))
+
+	for _, t := range tasks {
```
patrick-ogrady, line 210: You need to ensure the fetcher isn't stopped before doing this (otherwise it will panic for adding to a closed channel)

fetcher/fetcher.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// New creates a new [Fetcher]
+func New(numTxs int, concurrency int, im state.Immutable) *Fetcher {
+	f := &Fetcher{
+		im:          im,
+		cache:       make(map[string]*fetchData, numTxs),
+		keysToFetch: make(map[string]*keyData),
+		txnsToFetch: make(map[ids.ID]*sync.WaitGroup, numTxs),
+		fetchable:   make(chan *task),
```
patrick-ogrady, line 69: Interesting choice to make this a non-buffered channel. I don't think we should go this route (as it could really impact the concurrency of execution if `Lookup` is blocking).

fetcher/fetcher.go
wlawt: Oops I borrowed it when copying over the logic from https://github.com/ava-labs/hypersdk/blob/main/chain/processor.go#L94-L102 , however, I removed the comment since I think the `ok` check and what follows is clear

wlawt:
Please see #765 (no longer working off of the fork of the HyperSDK)

Issue #734, opened 2024-02-14
[x/programs] Improve Simulator User experience and optimize for automated tests alongside program code. 


richardpringle:
Currently, the Rust simulator client uses the simulator as a command line tool and parses JSON from the simulator's stdout. This isn't ideal if we want to log anything other than JSON output from the simulator while debugging and extending it. I'm open to ideas other than GRPC, but I think that would be the most convenient way to interop between a Go tool and the Rust tooling. The types for communication can simply be defined once with protobufs and the Go and Rust code can be generated from there. Right now, the interface (JSON) is all duplicated in both Rust and Go.  

Further, the ideal way to use the simulator is actually to write your tests in Rust directly alongside your program. Would it make more sense to just wrap the simulator code into a Rust library such that a separate binary doesn't need to be executed? One should just be able to run `cargo test` and get everything they need.

richardpringle:
Before we start working on this issue, there are a couple of things to discuss. 

Should the simulator be ephemeral, or should it be long-lived? Depending on how difficult it is to spin up a local Hyper(Chain|VM), I prefer that the simulator be ephemeral.

Ideally, the simulator would have a single API, meaning the CLI should use the same GRPC interface. That means that the simulator code used by the wasmlanche-sdk can spin up and tear down the GRPC server, AND the cli-client lives in the same crate (migrating the interface to Rust). IMO, this is the easiest way to maintain the simulator. When the protos are updated, the test-client has to be updated along with the cli; features will never land in one place before the other.

That brings me to another question: do we even need a CLI for the simulator? <- we will have to see, I think it's too early to tell at this point.

Issue #722, opened 2024-02-09
[x/programs] Cache state-values such that writes can happen lazily on context switch


richardpringle:
If `State` caches all key-value reads and writes, we can defer all the writes (to prevent unnecessary context switches). When an external program is called, you should have to pass the context into the external call, where we can make sure that all writes are flushed to the host. We can also trigger any pending writes on `Drop`.

Issue #572, opened 2023-10-18
[x/programs] Add program example: Multisig wallet 


exdx:
Add an example of a basic multisig wallet, entirely in Rust.

najeal:
@richardpringle I'm going to look at this one first ! I think it can be an interesting one.

If I understand well, two interesting examples would be:
1 - Using the Multisig wallet as an account:
- Transfering token to the Mulstisig Wallet using the ProgramdID as an address
- Then transfering from the Multisig Wallet to another address by calling the Token Program when minimum owners have confirmed. This needs program to program call (As you said I need to wait for Context Providing being implemented in an open PR)
2-  Using the MultisigWallet has the Token authority and a proxy to the managing commands like burn/mint etc...

Issue #505, opened 2023-10-03
[x/programs] Add program example: NFT Token


exdx:
Develop a basic NFT token example
* Based on the latest storage updates (#499)
* May need to be rebased on top of wasmtime runtime migration

The program will be exploratory in nature as several outstanding questions still remain about the structure of programs. It will likely be revised several times. 

The token will use the standard NFT metadata schema defined in EVM-based chains for simplicity: https://nftschool.dev/reference/metadata-schemas/#ethereum-and-evm-compatible-chains

PR #747, opened 2024-02-28
[x/programs] poc - simulator over grpc

najeal:
This Draft PR is related to #734 .
It does the following:

- create a new binary simulator_api alongside the simulator reusing same engine but add grpc transport over it.
- define the grpc proto files in simulator_api
- create a rust lib with generated client from the grpc proto file. (it cannot be embedded in `wasmlanche_sdk` because of some issue building dependencies with `wasm32` target
- define a short example of how to use the client in `counter` tests.

Implemented commands:

- [x] createKey
- [ ] createProgram
- [ ] run

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ members = [
     "x/programs/rust/sdk_macros",
     "x/programs/rust/wasmlanche_sdk",
     "x/programs/rust/examples/token",
-    "x/programs/rust/examples/counter",
+    "x/programs/rust/examples/counter", "x/programs/rust/simulator_client",
 ]
 resolver = "2"
 
--- a/x/programs/cmd/simulator/cmd/root.go
+++ b/x/programs/cmd/simulator/cmd/root.go
@@ -4,41 +4,12 @@
 package cmd
 
 import (
-	"context"
-	"encoding/json"
 	"fmt"
 	"os"
-	"path"
 
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
-
-	"github.com/ava-labs/avalanchego/api/metrics"
-	"github.com/ava-labs/avalanchego/ids"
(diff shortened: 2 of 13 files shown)


richardpringle:
Hey @najeal, thanks for jumping on #734! 

Unfortunately, that issue has yet to be fully fleshed out 😞. 

This PR also conflicts with #743, so even if #734 were fully fleshed out, I wouldn't be able to merge this first. 

I'll leave another comment on #734 with some of my thoughts in case you do end up taking that issue on when it's ready. In the meantime, want me to find you a different issue that would be helpful? 

Thanks again for the contribution! This actually looks pretty good; there are just a few more things about the simulator that need to be cleaned up before we take on this issue.

najeal:
No problem, I understand! I will close the PR for now.  I didn't want to spam you, that's why I picked this issue by myself.
Yes I would be glad if you find another issue @richardpringle  ! Thank !

richardpringle:
@najeal 
>I didn't want to spam you
Spam away, my friend. Open a discussion and `@` mention me, mention me on an issue, I'm always happy to respond! (Even if it takes me a couple of days).

As for issues that are isolated, there aren't many. https://github.com/ava-labs/hypersdk/issues/505 and https://github.com/ava-labs/hypersdk/issues/572 come to mind, but you won't really be able to get the API looking the way you want it until https://github.com/ava-labs/hypersdk/pull/710 is done (I'm going to pick that up next). 

Another interesting issue would be https://github.com/ava-labs/hypersdk/issues/722. I don't mind if you pick that one up, but you would have to be super patient with the review process as I'm not 100% how I want/need it to work yet. I think it would be pretty straightforward if you don't consider program to program calls, but the API might have to change to do that. 

Let me know! Feel free to open a discussion and `@`-mention me. Or feel free to comment on any of those issue.

PR #748, opened 2024-02-28
simulator tests parallel

richardpringle:
Follow up on https://github.com/ava-labs/hypersdk/pull/743. I added a `--cleanup` flag (open to new names) on the simulator to make the db-directory ephemeral. The simulator's node-id is appended to the db-directory name. 
This is needed to run tests from multiple examples at the same time. 

Future work: 
Add a flag to pass in the directory name so one can use the same directory between steps. Also might make sense to pass in directory name (or node-id) so that the same directory can be persisted throughout a single test.

### DIFF
--- a/x/programs/cmd/simulator/cmd/root.go
+++ b/x/programs/cmd/simulator/cmd/root.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -38,6 +39,7 @@ type simulator struct {
 	vm      *vm.VM
 	db      *state.SimpleMutable
 	genesis *genesis.Genesis
+	cleanup func()
 }
 
 func NewRootCmd() *cobra.Command {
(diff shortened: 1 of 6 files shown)


x/programs/cmd/simulator/src/lib.rs
```diff
@@ -190,39 +221,10 @@ impl Client {
     /// # Errors
     ///
     /// Returns an error if the if serialization or plan fails.
-    pub fn run<T>(&self, plan: &Plan) -> Result<Vec<T>, Box<dyn Error>>
-    where
-        T: serde::de::DeserializeOwned + serde::Serialize,
```
richardpringle, line 195: This was a premature abstraction. We shouldn't make things generic when there is only one type that can be used.

x/programs/cmd/simulator/cmd/root.go
```diff
@@ -81,6 +85,12 @@ func NewRootCmd() *cobra.Command {
 				)
 			}
 		}
+
+		// if the --cleanup flag is set, remove the simulator directory
```
richardpringle, line 89:
```suggestion
```

darioush (approved):

PR #751, opened 2024-02-29
[tstate] handle Allocate in Insert

### LINKED ISSUE #750: [tstate] handle Allocate separately from Write in tstate
Specifically in `Insert` we need to handle the appropriate key permissions (Write vs. Allocate).

wlawt:
Closes #750

### Overview
Within `Insert`, if the key doesn't exist, check to see if the key has `Write` and `Allocate` permissions.

### Tests
- Added unit tests in `tstate_test.go`
- Integration test pass (TokenVM + MorpheusVM)
- CI

### TODO
- [x] Tune key access in TokenVM and MorpheusVM actions


tstate/tstate_test.go
```diff
@@ -831,3 +831,80 @@ func TestUpdatingKeyPermission(t *testing.T) {
...
+		key         string
+		permission1 state.Permissions
+		permission2 state.Permissions
+		shouldFail  bool
+	}{
+		{
+			name:        "key has R then W",
+			key:         "test",
+			permission1: state.Read,
+			permission2: state.Write,
```
darioush, line 847: I don't think this test needs permisison1 and permission2, a single permission should be good enough.

darioush (dismissed):
lgtm

chain/transaction.go
```diff
@@ -149,7 +149,7 @@ func (t *Transaction) StateKeys(sm StateManager) (state.Keys, error) {
 	if t.Action.OutputsWarpMessage() {
 		p := sm.OutgoingWarpKeyPrefix(t.id)
 		k := keys.EncodeChunks(p, MaxOutgoingWarpChunks)
-		stateKeys.Add(string(k), state.Write)
+		stateKeys.Add(string(k), state.Write|state.Allocate)
```
patrick-ogrady, line 152: Nit: I think we should do Allocate|Write

tstate/tstate_view.go
```diff
@@ -234,12 +230,20 @@ func (ts *TStateView) Insert(ctx context.Context, key []byte, value []byte) erro
 			// No change, so this isn't an op.
 			return nil
 		}
+		// Updating just requires a Write permission
+		if !ts.checkScope(ctx, key, state.Write) {
```
patrick-ogrady, line 234: We need to check scope before calling `getValue`

tstate/tstate_view.go
```diff
@@ -213,8 +213,8 @@ func (ts *TStateView) isUnchanged(ctx context.Context, key string, nval []byte,
 // Insert allocates and writes (or just writes) a new key to [tstate]. If this
 // action returns the value of [key] to the parent view, it reverts any pending changes.
 func (ts *TStateView) Insert(ctx context.Context, key []byte, value []byte) error {
-	// Inserting requires a Write Permissions, so we pass state.Write
-	if !ts.checkScope(ctx, key, state.Write) {
+	// To insert, the key needs to have Allocate or Write permissions, or both
+	if !ts.checkScope(ctx, key, state.Allocate) && !ts.checkScope(ctx, key, state.Write) {
```
patrick-ogrady, line 217: Is there a way to just do Allocate | Write for this check rather than calling checkScope twice?

tstate/tstate_view.go
```diff
@@ -230,13 +230,21 @@ func (ts *TStateView) Insert(ctx context.Context, key []byte, value []byte) erro
 		pastWrites:    chunks(ts.writes, k),
 	}
 	if exists {
+		// Updating just requires a Write permission
```
patrick-ogrady, line 233: I wonder if we need to be more careful with the use of `getValue` in this function now that we aren't guaranteed to have read it...

I need to think about this a bit more.

patrick-ogrady: We can't really remove it because we need it for the exists check.

tstate/tstate_view.go
wlawt: Perhaps the `perm` parameter of `checkScope` could be variadic? I introduced a change in `checkScope` that uses this, but I'm not sure if that is the cleanest way. Another way is to maybe support a helper function called `Contains` in `state`

tstate/tstate_view.go
```diff
@@ -213,8 +222,8 @@ func (ts *TStateView) isUnchanged(ctx context.Context, key string, nval []byte,
 // Insert allocates and writes (or just writes) a new key to [tstate]. If this
 // action returns the value of [key] to the parent view, it reverts any pending changes.
 func (ts *TStateView) Insert(ctx context.Context, key []byte, value []byte) error {
-	// Inserting requires a Write Permissions, so we pass state.Write
-	if !ts.checkScope(ctx, key, state.Write) {
+	// To insert, the key needs to have Allocate or Write permissions, or both
+	if !ts.checkScope(ctx, key, state.Allocate, state.Write) {
```
patrick-ogrady, line 226: You can just check this using the `&` operator?

if `desired & provided == desired`

tstate/tstate_view.go
```diff
@@ -162,8 +162,17 @@ func (ts *TStateView) KeyOperations() (map[string]uint16, map[string]uint16) {
 }
 
 // checkScope returns whether [k] is in scope and has appropriate permissions.
-func (ts *TStateView) checkScope(_ context.Context, k []byte, perm state.Permissions) bool {
-	return ts.scope[string(k)].Has(perm)
+func (ts *TStateView) checkScope(_ context.Context, k []byte, perms ...state.Permissions) bool {
+	// At most only 2 key permissions are passed that we need to check. The case of
+	// 1 permission may be a single key permission or a combination (i.e., Allocate|Write).
+	switch len(perms) {
```
patrick-ogrady, line 168: Suggested way to use bit ops for this below...even if we did keep this, I would recommend iterating through a for loop here instead of writing through these cases.

tstate/tstate_view.go
```diff
@@ -230,13 +239,21 @@ func (ts *TStateView) Insert(ctx context.Context, key []byte, value []byte) erro
 		pastWrites:    chunks(ts.writes, k),
```
patrick-ogrady, line 239: You should leave a comment on `getValue` that says this is ok to call here because `state.Write` provide exclusive access to state. You may want to comment that as an invariant.

tstate/tstate_view.go
```diff
@@ -222,6 +222,8 @@ func (ts *TStateView) Insert(ctx context.Context, key []byte, value []byte) erro
 	}
 	valueChunks, _ := keys.NumChunks(value) // not possible to fail
 	k := string(key)
+	// Invariant: [getValue] is safe to call here because with [state.Write], it
```
patrick-ogrady, line 225: I wonder if it makes sense to just explicitly require read for writes?

patrick-ogrady: Given that we do so here?

examples/tokenvm/actions/mint_asset.go
```diff
@@ -37,8 +37,8 @@ func (*MintAsset) GetTypeID() uint8 {
 
 func (m *MintAsset) StateKeys(codec.Address, ids.ID) state.Keys {
 	return state.Keys{
-		string(storage.AssetKey(m.Asset)):         state.Read | state.Write,
-		string(storage.BalanceKey(m.To, m.Asset)): state.Read | state.Write,
+		string(storage.AssetKey(m.Asset)):         state.All,
```
patrick-ogrady, line 40: I believe this can remain R/W

tstate/tstate_view.go
patrick-ogrady: On the other hand, it seems like weird DX to have to specify a `Read` for something that doesn't require a `Get` call.

patrick-ogrady: I think we need to align this with the cost to the VM, which does indeed need to do a read to service this op.

state/keys.go
```diff
@@ -4,10 +4,9 @@
 package state
 
 const (
-	Read Permissions = 1 << iota
-	// TODO: Handle Allocate permission
-	Allocate
-	Write
+	Read     Permissions = 1
+	Allocate             = 2 | Read
+	Write                = 4 | Read
```
patrick-ogrady, line 9: I think this should be 3?

state/keys.go
```diff
@@ -4,10 +4,9 @@
 package state
 
 const (
-	Read Permissions = 1 << iota
-	// TODO: Handle Allocate permission
-	Allocate
-	Write
+	Read     Permissions = 1
```
patrick-ogrady, line 7: You need to keep the 1 << iota

state/keys.go
patrick-ogrady, line 9: nit: I would recommend keeping the bit shifts here instead of computing yourself:
1<<1
1<<2 | Read
1<<3 | Read

examples/tokenvm/actions/import_asset.go
```diff
@@ -46,30 +46,30 @@ func (i *ImportAsset) StateKeys(actor codec.Address, _ ids.ID) state.Keys {
 	if i.warpTransfer.Return {
 		assetID = i.warpTransfer.Asset
 		keys = state.Keys{
```
patrick-ogrady, line 48: I think these should just be RW on return (will test this)

examples/tokenvm/actions/import_asset.go
```diff
@@ -46,30 +46,30 @@ func (i *ImportAsset) StateKeys(actor codec.Address, _ ids.ID) state.Keys {
...
+			string(storage.AssetKey(assetID)):                      state.All,
+			string(storage.BalanceKey(i.warpTransfer.To, assetID)): state.All,
 		}
 	}
 
 	// If the [warpTransfer] specified a reward, we add the state key to make
 	// sure it is paid.
 	if i.warpTransfer.Reward > 0 {
-		keys.Add(string(storage.BalanceKey(actor, assetID)), state.Read|state.Write)
+		keys.Add(string(storage.BalanceKey(actor, assetID)), state.All)
```
patrick-ogrady, line 64: This should just be RW (actor must have a balance to issue)

examples/tokenvm/actions/import_asset.go
```diff
@@ -46,30 +46,30 @@ func (i *ImportAsset) StateKeys(actor codec.Address, _ ids.ID) state.Keys {
...
 	}
 
 	// If the [warpTransfer] requests a swap, we add the state keys to transfer
 	// the required balances.
 	if i.Fill && i.warpTransfer.SwapIn > 0 {
-		keys.Add(string(storage.BalanceKey(actor, i.warpTransfer.AssetOut)), state.Read|state.Write)
-		keys.Add(string(storage.BalanceKey(actor, assetID)), state.Read|state.Write)
-		keys.Add(string(storage.BalanceKey(i.warpTransfer.To, i.warpTransfer.AssetOut)), state.Read|state.Write)
+		keys.Add(string(storage.BalanceKey(actor, i.warpTransfer.AssetOut)), state.All)
```
patrick-ogrady, line 70: Again, this balance must exist

examples/tokenvm/actions/import_asset.go
```diff
@@ -46,30 +46,30 @@ func (i *ImportAsset) StateKeys(actor codec.Address, _ ids.ID) state.Keys {
...
+		keys.Add(string(storage.BalanceKey(actor, assetID)), state.All)
```
patrick-ogrady, line 71: Again, this balance must exist

patrick-ogrady (approved):

PR #761, opened 2024-03-06
Use better errors in wasmlanche-macros

richardpringle:
We shouldn't be panicking in macros, it doesn't provide good error messages. Here are the error messages with the changes.

image

The above is just a (flawed) example. The actual errors are in the new `.stderr` files.

### DIFF
--- a/x/programs/rust/sdk_macros/src/lib.rs
+++ b/x/programs/rust/sdk_macros/src/lib.rs
@@ -1,19 +1,13 @@
 extern crate proc_macro;
 
-use core::panic;
-
 use proc_macro::TokenStream;
-use proc_macro2::Span;
 use quote::quote;
 use syn::{
-    parse_macro_input, Fields, FnArg, Ident, ItemEnum, ItemFn, Pat, PatType, Type, Visibility,
+    parse_macro_input, parse_str, spanned::Spanned, Fields, FnArg, Ident, ItemEnum, ItemFn, Pat,
+    PatType, Path, Type, Visibility,
 };
 
-fn convert_param(param_name: &Ident) -> proc_macro2::TokenStream {
-    quote! {
-        unsafe { wasmlanche_sdk::memory::from_host_ptr(#param_name).expect("error serializing ptr") }
-    }
-}
+const PROGRAM_TYPE: &str = "wasmlanche_sdk::program::Program";
 
 /// An attribute procedural macro that makes a function visible to the VM host.
 /// It does so by wrapping the `item` tokenstream in a new function that can be called by the host.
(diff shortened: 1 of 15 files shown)


x/programs/rust/sdk_macros/tests/compile.rs
```diff
@@ -0,0 +1,5 @@
+#[test]
+fn ui() {
+    let t = trybuild::TestCases::new();
+    t.compile_fail("tests/ui/**/*.rs");
```
darioush, line 4: nit: not sure that "ui" is the best name here. maybe compilation_errors or something slightly more descriptive?

x/programs/rust/sdk_macros/tests/ui/ignore-second-param.rs
```diff
@@ -0,0 +1,12 @@
+use sdk_macros::public;
+
+// TODO:
+// We need to actually make sure this is the proper type
+// Currently, it's just a name check
+// and would break if you used a fully qualified name
```
darioush, line 6: nit: should this TODO be moved or replicated at the implementation point in `is_program` in `lib.rs`?

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -24,50 +16,127 @@ fn convert_param(param_name: &Ident) -> proc_macro2::TokenStream {
...
+                "Functions with the `#[public]` attribute cannot have a `self` parameter.",
+            )),
+            FnArg::Typed(PatType { pat, .. }) => match pat.as_ref() {
+                // TODO:
+                // we should likely remove this constraint. If we provide upgradability
+                // in the future, we may not want to change the function signature
+                // which means we might want wildcards in order to help produce stable APIs
+                Pat::Wild(_) => Err(syn::Error::new(
+                    fn_arg.span(),
+                    "Functions with the `#[public]` attribute can only ignore the first parameter.",
```
darioush, line 83: let's just remove it then? Why TODO for later?

darioush (approved):

x/programs/rust/sdk_macros/tests/compile.rs
wlawt: +1

richardpringle: Haha, I agree, but I also don't want to break convention:
https://docs.rs/trybuild/latest/trybuild/
https://github.com/serde-rs/serde/blob/master/test_suite/tests/compiletest.rs

x/programs/rust/sdk_macros/tests/ui/ignore-second-param.rs
richardpringle: Haha great nit! There's proof that you're correct in that this is the wrong place for this TODO... I actually (kind of) fixed this after I wrote the TODO. The fix was obviously in `lib.rs`. 

You've also made me realize that I actually need to add another test for passing in a type that's called `Program` but isn't the same `wasmlanche_sdk::program::Program`

x/programs/rust/sdk_macros/src/lib.rs
richardpringle: Didn't want to just make the executive decision. The TODO is me punting the discussion.

You still think I should just get rid of it? It's easier to delete later than to re-add, that's why I added the TODO (though re-adding is not hard).

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -24,50 +16,127 @@ fn convert_param(param_name: &Ident) -> proc_macro2::TokenStream {
...
+                    syn::Error::new(
+                        ty.span(),
+                        format!("The first paramter of a function with the `#[public]` attribute must be of type `{PROGRAM_TYPE}`"),
+                    )
+                }
+                Some(_) => {
+                    syn::Error::new(
+                        arg.span(),
+                        format!("The first paramter of a function with the `#[public]` attribute must be of type `{PROGRAM_TYPE}`"),
+                    )
```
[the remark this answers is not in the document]
richardpringle: `ty.span()` vs `arg.span()`

`ty.span()` produces this error message:
https://github.com/ava-labs/hypersdk/pull/761/files#diff-503501ed02d797e106c4a854848577049ae09241ee137c938f0643a1e2f7845eR4

`arg.span()` produces this error message:
https://github.com/ava-labs/hypersdk/pull/761/files#diff-b125ac62f89c3b56cba868dd72bbe58869948cc12d6b249d02674504644f77b6R4

richardpringle:
@darioush, changes since your last review https://github.com/ava-labs/hypersdk/compare/a11177f45c9ebbd59db8c47158b171333026cfe0..ee0acaa46831492af0a4ef0288da308e1fd46f1e

x/programs/rust/sdk_macros/tests/compile.rs
richardpringle: Going to keep this. I don't imagine too many program authors will look at the macro-tests. I would rather optimize for a Rust macro author

x/programs/rust/sdk_macros/tests/ui/ignore-second-param.rs
richardpringle: fixed

darioush (approved):

PR #765, opened 2024-03-07
[fetcher] pre-fetch StateKeys on-the-fly

### LINKED ISSUE #749: [chain/processor] address cache stampede

wlawt:
Closes #749 . This is a copy from #744 

### Overview
This PR addresses the cache stampede problem that occurs with concurrent reads to a key. The key may not be in the cache, resulting in us to go to disk. But, we only want to go to disk once and update the cache so that subsequent reads will fetch from cache. 

### Changes
We introduce the `Fetcher` package that will fetch values and store them into the cache on-the-fly. However, if the key is already in cache, then we can say we already fetched this value and reduce the count of the number of keys a given transaction is waiting on.

This also reduces the logic contained within `chain/processor.go` and is replaced by a simple `fetcher.Lookup(..)` and `fetcher.Get(..)` call.

### Review Guide
- `chain/processor.go`: new, reduced logic to lookup keys
- `fetcher/fetcher.go`: core logic 
   - `Lookup(..)`: enqueues keys for the workers to fetch
   - `Get(..)`: blocks a txn until all keys are fetched then return reads and storage from cache
   - `Stop()`: Don't call Lookup after calling Stop
   - `Wait()`: Wait until all workers finish and return any errors the workers may have encountered
- `fetcher/fetcher_tests.go`: unit tests

### Tests
- Added unit tests to `Fetcher` with ~94% coverage
- Integration tests with MorpheusVM and TokenVM pass
- CI

### TODO
- [x] More unit tests
- [x] Add `Stop()`
- [x] Handle bubbling up errors from workers


fetcher/fetcher.go
```diff
@@ -0,0 +1,249 @@
...
+			ctx: ctx,
+			id:  txID,
+			key: k,
+		}
+		tasks = append(tasks, t)
+	}
+	f.keyLock.Unlock()
+
+	if done := f.stopped(); done {
+		return wg // counter is zero
```
wlawt, line 186: Should we change the return value of the function to also include an error?

patrick-ogrady: I think it might just be easier to treat this as an invariant (which we satisfy in our usage) that says, "Don't call Lookup after calling Stop".

Then you can remove this extra locking.

chain/processor.go
```diff
@@ -138,21 +95,15 @@ func (b *StatelessBlock) Execute(
...
-				}
-				cacheLock.Unlock()
-			}
 			return nil
 		})
 	}
 	if err := e.Wait(); err != nil {
 		return nil, nil, err
 	}
+	if err := f.HandleErrors(); err != nil {
```
patrick-ogrady, line 104: I would change this function to `Wait()`, it is more common for this sort of thing.

patrick-ogrady, line 104: I would call this `Wait()`, it is more common in go.

chain/processor.go
```diff
@@ -55,53 +51,14 @@ func (b *StatelessBlock) Execute(
...
-					numChunks, ok := keys.NumChunks(v)
-					if !ok {
-						return ErrInvalidKeyValue
-					}
-					reads[k] = numChunks
-					toCache[k] = &fetchData{v, true, numChunks}
-					storage[k] = v
-				}
-			}
+			reads, storage := f.Wait(wg, stateKeys)
```
patrick-ogrady, line 61: Call this `Get` and use `Wait` for error handling at the end.

fetcher/fetcher.go
```diff
@@ -0,0 +1,249 @@
...
+
+	for {
+		select {
+		case t, ok := <-f.fetchable:
+			if !ok {
+				return
+			}
+
+			// Allows concurrent reads to cache.
+			if ok = f.check(t); ok {
```
patrick-ogrady, line 92: nit: `has` instead of `check`

fetcher/fetcher.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// Checks if a key is in the cache. If it's not in cache, check
+// if we were the first to request the key.
+func (f *Fetcher) check(t *task) bool {
+	f.keyLock.RLock()
+	defer f.keyLock.RUnlock()
+
+	_, exists := f.cache[t.key]
+	if !exists {
```
patrick-ogrady, line 131: Recommend doing:
```
if exists {
  return false
}

queue := f.keysToFetch[t.key].queue
if len(queue) == 0 || queue[0] != t.id {
	// Fetch other keys instead
	return false
}
return true
```

patrick-ogrady: I'm not actually sure the `queue` can be 0 here? you may not need to do that check

fetcher/fetcher.go
```diff
@@ -0,0 +1,249 @@
...
+
+// Updates the cache and dependencies
+func (f *Fetcher) update(k string, v []byte, exists bool, chunks uint16) {
+	f.keyLock.Lock()
+	defer f.keyLock.Unlock()
+
+	// Puts a key that was fetched from disk into cache
+	f.cache[k] = &fetchData{v, exists, chunks}
+
+	f.txnLock.Lock()
```
patrick-ogrady, line 149: nit: recommend releasing the key lock before doing this iteration (we'll want to release that lock asap). it would look something like:
```
	f.keyLock.Lock()
	// Puts a key that was fetched from disk into cache
	f.cache[k] = &fetchData{v, exists, chunks}
	queue := f.keysToFetch[k].queue
        f.keysToFetch[k].queue = nil
	f.keysToFetch[k].fetched = true
        f.keyLock.Unlock()

        f.txnLock.Lock()
	for _, id := range f.keysToFetch[k].queue {
		f.txnsToFetch[id].Done() // Notify all other txs
	}
	f.txnLock.Unlock()
```

fetcher/fetcher.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (f *Fetcher) Wait(wg *sync.WaitGroup, stateKeys state.Keys) (map[string]uint16, map[string][]byte) {
+	// Block until all keys for the tx are fetched
+	wg.Wait()
+
+	f.l.Lock()
+	f.completed++
+	if f.completed == f.numTxs {
+		f.done = true
+		close(f.fetchable)
```
patrick-ogrady, line 205: You may want to wait to close `f.fetchable` until after `f.wg.Wait()` returns?

patrick-ogrady: Then you can remove the `case t, ok := <-f.fetchable` and just do `t := <-f.fetchable`.

fetcher/fetcher.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	f.txnLock.Unlock()
+	f.keysToFetch[k].queue = nil
+	f.keysToFetch[k].fetched = true
+}
+
+// Lookup enqueues keys for the workers to fetch
+func (f *Fetcher) Lookup(ctx context.Context, txID ids.ID, stateKeys state.Keys) *sync.WaitGroup {
+	wg := &sync.WaitGroup{}
+	f.txnLock.Lock()
```
patrick-ogrady, line 161: You should add the tx at the end of this function, not here (as the wg isn't properly initialized yet)

fetcher/fetcher.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+}
+
+// Checks if a key is in the cache. If it's not in cache, check
+// if we were the first to request the key.
+func (f *Fetcher) shouldFetch(t *task) bool {
+	f.keyLock.RLock()
+	defer f.keyLock.RUnlock()
+
+	if f.keysToFetch[t.key].cache != nil {
```
wlawt, line 127: Couldn't remove `case t, ok := <-f.fetchable` because I would get a panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
```

fetcher/fetcher.go
```diff
@@ -0,0 +1,228 @@
...
+	f.stopOnce.Do(func() {
+		f.err = ErrStopped
+		close(f.stop)
+	})
+}
+
+// Wait until all the workers are done and return any errors
+func (f *Fetcher) Wait() error {
+	f.l.Lock()
+	if f.completed == f.numTxs {
```
patrick-ogrady, line 222: This isn't a stable cleanup. You should close `f.fetchable` once all items are enqueued, not by this condition (as most of the time you won't close it then). You don't need to track this.

fetcher/fetcher.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return wg
+}
+
+func (f *Fetcher) Get(wg *sync.WaitGroup, stateKeys state.Keys) (map[string]uint16, map[string][]byte) {
+	// Block until all keys for the tx are fetched
+	wg.Wait()
+
+	f.l.Lock()
+	f.completed++
```
patrick-ogrady, line 190: See below: you don't need to track this.

fetcher/fetcher.go
```diff
@@ -0,0 +1,228 @@
...
+	f.keyLock.Lock()
+	// Puts a key that was fetched from disk into cache
+	f.keysToFetch[k].cache = &fetchData{v, exists, chunks}
+	queue := f.keysToFetch[k].queue
+	f.keysToFetch[k].queue = nil
+	f.keyLock.Unlock()
+
+	f.txnLock.Lock()
+	for _, id := range queue {
+		f.txnsToFetch[id].Done() // Notify all other txs
```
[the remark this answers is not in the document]
patrick-ogrady: Ah, this is probably a race between `txnsToFetch` and `f.keysToFetch[k].queue = append(f.keysToFetch[k].queue, txID)`. This is why I suggested you could unify things behind a single lock instead of trying to use 2.

fetcher/fetcher.go
wlawt: Ah I realized, this is because `t` would be `nil` and it gets passed into here

fetcher/fetcher.go
```diff
@@ -210,12 +202,13 @@ func (f *Fetcher) Stop() {
...
+// Wait returns as soon as all enqueued items are fetched and
+// returns any errors.
+//
+// You should not call [Lookup] after [Wait] is called.
 func (f *Fetcher) Wait() error {
 	f.l.Lock()
-	if f.completed == f.numTxs {
-		close(f.fetchable)
-	}
+	close(f.fetchable)
```
patrick-ogrady, line 214: You don't need to close a channel in the lock, in fact I'm not sure if `f.l` is needed anymore

fetcher/fetcher.go
```diff
@@ -63,25 +63,23 @@ func New(numTxs int, concurrency int, im state.Immutable) *Fetcher {
...
 	for i := 0; i < concurrency; i++ {
-		f.wg.Add(1)
 		go f.runWorker()
 	}
 	return f
 }
 
 // Workers fetch individual keys
 func (f *Fetcher) runWorker() {
-	defer f.wg.Done()
```
patrick-ogrady: I think you'll still want to keep this.

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,258 @@
...
+
+	for i := 0; i < numTxs; i++ {
+		stateKeys := make(state.Keys, (i + 1))
+		for k := 0; k < i+1; k++ {
+			// Generate the same keys
+			stateKeys.Add(strconv.Itoa(k), state.Read)
+		}
+		txID := ids.GenerateTestID()
+		// We are fetching the same keys, so we should
+		// be getting subsequnt requests from cache
```
patrick-ogrady, line 114: nit: subsequent

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,258 @@
...
+func TestFetchDifferentKeys(t *testing.T) {
+	var (
+		require = require.New(t)
+		numTxs  = 100
+		f       = New(newTestDB(), numTxs, 4)
+		ctx     = context.TODO()
+		wg      sync.WaitGroup
+
+		l     sync.Mutex
+		cache = make(map[string]interface{})
```
patrick-ogrady, line 60: nit: may be better to use a `set.Set` for this

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,258 @@
...
+func newTestDB() *testDB {
+	return &testDB{
+		storage: make(map[string][]byte),
+	}
+}
+
+func newTestDBWithValue() *testDB {
+	db := testDB{storage: make(map[string][]byte)}
+	for i := 0; i < 100; i++ {
+		if i%2 == 0 {
```
patrick-ogrady, line 36: nit: you can just do `i+=2` above

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,258 @@
...
+	wg.Add(numTxs)
+	for i := 0; i < numTxs; i++ {
+		stateKeys := make(state.Keys, (i + 1))
+		for k := 0; k < i+1; k++ {
+			// Generate the same keys
+			stateKeys.Add(strconv.Itoa(k), state.Read)
+		}
+		txID := ids.GenerateTestID()
+		if i%2 == 0 {
+			time.Sleep(1 * time.Second)
```
patrick-ogrady, line 155: Usually, it is bad practice to sleep during a test. You should instead use a channel to replicate the behavior you want.

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,258 @@
...
+			l.Lock()
+			for k := range reads {
+				cache[k] = nil
+			}
+			for k := range storage {
+				cache[k] = nil
+			}
+			l.Unlock()
+			if i == 3 {
+				f.Stop()
```
patrick-ogrady, line 251: You can also require that some `f.Fetch` returns an error

chain/processor.go
```diff
@@ -138,21 +103,15 @@ func (b *StatelessBlock) Execute(
...
-				}
-				cacheLock.Unlock()
-			}
 			return nil
 		})
 	}
 	if err := e.Wait(); err != nil {
 		return nil, nil, err
 	}
+	if err := f.Wait(); err != nil {
```
patrick-ogrady, line 112: With the changes that have been made, it should now be safe to put `f.Wait` before `e.Wait` again: https://github.com/ava-labs/hypersdk/pull/744#discussion_r1513342923

fetcher/fetcher_test.go
```diff
@@ -150,12 +149,18 @@ func TestFetchSameKeysSlow(t *testing.T) {
...
 		}
 		txID := ids.GenerateTestID()
-		if i%2 == 0 {
-			time.Sleep(1 * time.Second)
-		}
-		require.NoError(f.Fetch(ctx, txID, stateKeys))
-		go func(txID ids.ID) {
+
+		delay := make(chan struct{})
+		go func(txID ids.ID, i int) {
```
patrick-ogrady, line 154: Usually you just take the scope from outside the goroutine and don't populate these args:

```
go func() {
....
}
```

fetcher/fetcher_test.go
```diff
@@ -150,12 +149,18 @@ func TestFetchSameKeysSlow(t *testing.T) {
...
 			defer wg.Done()
+			if i%2 == 0 {
+				// do some work to minimic timeout
+				for i := 0; i < 10000000; i++ {
+				}
+				close(delay)
```
patrick-ogrady, line 157: logic is cleaner if you just always close this after the `if` statement (because you always create it)..then you don't need to do the `continue` below.

fetcher/fetcher_test.go
```diff
@@ -154,9 +154,6 @@ func TestFetchSameKeysSlow(t *testing.T) {
 		go func(txID ids.ID, i int) {
 			defer wg.Done()
 			if i%2 == 0 {
-				// do some work to minimic timeout
-				for i := 0; i < 10000000; i++ {
-				}
 				close(delay)
```
patrick-ogrady, line 157: Wouldn't you want to do this after `f.Get` or at least after `f.Fetch`?

Would recommend writing out in comments what this `Slow` test is trying to do.

fetcher/fetcher_test.go
```diff
@@ -0,0 +1,272 @@
...
+				cache.Add(k)
+			}
+			for k := range storage {
+				cache.Add(k)
+			}
+			l.Unlock()
+			if i == 3 {
+				f.Stop()
+			}
+		}(i)
```
patrick-ogrady, line 267: I actually hadn't been familiar with this pattern but that's pretty cool.

patrick-ogrady (approved):

patrick-ogrady:
This pull request primarily introduces a new feature to the codebase: the ability to fetch state in a concurrent manner. This is achieved by adding a new `Fetcher` type that retrieves values on-the-fly and ensures that a value is only fetched from data once. Subsequent requests can be retrieved from cache. The most important changes include the addition of the `Fetcher` type, the modification of the `Execute` function in `chain/processor.go` to use the `Fetcher`, and the addition of the `GetStateFetchConcurrency` method in various interfaces and structures.

Addition of the `Fetcher` type:

* [`fetcher/errors.go`](diffhunk://#diff-3956f3ef1df4fa87ae1ef3b2c9e2afa7ef6985c6a34070397f70254ca3d50fa7R1-R12): Defined new error types for the `Fetcher`.
* [`fetcher/fetcher.go`](diffhunk://#diff-6327325397c53581c510fcbfbbc515a55d06a8c528aa09b5ee39bfb7437adeabR1-R250): Defined the `Fetcher` type and its methods.
* [`fetcher/fetcher_test.go`](diffhunk://#diff-d1e9691f7d41c678b1cf10ffd96a2043d725fdb68441fcbba3cef925016e2c31R1-R272): Added tests for the `Fetcher`.

Modifications to the `Execute` function:

* [`chain/processor.go`](diffhunk://#diff-ea8cc6fbeac112c3e93bf5ff4062bde4096c8a4123d55e32c05e2cafc3f9928fL43-R41): Modified the `Execute` function to use the `Fetcher` to fetch state keys from disk. This includes removing the old cache mechanism and replacing it with the `Fetcher`. [[1]](diffhunk://#diff-ea8cc6fbeac112c3e93bf5ff4062bde4096c8a4123d55e32c05e2cafc3f9928fL43-R41) [[2]](diffhunk://#diff-ea8cc6fbeac112c3e93bf5ff4062bde4096c8a4123d55e32c05e2cafc3f9928fR54-L104) [[3]](diffhunk://#diff-ea8cc6fbeac112c3e93bf5ff4062bde4096c8a4123d55e32c05e2cafc3f9928fL141-R111)

Addition of the `GetStateFetchConcurrency` method:

* [`chain/dependencies.go`](diffhunk://#diff-493fc040657d623dc557e54e46baf059d613bfe82adc14a8465721801cb63264R82): Added the `GetStateFetchConcurrency` method to the `VM` interface.
* [`config/config.go`](diffhunk://#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R24): Added the `GetStateFetchConcurrency` method to the `Config` structure.
* [`examples/morpheusvm/config/config.go`](diffhunk://#diff-b6f8c01fb0d3f7d871cf5ad93091993527a766d3b51629dab329f629d5c1707bR39): Added the `GetStateFetchConcurrency` method to the `Config` structure. [[1]](diffhunk://#diff-b6f8c01fb0d3f7d871cf5ad93091993527a766d3b51629dab329f629d5c1707bR39) [[2]](diffhunk://#diff-b6f8c01fb0d3f7d871cf5ad93091993527a766d3b51629dab329f629d5c1707bR98) [[3]](diffhunk://#diff-b6f8c01fb0d3f7d871cf5ad93091993527a766d3b51629dab329f629d5c1707bR112)
* [`examples/tokenvm/config/config.go`](diffhunk://#diff-21c5a3b81147838155f184eb9b89a938c384f3668c35e41ff8336037738705c5R41): Added the `GetStateFetchConcurrency` method to the `Config` structure. [[1]](diffhunk://#diff-21c5a3b81147838155f184eb9b89a938c384f3668c35e41ff8336037738705c5R41) [[2]](diffhunk://#diff-21c5a3b81147838155f184eb9b89a938c384f3668c35e41ff8336037738705c5R119) [[3]](diffhunk://#diff-21c5a3b81147838155f184eb9b89a938c384f3668c35e41ff8336037738705c5R134)
* [`vm/dependencies.go`](diffhunk://#diff-2bbdcab9c5b2c127504176864e55fac2ca3fe6816eb09a5dcf2d32835172329dR36): Added the `GetStateFetchConcurrency` method to the `Config` interface.
* [`vm/resolutions.go`](diffhunk://#diff-d9b2d35e026b4aabcdc2b3c18771992725c042540fe0e0879616a87f5c4e0eb2R488-R491): Added the `GetStateFetchConcurrency` method to the `VM` structure.

Other changes:

* [`chain/processor.go`](diffhunk://#diff-ea8cc6fbeac112c3e93bf5ff4062bde4096c8a4123d55e32c05e2cafc3f9928fL8-R14): Removed unused imports.

PR #769, opened 2024-03-11
[fetcher] Nits

patrick-ogrady:


fetcher/fetcher.go
```diff
@@ -79,144 +82,147 @@ func (f *Fetcher) runWorker() {
 
 	for {
 		select {
-		case t, ok := <-f.fetchable:
+		case t, ok := <-f.tasks:
 			if !ok {
 				return
 			}
 
-			if ok := f.shouldFetch(t); !ok {
```
patrick-ogrady, line 87: `shouldFetch` is no longer needed because we only enqueue a key the first time it is requested

fetcher/fetcher.go
```diff
@@ -79,144 +82,147 @@ func (f *Fetcher) runWorker() {
...
-	f.l.Lock()
-	if f.completed == f.numTxs {
-		close(f.fetchable)
-	}
-	f.l.Unlock()
+	f.waitOnce.Do(func() {
+		close(f.tasks)
+	})
 	f.wg.Wait()
 	return f.err
```
patrick-ogrady, line 227: We don't need a lock around `f.err` because it can never be updated after `f.wg.Wait()` returns.

chain/processor.go
```diff
@@ -55,10 +55,18 @@ func (b *StatelessBlock) Execute(
 			e.Stop()
 			return nil, nil, err
 		}
-		// Fetch keys from disk or check if it's in cache
-		wg := f.Lookup(ctx, tx.ID(), stateKeys)
+
+		// Prefetch state keys from disk
+		txID := tx.ID()
+		if err := f.Fetch(ctx, txID, stateKeys); err != nil {
```
patrick-ogrady, line 61: Both these cases require errors to ensure we don't deadlock in the case of a lookup error.

fetcher/fetcher.go
```diff
@@ -79,144 +83,162 @@ func (f *Fetcher) runWorker() {
...
 func (f *Fetcher) Wait() error {
+	f.waitOnce.Do(func() {
+		close(f.tasks)
+	})
+	f.wg.Wait()
 	f.l.Lock()
-	if f.completed == f.numTxs {
-		close(f.fetchable)
-	}
+	f.done = true
```
patrick-ogrady, line 241: It is ok to set this multiple times (otherwise we'd have a do once)

fetcher/fetcher.go
```diff
@@ -79,144 +83,162 @@ func (f *Fetcher) runWorker() {
...
+func (f *Fetcher) set(k string, v []byte, exists bool, chunks uint16) {
+	f.l.Lock()
+	defer f.l.Unlock()
+
+	// Puts a key that was fetched from data into cache
+	f.keys[k].cache = &data{v, exists, chunks}
+	for _, id := range f.keys[k].blocked {
+		f.txs[id].blockers--
+		if f.txs[id].blockers == 0 {
+			close(f.txs[id].waiter)
```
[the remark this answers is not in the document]
patrick-ogrady: It is a replacement for `wg` that we can `select` over (can't do with `sync.WaitGroup`)

fetcher/fetcher.go
```diff
@@ -79,144 +83,162 @@ func (f *Fetcher) runWorker() {
...
-	if f.keysToFetch[t.key].cache != nil {
-		return false
+func (f *Fetcher) set(k string, v []byte, exists bool, chunks uint16) {
+	f.l.Lock()
+	defer f.l.Unlock()
+
+	// Puts a key that was fetched from data into cache
+	f.keys[k].cache = &data{v, exists, chunks}
+	for _, id := range f.keys[k].blocked {
+		f.txs[id].blockers--
```
patrick-ogrady, line 118: nit: use `tx` object here rather than `f.txs[id]`

fetcher/fetcher.go
```diff
@@ -16,58 +17,61 @@ import (
...
-	l         sync.RWMutex
-	wg        sync.WaitGroup
+	wg       sync.WaitGroup
+	tasks    chan *task
+	waitOnce sync.Once
 
-	completed int
-	numTxs    int
+	stop     chan struct{}
+	stopOnce sync.Once
```
patrick-ogrady, line 36: nit: may be cleaner if we change to `doneOnce` and differ the `done` behavior whether we are "complete" or "stopping"

patrick-ogrady: Then we can remove this `done` var

patrick-ogrady: we can call `finishOnce`

fetcher/fetcher.go
```diff
@@ -79,144 +83,162 @@ func (f *Fetcher) runWorker() {
...
+// are only fetched once and fetch priority is done in the order [Fetch] is called.
+//
+// Fetch can be called concurrently.
+//
+// Invariant: Don't call [Fetch] afer calling [Stop] or [Wait]
+func (f *Fetcher) Fetch(ctx context.Context, txID ids.ID, stateKeys state.Keys) error {
+	f.l.Lock()
+	if fErr := f.err; fErr != nil {
+		f.l.Unlock()
+		return fErr // ensures we don't deadlock if encountering an error
```
patrick-ogrady, line 151: nit: can remove this copy because a non-nil `f.err` can never be modified

fetcher/fetcher.go
```diff
@@ -79,144 +81,158 @@ func (f *Fetcher) runWorker() {
...
-			key: k,
+
+		// Don't register that we are blocked if the key has already been fetched
+		if d.cache != nil {
+			continue
 		}
-		tasks = append(tasks, t)
+
+		// Register to get notified when the key is fetched
+		d.blocked = append(d.blocked, txID)
```
patrick-ogrady, line 167: We should still be adding to `blockers` here.

PR #779, opened 2024-03-14
[executor] handling read/allocate/write-keys in the execution engine

### LINKED ISSUE #701: [executor] parallelizing reads in the execution engine

wlawt:
Closes #701, #709 . This is a copy from #715 

### Overview

Please see the 2nd comment in #701 for the overall approach. Some modifications were made from the original idea. 

The executor keeps track of the last blocked Allocate/Write task in `edges`. It contains `data` which holds up to 1 Allocate/Write and possibly many Reads. This ensures that Reads can't happen before Writes, and Writes can't happen before Reads. Furthermore, this enforces the tasks to execute in the order they were enqueued, in the case of high contention.

### Review Guide

Within `executor.go` the following logic/structs would be important:

- `data`: Key information that is held within `edges`
- `Run(..)`: Updated algorithm to record dependencies.
- `runWorker()`: Removes any concurrent Reads that are executed so that the set only holds not-yet-executed Reads

The high-level algorithm of recording dependencies are as follows:

- Have we seen this key? 
   - If no, add it and set `allocateWrite` or `reads` based on the key permission.
- Have we requested an Allocate/Write key before? 
   - If no, then set `allocateWrite` if the key is an Allocate/Write key
- Have we executed the requested Allocate/Write?
   - If yes, update the last blocked task to our current task
   - If no, update the dependencies (reads are blocked by writes, writes are blocked by reads)

### Test Plan

- Added unit tests in `executor_test.go`, coverage is ~95%
- Integration tests in MorpheusVM and TokenVM pass
- CI
 
### TODOs

- [x] Add more unit tests
- [x] Handle GetValue caching

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -168,7 +168,7 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 	if t.dependencies != nil && t.dependencies.Len() > 0 {
 		// Ensure that we don't schedule a transaction that is
 		// blocked by another transaction currently running
-		for _, bID := range t.dependencies {
+		for bID := range t.dependencies {
 			// don't execute if any dependency task is in blocking
 			if e.blocking.Contains(bID) {
 				if e.metrics != nil {
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -5,7 +5,6 @@ package executor
 
 import (
 	"sync"
-	//"fmt"
 
 	"github.com/ava-labs/avalanchego/utils/set"
 
@@ -118,9 +117,6 @@ func (e *Executor) createWorker() {
 
 // Run executes [f] after all previously enqueued [f] with
 // overlapping [conflicts] are executed.
-// TODO: Handle read-only/write-only keys (currently the executor
-// treats everything still as ReadWrite, see issue below)
-// https://github.com/ava-labs/hypersdk/issues/709
 func (e *Executor) Run(conflicts state.Keys, f func() error) {
 	e.l.Lock()
 	defer e.l.Unlock()
@@ -146,11 +142,10 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 					lt.blocking = set.NewSet[int](defaultSetSize)
 				}
 				// key has ONLY a Read permission
-				// TODO: clean this up
-				if v.Has(state.Read) && !v.Has(state.Write) && !v.Has(state.Allocate) {
-					// lt contains a Write permission. If lt had only a Read
+				if v == state.Read {
+					// lt contains either Allocate or Write permission. If lt had only a Read
 					// permission, no dependency or blocking would be added
-					if latest.Permissions.Has(state.Write) {
+					if latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
 						t.dependencies.Add(lt.id) // t depends on lt to execute
 						lt.blocking.Add(id)       // lt is blocking this current task
 					}
@@ -158,8 +153,8 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 
 				// key contains a Write permission
 				if v.Has(state.Write) {
-					// lt contains a Read or Write permission
-					if latest.Permissions.Has(state.Read | state.Write) {
+					// lt contains either Read, Allocate, or Write
+					if latest.Permissions.Has(state.Read) || latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
 						t.dependencies.Add(lt.id)
 						lt.blocking.Add(id)
 					}
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -35,28 +35,12 @@ type Executor struct {
 	edges     map[string]*keyData
 }
 
-// keyData keeps track of the last blocked task, known
-// as the id, its Permissions, and any tasks that
-// are trying to read, but is blocked by the id,
-// known as ConcurrentReads.
-//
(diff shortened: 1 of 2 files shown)


executor/executor.go
```diff
@@ -41,104 +48,142 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
-				lt.blocking.Add(id)
+
+				if v == state.Read {
+					key.concurrentReads.Add(id)
+					recordDependencies(t, lt)
+					continue
+				}
+
+				if key.concurrentReads.Len() == 0 {
+					recordDependencies(t, lt)
```
wlawt, line 171: Wondering if there's a cleaner way for this

executor/executor.go
```diff
@@ -32,7 +35,13 @@ type Executor struct {
...
 	completed int
 	tasks     map[int]*task
-	edges     map[string]int
+	edges     map[string]*data
+}
+
+// We can do up to 1 Allocate/Write and many reads
+type data struct {
+	allocateWrite   int
+	concurrentReads set.Set[int]
```
wlawt, line 44: nit: maybe just name it `reads` ?

executor/executor.go
```diff
@@ -32,7 +35,13 @@ type Executor struct {
 	done      bool
 	completed int
 	tasks     map[int]*task
-	edges     map[string]int
+	edges     map[string]*data
+}
+
+// We can process up to 1 Allocate/Write and many reads
```
patrick-ogrady, line 41: nit: if you capitalize "Allocate/Write", capitalize "Read"

executor/executor.go
```diff
@@ -41,104 +50,140 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
-			lt := e.tasks[latest]
+			// No Allocate/Write has been requested yet
+			if key.allocateWrite == notSet {
+				if v == state.Read {
+					key.reads.Add(id)
+				} else {
+					key.allocateWrite = id
+					for b := range key.reads {
+						bt := e.tasks[b]
+						recordDependencies(t, bt)
```
patrick-ogrady, line 148: nit: just make this

```
recordDependencies(t, e.tasks[b])
```

executor/executor.go
```diff
@@ -41,104 +50,140 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
+			}
+
+			// Get last blocked Allocate/Write task
+			lt := e.tasks[key.allocateWrite]
 			if !lt.executed {
 				if t.dependencies == nil {
 					t.dependencies = set.NewSet[int](defaultSetSize)
 				}
-				t.dependencies.Add(lt.id)
 				if lt.blocking == nil {
```
patrick-ogrady, line 160: nit: I found this code was easier to reason about if we just create these structures in the initializer for `task` instead of here.

executor/executor.go
```diff
@@ -41,104 +50,140 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
-	}()
+	}
 }
 
 // Run executes [f] after all previously enqueued [f] with
 // overlapping [conflicts] are executed.
-// TODO: Handle read-only/write-only keys (currently the executor
-// treats everything still as ReadWrite, see issue below)
-// https://github.com/ava-labs/hypersdk/issues/709
 func (e *Executor) Run(conflicts state.Keys, f func() error) {
```
patrick-ogrady, line 123: nit: since we store this in `keys`, we might as well change this to `keys state.Keys`

executor/executor.go
```diff
@@ -41,104 +50,140 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
-				t.executed = true
-				e.completed++
-				if e.done && e.completed == len(e.tasks) {
-					// We will close here if there are unexecuted tasks
-					// when we call [Wait].
-					close(e.executable)
+			e.l.Lock()
+			// Update concurrent reads to only contain not-yet-executed reads
+			for k := range t.keys {
+				key := e.edges[k]
```
patrick-ogrady, line 95: I'm not sure this is guaranteed to be the right `edge` for that key? We probably want to put a pointer to that or store the ID in `t.keys`

executor/executor.go
```diff
@@ -32,7 +35,13 @@ type Executor struct {
 	done      bool
 	completed int
 	tasks     map[int]*task
-	edges     map[string]int
+	edges     map[string]*data
+}
+
+// We can process up to 1 Allocate/Write and many reads
+type data struct {
+	allocateWrite int
```
patrick-ogrady, line 43: not in love with the variable name, maybe we do:

```
type edge struct {
  exclusive int
  shared set.Set[int]
}
```

executor/executor.go
```diff
@@ -41,104 +50,140 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
-				}
-				t.blocking = nil // free memory
-				t.executed = true
-				e.completed++
-				if e.done && e.completed == len(e.tasks) {
-					// We will close here if there are unexecuted tasks
-					// when we call [Wait].
-					close(e.executable)
+			e.l.Lock()
+			// Update concurrent reads to only contain not-yet-executed reads
```
patrick-ogrady, line 93: nit: comment when we populate `reads` (when not allocate/write)

executor/executor.go
patrick-ogrady, line 148: There is an error here where the `bt` may be executed but you are now adding dependencies that won't be cleared.

wlawt:
Close in favor of #814

PR #781, opened 2024-03-14
More refactoring

richardpringle:
I keep running into errors in the add `Context` code that are a pain to deal with. This clean *some* of that up, but should make the code easier to extent in general.


x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-use wasmlanche_sdk::{params, program::Program, public, state_keys, types::Address};
+use wasmlanche_sdk::{params, public, state_keys, types::Address, Program};
```
richardpringle, line 1: Typically in Rust, it's only larger libraries (that have a bunch of feature flags) that force you to access commonly used types through modules. The most commonly used types should almost always be accessible from the root of a crate.

x/programs/rust/wasmlanche-sdk/src/errors.rs
richardpringle, line 1: Errors should be defined closer to the code that creates them. It's not usually a good practice to have a module with a single type.

x/programs/rust/wasmlanche-sdk/src/host/mod.rs
richardpringle, line 1: This module wasn't actually providing any value. I moved most things closer to the call-site as there was a bunch of stuff that was just being used in one place.

x/programs/rust/wasmlanche-sdk/src/lib.rs
```diff
@@ -1,13 +1,27 @@
 #![deny(clippy::pedantic)]
 
-pub mod errors;
-pub mod host;
```
richardpringle, line 4: Deleted these two

x/programs/rust/wasmlanche-sdk/src/lib.rs
```diff
@@ -1,13 +1,27 @@
...
-pub mod memory;
-pub mod program;
```
richardpringle, line 6: Made these two private and only exposed what's needed outside.

x/programs/rust/wasmlanche-sdk/src/lib.rs
```diff
@@ -1,13 +1,27 @@
...
 pub mod types;
 
+mod memory;
+mod program;
+
+pub use self::{
+    memory::from_host_ptr,
+    params::{serialize_param, Params},
+    program::Program,
+};
```
richardpringle, line 14: re-export `from_host_ptr`, `serialize_param`, `Params`, and `Program`

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -85,3 +113,55 @@ impl Key {
         Self(bytes)
     }
 }
+
+mod host {
```
richardpringle, line 117: This module was moved from the old public `host` module. It's now private and cannot be called outside of the `state` module.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -1,6 +1,6 @@
 use borsh::{BorshDeserialize, BorshSerialize};
 
-use crate::{errors::StateError, host::call_program, state::State};
+use crate::{memory::to_host_ptr, state::Error as StateError, state::State, Params};
```
darioush, line 3: should rename StateError's uses to Error instead of `use ... as`?

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -22,12 +50,12 @@ impl State {
     /// # Errors
     /// Returns an `StateError` if the key or value cannot be
```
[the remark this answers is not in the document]
richardpringle: Thanks! FYI, we're supposed to use `[Error]` in docs and it will automatically link. The docs will fail to build and CI will fail if there's a typo

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -85,3 +113,55 @@ impl Key {
...
+
+        #[link_name = "get"]
+        fn _get(caller: i64, key: i64) -> i64;
+
+        #[link_name = "delete"]
+        fn _delete(caller: i64, key: i64) -> i64;
+    }
+
+    /// Persists the bytes at `value_ptr` to the bytes at key ptr on the host storage.
+    pub(super) unsafe fn put_bytes<V>(caller: &Program, key: &Key, value: &V) -> Result<(), Error>
```
wlawt, line 134: just a question to gain more context: why do all these functions have `unsafe`?

richardpringle: That's a great question. I think there was a point in time when these functions were doing something unsafe beyond calling FFI functions. `unsafe` is supposed to be used to mark functions that could lead to undefined behaviour (UB), but I think since we've introduced borsh-serialization, there's less of that here. 

Marking a function `unsafe` forces you to use an unsafe block, `unsafe { /* function call goes here */ }`, to call it. Usually, that means you have to define your safety parameters/invariants outside of the function and add necessary tests to guarantee safety. 

Make sense? This was a great question because you've made me realize that I probably need to clean this code up. But I'm going to do that in another PR

darioush (approved):

wlawt (approved):

PR #793, opened 2024-03-18
[vryx] mempool rewrite (heaps are slow)

patrick-ogrady:
- [x] process incoming signatures async (to free up connection handler as soon as possible)
  - [ ] ~use `RPCAuth` for this (switch to `NonChunkAuth`)~ keep separate
- [x] track invalid txs in block
- [x] track invalid txs received in gossip
- [x] Change expired txs to recent (not all time) in cli
- [x] track invalid txs dropped during build
- [x] track invalid txs received over rpc
- [x] certs dropped on build path in a number of places...
  - [x] repeat chunk restore?
  - [x] mirror tx handling (remove during verify, re-add during reject)
- [x] Replace Canada with Cape Town + add Brazil (restore old distribution)
- [x] remove message latency tracking + better track bandwidth usage
- [ ] ~add fast start P-Chain epoch (set in block 1)~
- [ ] debug accepted block stall (likely a verification/engine error)
  - [x] track `fetcher` attempts
  - [x] record last executed epoch
  - [x] track dropped built chunks that weren't included in seen
  - [x] revisit epoch cache
  - [x] chunk re-inserted into build order at the back (if a cert keeps failing randomly, could never get to front) -> would also elevate ttf + backlog -> maintain sorting by when we first heard about chunk?
    - [x] drop certs that are too old 
  - [x] run executed ops in executed queue (signal a chunk is "final") -> otherwise a small race in that ExecutedBlock could be processed first
  - [ ] ~if `FilteredChunk` is being populated wrong, would never clear~
  - [ ] ~theory: `processJob` was erroring out and only send some of `FilteredChunks` to the `WebSocketServer`~
    - [x] need to send all chunk results to RPC so it can notify on invalid (otherwise will just treat as expired)
  - [x] properly enforce max chunk units in build/verify
- [x] include epoch in address partition (in devnet, p-chain height doesn't change) -> add `epoch` return to `getEpochHeight`
- [x] remove optimistic chunk verify (will be replaced by dedicated verification loop)
- [x] add metric for tx drops during gossip (because backlog too long)
- [x] track mempool submit errors from gossip
- [x] increase max chunk size for devnet
- [x] add executed units to metrics
- [x] Set mempool remaining to be Len, not just binary
- [x] Change chunk production frequency to 400ms
- [x] Don't log that we are not a validator as an error log
- [x] Enforce allocate in tstate + add allocate bool (#797)


patrick-ogrady:
What is this gapping dynamic? 
image

patrick-ogrady:
Even with no profiles, still jumps every ~minute:
image

patrick-ogrady:
Turns out we just weren't updating the gauge frequently enough: 
image

patrick-ogrady:
How are transactions expiring? I think they may be getting dropped rather than expiring...
image

patrick-ogrady:
Under load, it seems like sometimes chunk signatures are dominated by tx processing

patrick-ogrady:
Bug: not restoring chunks when results aren't ready...

patrick-ogrady:
Some sort of persistent verififcation error: 
image

patrick-ogrady:
Something clearly still broken: 
image

patrick-ogrady:
This should be 0:

image

patrick-ogrady:
```
[03-19|17:30:43.540] INFO <2hnuzmntih6aY6dceRUM1Je31q44mtE5JD8zqXvJ2ei7CJLDjZ Chain> vm/chunk_manager.go:773 unable to build chunk {"error": "executed epoch (28514483) is too far behind (28514490) to verify chunk"}
```

vm/chunk_manager.go
```diff
@@ -206,13 +218,17 @@ func NewChunkManager(vm *VM) *ChunkManager {
 func (c *ChunkManager) getEpochHeight(ctx context.Context, t int64) (uint64, error) {
 	r := c.vm.Rules(time.Now().UnixMilli())
 	epoch := utils.Epoch(t, r.GetEpochDuration())
+	if h, ok := c.epochHeights.Get(epoch); ok {
```
patrick-ogrady, line 221: Suspicious

patrick-ogrady:
```
[node3-bls] [03-19|11:26:50.207] WARN <2DtF9jP9wEV1zHQiGaU9ZzHfe7wAdCSoDQQHmhkVrX7zmHjbHH Chain> chain/engine.go:147 performed useless authorization {"count": 1}
```

patrick-ogrady:
```
[node4-bls] [03-19|11:27:37.499] WARN <2DtF9jP9wEV1zHQiGaU9ZzHfe7wAdCSoDQQHmhkVrX7zmHjbHH Chain> chain/engine.go:147 performed useless authorization {"count": 2}
```

patrick-ogrady:
Got unused chunk authorization with no expiry: 
image

patrick-ogrady:
image

as expected, results taking longer without any pre-signing when cert is ready

patrick-ogrady:
Issues: 
image

patrick-ogrady:
No mempool expiry: 
image

Chunk signature collection is fast:
image

Txs typically added to the mempool with 57s left for inclusion:
image

patrick-ogrady:
Still nothing expired by the mempool

```
[03-20|05:10:54.258] WARN <72ujQFtbhuSHKozE4cNA59qC8sxGCKpKazTBbR4ER5udeUTLx Chain> rpc/websocket_server.go:205 expired listeners {"count": 4882}
```

patrick-ogrady:
Found 1 fatal, just not the one we were looking for: 
```
[03-20|05:13:36.213] FATAL <72ujQFtbhuSHKozE4cNA59qC8sxGCKpKazTBbR4ER5udeUTLx Chain> vm/vm.go:1134 unable to prune block and chunks {"error": "remove /home/ubuntu/.avalanchego/chainData/72ujQFtbhuSHKozE4cNA59qC8sxGCKpKazTBbR4ER5udeUTLx/blobdb/c-1710911135000-2povkCWTePhLP9Ves9jLQy3Yjtxt4yRY8GWpCZ7CdEr2ajyvXz: no such file or directory"}
```

patrick-ogrady:
Latest observations:
image

Ideas:
* We process tx before adding tracking to listener
* We are dropping the tx in the gossip send logic
* Tx on Validator not getting into mempool

patrick-ogrady:
Potentially interesting correlation:
image
image
image

No known drops on incoming gossip:
image

patrick-ogrady:
👀 
image

patrick-ogrady:
```
[03-20|15:00:20.482] ^[[0;0mINFO^[[0;0m <iekzRnR2iSKN3cCrrszD8MrTWQE9oMAodqbVc55x7n8XRccTq Chain> vm/chunk_manager.go:844 unable to build chunk {"error": "packer has insufficient length for input"}
```

PR #794, opened 2024-03-18
Add benches for the merkle package

iFrostizz:
Addresses https://github.com/ava-labs/hypersdk/pull/689#issuecomment-1901577401

This PR adds a bench to measure the time of generation of the merkle root of transactions.
We execute it for 10, 100, 1000, and 10000 iterations, here are the results:

```sh
goos: linux
goarch: amd64
pkg: github.com/ava-labs/hypersdk/merkle
cpu: AMD Ryzen 9 5950X 16-Core Processor
BenchmarkMerkleTxRoot/10-32 	     973	   1855691 ns/op
BenchmarkMerkleTxRoot/100-32         	      15	  87687055 ns/op
BenchmarkMerkleTxRoot/1000-32        	       1	6838312818 ns/op
BenchmarkMerkleTxRoot/10000-32       	       1	618511353824 ns/op
PASS
ok  	github.com/ava-labs/hypersdk/merkle	628.683s
```

The average time it takes per operation doesn't scale linearly which makes sense given the tree data-structure.


merkle/merkle_test.go
```diff
@@ -0,0 +1,44 @@
...
+var res_root ids.ID
+var res_db merkledb.MerkleDB
+var res_err error
+
+func BenchmarkMerkleTxRoot(b *testing.B) {
+	for _, size := range []int{10, 100, 1000, 10000} {
+    ctx := context.TODO()
+    tracer := trace.Noop
+    merkleItems := make([][]byte, 0, size)
+    for i := 0; i < size; i++ {
```
patrick-ogrady, line 24: I'd recommend creating this prior to starting each benchmark.

I'd also recommend using crypto/rand to populate 32 bytes rather than this integer approach.

merkle/merkle_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"context"
+	"strconv"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/trace"
+	"github.com/ava-labs/avalanchego/x/merkledb"
+)
+
+var res_root ids.ID
```
patrick-ogrady, line 15: nit: variables in golang are camel case

patrick-ogrady:
```
BenchmarkMerkleTxRoot/1000-32        	       1	6838312818 ns/op
BenchmarkMerkleTxRoot/10000-32       	       1	618511353824 ns/op
```

We need to speed this up if we want to merge this. 618s for 10k items seems really wrong or broken.

merkle/merkle_test.go
```diff
@@ -0,0 +1,44 @@
...
+    for i := 0; i < size; i++ {
+      as_bytes := make([]byte, 32)
+      binary.BigEndian.PutUint32(as_bytes, uint32(i))
+      merkleItems = append(merkleItems, as_bytes)
+    }
+    var root ids.ID
+    var db merkledb.MerkleDB
+    var err error
+		b.Run(strconv.Itoa(size), func(b *testing.B) {
+			for n := 0; n < b.N; n++ {
```
patrick-ogrady, line 33: Can you also add `b.ReportAllocs()`

merkle/merkle.go
```diff
@@ -0,0 +1,52 @@
...
+
+	for _, item := range merkleItems {
+		key := utils.ToID(item)
+		batchOps = append(batchOps, database.BatchOp{
+			Key:   key[:],
+			Value: item,
+		})
+	}
+
+	db, err := merkledb.New(ctx, memdb.New(), merkledb.Config{
```
patrick-ogrady, line 27: We should explicitly populate all fields IMO: https://github.com/ava-labs/hypersdk/blob/5f47812e686042d774e06613f71578b9af19c351/vm/vm.go#L207-L218

Most importantly, we should allow the user to specify a root gen concurrency: https://github.com/ava-labs/hypersdk/blob/5f47812e686042d774e06613f71578b9af19c351/vm/vm.go#L210

merkle/merkle_test.go
iFrostizz: oops, a bad habit!

merkle/merkle_test.go
iFrostizz: done!

iFrostizz: For some reason I'm not able to see the memory allocation report with this activated on cd108423d752c9cf919ca964a29446bf8c73711c

iFrostizz: That only works when passing the `-test.benchmem` flag

chain/block.go
```diff
@@ -40,7 +41,8 @@ type StatefulBlock struct {
 	Tmstmp int64  `json:"timestamp"`
 	Hght   uint64 `json:"height"`
 
-	Txs []*Transaction `json:"txs"`
+	Txs     []*Transaction `json:"txs"`
+	TxsRoot ids.ID         `json:"txsRoot"`
```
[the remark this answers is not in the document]
iFrostizz: I moved the `block.go` file from the `chain` package to the `merkle` package written by @bianyuanop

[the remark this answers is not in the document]
iFrostizz: Actually @patrick-ogrady do you want to move all of the `StatelessBlock` machinery out of the `chain` package or just everything that is related to the merkle generation and verification but not all ?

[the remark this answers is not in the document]
iFrostizz: WIP at https://github.com/ava-labs/hypersdk/compare/main...iFrostizz:hypersdk:merkle_branch_wip

merkle/merkle.go
iFrostizz: `vm` seems to be meant to be at an upper level compared to `merkle` because importing `vm` in the `merkle` package creates a dependency cycle. Thus I have added a new parameter `config` with the type `merkledb.Config` so that the caller is free to provide any config they like.

merkle/merkle.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/database/memdb"
+	"github.com/ava-labs/avalanchego/trace"
+	"github.com/ava-labs/avalanchego/x/merkledb"
+	"github.com/ava-labs/hypersdk/utils"
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+// Generate merkle root for a set of items
+func GenerateMerkleRoot(ctx context.Context, config merkledb.Config, tracer trace.Tracer, merkleItems [][]byte, consumeBytes bool) (ids.ID, merkledb.MerkleDB, error) {
```
patrick-ogrady, line 15: tracer is provided in `merkledb.Config`, so this arg can be removed.

merkle/merkle.go
```diff
@@ -0,0 +1,45 @@
...
+	db, err := merkledb.New(ctx, memdb.New(), config)
+	if err != nil {
+		return ids.Empty, nil, err
+	}
+
+	view, err := db.NewView(ctx, merkledb.ViewChanges{BatchOps: batchOps, ConsumeBytes: consumeBytes})
+	if err != nil {
+		return ids.Empty, nil, err
+	}
+	if err := view.CommitToDB(ctx); err != nil {
```
patrick-ogrady, line 35: I think it'll be faster to just get the merkle root from the view...you can do that without calling `CommitToDB`

merkle/merkle_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func BenchmarkMerkleTxRoot(b *testing.B) {
+	for _, size := range []int{10, 100, 1000, 10000} {
+    ctx := context.TODO()
+    tracer := trace.Noop
+    merkleItems := make([][]byte, 0, size)
+    for i := 0; i < size; i++ {
+      item := make([]byte, 32)
+      _, err := rand.Read(item)
+      if err != nil {
```
patrick-ogrady, line 28: should be `require.NoError`

merkle/merkle_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/ava-labs/avalanchego/utils/units"
+	"github.com/ava-labs/avalanchego/x/merkledb"
+)
+
+var resRoot ids.ID
+var resDb merkledb.MerkleDB
+var resErr error
+
+func BenchmarkMerkleTxRoot(b *testing.B) {
+	for _, size := range []int{10, 100, 1000, 10000} {
```
patrick-ogrady, line 21: The indenting here is all messed up.

merkle/merkle.go
iFrostizz: removed!

merkle/merkle_test.go
iFrostizz: fixed, I was using spaces

merkle/merkle.go
iFrostizz: You're right, thanks! Updated

merkle/merkle.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/ava-labs/hypersdk/utils"
+)
+
+// Generate merkle root for a set of items
+func GenerateMerkleRoot(ctx context.Context, config merkledb.Config, merkleItems [][]byte, consumeBytes bool) (ids.ID, merkledb.MerkleDB, error) {
+	batchOps := make([]database.BatchOp, 0, len(merkleItems))
+
+	for _, item := range merkleItems {
+		key := utils.ToID(item)
```
patrick-ogrady, line 22: You should not hash this again

merkle/merkle_test.go
```diff
@@ -0,0 +1,64 @@
...
+			_, err := rand.Read(item)
+			require.NoError(b, err)
+			merkleItems = append(merkleItems, item)
+		}
+
+		var root ids.ID
+		var db merkledb.MerkleDB
+		var err error
+
+		defaultConfig := merkledb.Config{
```
patrick-ogrady, line 41: I don't really want to merge this until this config is largely removed. We need a "single instance" trie that is merkleDB-compatible for #950

merkle/merkle.go
iFrostizz: Removed the duplicated hashing!

PR #795, opened 2024-03-19
avalanchego v1.11.3 version bump

darioush:


vm/vm.go
```diff
@@ -369,6 +369,7 @@ func (vm *VM) Initialize(
 		vm.Logger(),
 		"",
 		syncRegistry,
+		nil,
```
patrick-ogrady, line 372: Can you add a TODO to populate this?

darioush: this is minVersion, I don't think it needs to be populated.

PR #796, opened 2024-03-19
Remove vm from simulator

### LINKED ISSUE #756: [x/programs] Remove the VM from the simulator
See @patrick-ogrady's comment here
https://github.com/ava-labs/hypersdk/pull/743#discussion_r1513329516

>I think there is a future where we could rip this out and just stub the host function to the WASM engine? If we did want to provide this "chain" feel, then it would make sense to keep it (but that may be better handled by an actual devnet instance of a VM with the WASM engine).

najeal:
This is related and closes #756 
It deletes the `vm` package and moved the important stuff in `internal`.
DB cleanup previously managed by the VM is now explicitly done in `cleanup` function of the simulator.


richardpringle:
@najeal going to take a little bit to get to this. I might also cause some conflicts. Sorry for the delay

najeal:
@richardpringle don't worry 🙂

najeal:
@richardpringle Do you think I can resolve conflicts?

richardpringle:
@najeal, I'm going to close this one since there were so many changes. 

I'm going to say that we shouldn't do this right away (at least for now). 

Thanks again for the contribution! Even if it didn't land, the PR will be here if someone wants to take a look in the future and we all really appreciate the efforts!

najeal:
@richardpringle Don’t hesitate to ping me when you want it, I would be glad to make it 😁

PR #799, opened 2024-03-21
[executor] use waiter pattern to record dependencies and enqueue tasks

### LINKED ISSUE #701: [executor] parallelizing reads in the execution engine

wlawt:
Another approach to the scheduling algorithm of #779 . Closes #701, #709.

### Overview 

Uses the `waiter` concurrency approach in the `Fetcher` package. 

`data` struct invariants:

- `data` only holds 1 Allocate/Write key OR many Read keys, but never both. This is represented using the bool flag `isAllocateWrite`

`waiter` invariants:

- the channel is made every time we update `e.edges` from its last blocked task.
- the channel is only closed when `blockers` is 0

### Tests

- Unit test coverage ~96.8%
- Integration test pass
- CI
- Benchmarks for performance

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -168,7 +168,7 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 	if t.dependencies != nil && t.dependencies.Len() > 0 {
 		// Ensure that we don't schedule a transaction that is
 		// blocked by another transaction currently running
-		for _, bID := range t.dependencies {
+		for bID := range t.dependencies {
 			// don't execute if any dependency task is in blocking
 			if e.blocking.Contains(bID) {
 				if e.metrics != nil {
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -5,7 +5,6 @@ package executor
 
 import (
 	"sync"
-	//"fmt"
 
 	"github.com/ava-labs/avalanchego/utils/set"
 
@@ -118,9 +117,6 @@ func (e *Executor) createWorker() {
 
 // Run executes [f] after all previously enqueued [f] with
 // overlapping [conflicts] are executed.
-// TODO: Handle read-only/write-only keys (currently the executor
-// treats everything still as ReadWrite, see issue below)
-// https://github.com/ava-labs/hypersdk/issues/709
 func (e *Executor) Run(conflicts state.Keys, f func() error) {
 	e.l.Lock()
 	defer e.l.Unlock()
@@ -146,11 +142,10 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 					lt.blocking = set.NewSet[int](defaultSetSize)
 				}
 				// key has ONLY a Read permission
-				// TODO: clean this up
-				if v.Has(state.Read) && !v.Has(state.Write) && !v.Has(state.Allocate) {
-					// lt contains a Write permission. If lt had only a Read
+				if v == state.Read {
+					// lt contains either Allocate or Write permission. If lt had only a Read
 					// permission, no dependency or blocking would be added
-					if latest.Permissions.Has(state.Write) {
+					if latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
 						t.dependencies.Add(lt.id) // t depends on lt to execute
 						lt.blocking.Add(id)       // lt is blocking this current task
 					}
@@ -158,8 +153,8 @@ func (e *Executor) Run(conflicts state.Keys, f func() error) {
 
 				// key contains a Write permission
 				if v.Has(state.Write) {
-					// lt contains a Read or Write permission
-					if latest.Permissions.Has(state.Read | state.Write) {
+					// lt contains either Read, Allocate, or Write
+					if latest.Permissions.Has(state.Read) || latest.Permissions.Has(state.Allocate) || latest.Permissions.Has(state.Write) {
 						t.dependencies.Add(lt.id)
 						lt.blocking.Add(id)
 					}
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -35,28 +35,12 @@ type Executor struct {
 	edges     map[string]*keyData
 }
 
-// keyData keeps track of the last blocked task, known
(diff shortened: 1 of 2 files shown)


executor/executor.go
```diff
@@ -41,118 +53,110 @@ func New(items, concurrency int, metrics Metrics) *Executor {
 		metrics:    metrics,
 		stop:       make(chan struct{}),
 		tasks:      make(map[int]*task, items),
-		edges:      make(map[string]int, items*2), // TODO: tune this
-		executable: make(chan *task, items),       // ensure we don't block while holding lock
+		edges:      make(map[string]*data, items*2), // TODO: tune this
```
wlawt, line 56: im not sure if `edges` is the best name now

executor/executor.go
```diff
@@ -32,7 +35,21 @@ type Executor struct {
...
+
+// invariant: [data] holds either 1 Allocate/Write or
+// a set of Reads, but never both
+//
+// invariant: [waiter] is made every time when setting
+// Allocate/Write or when adding the first Read. [waiter]
+// is closed only when [blockers] == 0
+type data struct {
+	allocateWrite int
+	reads         set.Set[int]
```
richardpringle, line 49: Something isn't adding up to me. It doesn't look like you're ever reading `allocateWrite` or `reads`, just adding to the `reads` set, or setting `allocateWrite` to an ID. Could you not accomplish the same thing with a bool? It's either blocking on reads (the number of which corresponds to `blockers`) or blocking on a single write. 

I'm not sure if you have plans to extend this, but you could definitely enforce your invariants by code via 
```suggestion
	isWrite bool
```
like I said, I could be missing something though

executor/executor_test.go
richardpringle, line 1: 👍

richardpringle (changes requested):
Can dismiss my review if I just don't understand

executor/executor.go
```diff
@@ -41,118 +58,115 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
+			// We can keep processing more Reads
+			if v == state.Read && key.reads != nil {
+				key.reads.Add(id)
+				key.blockers++
+				continue
+			}
+
+			// Allocate/Write was already set, or we're
+			// currently handling Reads, so we block
+			if key.allocateWrite != notSet || key.reads != nil {
```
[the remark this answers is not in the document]
wlawt: That's a good point, there's never a case where we don't hit this if statement. I think removing it will be easier to reason about as well

executor/executor.go
```diff
@@ -41,118 +50,108 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
-		t.dependencies = nil // free memory
-		e.executable <- t
-		if e.metrics != nil {
-			e.metrics.RecordExecutable()
+			// Block on write-after-write, read-after-write, or write-after-reads
+			if e.metrics != nil {
+				e.metrics.RecordBlocked()
+			}
+			e.l.Unlock()
+			<-key.waiter
```
wlawt, line 140: Do we want to block in `runWorker` instead?

executor/executor.go
```diff
@@ -41,118 +50,108 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
+			e.l.Lock()
+		}
+		// Key doesn't exist or we just processed Allocate/Write or many Reads
+		d := &data{isAllocateWrite: false, waiter: make(chan struct{}), blockers: 1}
+		if v.Has(state.Allocate) || v.Has(state.Write) {
+			d.isAllocateWrite = true
```
darioush, line 146:
```suggestion
		d := &data{
			isAllocateWrite: v.Has(state.Allocate) || v.Has(state.Write), 
			waiter: make(chan struct{}), 
			blockers: 1,
		}
```

darioush (commented):
this seems correct, do we have any benchmarks for the performance changes? (adds concurrency but also bookkeeping overhead)

wlawt:
@darioush will add benchmarks!

executor/executor.go
```diff
@@ -41,118 +50,108 @@ func New(items, concurrency int, metrics Metrics) *Executor {
...
+	}
 }
 
 // Run executes [f] after all previously enqueued [f] with
 // overlapping [conflicts] are executed.
-// TODO: Handle read-only/write-only keys (currently the executor
-// treats everything still as ReadWrite, see issue below)
-// https://github.com/ava-labs/hypersdk/issues/709
 func (e *Executor) Run(conflicts state.Keys, f func() error) {
 	e.l.Lock()
```
wlawt, line 114: From: https://github.com/ava-labs/hypersdk/blob/3c56945e64f0b76415b27bac30341c10772644bb/executor/executor.go 

I think this can applied here as well

executor/executor.go
```diff
@@ -32,7 +28,20 @@ type Executor struct {
...
+}
+
+// invariant: [data] holds either 1 Allocate/Write or
+// a set of Reads, but never both
+//
+// invariant: [waiter] is made every time when setting
+// Allocate/Write or when adding the first Read. [waiter]
+// is closed only when [blockers] == 0
+type data struct {
+	isAllocateWrite bool
```
richardpringle, line 41: 🚀

executor/executor.go
```diff
@@ -32,7 +28,20 @@ type Executor struct {
...
-	edges     map[string]int
+	edges     map[string]*data
+}
+
+// invariant: [data] holds either 1 Allocate/Write or
+// a set of Reads, but never both
+//
+// invariant: [waiter] is made every time when setting
+// Allocate/Write or when adding the first Read. [waiter]
+// is closed only when [blockers] == 0
```
richardpringle, line 39: This needs to be updated now

executor/executor.go
patrick-ogrady, line 140: We can't block like this in `Run`

patrick-ogrady: This could, for example, prevent us from enqueing other keys that don't block.

PR #800, opened 2024-03-22
[vryx] reduce locking in `executor` and `tstate`

patrick-ogrady:
- [x] Don't use lock for key operations array (just assign entry based on when kicking off execution)
- [x] Migrate to a sync.Map or multi-map for changed keys 
- [x] review locking in executor


patrick-ogrady:
processor:
image

wait exec: 
image

(if there was no bottleneck, wait exec would be the whole amount)

executor/executor.go
```diff
@@ -53,79 +57,93 @@ type task struct {
...
-					e.stopOnce.Do(func() {
-						e.err = err
-						close(e.stop)
-					})
-					return
-				}
+	t.l.Lock()
+	for b := range t.blocking { // works fine on non-initialized map
+		bt := e.tasks[b]
+		bt.l.Lock() // BUG: what if need to update [t], but holding outer (a waiting for b and b waiting for a)?
```
patrick-ogrady, line 86: Can change lock to only contain an atomic.Int (don't track dependencies).

patrick-ogrady:
interesting:

image

PR #805, opened 2024-03-24
[vryx] Add sponsor sub-partitions

patrick-ogrady:
Note: this approach does not guarantee distinct partitions


patrick-ogrady:
Smoothed out sent/mempools:
image
image

Increased "executable rate":
image

No impact on execution latency (just more stable):
image

PR #814, opened 2024-04-03
[executor] Introduce on-the-fly scheduling for Read/Allocate/Write keys

### LINKED ISSUE #701: [executor] parallelizing reads in the execution engine

wlawt:
Closes #701, #709 . 

### Overview

The executor keeps track of a dependency graph to determine if a `task` is executable or not. 

Properties of the dependency graph: 

- There is a graph for every key. 
- The nodes in the graph represent the `task` and the type of access its requesting for that key (Read/Allocate/Write).
- Generally, nodes in the graph are Write keys. Unless, if, it's the first node in the graph (i.e., the graph was empty), and it is a Read access, then it is also a node.
- An edge is formed between Write-Write, Read-Write, or Write-Read nodes.
- An edge is formed to the same node (self-edge) every time a Read is requested

### Test Plan

- Added unit tests in `executor_test.go`, coverage is ~97.5%
- Integration tests pass
- CI (with load & sync testing)

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -120,13 +120,15 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 
 	// Record dependencies
 	previousDependencies := set.NewSet[int](len(keys))
-	for k := range keys {
+	for k, v := range keys {
 		latest, ok := e.nodes[k]
 		if ok {
 			lt := e.tasks[latest]
 			if !lt.executed {
 				previousDependencies.Add(latest) // we may depend on the same task multiple times
-				lt.blocking[id] = t
+				if v.Has(state.Allocate) || v.Has(state.Write) {
+					lt.blocking[id] = t
+				}
 			}
 		}
 		e.nodes[k] = id
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -161,8 +161,18 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 					}
 					lt.blocking[id] = t
 				case (v.Has(state.Allocate) || v.Has(state.Write)) && n.isAllocateWrite:
-					t.dependencies.Add(int64(1))
-					lt.blocking[id] = t					
+					if len(lt.blocking) == 0 {
+						t.dependencies.Add(int64(1))
+						lt.blocking[id] = t
+					} else {
+						t.dependencies.Add(int64(len(lt.blocking)))
+						for b := range lt.blocking {
+							bt := e.tasks[b]
+							bt.l.Lock()
+							bt.blocking[id] = t
+							bt.l.Unlock()
+						}
+					}
 				}
 			}
 			lt.l.Unlock()
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -117,7 +117,7 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 	t := &task{
 		id:       id,
 		f:        f,
-		blocking: map[int]*task{},
+		blocking: make(map[int]*task),
 	}
 	e.tasks[id] = t
 
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -150,9 +150,16 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 					lt.blocking[id] = t
 					lt.l.Unlock()
 					continue
-				/*case (v.Has(state.Allocate) || v.Has(state.Write)) && !n.isAllocateWrite:
+				case (v.Has(state.Allocate) || v.Has(state.Write)) && !n.isAllocateWrite:
 					// blocked by all reads
-					t.dependencies.Add(int64(len(lt.blocking)))*/
+					t.dependencies.Add(int64(len(lt.blocking) + 1))
+					for b := range lt.blocking {
+						bt := e.tasks[b]
+						bt.l.Lock()
+						bt.blocking[id] = t
+						bt.l.Unlock()
+					}
+					lt.blocking[id] = t
 				case (v.Has(state.Allocate) || v.Has(state.Write)) && n.isAllocateWrite:
 					t.dependencies.Add(int64(1))
 					lt.blocking[id] = t					
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -151,6 +151,9 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 				case (v.Has(state.Allocate) || v.Has(state.Write)) && !n.isAllocateWrite:
 					// blocked by all reads
 					t.dependencies.Add(int64(len(lt.blocking)))*/
+				case (v.Has(state.Allocate) || v.Has(state.Write)) && n.isAllocateWrite:
+					t.dependencies.Add(int64(1))
+					lt.blocking[id] = t					
 				}
 			}
 			lt.l.Unlock()
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -145,10 +145,12 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 					lt.blocking[id] = t
 					lt.l.Unlock()
 					continue
-				/*case v == state.Read && n.isAllocateWrite:
+				case v == state.Read && n.isAllocateWrite:
 					t.dependencies.Add(int64(1))
 					lt.blocking[id] = t
-				case (v.Has(state.Allocate) || v.Has(state.Write)) && !n.isAllocateWrite:
+					lt.l.Unlock()
+					continue
+				/*case (v.Has(state.Allocate) || v.Has(state.Write)) && !n.isAllocateWrite:
 					// blocked by all reads
 					t.dependencies.Add(int64(len(lt.blocking)))*/
 				case (v.Has(state.Allocate) || v.Has(state.Write)) && n.isAllocateWrite:
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -7,7 +7,7 @@ import (
 	"sync"
 	"sync/atomic"
 
-	"github.com/ava-labs/avalanchego/utils/set"
+	// "github.com/ava-labs/avalanchego/utils/set"
 
 	"github.com/ava-labs/hypersdk/state"
 )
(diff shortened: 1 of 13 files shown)


executor/executor.go
```diff
@@ -20,160 +19,181 @@ const defaultSetSize = 8
...
+
+	tasks map[int]*task
+	nodes map[string]*node
+
+	err uatomic.Error
+}
+
+type node struct {
+	id              int
+	isAllocateWrite bool
```
patrick-ogrady, line 37: `isAllocateWrite` -> `modification`

executor/executor.go
```diff
@@ -20,160 +19,181 @@ const defaultSetSize = 8
...
-					close(e.executable)
-				}
-				e.l.Unlock()
-			case <-e.stop:
-				return
-			}
+	t.l.Lock()
+	for _, bt := range t.blocking {
+		// Reads would have zero dependencies
+		if bt.dependencies.Load() == 0 || bt.dependencies.Add(-1) > 0 {
```
patrick-ogrady, line 97: I would guess there is a bug here

executor/executor.go
```diff
@@ -5,175 +5,228 @@ package executor
...
 	"github.com/ava-labs/hypersdk/state"
+
+	uatomic "go.uber.org/atomic"
 )
 
-const defaultSetSize = 8
+// baseDependencies must be greater than the maximum number of dependencies
+// any single task could have. This is used to ensure a dependent task
+// does not begin executing a task until all dependencies have been enqueued.
+const baseDependencies = 100_000_000
```
wlawt, line 20: nit: maybe `maxDependencies`? since this number is an upper bound that a task can have and we only decrement our count?

executor/executor.go
```diff
@@ -5,175 +5,228 @@ package executor
...
 
 // New creates a new [Executor].
 func New(items, concurrency int, metrics Metrics) *Executor {
 	e := &Executor{
 		metrics:    metrics,
-		stop:       make(chan struct{}),
-		tasks:      make(map[int]*task, items),
-		edges:      make(map[string]int, items*2), // TODO: tune this
-		executable: make(chan *task, items),       // ensure we don't block while holding lock
+		nodes:      make(map[string]*task, items*2), // TODO: tune this
```
wlawt, line 50: would `graph` be a better name? not sure

executor/executor.go
patrick-ogrady: We could also just make the `items` argument into the executor have a much stronger invariant instead of using this `const`? Seems like that might be less arbitrary.

executor/executor.go
```diff
@@ -5,175 +5,228 @@ package executor
...
+		id:      id,
+		f:       f,
+		reading: []*task{},
+
+		blocked: make(map[int]*task),
+		readers: make(map[int]*task),
 	}
-	e.tasks[id] = t
+
+	// Add fake dependencies to ensure we don't execute the task
```
wlawt, line 156:
```suggestion
	// Add maximum number of dependencies to ensure we don't execute the task
```

executor/executor.go
```diff
@@ -5,175 +5,228 @@ package executor
...
-		e.edges[k] = id
+		e.nodes[k] = t
 	}
 
-	// Start execution if there are no blocking dependencies
-	if t.dependencies == nil || t.dependencies.Len() == 0 {
-		t.dependencies = nil // free memory
-		e.executable <- t
+	// Adjust dependency traker and execute if necessary
+	extraDependencies := baseDependencies - int64(dependencies.Len())
```
wlawt, line 205: nit: 
```suggestion
	difference := maxDependencies - int64(dependencies.Len())
```

executor/executor.go
wlawt: oh do you mean something like #577 (limiting number of state keys that a tx can specify) ? feel like that would be a stronger invariant

executor/executor.go
wlawt: yeah i was trying to think of a better name, but it also made sense with the right amount of verbosity & purpose of the name

executor/executor.go
```diff
@@ -5,175 +5,228 @@ package executor
...
+		}
+		e.runTask(t)
+	}
+}
+
 type task struct {
 	id int
 	f  func() error
+	// reading are the tasks that this task is using non-exclusively.
+	reading []*task
```
wlawt, line 76: maybe `shared`?

executor/executor.go
```diff
@@ -5,175 +5,228 @@ package executor
...
 
-	dependencies set.Set[int]
-	blocking     set.Set[int]
-
+	l        sync.Mutex
 	executed bool
+	// blocked are the tasks that are waiting on this task to execute
+	// before they can be executed. tasks should not be added to
+	// [blocked] after the task has been executed.
+	blocked map[int]*task
```
wlawt, line 83: maybe `exclusive`? I don't really like `exclusive` replacing `blocked` but I figured I'd suggest it, as it then makes the exclusive <> shared key relationship explicit and clear

patrick-ogrady: I don't think `exclusive` makes sense for this because non-exclusive requests are also stored here.

wlawt:
I think there is still a correctness bug somewhere (tokenVM sync test is failing)

executor/executor_test.go
```diff
@@ -5,16 +5,28 @@ package executor
 
 import (
 	"errors"
+	"math/rand"
```
wlawt, line 8: use `crypto/rand` instead

executor/executor_test.go
```diff
@@ -194,6 +225,749 @@ func TestStop(t *testing.T) {
...
+		// mix of unique and conflict keys in each tx
+		s := make(state.Keys, (numKeys + 1))
+
+		// random size of conflict keys to add
+		min := 1
+		max := 6
+		setSize := rand.Intn(max-min+1) + min             //nolint:gosec
+		randomConflictingKeys := set.NewSet[int](setSize) // indices of [conflictKeys]
+		for {
+			if randomConflictingKeys.Len() == setSize {
```
patrick-ogrady, line 911: typically, you'd just make this the termination condition of the for loop...this `if` immediately inside of a `for` is odd

patrick-ogrady: I think you do this a few places

executor/executor_test.go
```diff
@@ -194,6 +225,749 @@ func TestStop(t *testing.T) {
...
+
+	// Generate txs
+	for i := 0; i < numTxs; i++ {
+		// mix of unique and conflict keys in each tx
+		s := make(state.Keys, (numKeys + 1))
+
+		// random size of conflict keys to add
+		min := 1
+		max := 6
+		setSize := rand.Intn(max-min+1) + min             //nolint:gosec
```
patrick-ogrady, line 908: I think you can just do `max(rand.Intn(max), min)`

executor/executor_test.go
```diff
@@ -194,6 +225,749 @@ func TestStop(t *testing.T) {
...
+		for {
+			if randomConflictingKeys.Len() == setSize {
+				break
+			}
+			randomConflictingKeys.Add(rand.Intn(numKeys)) //nolint:gosec
+		}
+
+		// randomly pick if these conflict keys are
+		// going to be Read/Write
+		conflictMode := rand.Intn(2) //nolint:gosec
```
patrick-ogrady, line 919: I would do this per conflict key

executor/executor_test.go
```diff
@@ -194,6 +225,749 @@ func TestStop(t *testing.T) {
...
+	for {
+		if blocking.Len() == 10000 {
+			break
+		}
+		blocking.Add(rand.Intn(100_000)) //nolint:gosec
+	}
+
+	// make the conflict keys
+	for k := 0; k < numKeys; k++ {
+		conflictKeys = append(conflictKeys, ids.GenerateTestID().String())
```
patrick-ogrady, line 897: I recommend making the set of conflict keys larger (maybe 100) but still only using ~1-5 per tx

wlawt: Do we want to increase the total number of keys to be 100, as well?

patrick-ogrady: No. It is critical IMO that the number of keys per tx is small.

executor/executor.go
```diff
@@ -5,175 +5,228 @@ package executor
...
+				t.shared = append(t.shared, lt)
+				lt.readers[id] = t
+			} else {
+				// If we do need exclusive access to a key, we need to
+				// mark ourselves blocked on all readers ahead of us.
+				//
+				// If a task is a reader, that means it is not executed yet
+				// and can't mark itself as executed until all [shared] are
+				// cleared (which can't be done while we hold the lock for [lt]).
+				for _, rt := range lt.readers {
```
wlawt, line 182: I _think_ there might be a bug here, but I could be wrong. Consider two transactions, T1 and T2 both accessing key A and B. Suppose in the slow path, T1 and T2 have not executed.

T1 requests: A read, B write

T2 requests: A read, B write

In T1, these keys don't exist in `e.nodes` so we'll just add it

In T2, when we encounter A, `lt` is T1

We add ourselves, T2 to `lt.readers` for key A, and note that T1 hasn't executed so T1 is blocking T2. 

For key B in T2, we go through the `lt.readers` now since we're a Write, and one of the `rt` is us, T2. So we're saying T2 is blocked on ourself

wlawt: See `TestTwoConflictKeys` for a simplified unit test and see `TestLargeRandomReadsAndWrites` for arbitrary requesting

executor/executor.go
```diff
@@ -5,175 +5,235 @@ package executor
...
 	"github.com/ava-labs/hypersdk/state"
+
+	uatomic "go.uber.org/atomic"
 )
 
-const defaultSetSize = 8
+// maxDependencies must be greater than the maximum number of dependencies
+// any single task could have. This is used to ensure a dependent task
+// does not begin executing a task until all dependencies have been enqueued.
+const maxDependencies = 100_000_000
```
patrick-ogrady, line 20: I think the last tweak we should make is to make this a param you provide in `New`. I think this const is "jank".

executor/executor.go
```diff
@@ -14,21 +14,17 @@ import (
...
 // tasks with arbitrary conflicts on-the-fly.
 //
 // Executor ensures that conflicting tasks
 // are executed in the order they were queued.
 // Tasks with no conflicts are executed immediately.
 //
-// It is assumed that no single task has more than [maxDependencies]. If
-// this invariant is violated, some tasks will never execute and this code
-// could deadlock.
+// It is assumed that no single task has more than [maxDependencies].
```
patrick-ogrady, line 24: This comment should be above `New` imo.

patrick-ogrady (approved):

PR #829, opened 2024-04-11
[appenddb] single map

patrick-ogrady:


patrick-ogrady:
Not fully optimized yet but good enough for now:
image
image
image
image

PR #832, opened 2024-04-13
[vilmo] Remove File Moves + Fix Complex Reload Case

patrick-ogrady:
- [x] use cockroach/pebbledb (don't need latest)
- [x] remove batch pre-allocation in pebbledb (356TB of unneeded allocations over 1.8B txs)
- [x] test mmap on growing file
- [x] add failing test for complex reload
  - [x] add test for 2 dbs where we close/restart one and ensure checksums are the same
- [x] remove rename from `vilmo` (should also fix failing test)
  - [x] don't rewrite each record whenever a new batch is created
- [x] handle partially written file recovery (doesn't end in valid checksum)
- [x] remove batch map lookup when removing other keys (use pointer)
- [x] only update pointers on alive keys when rewrite to new file
- [x] Use value from alive iteration rather than calling db.Get (will access keys map)
- [x] Add open bytes when performing nullifications
- [x] Break if corrupt != nil


patrick-ogrady:
oops: 
image

patrick-ogrady:
Somewhat jumpy...seems like there may be room to reuse allocs here: 
image

PR #837, opened 2024-04-16
[executor] multi-key conflict bug

wlawt:
Consider two transactions, T1 and T2 both accessing key A and B. Suppose in the slow path, T1 and T2 have not executed.

T1 requests: A read, B write

T2 requests: A read, B write

In T1, these keys don't exist in e.nodes so we'll just add it

In T2, when we encounter A, lt is T1

We add ourselves, T2 to lt.readers for key A, and note that T1 hasn't executed so T1 is blocking T2.

For key B in T2, we go through the lt.readers now since we're a Write, and one of the rt is us, T2. So we're saying T2 is blocked on ourself

### DIFF
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -170,6 +170,9 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 			if v == state.Read {
 				// If we don't need exclusive access to a key, just mark
 				// that we are reading it and that we are a reader of it.
+				//
+				// We use a map for [reading] because a single task can have
+				// different keys that are all just readers of another task.
 				t.reading[lt.id] = lt
 				lt.readers[id] = t
 			} else {
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -73,7 +73,7 @@ type task struct {
 	id int
 	f  func() error
 	// reading are the tasks that this task is using non-exclusively.
-	reading []*task
+	reading map[int]*task
 
 	l        sync.Mutex
 	executed bool
@@ -147,7 +147,7 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 	t := &task{
 		id:      id,
 		f:       f,
-		reading: []*task{},
+		reading: make(map[int]*task),
 
 		blocked: make(map[int]*task),
 		readers: make(map[int]*task),
@@ -170,7 +170,9 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 			if v == state.Read {
 				// If we don't need exclusive access to a key, just mark
 				// that we are reading it and that we are a reader of it.
-				t.reading = append(t.reading, lt)
+				// TODO: currently a no-op when we add ourself multiple times
+				// but find a better way of handling [reading]
+				t.reading[lt.id] = lt
 				lt.readers[id] = t
 			} else {
 				// If we do need exclusive access to a key, we need to
--- a/executor/executor.go
+++ b/executor/executor.go
@@ -180,6 +180,7 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 				// and can't mark itself as executed until all [shared] are
 				// cleared (which can't be done while we hold the lock for [lt]).
 				for _, rt := range lt.readers {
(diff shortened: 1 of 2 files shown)


executor/executor.go
```diff
@@ -180,6 +180,10 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 				// and can't mark itself as executed until all [shared] are
 				// cleared (which can't be done while we hold the lock for [lt]).
 				for _, rt := range lt.readers {
+					// Don't block on ourself if we just recorded us reading
```
patrick-ogrady, line 183: As mentioned in DMs, I think there is still a weird scenario this doesn't address (assuming neither tx1 nor tx2 are executed):
```
tx1: A:read, B:write, C:read
tx2: A:read, B:write, C:read
```

executor/executor.go
patrick-ogrady, line 1: I think we should align `shared/readers`...they no longer sound like opposite ends of a single relationship.

executor/executor.go
```diff
@@ -72,8 +72,8 @@ func (e *Executor) work() {
 type task struct {
 	id int
 	f  func() error
-	// shared are the tasks that this task is using non-exclusively.
-	shared []*task
+	// reading are the tasks that this task is using non-exclusively.
+	reading map[int]*task
```
wlawt, line 76: maybe we can have a bool flag like `accessed` to see if any keys in a task have used a key non-exclusively. this would let us keep the array approach by not adding ourselves multiple times, and is better than the map approach since we aren't doing no-ops on that action

executor/executor.go
```diff
@@ -170,7 +170,9 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 			if v == state.Read {
 				// If we don't need exclusive access to a key, just mark
 				// that we are reading it and that we are a reader of it.
-				t.reading = append(t.reading, lt)
+				// TODO: currently a no-op when we add ourself multiple times
```
patrick-ogrady, line 173: I'm not sure this is that bad and/or deserves a TODO.

The alternative might be trying to track this across keys but that'll get complex with locking imo.

executor/executor.go
```diff
@@ -170,16 +170,22 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
...
 			} else {
 				// If we do need exclusive access to a key, we need to
 				// mark ourselves blocked on all readers ahead of us.
 				//
 				// If a task is a reader, that means it is not executed yet
-				// and can't mark itself as executed until all [shared] are
+				// and can't mark itself as executed until all [reading] are
 				// cleared (which can't be done while we hold the lock for [lt]).
 				for _, rt := range lt.readers {
+					// Don't block on ourself if we just recorded us reading
```
patrick-ogrady, line 185: nit: `Don't block on ourself if we already marked ourself as a reader`

executor/executor.go
```diff
@@ -170,8 +170,6 @@ func (e *Executor) Run(keys state.Keys, f func() error) {
 			if v == state.Read {
 				// If we don't need exclusive access to a key, just mark
 				// that we are reading it and that we are a reader of it.
-				// TODO: currently a no-op when we add ourself multiple times
-				// but find a better way of handling [reading]
 				t.reading[lt.id] = lt
```
patrick-ogrady, line 173: I think we should note clearly the scenario for why we use a map here (i.e. that a single task can have different keys that are all just readers of another task)

executor/executor_test.go
```diff
@@ -879,12 +874,10 @@ func TestLargeRandomReadsAndWrites(t *testing.T) {
...
 
-		// randomly pick if these conflict keys are
-		// going to be Read/Write
-		conflictMode := rand.Intn(2) //nolint:gosec
-
 		// add the random keys to tx
 		for k := range randomConflictingKeys {
+			// randomly pick if conflict key is Read/Write
+			conflictMode := rand.Intn(2) //nolint:gosec
 			switch conflictMode {
```
patrick-ogrady, line 881: You can just do `switch rand.Intn(2) {`

patrick-ogrady (approved):

PR #840, opened 2024-04-18
Cached state values in the program state

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/722

Add a cache to the program state in order to be able to return values that were previously written / read instead of always getting it from the runtime to benefit from a performance boost.


x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -155,7 +155,7 @@ pub fn state_keys(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut item_enum = parse_macro_input!(item as ItemEnum);
     // add default attributes
     item_enum.attrs.push(syn::parse_quote! {
-         #[derive(Clone, Copy, Debug)]
+         #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
```
richardpringle, line 158: 👍

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -1,5 +1,8 @@
+use std::hash::Hash;
+
 use borsh::{BorshDeserialize, BorshSerialize};
 
+use crate::state::Key;
```
richardpringle, line 5: nit: 
Can we combine this with the imports below, please?

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -1,5 +1,9 @@
-use crate::{memory::from_host_ptr, program::Program};
-use borsh::{BorshDeserialize, BorshSerialize};
+use crate::memory::into_bytes;
+use crate::program::Program;
+use crate::state::Error as StateError;
```
richardpringle, line 3: Can we merge these

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -1,5 +1,9 @@
...
+use borsh::{from_slice, to_vec, BorshDeserialize, BorshSerialize};
+use std::collections::HashMap;
+use std::hash::Hash;
```
richardpringle, line 6: And these two with the line below

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -35,27 +39,51 @@ pub enum Error {
...
+    cache: HashMap<K, Vec<u8>>,
 }
 
-impl State {
+impl<K> Drop for State<K>
+where
+    K: Into<Key> + Hash + PartialEq + Eq + Clone,
+{
+    fn drop(&mut self) {
+        if !self.cache.is_empty() {
```
richardpringle, line 55: 👍

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -68,30 +96,47 @@ impl State {
...
+
+            // Wrap in OK for now, change from_raw_ptr to return Result
+            into_bytes(val_ptr)
+        };
 
-        // Wrap in OK for now, change from_raw_ptr to return Result
-        unsafe { from_host_ptr(val_ptr) }
+        self.cache.insert(key, val_bytes.clone());
+
+        from_slice::<V>(&val_bytes).map_err(|_| StateError::Deserialization)
```
richardpringle, line 117: Did you try using the `HashMap::entry` method? It should enable you do get `val_bytes` without cloning anywhere.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -68,30 +96,47 @@ impl State {
...
     /// Delete a value from the hosts's storage.
     /// # Errors
     /// Returns an [Error] if the key cannot be serialized
     /// or if the host fails to delete the key and the associated value
-    pub fn delete<K>(&self, key: K) -> Result<(), Error>
-    where
-        K: Into<Key>,
-    {
+    pub fn delete(&mut self, key: K) -> Result<(), Error> {
+        self.cache.remove(&key);
```
richardpringle, line 125: 🤔, can you add a TODO comment to cache deletes as well. We want to defer the deletes to when the `State` is dropped as well, but that requires some changes to host functionality.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -68,30 +96,47 @@ impl State {
...
+
         unsafe { host::delete_bytes(&self.program, &key.into()) }
     }
+
+    /// Apply all pending operations to storage and mark the cache as flushed
+    fn flush(&mut self) -> Result<(), Error> {
+        for (key, value) in self.cache.drain() {
```
richardpringle, line 132: Can you add a TODO here as well? We should be batching this but obviously can't yet.

richardpringle (changes requested):
Nothing major

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Yes, thanks!

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: Sure! Done!

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: I did tried, but the entry API doesn't provide any convenient method other than chaining `and_modify` with `or_insert`, moreover the value of the `HashMap` is owned so I'm not sure if we can get around that.

iFrostizz: By permutating the line that calls `from_slice` with the one that inserts in the map, I could get around that clone.

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: Wrote it, why would we need to change the host functionality ? Can't we modify the `get` method to check if the value was deleted and return from the cache if not, and lastly from the runtime if not present in the rw cache ?

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: Done! Thanks!

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle: It would be better to pass all updates to the host at the end of the program-instance execution instead of explicitly calling `delete`. The goal is to minimize hopping back and forth when we don't need to. However, the host currently needs the functionality to handle that.

richardpringle:
🤔... looks like there's a failure somewhere.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -68,30 +96,47 @@ impl State {
     /// the host fails to read the key and value.
     /// # Panics
     /// Panics if the value cannot be converted from i32 to usize.
-    pub fn get<T, K>(&self, key: K) -> Result<T, Error>
```
richardpringle, line 71: 🤔... I'm just realizing that this API is flawed. Can you open another PR that fixes the return type here? We don't handle the case where there's nothing stored at the key. This function should return a `Result<Option<T>, Error>`.

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle: Not sure what you mean by "permutating".

I think it should look something more like this:
```rust
    pub fn get<V>(&mut self, key: K) -> Result<Option<V>, Error>
    where
        V: BorshDeserialize,
    {
        if let Some(val) = self.cache.get(&key) {
            return from_slice(&val)
                .map_err(|_| Error::Deserialization)
                .map(Some);
        }

        let host_ptr = unsafe {
            // cloning the key here isn't ideal, but we should be able to fix this later
            host::get_bytes(&self.program, &key.clone().into()).map_err(|_| Error::Read)?
        };

        // I'm not 100% that this is how the host handles things when a key doesn't exist
        if host_ptr == 0 {
            return Ok(None);
        }

        let bytes = into_bytes(host_ptr);
        let val = self.cache.entry(key).or_insert(bytes);

        from_slice(&val)
            .map_err(|_| Error::Deserialization)
            .map(Some)
    }
```

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle: https://github.com/ava-labs/hypersdk/issues/868

richardpringle (changes requested):
Just need to change the `V` back to a `&V`.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -71,30 +99,46 @@ impl State {
...
-
-        // Wrap in OK for now, change from_raw_ptr to return Result
-        from_host_ptr(val_ptr)
+        let val_bytes = if let Some(val) = self.cache.get(&key) {
+            val
+        } else {
+            let val_ptr = unsafe { host::get_bytes(&self.program, &key.clone().into())? };
+            if val_ptr < 0 {
+                return Err(Error::Read);
+            }
```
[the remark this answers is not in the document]
richardpringle: Nah, do this in a follow up.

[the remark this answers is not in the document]
iFrostizz: Yeah this is being tackled in https://github.com/ava-labs/hypersdk/pull/860

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -71,30 +99,46 @@ impl State {
...
+
+            // Wrap in OK for now, change from_raw_ptr to return Result
```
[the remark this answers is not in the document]
iFrostizz: Done!

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -38,27 +42,51 @@ pub enum Error {
...
+        }
     }
 
     /// Store a key and value to the host storage. If the key already exists,
     /// the value will be overwritten.
     /// # Errors
     /// Returns an [Error] if the key or value cannot be
     /// serialized or if the host fails to handle the operation.
-    pub fn store<K, V>(&self, key: K, value: &V) -> Result<(), Error>
+    pub fn store<V>(&mut self, key: K, value: V) -> Result<(), Error>
```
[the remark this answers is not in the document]
iFrostizz: Good suggestion, reverting that

richardpringle (approved):

PR #856, opened 2024-04-23
Remove Warp Integration

### LINKED ISSUE #855: Remove Warp integration
For many reasons, we will temporarily remove Warp integration. 

This will allow us to agree and solidify the Action <> Transaction relationship to enable Multi-Action support (#335)

Once Multi-Action is in, we will re-introduce Warp integration

wlawt:
Closes #855


examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -436,19 +432,19 @@ var _ = ginkgo.Describe("[Tx Processing]", func() {
 			// read: 2 keys reads, 1 had 0 chunks
 			// allocate: 1 key created with 1 chunk
 			// write: 2 keys modified (new + old)
-			transferTxConsumed := fees.Dimensions{191, 7, 12, 25, 26}
+			transferTxConsumed := fees.Dimensions{187, 7, 12, 25, 26}
```
wlawt, line 435: Removed https://github.com/ava-labs/hypersdk/blob/203e78b7c65a8b7c875fa4941f974292ce989f52/chain/transaction.go#L76C9-L76C17 which returns an `int` which is 4 bytes 

Which also resulted in updating the numbers below this line for the rest of the file as well

examples/tokenvm/tests/integration/integration_test.go
```diff
@@ -437,19 +432,19 @@ var _ = ginkgo.Describe("[Tx Processing]", func() {
 			// read: 2 keys reads, 1 had 0 chunks
 			// allocate: 1 key created
 			// write: 1 key modified, 1 key new
-			transferTxConsumed := fees.Dimensions{227, 7, 12, 25, 26}
+			transferTxConsumed := fees.Dimensions{223, 7, 12, 25, 26}
```
wlawt, line 435: same comment as above

chain/block.go
```diff
@@ -112,10 +99,7 @@ type StatelessBlock struct {
 	bytes  []byte
 	txsSet set.Set[ids.ID]
 
-	warpMessages map[ids.ID]*warpJob
-	containsWarp bool // this allows us to avoid allocating a map when we build
-	bctx         *block.Context
-	vdrState     validators.State
+	bctx *block.Context
```
patrick-ogrady, line 102: Haven't looked through the changes yet but I'd imagine you can remove this.

chain/block.go
```diff
@@ -298,12 +253,12 @@ func (b *StatelessBlock) initializeBuilt(
...
 // implements "block.WithVerifyContext"
-func (b *StatelessBlock) ShouldVerifyWithContext(context.Context) (bool, error) {
-	return b.containsWarp, nil
+func (*StatelessBlock) ShouldVerifyWithContext(context.Context) (bool, error) {
+	return false, nil
 }
 
 // implements "block.WithVerifyContext"
-func (b *StatelessBlock) VerifyWithContext(ctx context.Context, bctx *block.Context) error {
+func (b *StatelessBlock) VerifyWithContext(ctx context.Context, _ *block.Context) error {
```
patrick-ogrady, line 261: I think we can remove the usage of:
* `BuildBlockWithContext`
* `ShouldVerifyWithContext`
* `VerifyWithContext`

chain/dependencies.go
```diff
@@ -267,12 +255,7 @@ type Action interface {
 		timestamp int64,
 		actor codec.Address,
 		txID ids.ID,
-		warpVerified bool,
-	) (success bool, computeUnits uint64, output []byte, warpMessage *warp.UnsignedMessage, err error)
-
-	// OutputsWarpMessage indicates whether an [Action] will produce a warp message. The max size
-	// of any warp message is [MaxOutgoingWarpChunks].
-	OutputsWarpMessage() bool
```
patrick-ogrady, line 275: Make sure to mirror this change in the README

chain/result.go
```diff
@@ -17,17 +15,11 @@ type Result struct {
...
 }
 
 func (r *Result) Size() int {
 	size := consts.BoolLen + codec.BytesLen(r.Output) + fees.DimensionsLen + consts.Uint64Len
-	if r.WarpMessage != nil {
-		size += codec.BytesLen(r.WarpMessage.Bytes())
-	} else {
-		size += codec.BytesLen(nil)
-	}
+	size += codec.BytesLen(nil)
```
patrick-ogrady, line 22: This is incorrect. You should add nothing here.

patrick-ogrady: (you can just return the line above)

chain/result.go
```diff
@@ -73,15 +60,6 @@ func UnmarshalResult(p *codec.Packer) (*Result, error) {
 	}
 	result.Consumed = consumed
 	result.Fee = p.UnpackUint64(false)
```
patrick-ogrady, line 62: I think we should also check that there are no bytes left to read here (seems I forgot to add that check previously)...it would've caught the bug I commented on above.

examples/tokenvm/actions/outputs.go
```diff
@@ -4,37 +4,33 @@
...
+	OutputValueMisaligned    = []byte("value is misaligned")
+	OutputSymbolEmpty        = []byte("symbol is empty")
+	OutputSymbolIncorrect    = []byte("symbol is incorrect")
+	OutputSymbolTooLarge     = []byte("symbol is too large")
+	OutputDecimalsIncorrect  = []byte("decimal is incorrect")
+	OutputDecimalsTooLarge   = []byte("decimal is too large")
+	OutputMetadataEmpty      = []byte("metadata is empty")
+	OutputMetadataTooLarge   = []byte("metadata is too large")
+	OutputSameInOut          = []byte("same asset used for in and out")
+	OutputAnycast            = []byte("anycast output")
```
patrick-ogrady, line 32: Is this still used?

examples/tokenvm/storage/storage.go
```diff
@@ -357,9 +351,6 @@ func SetAsset(
 	binary.BigEndian.PutUint64(v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len+metadataLen:], supply)
 	copy(v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len+metadataLen+consts.Uint64Len:], owner[:])
 	b := byte(0x0)
```
patrick-ogrady, line 353: You no longer need `b`. You can remove it and remove the unnecessary alloc from this array (i.e. it should be a "boolean shorter").

vm/resolutions.go
```diff
@@ -175,38 +175,6 @@ func (vm *VM) processAcceptedBlock(b *chain.StatelessBlock) {
 		vm.Fatal("accepted processing failed", zap.Error(err))
 	}
 
-	// Sign and store any warp messages (regardless if validator now, may become one)
-	results := b.Results()
-	for i, tx := range b.Txs {
-		// Only cache auth for accepted blocks to prevent cache manipulation from RPC submissions
-		vm.cacheAuth(tx.Auth)
```
patrick-ogrady, line 182: This should still be done. This is unrelated to warp.

patrick-ogrady: That being said, this was a remnant of earlier crypto libs we used and it may be better to just remove this functionality from the auth interface.

In either case, we should do one or the other.

wlawt: I kept the `vm.cacheAuth`

examples/tokenvm/storage/storage.go
```diff
@@ -350,8 +350,6 @@ func SetAsset(
 	copy(v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len:], metadata)
 	binary.BigEndian.PutUint64(v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len+metadataLen:], supply)
 	copy(v[consts.Uint16Len+symbolLen+consts.Uint8Len+consts.Uint16Len+metadataLen+consts.Uint64Len:], owner[:])
-	b := byte(0x0)
```
patrick-ogrady, line 359: You still need to reduce the size of the allocation of `v` above ^

README.md
patrick-ogrady, line 1: I think there may be equivalent cleanups to do in the `tokenvm` README.

vm/vm.go
```diff
@@ -753,19 +753,6 @@ func (vm *VM) BuildBlock(ctx context.Context) (snowman.Block, error) {
 	return vm.buildBlock(ctx, nil)
```
patrick-ogrady, line 753: You should be able to remove the context arg from `buildBlock`

vm/resolutions.go
```diff
@@ -174,6 +174,11 @@ func (vm *VM) processAcceptedBlock(b *chain.StatelessBlock) {
 		vm.Fatal("accepted processing failed", zap.Error(err))
 	}
 
+	for i, tx := range b.Txs {
+		// Only cache auth for accepted blocks to prevent cache manipulation from RPC submissions
+		vm.cacheAuth(tx.Auth)
```
patrick-ogrady, line 179: Can you add a `TODO` to potentially remove this capability as it is unused and requires an extra iteration.

chain/block.go
```diff
@@ -252,34 +251,6 @@ func (b *StatelessBlock) initializeBuilt(
...
-		),
-	)
-	defer span.End()
-
-	// Proceed with normal verification
-	return b.verify(ctx, stateReady)
-}
-
 // implements "snowman.Block"
 func (b *StatelessBlock) Verify(ctx context.Context) error {
```
patrick-ogrady, line 255: We may want to unify `Build`/`build` and `Verify`/`verify` for now but not sure where we use them.

patrick-ogrady (approved):

Issue #359, opened 2023-08-15
[chain] Allow Transactions to output Multiple `Outputs`


patrick-ogrady:
This will allow `Programs` to emit arbitrary events that can be indexed by external providers. It should be possible to emit multiple `Outputs` because `Actions` may invoke multiple `Programs` and/or there may be multiple `Actions` in a `Transaction` (#335).

PR #858, opened 2024-04-24
[chain] Introduce Multi-Actions and Outputs in Transactions

wlawt:
Resolves: #335 , #359

### Overview

This PR introduces the ability to specify multiple actions within a single transaction. We limit the number of actions that can be specified within a transaction using `MaxActionsPerTx`. 

Furthermore, this PR introduces the ability to specify multiple outputs in `Result`. This lays the groundwork to enable outputting multiple Warp messages in an Action/Transaction. 

### Review Guide

- `chain/transaction.go` - Core logic to handle multiple actions
- `chain/result.go` - Enabling multiple outputs
- Integration tests for both MorpheusVM and TokenVM

### TODO

- [x] Add `Rule` to limit number of `Actions` per `Transaction`
- [x] Make Result.Output [][][]byte
- [x] Add multi-action tests in TokenVM integration
- [x] Add `ActionID` (`hash(txID, actionIndex)`)

### Tests

- CI
- Added more integration tests in TokenVM to include multi-action


chain/dependencies.go
```diff
@@ -152,6 +152,8 @@ type Rules interface {
 	GetStorageValueWriteUnits() uint64 // per chunk
 
 	FetchCustom(string) (any, bool)
+
+	GetMaxActionsPerTx() int
```
wlawt, line 156: should maybe be an `uint8`?

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -432,19 +432,19 @@ var _ = ginkgo.Describe("[Tx Processing]", func() {
 			// read: 2 keys reads, 1 had 0 chunks
 			// allocate: 1 key created with 1 chunk
 			// write: 2 keys modified (new + old)
-			transferTxConsumed := fees.Dimensions{187, 7, 12, 25, 26}
+			transferTxConsumed := fees.Dimensions{191, 7, 12, 25, 26}
```
wlawt, line 435: updated back to 191 because we do `p.PackInt(len(t.Actions))` which is 4 bytes

cli/prompt.go
```diff
@@ -71,24 +71,21 @@ func (h *Handler) PromptAsset(label string, allowNative bool) (ids.ID, error) {
 			if allowNative && input == symbol {
 				return nil
 			}
-			_, err := ids.FromString(input)
-			return err
+			_ = codec.LIDFromString(label, input)
```
wlawt, line 74: I don't think this is right

codec/address.go
```diff
@@ -78,3 +90,19 @@ func ParseAddressBech32(hrp, saddr string) (Address, error) {
...
+	return addr
+}
+
+func LIDFromString(hrp string, lid string) LID {
+	addr, err := ParseAddressBech32(hrp, lid)
+	if err != nil {
+		panic(err)
+	}
+	return LID(addr)
+}
```
wlawt, line 108: Need to fix this

codec/packer.go
```diff
@@ -81,6 +81,17 @@ func (p *Packer) UnpackAddress(dest *Address) {
...
+	p.p.PackFixedBytes(a[:])
+}
+
+func (p *Packer) UnpackActionID(required bool, dest *LID) {
+	copy((*dest)[:], p.p.UnpackFixedBytes(AddressLen))
+	if required && *dest == EmptyAddress {
+		p.addErr(fmt.Errorf("%w: LID field is not populated", ErrFieldNotPopulated))
+	}
+}
+
```
wlawt, line 94: rename to use `LID` over `ActionID`. and also, is there a way to combine with `PackAddress` and `UnpackAddress`?

eheap/eheap.go
```diff
@@ -22,22 +22,22 @@ type Item interface {
 // instead of grouping by expiry to support this feature, which makes it
 // less efficient).
 type ExpiryHeap[T Item] struct {
-	minHeap *heap.Heap[T, int64]
+	minHeap *heap.Heap[ids.ID, T, int64]
```
wlawt, line 25: need to think of a better way

heap/heap.go
```diff
@@ -6,74 +6,72 @@ package heap
...
-	"github.com/ava-labs/avalanchego/ids"
 )
 
 // Heap[I,V] is used to track objects of [I] by [Val].
 //
 // This data structure does not perform any synchronization and is not
 // safe to use concurrently without external locking.
-type Heap[I any, V cmp.Ordered] struct {
-	ih *innerHeap[I, V]
+type Heap[T comparable, I any, V cmp.Ordered] struct {
```
wlawt, line 15: see https://github.com/ava-labs/hypersdk/pull/863#discussion_r1581335275

cli/prompt.go
```diff
@@ -277,6 +274,25 @@ func (*Handler) PromptID(label string) (ids.ID, error) {
 	return id, nil
 }
 
+func (*Handler) PromptLID(label string) (codec.LID, error) {
```
wlawt, line 277: This is used the in `token-cli` (token-cli/cmd/action.go)

wlawt:
For now, ignore `x/*`

chain/dependencies.go
```diff
@@ -217,6 +219,10 @@ type Object interface {
 type Action interface {
 	Object
 
+	// GetActionID returns the LID for an [Action] in a [Transaction]. There may be
+	// multiple [Action]s, so we pass its index in the [Action] array along with the txID.
+	GetActionID(i uint8, txID ids.ID) codec.LID
```
patrick-ogrady, line 224: This should just be a helper (not on the action) -> no data in `Action` is used to construct this.

patrick-ogrady: also prefer `txID|index`

patrick-ogrady: Or you should comment that we do this to stay consistent with how address construction works

README.md
```diff
@@ -772,6 +772,8 @@ type Rules interface {
 	GetStorageValueWriteUnits() uint64 // per chunk
 
 	FetchCustom(string) (any, bool)
+
+	GetMaxActionsPerTx() uint8
```
patrick-ogrady, line 776: I think this should be right under `GetValidityWindow`. I purposely tried to leave `FetchCustom` as the last function in `Rules`.

chain/dependencies.go
```diff
@@ -248,14 +250,16 @@ type Action interface {
 	//
 	// An error should only be returned if a fatal error was encountered, otherwise [success] should
 	// be marked as false and fees will still be charged.
+	//
+	// TODO: Consider limiting number of outputs an [Action] can have
```
patrick-ogrady, line 254: If you have things well-architected, seems like an easy addition to `Rules`?

chain/result.go
```diff
@@ -11,19 +11,37 @@ import (
...
 
 func (r *Result) Marshal(p *codec.Packer) error {
 	p.PackBool(r.Success)
-	p.PackBytes(r.Output)
+
+	numOutputs := 0
+	for _, action := range r.Outputs {
+		numOutputs += len(action)
+	}
+	p.PackInt(len(r.Outputs))
```
patrick-ogrady, line 37: This is not included in the size calculation above?

chain/transaction.go
```diff
@@ -29,9 +29,8 @@ var (
 type Transaction struct {
 	Base *Base `json:"base"`
 
-	// TODO: turn [Action] into an array (#335)
-	Action Action `json:"action"`
-	Auth   Auth   `json:"auth"`
+	Actions []Action `json:"action"`
```
patrick-ogrady, line 32: nit `actions` instead of `action`

chain/transaction.go
```diff
@@ -102,17 +107,19 @@ func (t *Transaction) StateKeys(sm StateManager) (state.Keys, error) {
...
-				return nil, ErrInvalidKeyValue
+	for i, action := range t.Actions {
+		actionKeys := action.StateKeys(t.Auth.Actor(), codec.CreateLID(uint8(i), t.id))
+		sponsorKeys := sm.SponsorStateKeys(t.Auth.Sponsor())
+		for _, m := range []state.Keys{actionKeys, sponsorKeys} {
+			for k, v := range m {
+				if !keys.Valid(k) {
+					return nil, ErrInvalidKeyValue
+				}
+				stateKeys.Add(k, v)
```
patrick-ogrady, line 121: You need to union the access here, not just use the last access rule.

chain/transaction.go
```diff
@@ -102,17 +107,19 @@ func (t *Transaction) StateKeys(sm StateManager) (state.Keys, error) {
...
-	actionKeys := t.Action.StateKeys(t.Auth.Actor(), t.ID())
-	sponsorKeys := sm.SponsorStateKeys(t.Auth.Sponsor())
-	stateKeys := make(state.Keys)
-	for _, m := range []state.Keys{actionKeys, sponsorKeys} {
-		for k, v := range m {
-			if !keys.Valid(k) {
-				return nil, ErrInvalidKeyValue
+	for i, action := range t.Actions {
+		actionKeys := action.StateKeys(t.Auth.Actor(), codec.CreateLID(uint8(i), t.id))
+		sponsorKeys := sm.SponsorStateKeys(t.Auth.Sponsor())
```
patrick-ogrady, line 115: You can do sponsor keys after the action iteration.

chain/transaction.go
```diff
@@ -240,14 +249,19 @@ func (t *Transaction) PreExecute(
...
-		return ErrActionNotActivated
+	if len(t.Actions) > int(r.GetMaxActionsPerTx()) {
+		return ErrTooManyActions
 	}
-	if end >= 0 && timestamp > end {
-		return ErrActionNotActivated
+	for _, action := range t.Actions {
+		start, end := action.ValidRange(r)
+		if start >= 0 && timestamp < start {
+			return ErrActionNotActivated
```
patrick-ogrady, line 258: nit: it would probably be a lot more helpful to turn these into `fmt.Errorf` with info on which action or which index?

chain/transaction.go
```diff
@@ -305,106 +319,128 @@ func (t *Transaction) Execute(
...
-		return handleRevert(err)
-	}
-	if len(output) == 0 && output != nil {
-		// Enforce object standardization (this is a VM bug and we should fail
-		// fast)
-		return handleRevert(ErrInvalidObject)
-	}
-	if !success {
-		ts.Rollback(ctx, actionStart)
+		return &Result{false, [][][]byte{{utils.ErrBytes(rerr)}}, maxUnits, maxFee}, nil
```
patrick-ogrady, line 322: We probably want to document somewhere that a revert will not return an error per action? Otherwise, I could imagine there being some expectation that an err is returned per action?

chain/transaction.go
```diff
@@ -305,106 +319,128 @@ func (t *Transaction) Execute(
...
-	// Because we compute the fee before [Auth.Refund] is called, we need
-	// to pessimistically precompute the storage it will change.
-	for key := range s.SponsorStateKeys(t.Auth.Sponsor()) {
-		// maxChunks will be greater than the chunks read in any of these keys,
-		// so we don't need to check for pre-existing values.
-		maxChunks, ok := keys.MaxChunks([]byte(key))
-		if !ok {
-			return handleRevert(ErrInvalidKeyValue)
+		if !success {
+			txSuccess = false
```
patrick-ogrady, line 346: We need to skip all following actions if one is unsuccessful.

chain/transaction.go
```diff
@@ -305,106 +319,128 @@ func (t *Transaction) Execute(
...
-	// Check to see if the units consumed are greater than the max units
-	//
-	// This should never be the case but erroring here is better than
-	// underflowing the refund.
-	if !maxUnits.Greater(used) {
-		return nil, fmt.Errorf("%w: max=%+v consumed=%+v", ErrInvalidUnitsConsumed, maxUnits, used)
-	}
+		// Because the key database is abstracted from [Auth]/[Actions], we can compute
+		// all storage use in the background. KeyOperations is unique to a view.
+		allocates, writes := ts.KeyOperations()
```
patrick-ogrady, line 361: We should wait to do fee computation until the end because we shouldn't charge more to access the same key twice across different actions?

chain/transaction.go
```diff
@@ -305,106 +319,128 @@ func (t *Transaction) Execute(
...
+		//
+		// To avoid storage abuse of [Auth.Refund], we precharge for possible usage.
+		feeRequired, err := feeManager.MaxFee(used)
+		if err != nil {
 			return handleRevert(err)
 		}
+		totalFeeRequired += feeRequired
+
+		refund := maxFee - feeRequired
+		if refund > 0 {
```
patrick-ogrady, line 429: This should only ever be done once...we can't offer a refund per action from the tx max fee?

cli/prompt.go
patrick-ogrady: You should be decoding the entire input as cb58, don't include label here? https://github.com/ava-labs/avalanchego/blob/67a8b10abd9cf8644e819c0b28f7b25894cec7fc/ids/id.go#L42

cli/prompt.go
```diff
@@ -71,24 +71,21 @@ func (h *Handler) PromptAsset(label string, allowNative bool) (ids.ID, error) {
...
 	if asset != symbol {
-		assetID, err = ids.FromString(asset)
-		if err != nil {
-			return ids.Empty, err
-		}
+		assetID = codec.LIDFromString(label, asset)
 	}
-	if !allowNative && assetID == ids.Empty {
-		return ids.Empty, ErrInvalidChoice
+	if !allowNative && assetID == codec.EmptyAddress {
```
patrick-ogrady, line 87: we should not be comparing an `assetID` to a constant that is `EmptyAddress`

cli/spam.go
```diff
@@ -110,8 +110,8 @@ func (h *Handler) Spam(
 	if err != nil {
 		return err
 	}
-	action := getTransfer(keys[0].Address, 0)
-	maxUnits, err := chain.EstimateMaxUnits(parser.Rules(time.Now().UnixMilli()), action, factory)
+	actions := getTransfer(keys[0].Address, 0)
+	maxUnits, err := chain.EstimateMaxUnits(parser.Rules(time.Now().UnixMilli()), actions[0], factory)
```
patrick-ogrady, line 114: this should support `[]chain.Action`, the maxUnits is computed over all inputs?

cli/spam.go
```diff
@@ -492,7 +492,11 @@ func startIssuer(cctx context.Context, issuer *txIssuer) {
 				if result.Success {
 					confirmedTxs++
 				} else {
-					utils.Outf("{{orange}}on-chain tx failure:{{/}} %s %t\n", string(result.Output), result.Success)
+					for i := 0; i < len(result.Outputs); i++ {
```
patrick-ogrady, line 495: You should respect the invariant you implicitly created before (where any revert just populates the first output of the first action) or ensure each action output has a value

codec/address.go
```diff
@@ -21,10 +21,28 @@ const (
 	maxBech32Size = 90
 )
 
-type Address [AddressLen]byte
+type (
+	LID     [AddressLen]byte // Long ID
```
patrick-ogrady, line 25: You should not use `AddressLen` to define `LID`...there should be a specific constant for LID

codec/address.go
```diff
@@ -21,10 +21,28 @@ const (
...
+// CreateLID returns [LID] made from concatenating
+// some [i] with an [id].
+//
+// This is commonly used for creating an ActionID. We
+// keep the index|txID format to keep consistency with
+// how address construction works. Index is the index
+// in the [Action] array of a transaction. txID is the
+// ID of that transaction.
+func CreateLID(i uint8, id ids.ID) LID {
+	a := make([]byte, AddressLen)
```
patrick-ogrady, line 40: Likewise, use a dedicated constant.

codec/address.go
```diff
@@ -21,10 +21,28 @@ const (
 	maxBech32Size = 90
 )
 
-type Address [AddressLen]byte
+type (
+	LID     [AddressLen]byte // Long ID
+	Address LID
+)
 
 var EmptyAddress = [AddressLen]byte{}
```
patrick-ogrady, line 29: Once creating LID, you should not bypass it like this when aliasing? I think you'd want to do `EmptyAddress = LID{}`?

codec/packer.go
```diff
@@ -81,6 +81,17 @@ func (p *Packer) UnpackAddress(dest *Address) {
 	}
 }
 
+func (p *Packer) PackActionID(a LID) {
```
patrick-ogrady, line 84: You should make this generic and always use PackLID

codec/packer.go
patrick-ogrady: This is unneeded. Just always use the LID constant.

Even if you did want to do this, you probably wouldn't use a generic.

examples/morpheusvm/cmd/morpheus-cli/cmd/resolutions.go
```diff
@@ -57,26 +57,34 @@ func sendAndWait(
...
-		switch action := tx.Action.(type) { //nolint:gocritic
-		case *actions.Transfer:
-			summaryStr = fmt.Sprintf("%s %s -> %s", utils.FormatBalance(action.Value, consts.Decimals), consts.Symbol, codec.MustAddressBech32(consts.HRP, action.To))
+	}
+	for i := 0; i < len(result.Outputs); i++ {
+		for j := 0; j < len(result.Outputs[i]); j++ {
+			actor := tx.Auth.Actor()
+			for _, action := range tx.Actions {
+				summaryStr := string(result.Outputs[i][j])
+				if result.Success {
```
patrick-ogrady, line 69: You can be smarter about this I think....if the tx is not successful, there is only 1 error (according to your earlier handling).

chain/transaction.go
patrick-ogrady: I would actually not recommend deleting all previous outputs and overriding them with the revert error. Instead, I would keep what existed and then just include this error in the last action?

x/programs/examples/storage/storage.go
```diff
@@ -20,19 +19,19 @@ const (
 )
 
 func ProgramPrefixKey(id []byte, key []byte) (k []byte) {
-	k = make([]byte, consts.IDLen+1+len(key))
+	k = make([]byte, codec.AddressLen+1+len(key))
```
patrick-ogrady, line 22: This is not the right const to use here.

wlawt:
PRs that are merging into this one: #889 and #894

wlawt:
PRs that are merging into this one: https://github.com/ava-labs/hypersdk/pull/889 and https://github.com/ava-labs/hypersdk/pull/894

Patrick354 (approved):

README.md
```diff
@@ -648,24 +648,23 @@ type Action interface {
...
+	// If [Execute] returns an error, execution will halt and any state changes will revert.
 	Execute(
 		ctx context.Context,
 		r Rules,
 		mu state.Mutable,
 		timestamp int64,
 		actor codec.Address,
-		txID ids.ID,
-	) (success bool, computeUnits uint64, output []byte, err error)
+		actionID codec.LID,
```
patrick-ogrady, line 666: Hash `txID` + `index` to retain 32-byte

PR #863, opened 2024-04-25
[Heap] Make map index generic

wlawt:
Instead of `ids.ID` we make this a generic. This will unblock #858 and let us use `ActionID` as a comparable type.

### DIFF
--- a/examples/tokenvm/orderbook/orderbook.go
+++ b/examples/tokenvm/orderbook/orderbook.go
@@ -35,7 +35,7 @@ type OrderBook struct {
 	// dust orders from filling the heap.
 	//
 	// TODO: Allow operator to specify min creation supply per pair to be tracked
-	orders           map[string]*heap.Heap[*Order, float64]
+	orders           map[string]*heap.Heap[ids.ID, *Order, float64]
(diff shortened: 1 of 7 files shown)


examples/tokenvm/orderbook/orderbook.go
```diff
@@ -86,10 +86,10 @@ func (o *OrderBook) Add(txID ids.ID, actor codec.Address, action *actions.Create
 		return
 	case !ok && o.trackAll:
 		o.c.Logger().Info("tracking order book", zap.String("pair", pair))
-		h = heap.New[*Order, float64](o.maxOrdersPerPair+1, true)
+		h = heap.New[ids.ID, *Order, float64](o.maxOrdersPerPair+1, true)
 		o.orders[pair] = h
 	}
-	h.Push(&heap.Entry[*Order, float64]{
+	h.Push(&heap.Entry[ids.ID, *Order, float64]{
 		ID:    order.ID,
```
patrick-ogrady, line 93: I don't know if there is a way around this, but this is not very golang-y imo (to have all three of these values populated by `order` derivatives)

wlawt:
Close because I cherry-picked the commits into #858

PR #866, opened 2024-04-26
[x/programs] Fix program calling another program on the Rust side

iFrostizz:
The programs calling another contract functionality was fully broken, here are the steps taken
- ~~Support of a new step parameter called `Program`, the step ID should be passed the reason is that the previously ID was not decoding the parameter to the right format~~ Normalize all data from call params as a program ID when relevant
- Make the `storage/storage.go` temporarly fallback to the duplicated `examples/storage. This is a hack while waiting for @dboehm-avalabs rewrite and lets us read data from the cache
- Test of the counter on the Rust side to prove this works


x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -66,3 +66,241 @@ pub fn get_value_external(_: Context, target: Program, max_units: i64, of: Addre
...
+            max_units: 1000000,
+            params: vec![Param::String(PROGRAM_PATH.into())],
+            require: None,
+        });
+
+        plan.add_step(Step {
+            endpoint: Endpoint::Execute,
+            method: "initialize_address".into(),
+            max_units: 1000000,
+            params: vec![counter1_id.into(), alice_key.clone()],
```
richardpringle, line 107: Don't need to clone `alice_key` here... but I'll allow it

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -66,3 +66,241 @@ pub fn get_value_external(_: Context, target: Program, max_units: i64, of: Addre
...
+            require: None,
+        });
+
+        // run plan
+        let plan_responses = simulator.run_plan(&plan).unwrap();
+
+        // ensure no errors
```
richardpringle, line 114: I think both these steps are super clear, no need to add the extra comments.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -66,3 +66,241 @@ pub fn get_value_external(_: Context, target: Program, max_units: i64, of: Addre
...
+        assert!(
+            plan_responses.iter().all(|resp| resp.error.is_none()),
+            "error: {:?}",
+            plan_responses
+                .iter()
+                .filter_map(|resp| resp.error.as_ref())
+                .next()
+        );
+    }
+}
```
richardpringle, line 306: 👏

This is the way testing should always be done! 🥇

richardpringle (approved):
Well done

x/programs/examples/storage/storage.go
```diff
@@ -15,7 +15,8 @@ import (
 )
 
 const (
-	programPrefix = 0x0
+	programPrefix        = 0x0
+	ProgramChunks uint16 = 1
```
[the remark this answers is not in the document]
iFrostizz: Nope, it was for mimicking a block of the `vm/storage/storage.go` file but it ended up being removed apparently, will fix in a followup!

Issue #862, opened 2024-04-25
[x/programs] All host functions (that get linked) should follow the same two argument calling convention


richardpringle:
Similarly to #638, we want to serialize all the arguments into a single `Vec<u8>`. Further, **we need** to make sure that the bytes aren't dropped before they're read by the host. 

I'm going to propose that _all_ host functions (on the Rust side) have a safe wrapper and that their actual definition is written with a macro instead of being called directly. 

Right now, the code looks something like this:
```rust
#[link(wasm_import_module = "state")]
extern "C" {
    #[link_name = "put"]
    fn _put(caller: i64, key: i64, value: i64) -> i64;
}

pub unsafe fn put_bytes<V>(caller: &Program, key: &Key, value: &V) -> Result<(), Error>
where
    V: BorshSerialize,
{
    let value_bytes = borsh::to_vec(value).map_err(|_| Error::Serialization)?;
    let caller = to_host_ptr(caller.id())?;
    let value = to_host_ptr(&value_bytes)?;
    let key = to_host_ptr(key)?;

    match unsafe { _put(caller, key, value) } {
        0 => Ok(()),
        _ => Err(Error::Write),
    }
}
```
But this functionality should look something like this:
```rust
pub fn put_bytes<V>(caller: &Program, key: &Key, value: &V) -> Result<(), Error>
where
    V: BorshSerialize,
{
    call_host_fn! {
        wasm_import_module = "state";
        link_name = "put";
        args = (caller, key, value);
    }
}
```
Here the macro would actually define the external function. Alternatively, if there is no module, there should be another pattern to patch the following:
```rust
pub fn put_bytes<V>(caller: &Program, key: &Key, value: &V) -> Result<(), Error>
where
    V: BorshSerialize,
{
    call_host_fn!("put", (caller, key, value))
}
```
Inside the macro, the actual name of the function inside the `extern` block doesn't matter; it's only the link name that does something. 

Could also fix https://github.com/ava-labs/hypersdk/issues/849 while fixing this.

richardpringle:
Actually assigned to @iFrostizz

iFrostizz:
TODO: after merging https://github.com/ava-labs/hypersdk/pull/872, we should serialize the parameters being `(ptr_offset, data_len)*` with Borsh and just pass one of these `CPointer` which will simplify the API a lot.

richardpringle:
[...]

PR #869, opened 2024-04-26
[x/programs] safe wrapper around C ffi interface

iFrostizz:
Mentioning https://github.com/ava-labs/hypersdk/issues/862

In an effort to avoid repetitions when defining the C interface and make it safer, create a macro that should be used when defining new hosts.

TODO in a followup PR: Use `*const u8` for the pointer offset and `usize` for the pointer length instead of `i64` for everything

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/state.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/state.rs
@@ -202,14 +202,6 @@ macro_rules! call_host_fn {
 
         unsafe { ffi($caller, $key, $value) }
     }};
-
-    ($mod:literal, ($caller:ident, $key:ident)) => {
-        // TODO
-    };
-
-    ($mod:literal, ($caller:ident, $key:ident, $value:ident)) => {
-        // TODO
-    };
 }
 
 mod host {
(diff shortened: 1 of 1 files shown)


x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -117,52 +117,96 @@ impl Key {
...
+        unsafe { ffi($caller, $key, $value) }
+    }};
+
+    ($mod:literal, ($caller:ident, $key:ident)) => {
+        todo!()
+    };
+
+    ($mod:literal, ($caller:ident, $key:ident, $value:ident)) => {
+        todo!()
+    };
```
richardpringle, line 172:
```suggestion
```

richardpringle (changes requested):

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: Replaced the todo! with harmless TODO comments so it gives a starting point for further implementations

richardpringle (approved):

PR #872, opened 2024-04-29
[x/programs] Remove `SmartPtr` and `HostPtr`

iFrostizz:
Note: please first merge https://github.com/ava-labs/hypersdk/pull/866

In the goal of simplifying the go <> rust operations, the `SmartPtr` and `HostPtr` are replaced by a tuple of `*const u8` and `usize`.


x/programs/cmd/simulator/cmd/program.go
```diff
@@ -144,7 +144,6 @@ func programExecuteFunc(
 ) (ids.ID, []int64, uint64, error) {
 	// simulate create program transaction
 	programTxID, err := generateRandomID()
-
```
iFrostizz, line 147: This is the result of linting, did we changed rules recently ?

x/programs/examples/imports/program/program.go
```diff
@@ -59,12 +59,15 @@ func (i *Import) Register(link *host.Link, callContext program.Context) error {
...
 		wasmCaller *wasmtime.Caller,
-		programID int64,
-		function int64,
-		args int64,
+		programPtr int32,
+		programLen int32,
+		functionPtr int32,
+		functionLen int32,
+		argsPtr int32,
+		argsLen int32,
```
richardpringle, line 70: Instead of casting all of these, I think you can just make them `uint32` to begin with, can't you? 

Don't worry about it for this PR, but see if you can avoid casting when tackle the rest of #862

x/programs/examples/token.go
```diff
@@ -386,13 +386,13 @@ func (t *Token) RunShort(ctx context.Context) error {
...
 
-func (t *Token) GetUserBalanceFromState(ctx context.Context, programID ids.ID, userPublicKey ed25519.PublicKey) (res int64, err error) {
+func (t *Token) GetUserBalanceFromState(ctx context.Context, programID ids.ID, userPublicKey ed25519.PublicKey) (res uint32, err error) {
 	key := storage.ProgramPrefixKey(programID[:], append([]byte{uint8(Balance)}, userPublicKey[:]...))
 	b, err := t.db.GetValue(ctx, key)
 	if err != nil {
 		return 0, err
 	}
-	err = borsh.Deserialize(&res, b)
+	err = borsh.Deserialize(&res, b[4:])
```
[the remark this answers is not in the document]
richardpringle: Not quite sure, you should be able to write a test in `token/src/lib.rs` if you want to test it out (if I'm understanding you're question right).

x/programs/program/memory.go
```diff
@@ -158,37 +157,7 @@ func WriteBytes(m *Memory, buf []byte) (uint32, error) {
...
-	// read the range of PtrOffset + length from memory
-	bytes, err := memory.Range(s.PtrOffset(), s.Len())
-	if err != nil {
-		return nil, err
-	}
-
-	return bytes, nil
-}
-
-// AllocateBytes writes [bytes] to memory and returns the resulting SmartPtr.
```
richardpringle, line 191: 🙌

x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -47,17 +19,16 @@ thread_local! {
 /// Returns an [`StateError`] if the pointer or length of `args` exceeds
 /// the maximum size of a u32.
 #[allow(clippy::cast_possible_truncation)]
-pub fn to_host_ptr(arg: &[u8]) -> Result<HostPtr, StateError> {
-    let ptr = arg.as_ptr() as usize;
+pub fn to_ffi_ptr(arg: &[u8]) -> Result<CPointer, StateError> {
```
richardpringle, line 22: Can do this in a follow up since it would be a small change, but shouldn't the `CPointer` be defined in this module? Pointers are related to `memory`, not really a `program`.

x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -82,10 +53,10 @@ where
 /// Reconstructs the vec from the pointer with the length given by the store
 /// `host_ptr` is encoded using Big Endian as an i64.
 #[must_use]
-fn into_bytes(ptr: HostPtr) -> Option<Vec<u8>> {
+fn into_bytes(ptr: *const u8) -> Option<Vec<u8>> {
     GLOBAL_STORE
-        .with_borrow_mut(|s| s.remove(&(ptr as *const u8)))
-        .map(|len| unsafe { std::vec::Vec::from_raw_parts(ptr as *mut u8, len, len) })
+        .with_borrow_mut(|s| s.remove(&ptr))
+        .map(|len| unsafe { std::vec::Vec::from_raw_parts(ptr.cast_mut(), len, len) })
```
richardpringle, line 59: 👍

x/programs/tests/fixture/token.wasm
richardpringle, line 1: Oh wow... this has more than doubled in size, and that's not good. Not something we need to worry about right now, but we should definitely be aware.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -235,7 +237,7 @@ mod host {
     /// Deletes the bytes at key ptr from the host storage
     pub(super) unsafe fn delete_bytes(caller: &Program, key: &Key) -> Result<(), Error> {
         match call_host_fn! {
-            wasm_import_module = "state"
+             wasm_import_module = "state"
```
richardpringle, line 240: nit
```suggestion
            wasm_import_module = "state"
```

richardpringle (approved):

x/programs/tests/fixture/token.wasm
iFrostizz: I think that it's a release build that I've stored, could it explain that ?

x/programs/examples/imports/program/program.go
iFrostizz: I don't think that it's something that works in WASM, the only supported primitive types are `i32`, `i64`, `f32`, `f64`

iFrostizz: I'm getting the following error: "function types incompatible: expected func of type `(i32, i32, i32, i32, i32, i32, i64) -> (i64)`, found func of type `(externref, externref, externref, externref, externref, externref, i64) -> (i64)"

x/programs/tests/fixture/token.wasm
iFrostizz: debug*, I'll update soon

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: Good catch!

x/programs/rust/wasmlanche-sdk/src/memory.rs
iFrostizz: I updated it in this PR!

x/programs/tests/fixture/token.wasm
iFrostizz: Fixed! The token is still twice bigger!

x/programs/examples/imports/program/program.go
richardpringle: That's super weird. When using `wasmtime` directly in Rust, the only part of the type check that matters is the size. I just tried it out on the `public_functions.rs` code, and I swapped the `allocate` type signature to use `u32` instead of `i32`, and it worked exactly as expected without changing the actual wasm.

x/programs/host/link.go
```diff
@@ -136,12 +136,12 @@ func (l *Link) RegisterImportWrapFn(module, name string, paramCount int, f func(
...
 	valType := make([]*wasmtime.ValType, paramCount)
 	for i := 0; i < paramCount; i++ {
-		valType[i] = wasmtime.NewValType(wasmtime.KindI64)
+		valType[i] = wasmtime.NewValType(wasmtime.KindI32)
 	}
 
 	funcType := wasmtime.NewFuncType(
 		valType,
-		[]*wasmtime.ValType{wasmtime.NewValType(wasmtime.KindI64)},
+		[]*wasmtime.ValType{wasmtime.NewValType(wasmtime.KindI32)},
```
richardpringle, line 144: Ah, this is why you can't use `uint32`... wasmtime-go doesn't support it. Casting should be fine in that case

x/programs/examples/imports/program/program.go
iFrostizz: Indeed, I just swapped the function parameters of the `call_program` linked function, I don't think that there is anything else to do ?

richardpringle (approved):

PR #881, opened 2024-05-02
Host calls deserialize a single struct

### LINKED ISSUE #848: [x/programs] `state::host::get_bytes` should return an `Option<Vec<u8>>` in the `wasmlanche-sdk`
If we're handling converting the key into a host-pointer, we should also handle converting the return host-pointer into the `Vec<u8>` and handle the null-pointer.

richardpringle:
All host functions exposed to the wasm module should take the same arguments: a pointer and a length (both `int32` in Go, `*const u8` and `usize` in Rust). All functions will have their arguments serialized and deserialized to and from a struct. 

While this new calling convention may be slower for the case where the host only needs wasm primitives (`i32` or `i64`), it shouldn't be any slower when the host has to read from linear memory. Every case that we've come across so far is the latter.

Also closes #848


x/programs/examples/imports/pstate/pstate.go
```diff
@@ -83,32 +93,34 @@ func (i *Import) putFn(caller *program.Caller, idPtr int32, idLen int32, keyPtr
...
-	keyBytes, err := memory.Range(uint32(keyPtr), uint32(keyLen))
+	tmp := bytes[:4]
 	if err != nil {
-		i.log.Error("failed to read key from memory",
-			zap.Error(err),
-		)
 		return nil, err
 	}
+	l := int(binary.LittleEndian.Uint32(tmp))
+	fmt.Fprintf(os.Stderr, "l: %v\n", l)
```
richardpringle, line 110: deleteme

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -2,6 +2,8 @@ use crate::{memory::into_bytes, program::Program, state::Error as StateError};
 use borsh::{from_slice, to_vec, BorshDeserialize, BorshSerialize};
 use std::{collections::HashMap, hash::Hash, ops::Deref};
 
+use self::host::PutArgs;
```
richardpringle, line 5: deleteme

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -159,87 +186,65 @@ impl Key {
...
-
-    (
-        wasm_import_module = $mod:literal
-        link_name = $link:literal
-        args = ($caller:ident, $key:ident, $value:ident)
-    ) => {{
-        ffi_linker!($mod, $link, $caller, $key, $value);
-
-        unsafe { ffi($caller, $key, $value) }
-    }};
```
richardpringle, line 205: Don't worry @iFrostizz, this will come back, it'll just look a little different

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -159,87 +186,65 @@ impl Key {
...
+#[derive(BorshSerialize)]
+pub(super) struct GetAndDeleteArgs {
+    pub(super) caller: Program,
+    pub(super) key: Vec<u8>,
```
richardpringle, line 199: Wrote this in a different module, didn't edit when I moved. 
```suggestion
struct GetAndDeleteArgs {
    caller: Program,
    key: Vec<u8>,
```

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -159,87 +186,65 @@ impl Key {
...
-
-        let $caller = to_ffi_ptr($caller.id())?;
-        let $key = to_ffi_ptr($key)?;
-        let $value = to_ffi_ptr($value)?;
-    };
+#[derive(BorshSerialize)]
+pub(super) struct PutArgs {
+    pub(super) caller: Program,
+    pub(super) key: Vec<u8>,
+    pub(super) bytes: Vec<u8>,
```
richardpringle, line 193: Wrote this in a different module, didn't edit when I moved.
```suggestion
struct PutArgs {
    caller: Program,
    key: Vec<u8>,
    bytes: Vec<u8>,
```

dboehm-avalabs (approved):

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -45,7 +45,7 @@ pub fn inc(context: Context, to: Address, amount: i64) -> bool {
 #[public]
 pub fn inc_external(_: Context, target: Program, max_units: i64, of: Address, amount: i64) -> i64 {
     let params = params!(&of, &amount).unwrap();
```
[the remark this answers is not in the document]
iFrostizz: What would you change ?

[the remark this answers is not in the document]
richardpringle: If you look at the `params!` definition here:
https://github.com/ava-labs/hypersdk/blob/3fdf8bfdddd50496e45956efb3de1ca4aa0edb61/x/programs/rust/wasmlanche-sdk/src/params.rs#L7-L13

You can see that each param is serialized eagerly, then all the serialized params are put into a `Vec<Param>`, which is then also serialized into `Params(Vec<u8>)`. But with the changes in this PR, we serialized this _again_  ... it's super inefficient. It kind of made sense before, but it definitely doesn't make sense to do this now. 

I'm thinking that `Params(Vec<u8>)` -> `Vec<Box<dyn BorshSerialize>>` , and serialize it in one shot here:
https://github.com/ava-labs/hypersdk/pull/881/files/3fdf8bfdddd50496e45956efb3de1ca4aa0edb61#diff-df55dbd18bc9cea6bdf585205a2306b353f57c2a679a4e6c735a64d043aed93eR63

Make sense? Can clarify further if you need.

dboehm-avalabs (approved):

x/programs/rust/examples/counter/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: That does make sense, thanks for the clarification ! 
I think that it would be a sensible change, and after looking at the `serialize_param` it's doing exactly what Borsh does when treating dynamically sized arrays, if you look at the specification on https://borsh.io/
https://github.com/ava-labs/hypersdk/blob/f002156f55e422c79631904582f77ed0811d7419/x/programs/rust/wasmlanche-sdk/src/params.rs#L46-L51

edit: that won't work because `borsh::to_vec` takes as an input a generic which implements `BorshSerialize + ?Sized`

x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -16,24 +16,6 @@ thread_local! {
     static GLOBAL_STORE: RefCell<HashMap<*const u8, usize>> = RefCell::new(HashMap::new());
 }
 
-/// Converts a pointer to a i64 with the first 4 bytes of the pointer
-/// representing the length of the memory block.
-/// # Errors
-/// Returns an [`StateError`] if the pointer or length of `args` exceeds
-/// the maximum size of a u32.
```
iFrostizz, line 23: If this function is ever coming back again, we should make sure to write a correct comment

x/programs/rust/wasmlanche-sdk/src/params.rs
```diff
@@ -22,9 +19,11 @@ pub struct Param(Vec<u8>);
...
 pub struct Params(Vec<u8>);
 
-impl Params {
-    pub(crate) fn into_ffi_ptr(self) -> Result<CPointer, StateError> {
-        to_ffi_ptr(&self.0)
+impl Deref for Params {
+    type Target = Vec<u8>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
```
iFrostizz, line 26: good idea!

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -45,25 +44,32 @@ impl Program {
...
     ) -> Result<i64, StateError> {
-        // flatten the args into a single byte vector
-        let target = to_ffi_ptr(self.id())?;
-        let function = to_ffi_ptr(function_name.as_bytes())?;
-        let args = args.into_ffi_ptr()?;
+        #[link(wasm_import_module = "program")]
+        extern "C" {
+            #[link_name = "call_program"]
+            fn ffi(ptr: *const u8, len: usize) -> i64;
+        }
```
iFrostizz, line 54: This will play nicely with the macro when rewritten

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -124,15 +134,31 @@ where
...
-            }
+        let args_iter = self
+            .cache
+            .drain()
+            .map(|(key, val)| (key.into(), val))
+            .map(|(key, val)| PutArgs {
+                caller: self.program,
+                key: key.0,
+                bytes: val,
+            })
```
iFrostizz, line 157: nit

```rust
//.         .map(|(key, val)| (key.into(), val))
            .map(|(key, val)| PutArgs {
                caller: self.program,
                key: key.into().0,
                bytes: val,
            })
```

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -124,15 +134,31 @@ where
...
+            .map(|args| borsh::to_vec(&args).map_err(|_| StateError::Serialization));
+
+        for args in args_iter {
+            host::put_bytes(&args?)?;
```
iFrostizz, line 161: Can we consider calling `.collect()` to fail fast in case of a serialization error ?
It also makes the iteration below a little bit cleaner imo

```rust
    /// Apply all pending operations to storage and mark the cache as flushed
    fn flush(&mut self) -> Result<(), Error> {
        let args_iter: Result<Vec<Vec<u8>>, StateError> = self
            .cache
            .drain()
            .map(|(key, val)| (key.into(), val))
            .map(|(key, val)| PutArgs {
                caller: self.program,
                key: key.0,
                bytes: val,
            })
            .map(|args| borsh::to_vec(&args).map_err(|_| StateError::Serialization))
            .collect();

        for args in args_iter? {
            host::put_bytes(&args)?;
        }

        Ok(())
    }
```

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: Can you not just replace all of this with

```rust
macro_rules! call_host_fn {
    (
        wasm_import_module = $mod:literal
        link_name = $link:literal
        args = $value:ident
    ) => {{
        #[link(wasm_import_module = $mod)]
        extern "C" {
            #[link_name = $link]
            fn ffi(ptr: *const u8, len: usize) -> usize;
        }

        unsafe { ffi($value.as_ptr(), $value.len()) }
    }};
}
```

Then call it with

```rust
    /// Persists the bytes at key on the host storage.
    pub(super) fn put_bytes(bytes: &[u8]) -> Result<(), Error> {
        match call_host_fn! {
            wasm_import_module = "state"
            link_name = "put"
            args = bytes
        } {
            0 => Ok(()),
            _ => Err(Error::Write),
        }
    }
```

richardpringle: Yes! I want to figure out a better way to handle the return type instead of just hardcoding a size and then casting, so I decided to leave this for another PR. I want to figure out a way to normalize the return type for all host-calls. it's easy for most calls, but distinguishing between no-value and an error at an existing key becomes a little harder

x/programs/rust/examples/counter/src/lib.rs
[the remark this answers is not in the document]
richardpringle: First of all, I suggested something that wouldn't work; `BorshSerialize` isn't object-safe because its only method is generic. There might be some kind of workaround though. In any case, I'll create an issue.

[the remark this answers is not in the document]
richardpringle: New issue: https://github.com/ava-labs/hypersdk/issues/885

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -102,15 +102,25 @@ where
...
+
+            let bytes = into_bytes(ptr).ok_or(Error::InvalidPointer)?;
+
+            // TODO:
+            // should be able to do something like the following
+            // `let key = Key(args.key);`
+            // to avoid cloning. The problem is we convert into a Key without knowing
+            // that we can convert back into a K.
+            // Either we need the key to actually be `Key` instead of `Into<Key>`
+            // or we put the bound `K: From<Key>` as well.
```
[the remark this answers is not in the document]
richardpringle: https://github.com/ava-labs/hypersdk/issues/886

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -43,38 +44,44 @@ func (*Import) Name() string {
...
+	if len(args) != 2 {
+		return nil, errors.New("expected 2 arguments")
 	}
-	return i.deleteFn(caller, args[0], args[1], args[2], args[3])
+	return i.deleteFn(caller, args[0], args[1])
 }
 
-func (i *Import) putFn(caller *program.Caller, idPtr int32, idLen int32, keyPtr int32, keyLen int32, valuePtr int32, valueLen int32) (*types.Val, error) {
+type putArgs struct {
+	ProgramID [32]byte
```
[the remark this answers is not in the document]
richardpringle: Same response as above about the re-write

x/programs/rust/wasmlanche-sdk/src/memory.rs
richardpringle: I don't think it's ever coming back. The current version of `wasmtime` we're using only supports `size_of::<usize>() == size_of::<u32>()`, so the safety condition here isn't relevant. 

If there comes a point in time when we upgrade to support a 64-bit runtime, we'll have to do something to handle existing user code anyway.

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle: Iterators are lazy, if there's a failure in serialization, none of the `map` code will ever get called.

PR #883, opened 2024-05-04
x/simulator: plan using not created key returns an error

### LINKED ISSUE #861: [x/programs] Simulator should fail if a key that is used was not created
1. Change the `counter` program
```rust
#[public]
pub fn init(_: Context, _addr: Address) {
    //
}

#[public]
pub fn init2(_: Context) {
    //
}
```

2. Add this test in the `counter` program
```rust
    #[test]
    fn init_program() {
        let simulator = simulator::Client::new();

        let owner_key = String::from("owner");
        let [alice_key, bob_key] = ["alice", "bob"]
            .map(String::from)
            .map(Key::Ed25519)
            .map(Param::Key);

        let mut plan = Plan::new(owner_key.clone());

        plan.add_step(Step::create_key(Key::Ed25519(owner_key)));

        let program_id = plan.add_step(Step {
            endpoint: Endpoint::Execute,
            method: "program_create".into(),
            max_units: 1000000,
            params: vec![Param::String(PROGRAM_PATH.into())],
            require: None,
        });

        plan.add_step(Step {
            endpoint: Endpoint::Execute,
            method: "init2".into(),
            max_units: 10000000000,
            params: vec![program_id.into()],
            require: None,
        });

        plan.add_step(Step {
            endpoint: Endpoint::Execute,
            method: "init".into(),
            max_units: 10000000000,
            params: vec![program_id.into(), alice_key.clone()],
            require: None,
        });

        // run plan
        let plan_responses = simulator.run_plan(&plan).unwrap();

        // ensure no errors
        assert!(
            plan_responses.iter().all(|resp| resp.error.is_none()),
            "error: {:?}",
            plan_responses
                .iter()
                .filter_map(|resp| resp.error.as_ref())
                .next()
        );
    }
```

The test fails with an unreachable error from WASM, but that's only happening at the call to the `init` function, and it only happens when calling a function with more than one parameter. The fix here is to register the key:

```rust
        plan.add_step(Step {
            endpoint: Endpoint::Key,
            method: "key_create".into(),
            params: vec![alice_key.clone()],
            max_units: 0,
            require: None,
        });
```

We could try to error if the key has not been registered before.

najeal:
Closes #861 

@iFrostizz using a key that has not been created will fail except for `key` endpoint (endpoint for key creation)

### DIFF
--- a/x/programs/cmd/simulator/cmd/plan.go
+++ b/x/programs/cmd/simulator/cmd/plan.go
@@ -175,7 +175,7 @@ func (c *runCmd) Run(ctx context.Context) error {
 			zap.Any("params", step.Params),
 		)
 
-		params, err := c.createCallParams(ctx, c.db, step.Params)
+		params, err := c.createCallParams(ctx, c.db, step.Params, step.Endpoint)
 		if err != nil {
 			return err
 		}
@@ -270,7 +270,7 @@ func runStepFunc(
 }
 
 // createCallParams converts a slice of Parameters to a slice of runtime.CallParams.
-func (c *runCmd) createCallParams(ctx context.Context, db state.Immutable, params []Parameter) ([]actions.CallParam, error) {
+func (c *runCmd) createCallParams(ctx context.Context, db state.Immutable, params []Parameter, endpoint Endpoint) ([]actions.CallParam, error) {
 	cp := make([]actions.CallParam, 0, len(params))
 	for _, param := range params {
 		switch param.Type {
@@ -313,12 +313,15 @@ func (c *runCmd) createCallParams(ctx context.Context, db state.Immutable, param
 			key := val
 			// get named public key from db
 			pk, ok, err := storage.GetPublicKey(ctx, db, val)
-			if ok {
-				// otherwise use the public key address
-				key = string(pk[:])
-			}
 			if err != nil {
 				return nil, err
+			}
+			if !ok && endpoint != EndpointKey {
+				// using not stored named public key in other context than key creation
+				return nil, fmt.Errorf("%s: %s", ErrNamedKeyNotFound, val)
+			}
+			if ok {
+				key = string(pk[:])
 			}
 			cp = append(cp, actions.CallParam{Value: key})
 		case Uint64:


iFrostizz:
Hey @najeal , I think that it's a very reasonable fix, thank you very much!
Do you want to write a test that shows that the error is returned when not interacting with the key endpoint and that the key was not registered and it is not anymore after registering the key ?
If you don't want, I can too! As you want :)

najeal:
Thank for your answer !
I tried to make a unit test instead of creating one test over the CLI. I also wrapped a cleanup function (`manageCleanup()`) to reuse it in the test but the function has approximately the same name than `cleanupFn`.

I would also prefer to avoid initialize a VM but #796 is not yet merged

x/programs/cmd/simulator/cmd/config_test.go
```diff
@@ -29,7 +30,9 @@ func TestValidateAssertion(t *testing.T) {
 	}
 
 	for _, tt := range tests {
-		result, err := validateAssertion(tt.actual, &tt.assertion)
+		actual, err := borsh.Serialize(tt.actual)
+		require.NoError(err)
```
iFrostizz, line 34: It's much nicer being more explicit about the encoding, thanks!

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -328,13 +328,16 @@ func (c *runCmd) createCallParams(ctx context.Context, db state.Immutable, param
...
 			if err != nil {
 				return nil, err
 			}
+			if !ok && endpoint != EndpointKey {
+				// using not stored named public key in other context than key creation
+				return nil, fmt.Errorf("%s: %s", ErrNamedKeyNotFound, val)
+			}
+			if ok {
+				key = string(pk[:])
+			}
```
iFrostizz, line 340: It's interesting that this is silently failing in the case of the key not being found in the database, we should probably do something about this later

x/programs/cmd/simulator/cmd/plan_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateCallParams(t *testing.T) {
+	// Using a key not yet created must return an error
+	// This test creates a simulator to initialize the db we need to use
+	ctx := context.Background()
+	newSimulator := func() *Simulator {
+		logLevel := "info"
```
iFrostizz, line 15: Can we please avoid printing to the terminal when running the test ?

```suggestion
		logLevel := "error"
```

iFrostizz: It should print to stdout, not stderr by the way but we will fix this later.

x/programs/cmd/simulator/cmd/program.go
```diff
@@ -49,7 +49,7 @@ func (c *programCreateCmd) Run(ctx context.Context, log logging.Logger, db *stat
 		return newResponse(0), err
 	}
 	if !exists {
-		return newResponse(0), fmt.Errorf("%w: %s", ErrNamedKeyNotFound, c.keyName)
+		return newResponse(0), fmt.Errorf("%w: %s", ErrNamedKeyNotFound, *c.keyName)
```
iFrostizz, line 52: @najeal do you know why this is not caught at compile-time ? Isn't Go statically typed ?

iFrostizz (commented):
Small nit, very nice work thanks a lot!

x/programs/cmd/simulator/cmd/config_test.go
najeal: `go test` was not compiling before using `borsh.Serialize` I think simulator go tests are not run by the CI.
Not sure to understander your comment. Are you just saying it is better right now, or are you asking to be more explicit? 😀

x/programs/cmd/simulator/cmd/program.go
najeal: Yes Go is statically typed. I think validating the type over a string verb (ex: "%s") is out of scope of the compiler.
It is caught by `go vet` tool when running `go test`. The tool is enabled by default when using `go test` (you can disable it with `-vet=off`).

x/programs/cmd/simulator/cmd/plan_test.go
najeal: Why do we want to potentially print logs into stdout? Stdout already prints `Response` object into stdout. Mixing both we will not be able to Parse stdout to analyse result when running the simulator.

iFrostizz: Communicating over stdin/out is an anti-pattern and we want to avoid that, we worked towards building an abstract layer of communication and will switch to grpc and / or ipc later on

x/programs/cmd/simulator/cmd/program.go
iFrostizz: Thanks!

x/programs/cmd/simulator/cmd/plan_test.go
najeal: Ok it makes sense 👍

x/programs/cmd/simulator/cmd/plan.go
iFrostizz: of course 🤦 thanks

iFrostizz (approved):
This is looking good to me, looking for an additional approval

x/programs/cmd/simulator/cmd/config_test.go
iFrostizz: It was a CI issue, some static checks were not done and it led me to think that go was inferring the type, it was an actual bug that was apparently ignored! I'm not asking you to change anything

x/programs/cmd/simulator/cmd/program.go
iFrostizz: Yep, again an issue with CI

richardpringle (approved):

richardpringle:
@najeal, just need to update the branch and fix the lint-issue

najeal:
@richardpringle lint is fixed and branch is up-to-date with `main` 👍

richardpringle (approved):

Issue #880, opened 2024-05-02
[x/programs] Synchronous communication with the simulator through stdin/out


iFrostizz:
The current design of the simulator makes it relatively impractical to interact with programs because plans have to be a list of steps, some issues are arising from that:
- It's harder to debug program execution when writing plans on the Rust side because when it's failing, we only know that it happened somewhere in the plan
- There is no way to execute and assert that the return value of a public function is `x` more than once in a plan

iFrostizz:
TODO: 
- use a different communication "protocol" such as os sockets or a proper rpc server, logs should probably be printed to stdout and this prevents it
- write tests that shows the new capabilities of the simulator (do stuff on an intermediary state)
- add a "forking" capability in order to be able to write fixtures for programs tests

PR #884, opened 2024-05-06
[x/programs] Interpreter mode to allow intermediate state fetching through plan run

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/880

Changes:
- Removal of the idea of plan on the Go side, we can now execute only one step and still have the binary running
- New `interpreter` subcommand which parses other commands from stdin from the time it is called
- Switch from cobra to argparse in order to be able to parse arguments from a `string`


x/programs/cmd/simulator/src/lib.rs
```diff
@@ -199,15 +219,9 @@ pub struct Client {
...
-impl Default for Client {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl Client {
     #[must_use]
-    pub fn new() -> Self {
+    pub fn new() -> Result<Self, Box<dyn Error>> {
```
richardpringle, line 224: I don't love this, I would rather just panic

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -220,29 +234,43 @@ impl Client {
...
+            .arg("interpreter")
+            .arg("--cleanup")
+            .arg("--log-level")
+            .arg("error") // TODO control this through an env variable
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .spawn()?;
+
+        let stdin = &mut child.stdin.unwrap();
+        let stdout = &mut child.stdout.unwrap();
```
richardpringle, line 255:
```suggestion
        let (stdin, stdout) = (&mut child.stdin, &mut child.stdout) else {
            panic!("could not attach to the child process stdio");
        };
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -271,58 +299,82 @@ where
...
+    for step in &plan.steps {
+        let resp = run_step::<T>(
+            stdin,
+            stdout,
+            &Plan {
+                caller_key: plan.caller_key.clone(),
+                steps: vec![step.clone()],
+            },
+        )?;
+        items.push(resp);
```
richardpringle, line 320:
```rust
let items: Result<Vec<T>, _> = plan.steps
    .iter()
    .map(|step| Plan { caller_key: plan.caller_key.clone(), steps: step.clone() })
    .map(|plan| run_step(stdin, stdout, &plan))
    .collect()
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -271,58 +299,82 @@ where
...
-        for line in String::from_utf8_lossy(&output.stderr).lines() {
-            println!("{line}");
-        }
-        println!("stdout");
-        for line in String::from_utf8_lossy(&output.stdout).lines() {
-            println!("{line}");
-        }
-        return Err(String::from_utf8(output.stdout)?.into());
-    }
+    let run_command = "run --stdin\n".to_string();
```
richardpringle, line 333:
```suggestion
    let run_command = "run --stdin\n";
```

x/programs/tests/fixture/counter.wasm
richardpringle, line 1: I don't think we use this anymore, do we?

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -271,58 +299,82 @@ where
...
-        }
-        return Err(String::from_utf8(output.stdout)?.into());
+    for step in &plan.steps {
+        let resp = run_step::<T>(
+            stdin,
+            stdout,
+            &Plan {
+                caller_key: plan.caller_key.clone(),
+                steps: vec![step.clone()],
+            },
```
richardpringle, line 318: Something seems off here. I think you moved the caller-key, right? I think you could abstract this a little better.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Yeah I did some changes and forgot to remove this one, thanks!

x/programs/cmd/simulator/cmd/root.go
```diff
@@ -32,70 +35,145 @@
...
-	cmd.CompletionOptions.HiddenDefaultCmd = true
-	cmd.DisableAutoGenTag = true
-	cmd.SilenceErrors = true
-	cmd.SetHelpCommand(&cobra.Command{Hidden: true})
-	cmd.PersistentFlags().StringVar(&s.logLevel, "log-level", "info", "log level")
+	if !interpreterMode {
+		s.Init()
+	}
+	s.log.Debug("simulator args", zap.Any("args", args))
+[]
```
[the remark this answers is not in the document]
iFrostizz: nit, fixed

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Yes, the idea is to be able to change the caller from step to step, while still being able to define a plan with the same caller for each step in it. It is now possible to also define a plan with caller A and another one with caller B in the same simulation. I have used a `Step` here instead now.

x/programs/tests/fixture/counter.wasm
iFrostizz: Yep, removing!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Nice, thanks

x/programs/cmd/simulator/src/lib.rs
iFrostizz: I ended up defining a new `SimulatorStep`, please take a look https://github.com/ava-labs/hypersdk/pull/884/files#diff-33034c207f9554212c375b13284c24b46f009907280babee29791aa8c0ec11baR296-R314

x/programs/cmd/simulator/src/lib.rs
iFrostizz: I love this pattern thank you so much

x/programs/cmd/simulator/cmd/interpreter.go
```diff
@@ -0,0 +1,29 @@
...
+type InterpreterCmd struct {
+	cmd *argparse.Command
+}
+
+func (s *InterpreterCmd) New(parser *argparse.Parser) {
+	s.cmd = parser.NewCommand("interpreter", "Read input from a buffered stdin")
+}
+
+func (s *InterpreterCmd) Run(ctx context.Context, log logging.Logger, args []string) (*Response, error) {
+	resp := newResponse(0)
```
iFrostizz, line 22: The response has an ID of 0 because we usually don't care about this return, I'm thinking about if it would be necessary to assign it a correct identifier by modifying the `Run` function so that it takes a `startId` parameter.

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -4,86 +4,108 @@
...
+	c.programIDStrMap = programIDStrMap
+	c.cmd = parser.NewCommand("run", "Run a HyperSDK program simulation plan")
+	c.stdin = c.cmd.Flag("", "stdin", &argparse.Options{
+		Help:     "name of the key to use to deploy the program",
+		Required: false,
+		Default:  false,
+	})
+	c.file = c.cmd.String("", "file", &argparse.Options{
+		Required: false,
+	})
```
iFrostizz, line 60: Write a test that uses the file parameter, let's open an issue

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -4,86 +4,108 @@
...
+	c.reader = reader
+}
+
+func (c *runCmd) Run(ctx context.Context, log logging.Logger, args []string) (*Response, error) {
+	c.log = log
+	resp1 := newResponse(0)
+	resp1.setTimestamp(time.Now().Unix())
+	err := resp1.Print()
+	if err != nil {
+		return newResponse(0), err
```
iFrostizz, line 71: Same as above about the ID of 0 by default. We also return a "default" response with a missing timestamp in case of an error because this response will not be printed anyway, is that the correct way to return a kind of "nil" reponse ?

x/programs/cmd/simulator/cmd/interpreter.go
richardpringle: 🤷 

Seems fine for now. This will be easy to change later if we want to

x/programs/cmd/simulator/cmd/plan.go
richardpringle: What is the file-parameter? To use a file for input? Honestly, I wouldn't do this. It should be super easy to just read a file into stdin or pipe stdout to a new file.

x/programs/cmd/simulator/cmd/plan.go
richardpringle: That's a good question. I'm inclined to think that we should just print to stderr and exit with a non-0 exit code. Maybe there's a case where we should handle the error though...

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -4,86 +4,108 @@
...
-				return err
-			}
-			return r.Run(cmd.Context())
-		},
+	err = c.Init()
+	if err != nil {
+		return newResponse(0), err
 	}
+	err = c.Verify()
+	if err != nil {
```
richardpringle, line 78:
```suggestion

	if err := c.Verify(); err != nil {
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -223,109 +225,99 @@ impl Client {
         Self { path }
     }
 
+    fn read_response(
```
richardpringle, line 228: I don't think this should be a static method, you should just move it to be a normal function as it's private anyway. To be honest, I'm not sure I would put the logic in its own function to begin with. I'll come back here when I see all the places it's being used and leave another comment

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -223,109 +225,99 @@ impl Client {
...
-            .map_err(|e| format!("failed to write to stdin: {e}"))?;
-    }
+        let res = Self::run_steps(stdin, lines, plan)?;
 
-    child
-        .wait_with_output()
-        .map_err(|e| format!("failed to wait for child: {e}").into())
-}
+        // kill the process to avoid it to read an EOF from stdin
+        child.kill().unwrap();
```
richardpringle, line 278: Use `expect` instead of `unwrap`

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -223,109 +225,99 @@ impl Client {
...
+    fn run_steps(
+        stdin: &mut ChildStdin,
+        lines: &mut Lines<BufReader<&mut ChildStdout>>,
+        plan: &Plan,
+    ) -> Result<Vec<PlanResponse>, Box<dyn Error>> {
+        let mut items: Vec<PlanResponse> = Vec::new();
+        for step in &plan.steps {
+            let resp = Self::run_step(stdin, lines, plan.caller_key.clone(), step.clone())?;
+            items.push(resp);
+        }
```
richardpringle, line 292: Please use iterators with `collect` instead of pushing to a `Vec`. That way `items` doesn't ever need to be mutable. 

In theory, it's possible that the compiler doesn't figure out the proper size of the `Vec` ahead of time, which means that every time you reach _capacity_, you have to re-allocate the whole vec with new space. If you just use iterators, you never have to worry about this because the compiler knows how to create `items` with the exact capacity it needs (since it should be immutable). Always favour iterators except in the case of async code or where nested results make it hard to early return.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -223,109 +225,99 @@ impl Client {
...
+            .ok_or("EOF")?
+            .map_err(|e| format!("failed to read from stdout: {e}"))?;
+
+        let resp = serde_json::from_str(&text)
+            .map_err(|e| format!("failed to parse output to json: {e}"))?;
+
+        Ok(resp)
+    }
+
+    fn write_stdin(stdin: &mut ChildStdin, bytes: &[u8]) -> Result<(), String> {
```
richardpringle, line 242: Same with this, there's no reason to have this as a static method.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -223,109 +225,99 @@ impl Client {
...
-        .lines()
-        .filter(|line| !line.trim().is_empty())
-    {
-        let item = serde_json::from_str(line)
-            .map_err(|e| format!("failed to parse output to json: {e}"))?;
-        items.push(item);
+    fn run_steps(
+        stdin: &mut ChildStdin,
+        lines: &mut Lines<BufReader<&mut ChildStdout>>,
+        plan: &Plan,
```
richardpringle, line 286: Just consume the plan so you don't have to clone steps

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -223,109 +225,99 @@ impl Client {
...
+            let resp = Self::run_step(stdin, lines, plan.caller_key.clone(), step.clone())?;
+            items.push(resp);
+        }
+        Ok(items)
     }
 
-    Ok(items)
+    fn run_step(
+        stdin: &mut ChildStdin,
+        lines: &mut Lines<BufReader<&mut ChildStdout>>,
```
richardpringle, line 298: When you're always passing in the things, this should be a good hint that you should just create a new type with methods like `write`, `run_step`, and `run_plan`.

richardpringle: Also, you shouldn't need to use this `Lines` type directly. You should be dealing with `Iterator<Item = Result<Resposne, Error>>` instead.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -303,4 +303,66 @@ mod tests {
                 .next()
         );
     }
+
+    #[test]
+    fn wrong_assertion() {
```
richardpringle, line 308:
```suggestion
    fn assertion_can_fail() {
```
We shouldn't be testing this here though. There should be tests on the Go side for this.

richardpringle (changes requested):
Overall architecture is good! 

I would change up the implementation such that you can have a clean abstraction with for calling `run_step` and `run_plan` though. I would wrap `stdin` and an iterator over results of responses. I think if you'll do that, the code will be a lot easier to follow

x/programs/cmd/simulator/cmd/interpreter.go
iFrostizz: Agree!

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: Would you remove this ? This is a property that I broke during the writing of this feature and would have been great to figure out what is missing more quickly with an already written test. If it's written on the Go side, I guess that it would be interacting more directly with the simulator and anything that is between Go and Rust (so essentially communication) is left and could potentially be buggy.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: That's a fair point, my reasoning was that interacting with the simulator should only be done in the `Client`

iFrostizz: It's used in 3 places already and it's not looking very good when inlined.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: done!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Thanks, that makes sense !

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Moving this out

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Good point, thanks

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: renamed the test

x/programs/cmd/simulator/src/lib.rs
iFrostizz: I think that we could just refactor the `Client` a bit to do that and have the over-used parameters as fields of the struct

iFrostizz: Yeah you're right, a client is should not be tied to a simulator process. So I will create a new struct that denotes a single simulator process.

iFrostizz: updated by creating a `SimulatorChild` which manages all that.

x/programs/cmd/simulator/cmd/plan.go
iFrostizz: done and updated to other occurences !

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +228,95 @@ impl Client {
...
+        step: &Step,
+    ) -> Result<PlanResponse, Box<dyn Error>> {
+        let run_command = "run --stdin\n";
+        self.write_stdin(run_command.as_bytes())?;
+        let _ = self.read_response()?;
+
+        let step = SimulatorStep { caller_key, step };
+        let mut input =
+            serde_json::to_string(&step).map_err(|e| format!("failed to serialize json: {e}"))?;
+        input.push('\n');
```
richardpringle, line 290: Performance isn't really a concern here, but if you're always using best performance practices, these things will be second nature when performance does matter. 

```rust
let step = serde_json::to_vec(&step)?;
self.writer.write_all(&step)?;
self.writer.write_all(b"\n")?;
self.writer.flush()?;
self.responses.next().unwrap_or(Err("EOF"))
```
This way, we give `serde_json` the ability to allocate the exact size for the `Vec`. The bytes are also copied regardless, There's no need to have to allocate constant data like the newline character. 

While the compiler might optimize both implementations to the same thing, you would have to verify to make sure that that's the case (and things could change between Rust versions). It's better to give the compiler the easiest thing to optimize, especially if it doesn't look any worse.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +228,95 @@ impl Client {
...
+        self.stdin
+            .write_all(bytes)
+            .map_err(|e| format!("failed to write to stdin: {e}"))?;
+        self.stdin
+            .flush()
+            .map_err(|e| format!("failed to flush from stdin: {e}"))?;
 
-    Ok(resp)
+        Ok(())
+    }
```
richardpringle, line 319: With this, you're forced to flush every time you write. It's better to just deal with something that implements the `Write` trait directly so that you can call flush when you want.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +228,95 @@ impl Client {
...
+        input.push('\n');
+        self.write_stdin(input.as_bytes())?;
+        let resp = self.read_response()?;
+
+        Ok(resp)
     }
 
-    Ok(items)
-}
+    fn read_response(&mut self) -> Result<PlanResponse, Box<dyn Error>> {
```
richardpringle, line 297: Instead of calling `format!` inline for errors, we should really just use `thiserror`.

x/programs/rust/examples/counter/src/lib.rs
richardpringle: The problem is that the test is in the counter example, but you're not testing the counter example, you're testing the simulator itself. You're coupling two crates together instead of testing at the boundaries.

richardpringle (changes requested):

x/programs/cmd/simulator/cmd/key.go
```diff
@@ -6,72 +6,49 @@ package cmd
...
+	if err != nil {
+		return resp, err
 	}
-}
 
-func (c *keyCreateCmd) Init(args []string) error {
-	c.name = args[0]
-	return nil
-}
+	utils.Outf("{{green}}key create successful: {{/}}%x\n", pkey)
```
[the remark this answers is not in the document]
iFrostizz: Indeed, not sure why it didn't collide but we should probably just log this as debug.

x/programs/cmd/simulator/cmd/key.go
```diff
@@ -6,72 +6,49 @@ package cmd
...
+	db   **state.SimpleMutable
+	name *string
 }
 
-func newKeyCreateCmd(log logging.Logger, db *state.SimpleMutable) *cobra.Command {
-	c := &keyCreateCmd{
-		log: log,
-		db:  db,
-	}
+func (c *keyCreateCmd) New(parser *argparse.Parser, db **state.SimpleMutable) {
```
iFrostizz, line 30: This double pointer `**` is used because the `Simulator.Init` is called after the subcommands have been generated, and the db has to be passed before that the command line arguments have been parsed. Kind of a chicken-egg proble; but this is how I overcome it, if you have a better-looking solution I'm all ears!

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -303,4 +303,66 @@ mod tests {
...
+            endpoint: Endpoint::Execute,
+            method: "program_create".into(),
+            max_units: 1000000,
+            params: vec![Param::String(PROGRAM_PATH.into())],
+            require: None,
+        });
+        plan.add_step(Step {
+            endpoint: Endpoint::Execute,
+            method: "initialize_address".into(),
+            max_units: 1000000,
```
iFrostizz, line 336: While writing this PR, @richardpringle (thanks!) found that I broke assertions. Having this test would have caught it

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +228,95 @@ impl Client {
...
+            .spawn()?;
+
+        let (Some(stdin), Some(stdout)) = (child.stdin, child.stdout) else {
+            panic!("could not attach to the child process stdio");
+        };
+        let lines = BufReader::new(stdout).lines();
+
+        let mut sim = Self { stdin, lines };
+
+        let _ = sim.read_response()?; // intepreter ready
```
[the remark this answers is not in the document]
iFrostizz: Yeah, now that the bug is fixed I removed it!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Got it, thanks a lot for the explanation

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Done, `eyre` is great too because it gives meaningful backtraces when it crashes.

iFrostizz:
As a side note, I'm not actually sure that it should be called an interpreter, it's rather an REPL

x/programs/cmd/simulator/Cargo.toml
```diff
@@ -8,3 +8,4 @@ include = ["**/*.go"]
 serde = { version = "1.0.196", features = ["derive"] }
 serde_json = "1.0.113"
 serde_with = "3.6.1"
+thiserror = "1.0.46"
```
richardpringle, line 11: newline, please!

x/programs/cmd/simulator/cmd/config.go
```diff
@@ -16,14 +16,9 @@ const (
 	ProgramExecute = "execute"
 )
 
-type Plan struct {
-	// The key of the caller used in each step of the plan.
-	CallerKey string `yaml:"caller_key" json:"callerKey"`
-	// Steps to performed during simulation.
-	Steps []Step `json,yaml:"steps"`
-}
```
richardpringle, line 24: 👋

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +239,70 @@ impl Client {
...
-        println!("stderr");
-        for line in String::from_utf8_lossy(&output.stderr).lines() {
-            println!("{line}");
-        }
-        println!("stdout");
-        for line in String::from_utf8_lossy(&output.stdout).lines() {
-            println!("{line}");
-        }
-        return Err(String::from_utf8(output.stdout)?.into());
+    fn run_step(&mut self, caller_key: &String, step: &Step) -> Result<PlanResponse, ClientError> {
```
richardpringle, line 289:
```suggestion
    fn run_step(&mut self, caller_key: &str, step: &Step) -> Result<PlanResponse, ClientError> {
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -48,6 +46,15 @@ pub struct Step {
     pub require: Option<Require>,
 }
 
+#[derive(Debug, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct SimulatorStep<'a> {
+    /// The key of the caller used in each step of the plan.
+    pub caller_key: &'a String,
```
richardpringle, line 53: `String` is a smart-pointer to a heap allocation that effectively looks like this
```rust
struct String {
    ptr: *const u8,
    len: usize,
    capacity: usize,
}
```
(It doesn't actually look like that, but you can think of it as such)
`&str` is effectively a pointer and a length (`size_of::<&str>()` is 1/3 smaller).
`&String` is a pointer to a struct on the stack that points to a heap allocation; you have an extra indirection. So it's always better to use `&str` and `&[T]` instead of `&String` and `&Vec<T>`
```suggestion
    pub caller_key: &'a str,
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +239,70 @@ impl Client {
...
+        let step = SimulatorStep { caller_key, step };
+        let input = serde_json::to_vec(&step)?;
+        self.writer.write_all(&input)?;
+        self.writer.write_all(b"'\n")?;
+        self.writer.flush()?;
+
+        let resp = self
+            .responses
+            .next()
+            .ok_or(ClientError::Custom("EOF".into()))??;
```
richardpringle, line 302: You should make `EOF` a variant of `ClientError` instead of unnecessarily allocating a `String` here.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -193,20 +200,25 @@ pub struct PlanResult {
...
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ClientError {
+    #[error("read")]
+    Read(#[from] std::io::Error),
+    #[error("serde_json error")]
+    Deserialization(#[from] serde_json::Error),
+    #[error("custom")]
+    Custom(String),
```
richardpringle, line 212: This isn't a great practice. You're allocating a string for the error but not using it in the display of the error (only when debugging). It's better not to have "custom" errors but to have well-defined errors (or at least display the custom error).

I would change this to
 ```suggestion
    #[error("EOF")]
    Eof,
```

File: x/programs/cmd/simulator/src/lib.rs [hunk 31]
```diff
@@ -193,20 +200,25 @@ pub struct PlanResult {
     pub response: Option<Vec<i64>>,
 }
 
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ClientError {
+    #[error("read")]
```
richardpringle, line 207:
```suggestion
    #[error("Read error: {0}")]
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -193,20 +200,25 @@ pub struct PlanResult {
...
+    Read(#[from] std::io::Error),
+    #[error("serde_json error")]
```
richardpringle, line 209:
```suggestion
    #[error("Deserialization error: {0}")]
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +239,70 @@ impl Client {
...
-    ///
-    /// Returns an error if the serialization or single-[Step]-[Plan] fails.
-    pub fn execute_step(&self, key: &str, step: Step) -> Result<PlanResponse, Box<dyn Error>> {
-        let plan = &Plan {
-            caller_key: key.into(),
-            steps: vec![step],
-        };
-
-        run_step(self.path, plan)
+    pub fn run_plan(self, plan: Plan) -> Result<Vec<PlanResponse>, ClientError> {
```
richardpringle, line 242: Taking ownership here makes sense, 👍!

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +239,70 @@ impl Client {
...
+            .arg("interpreter")
+            .arg("--cleanup")
+            .arg("--log-level")
+            .arg("error")
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .spawn()?;
+
+        let writer = child.stdin.take().ok_or("failed to open stdin").unwrap();
+        let reader = child.stdout.take().ok_or("failed to open stdout").unwrap();
```
richardpringle, line 253: This is a nit, but it's a good practice to perform the minimal number of mutations necessary. 
```suggestion
        let writer = child.stdin.as_mut().ok_or(ClientError::StdIo)?;
        let reader = child.stdout.as_mut().ok_or(ClientError::StdIo)?;
```
You might be able to come up with a better name for the new error variant though.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +239,70 @@ impl Client {
...
+
+        let responses = BufReader::new(reader)
+            .lines()
+            .map(|line| -> Result<_, ClientError> {
+                let line = line.map_err(ClientError::Read)?;
+                let res = serde_json::from_str::<PlanResponse>(&line)
+                    .map_err(ClientError::Deserialization)?;
+                Ok(res)
+            });
```
richardpringle, line 262: Use inference as much as possible. 
```suggestion
        let responses = BufReader::new(reader)
            .lines()
            .map(|line| serde_json::from_str(&line?).map_err(Into::into));
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -227,105 +239,70 @@ impl Client {
...
+
+        let mut child = SimulatorChild { writer, responses };
+
+        plan.steps
+            .iter()
+            .map(|step| child.run_step(&plan.caller_key, step))
+            .collect::<Result<_, ClientError>>()
```
richardpringle, line 269:
```suggestion
            .collect()
```

richardpringle (changes requested):

x/programs/cmd/simulator/cmd/key.go
dboehm-avalabs: It looks like you could just move s.Init up in the function.

x/programs/cmd/simulator/cmd/config.go
iFrostizz: See ya!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: I see, so `&String` is most of the time an anti-pattern ! Thank you very much!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Done, thanks for the explanation below!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Indeed, that's much better now thanks !

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Done, thanks for the below further explanation !

x/programs/cmd/simulator/src/lib.rs
iFrostizz: I was a bit careless with those messages looks like

x/programs/cmd/simulator/cmd/key.go
dboehm-avalabs: Looks like the db is used in most of the commands, could just add it to the Cmd.Run() signature as a parameter

x/programs/cmd/simulator/src/lib.rs
iFrostizz: This is looking much better!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Thanks!

x/programs/cmd/simulator/cmd/key.go
dboehm-avalabs: Is the parser also required to be run before the simulator inits?  I'm assuming most initialization isn't dependent on the args anyway, so it seems like you could run the init before you create the parsing commands.

x/programs/cmd/simulator/cmd/key.go
```diff
@@ -6,72 +6,49 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"time"
 
-	"github.com/spf13/cobra"
```
dboehm-avalabs, line 10: Any particular reason to switch to arg parse over cobra?

x/programs/cmd/simulator/src/lib.rs
richardpringle: FYI, I wasn't suggesting you inline the function, I was suggesting to move the function to the module instead of the `mod::read_response` instead of `mod::Client::read_response`.

x/programs/cmd/simulator/src/lib.rs
richardpringle: Could also do something like:
```rust
let Child { stdin, stdout, .. } = Command::new(self.path) // ...
```

x/programs/cmd/simulator/cmd/key.go
iFrostizz: Yeah, the `interpreter` new subcommand parses commands iteratively and it requires to have access to a function that does the parsing from a string, which cobra didn't offered (I don't think)

x/programs/cmd/simulator/src/lib.rs
iFrostizz: I like this variant, I'm not feeling very inspired by other error names so `StdIo` will do

x/programs/cmd/simulator/cmd/plan.go
iFrostizz: @dboehm-avalabs Is that fine to use a partially incorrect `Response` type since it won't be used anyway as a "default" ?

x/programs/cmd/simulator/cmd/key.go
iFrostizz: Yes, there is the `logLevel` flag which is ran in the initialization and has to be parsed from the args, I will pass the db to the `Run` command though, that makes much more sense

iFrostizz: Yep passing the `db` to the `Run` func is much better. Thanks !

richardpringle (approved):

PR #889, opened 2024-05-06
[chain] fix result size

wlawt:
Fixes the following:

- Packs the number of Actions in result.Outputs
- Packs the number of Outputs in each Action. This fixes the problem when unmarshaling we know where the outputs should be nested


patrick-ogrady:
Going to consolidate into a single branch to make reviews easier.

Issue #595, opened 2023-10-31
[x/programs] external program calls should return result


hexfusion:
Currently exported `public` Program functions all return a i64 and most errors are panics. This issue is for returning a result and handling the custom error types.

I think it would be easier for the user to reason about something like this.
```rust
enum ProgramError {
     Fatal(String), // <- panic
     OhhNotSooBad(String), // some message that gets handled by result
}
```

possible usage

```rust
/// Initializes the program with a name, symbol, and total supply.
#[public]
pub fn init(program: Program) -> Result<(), ProgramError>{
    // set total supply
    program
        .state()
        .store(StateKey::TotalSupply.to_vec(), &123456789_i64)?;
[....]
```

Sanghren:
As a user, I would definitely like something like this. It feels weird to have that `-> i64`. Intuitively while testing one of the example, I tried to return a `String` in one of the function and it gave me an error that had me puzzled for a while.

 `(Get name of token : PlanResponse { id: 0, result: PlanResult { id: None, msg: None, timestamp: 1698855794, response: None }, error: Some("program execution failed: invalid parameter count for function get_name: 2 expected: 3") }`.

hexfusion:
Yeah I am working on this now. Ideally you could return whatever object you wanted as long as it was properly serializable. Stay tuned.

richardpringle:
@hexfusion, were you thinking that `ProgramError` would be defined by the user or in the SDK?

hexfusion:
> @hexfusion, were you thinking that `ProgramError` would be defined by the user or in the SDK?

I am still working through this but as @Sanghren noted it is supported functionality of Solidity and logically it seems that there is no reason the user can not define errors to a certain extent. Perhaps for simplicity we have concrete types exposed by the SDK that the user can chose and pass a custom `String`.

On the `HyperSDK` side of this we still need the result.Object API (not built) to adhere to some constraints as well.

richardpringle:
> > @hexfusion, were you thinking that `ProgramError` would be defined by the user or in the SDK?
> 
> I am still working through this but as @Sanghren noted it is supported functionality of Solidity and logically it seems that there is no reason the user can not define errors to a certain extent. Perhaps for simplicity we have concrete types exposed by the SDK that the user can choose and pass a custom error `String`.
> 
> On the `HyperSDK` side of this we still need the result.Object API (not built) to adhere to some constraints as well.
> 
> What would you recommend?

[...]

PR #890, opened 2024-05-06
Make delete return the old data

richardpringle:
Part 1 of a bigger fixup of the host-call signature normalization

### DIFF
--- a/x/programs/cmd/simulator/cmd/plan.go
+++ b/x/programs/cmd/simulator/cmd/plan.go
@@ -240,7 +240,15 @@ func runStepFunc(
 
 			return nil
 		}
-		id, _, balance, err := programExecuteFunc(ctx, log, db, params, method, maxUnits)
+		id, response, balance, err := programExecuteFunc(ctx, log, db, params, method, maxUnits)
+		if err != nil {
+			return err
+		}
+		resp.setResponse(response)
+		ok, err := validateAssertion(response[0], require)
+		if !ok {
+			return fmt.Errorf("%w", ErrResultAssertionFailed)
+		}
 		if err != nil {
 			return err
 		}
--- a/x/programs/examples/imports/pstate/pstate.go
+++ b/x/programs/examples/imports/pstate/pstate.go
@@ -217,9 +217,45 @@ func (i *Import) deleteFn(caller *program.Caller, memOffset int32, size int32) (
 	}
 
 	k := storage.ProgramPrefixKey(args.ProgramID[:], args.Key)
-	if err := i.mu.Remove(context.Background(), k); err != nil {
-		i.log.Error("failed to remove from storage", zap.Error(err))
-		return types.ValI32(-1), nil
+	bytes, err = i.mu.GetValue(context.Background(), k)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			val, err := program.WriteBytes(memory, []byte{0})
+			if err != nil {
+				i.log.Error("failed to write to memory",
+					zap.Error(err),
+				)
+				return nil, err
+			}
+
+			return types.ValI32(int32(val)), nil
+		}
+
+		i.log.Error("failed to get value from storage",
+			zap.Error(err),
+		)
+		return nil, err
 	}
-	return types.ValI32(0), nil
+
+	// prepend 1 to val
+	bytes = append([]byte{1}, bytes...)
+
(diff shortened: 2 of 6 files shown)


x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -240,7 +240,15 @@ func runStepFunc(
 
 			return nil
 		}
-		id, _, balance, err := programExecuteFunc(ctx, log, db, params, method, maxUnits)
+		id, response, balance, err := programExecuteFunc(ctx, log, db, params, method, maxUnits)
+		if err != nil {
+			return err
+		}
+		resp.setResponse(response)
+		ok, err := validateAssertion(response[0], require)
```
iFrostizz, line 248: Is there any reason why is `response` an array ? I know that it's an initial design choice, but it doesn't seem used and we need to access some *potentially* out of bounds elements with this `0`, maybe that it will be a change of the fix for https://github.com/ava-labs/hypersdk/issues/595#event-12716725615 though

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -217,9 +217,45 @@ func (i *Import) deleteFn(caller *program.Caller, memOffset int32, size int32) (
 	}
 
 	k := storage.ProgramPrefixKey(args.ProgramID[:], args.Key)
+	bytes, err = i.mu.GetValue(context.Background(), k)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			val, err := program.WriteBytes(memory, []byte{0})
```
iFrostizz, line 223: What is the reasoning behind writing a single `0` byte to memory, is that to get a pointer ?

x/programs/examples/imports/pstate/pstate.go
```diff
@@ -217,9 +217,45 @@ func (i *Import) deleteFn(caller *program.Caller, memOffset int32, size int32) (
...
+		return nil, err
+	}
+
 	if err := i.mu.Remove(context.Background(), k); err != nil {
-		i.log.Error("failed to remove from storage", zap.Error(err))
-		return types.ValI32(-1), nil
+		i.log.Error("failed to delete value from storage",
+			zap.Error(err),
+		)
+		return nil, err
```
iFrostizz, line 257: Since you have set the new pointer type to a `NonNull<_>`, returning a `nil` value would violate the property of this type: `Unlike *mut T, the pointer must always be non-null, even if the pointer is never dereferenced. This is so that enums may use this forbidden value as a discriminant – Option<NonNull<T>> has the same size as *mut T. However the pointer may still dangle if it isn’t dereferenced.` but I just wanted to make sure that returning a non-nil error from these linked functions would indeed cause a panic in the host, it seems to be what is happening in wasmtime https://pkg.go.dev/github.com/bytecodealliance/wasmtime-go#NewFunc can you please confirm my understanding ?

x/programs/cmd/simulator/cmd/plan.go
richardpringle: >Is there any reason why is response an array? 

No idea

>but it doesn't seem used and we need to access some potentially out of bounds elements with this 0

🤷  haha... not an issue yet, let's fix it when it becomes an issue.

>but it doesn't seem used and we need to access some potentially out of bounds elements with this 0

Sure

x/programs/examples/imports/pstate/pstate.go
richardpringle: Good question, I need to add a comment

richardpringle: Added, let me know if that's good enough or if you think I need to add more detail

x/programs/examples/imports/pstate/pstate.go
richardpringle: I believe our linters force us to handle all errors. If they don't they should. There are a number of places in the this code that return the `nil, err` tuple. I'm just following the convention that was already set up

x/programs/examples/imports/pstate/pstate.go
iFrostizz: It all makes sense, thanks !

richardpringle:
> Nice, we could write some tests for the low-level `delete_bytes` function in the future to make sure that it is working as intended.

I have one in the token-test, but I agree, I could add something to `public_functions.rs`

PR #893, opened 2024-05-08
Make `#[public]` ffi functions private and unsafe

richardpringle:
Make it harder to call the `extern` functions written by the `#[public]` macro

### DIFF
--- a/x/programs/rust/sdk_macros/Cargo.toml
+++ b/x/programs/rust/sdk_macros/Cargo.toml
@@ -12,5 +12,6 @@ quote = "1.0.33"
 syn = { version = "2.0.37", features = ["full", "extra-traits"] }
 
 [dev-dependencies]
+borsh = "1.5.0"
 trybuild = "1.0.89"
 wasmlanche-sdk = { path = "../wasmlanche-sdk" }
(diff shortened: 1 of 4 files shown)


x/programs/rust/sdk_macros/Cargo.toml
richardpringle, line 1: Needed for the `ui` test

x/programs/rust/sdk_macros/src/lib.rs
richardpringle, line 1: Swap the user-specified type for the fully qualified type. If the type is actually a user-defined type with the same name, we now get the proper error.

x/programs/rust/sdk_macros/src/lib.rs
```diff
@@ -64,11 +68,11 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
         }
     };
 
-    let param_idents = input_args
-        .iter()
-        .enumerate()
-        .skip(1)
-        .map(|(i, fn_arg)| match fn_arg {
+    let (input, context_type) = (input, context_type);
```
iFrostizz, line 71: nit: this is a noop

iFrostizz (approved):
nice!

x/programs/rust/sdk_macros/src/lib.rs
richardpringle: Not quite, it's a move from mutable to immutable... but it's not obvious, so I'll make a change here

iFrostizz (approved):

PR #894, opened 2024-05-09
[chain] Remove execute success

wlawt:
When an `Action` returns an `err`, it is considered unsuccessful (not fatal).


chain/transaction.go
```diff
@@ -348,10 +348,7 @@ func (t *Transaction) Execute(
...
 		}
 		actionID := codec.CreateLID(uint8(i), t.id)
-		success, actionCUs, outputs, err := action.Execute(ctx, r, ts, timestamp, t.Auth.Actor(), actionID)
-		if err != nil {
-			return handleRevert(err)
-		}
+		success, actionCUs, outputs := action.Execute(ctx, r, ts, timestamp, t.Auth.Actor(), actionID)
 		if len(outputs) == 0 && outputs != nil {
 			// Enforce object standardization (this is a VM bug and we should fail
 			// fast)
```
patrick-ogrady, line 354: We should give the transaction a dedicated place to return an error instead of just appending to the last action?

chain/transaction.go
```diff
@@ -348,10 +348,7 @@ func (t *Transaction) Execute(
 			break
 		}
 		actionID := codec.CreateLID(uint8(i), t.id)
-		success, actionCUs, outputs, err := action.Execute(ctx, r, ts, timestamp, t.Auth.Actor(), actionID)
-		if err != nil {
-			return handleRevert(err)
-		}
+		success, actionCUs, outputs := action.Execute(ctx, r, ts, timestamp, t.Auth.Actor(), actionID)
 		if len(outputs) == 0 && outputs != nil {
```
patrick-ogrady, line 352: We may want the opposite of this check

Issue #885, opened 2024-05-06
[x/programs] Lazily Serialize Params


richardpringle:
If you look at the `params!` definition here:
https://github.com/ava-labs/hypersdk/blob/3fdf8bfdddd50496e45956efb3de1ca4aa0edb61/x/programs/rust/wasmlanche-sdk/src/params.rs#L7-L13

You can see that each param is serialized eagerly, then all the serialized params are put into a `Vec<Param>`, which is then also serialized into `Params(Vec<u8>)`. But with the changes in this PR, we serialized this _again_  ... it's super inefficient. It kind of made sense before, but it definitely doesn't make sense to do this now. 

I'm thinking that `Params(Vec<u8>)` -> `Vec<Box<dyn BorshSerialize>>` , and serialize it in one shot here:
https://github.com/ava-labs/hypersdk/blob/3fdf8bfdddd50496e45956efb3de1ca4aa0edb61/x/programs/rust/wasmlanche-sdk/src/program.rs#L63

_Originally posted by @richardpringle in https://github.com/ava-labs/hypersdk/pull/881#discussion_r1589622876_

iFrostizz:
https://github.com/ava-labs/hypersdk/pull/954

PR #900, opened 2024-05-14
[WIP] Simplify Programs

dboehm-avalabs:
A first pass at simplifying the programs code.  

[WIP] list
- Not all functionality is complete such as actor, callstack, how much fuel is left, etc
- Breaks the simulator


x/programs/v2/config.go
```diff
@@ -0,0 +1,321 @@
...
+	defaultFuelMetering                 = true
+	defaultWasmMultiMemory              = false
+	defaultWasmMemory64                 = false
+	defaultCompilerStrategy             = wasmtime.StrategyCranelift
+	defaultEpochInterruption            = true
+	defaultNaNCanonicalization          = "true"
+	defaultCraneliftOptLevel            = wasmtime.OptLevelSpeed
+	defaultEnableCraneliftDebugVerifier = false
+	defaultEnableDebugInfo              = false
+)
```
richardpringle, line 38: Unrelated to this PR... we need to figure out a way to disable floats... The underlying engine can definitely be configured not to use floats, but not sure if that configuration option is exposed (in Rust or Go)

x/programs/v2/import_state.go
```diff
@@ -0,0 +1,117 @@
...
+				return nil, callInfo.State.Remove(ctx, writeKey)
+			}),
+			"log": FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
+				return log("INFO", input)
+			}),
+		},
+	}
+}
+
+type StateAccessList interface {
```
patrick-ogrady, line 87: These are also managed by the underlying `tstate`. It will revert if you attempt to access something that wasn't previously specified regardless of the output of the action.

Removing this tracking may help further simplify.

dboehm-avalabs: ahh, didn't realize, that will be good to remove then.

patrick-ogrady: This happens here: https://github.com/ava-labs/hypersdk/blob/0d108cd526282a289ac530dfaf84567d790ed160/tstate/tstate_view.go#L164-L167

x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -90,7 +88,7 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
                     fn_arg.span(),
                     "Functions with the `#[public]` attribute can only ignore the first parameter.",
                 )),
-                _ => Ok(Ident::new(&format!("param_{i}"), fn_arg.span())),
+                _ => Ok(pat),
```
richardpringle, line 91: 🤔... we might want to put this back. I think a pattern match might break this:
```rust
fn add((a, b): (usize, usize)) -> usize {
    a + b
}
```
(this is a completely valid function signature)

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -21,19 +21,14 @@ impl Program {
...
-
     /// Returns a State object that can be used to interact with persistent
     /// storage exposed by the host.
     #[must_use]
     pub fn state<K>(&self) -> State<K>
     where
         K: Into<Key> + Hash + PartialEq + Eq + Clone,
     {
-        State::new(Program::new(*self.id()))
+        State::new()
```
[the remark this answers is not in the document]
richardpringle: Okay, looking into things further, I think what we actually want here is a `Context` type that's "augmented". We want to initialize this `State` object on initialization. We're going to use the `Context` type to do a little magic, I don't think we're strictly going to treat it as data.

x/programs/v2/import_log.go
```diff
@@ -0,0 +1,28 @@
+package v2
+
+import (
+	"fmt"
+	"os"
+)
+
+func NewLogModule() *ImportModule {
+	return &ImportModule{name: "log",
```
richardpringle, line 9: 🤔 I think this is overkill, but I could be convinced otherwise. What we have right now is something that will log strings. It's super useful, because you want use the `wasmlanche_sdk::dbg!` macro that works just like the `std::dbg!` macro. It will inject the file-path and line-number as a prefix to whatever you're logging. It's not mean to be used with any environment variable log-level type thing.

Might need to discuss this a little more regardless.

On another node, is it possible to conditionally compile in Go? There should be no logging on a production VM.

x/programs/v2/imports.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+type Imports struct {
+	modules      map[string]*ImportModule
+	totalImports int
+}
+
+type ImportModule struct {
+	name  string
+	funcs map[string]HostFunction
```
richardpringle, line 15: 😱  I see what you're doing here... wild.

x/programs/v2/program.go
```diff
@@ -0,0 +1,112 @@
...
+	c.inst.store.ConsumeFuel(fuel)
+}
+
+type Program struct {
+	module    *wasmtime.Module
+	programID ids.ID
+}
+
+type ProgramInstance struct {
+	*Program
```
richardpringle, line 56: What's going on here... is this a nameless property?

richardpringle (commented):
Initial review... a bunch of those comments are for me

x/programs/rust/sdk-macros/src/lib.rs
dboehm-avalabs: Feel free to back if you fix the struct construction.  I was struggling to get it to generate correctly.

x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -126,18 +124,53 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
+           param_struct.#param_name
         }
     });
 
-    let param_types = std::iter::repeat(quote! { *const u8 }).take(param_names.len());
     let name = &input.sig.ident;
+    let struct_name = Ident::new("Input", name.span());
+
+    let args = input.sig.inputs.iter().skip(1).map(|fn_arg| match fn_arg {
+        FnArg::Receiver(_) => quote! {},
```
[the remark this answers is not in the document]
dboehm-avalabs: it gets caught in the other iteration of the inputs that construct the converted params

x/programs/rust/wasmlanche-sdk/src/logging.rs
```diff
@@ -42,7 +42,7 @@ pub(super) fn log_bytes(bytes: &[u8]) {
     #[link(wasm_import_module = "state")]
```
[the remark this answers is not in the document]
dboehm-avalabs: it doesn't have to be, I added it in both to preserve as much of the old functioning as I could.  I'd like to delete it from state and have a separate logging import.  Ideally I'd also like to have the debug version become a noop when the rust is compiled in debug, assuming that isn't already the case.

x/programs/v2/program.go
dboehm-avalabs: yes/no.  I don't recall why that was on there in the first place, so I'll delete it.  What it is used for is embedding the struct into another, while preserving the ability to treat the enclosing struct as if it were of the same type as the embedded struct.  Its mostly syntactic sugar so you can call the fields and methods on the bigger struct instead of having to do inst.program.method.

x/programs/v2/import_log.go
dboehm-avalabs: `// +build tag`
makes the code only compiles when 'tag' is attached to the build

x/programs/v2/program.go
dboehm-avalabs: https://gobyexample.com/struct-embedding

x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -126,18 +124,53 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
+
+    let struct_deserialization = if has_no_args {
+        quote! {}
+    } else {
+        quote! {
+            let param_struct: #struct_name = unsafe {
+                wasmlanche_sdk::from_host_ptr(param_struct_ptr).expect("error serializing ptr")
+            };
+        }
+    };
```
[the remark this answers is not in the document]
dboehm-avalabs: the deserialization seemed like it was needed, since there would be no param struct pointer in this case.

x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -126,18 +124,53 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
-            unsafe {
-                wasmlanche_sdk::from_host_ptr(#param_name).expect("error serializing ptr")
-            }
+           param_struct.#param_name
         }
     });
 
-    let param_types = std::iter::repeat(quote! { *const u8 }).take(param_names.len());
     let name = &input.sig.ident;
+    let struct_name = Ident::new("Input", name.span());
```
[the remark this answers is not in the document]
iFrostizz: @dboehm-avalabs is this struct the plan to replace serialized parameters https://github.com/ava-labs/hypersdk/issues/885 ?

x/programs/rust/wasmlanche-sdk/src/logging.rs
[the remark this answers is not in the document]
iFrostizz: Do you mean no instruction and runtime penalty when ran in **release** mode ? If so, yes that is the case there is some conditional compilation in the `dbg!` macro

x/programs/rust/sdk-macros/src/lib.rs
iFrostizz: Yes and it is also going to be broken for destructured function parameters:

```rust
struct Hi {
    name: String,
    age: u64
}

fn third(one: u64, Hi { name, .. }: Hi) {}
```

For now, there could be a compile-time error that makes sure that the pattern of the type is unary.

x/programs/rust/sdk-macros/src/lib.rs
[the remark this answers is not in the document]
dboehm-avalabs: Yes, this reduces it to a single deserialization.

x/programs/v2/import_program.go
```diff
@@ -0,0 +1,49 @@
...
+	return &ImportModule{name: "program",
+		funcs: map[string]HostFunction{
+			"call_program": FunctionWithOutput(func(callInfo *CallInfo, input []byte) ([]byte, error) {
+				newInfo := *callInfo
+				parsedInput := &callProgramInput{}
+				if err := borsh.Deserialize(parsedInput, input); err != nil {
+					return nil, err
+				}
+
+				newInfo.ProgramID = ids.ID(parsedInput.ProgramID)
```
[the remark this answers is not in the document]
dboehm-avalabs: This will be fixed, just setup this way to make the deserialization easier for initial test.

x/programs/v2/import_log.go
dboehm-avalabs: I've moved this to single entry, but I'm not going to worry about noop'ing this on the host side.  Just calling into the host is expensive, so I'll leave it is up to the guest code to noop in release.

richardpringle (approved):

x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -126,18 +124,53 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
+    let struct_declaration = if has_no_args {
+        quote! {}
+    } else {
+        quote! {
+            #[derive(borsh::BorshDeserialize)]
+            struct #struct_name {
+                #(#args),*
+            }
+        }
+    };
```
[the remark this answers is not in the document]
richardpringle: >I think that this whole part is a bit messy, this is way more readable:

There's a ton of duplication in what you're suggesting which makes edits a lot harder. In any case, we generically handle 0-length now. 

I further want to combine `Context` and `Args` such that the calls only take a single pointer and do a single deserialization.

richardpringle (approved):

x/programs/v2/import_program.go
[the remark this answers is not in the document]
richardpringle: https://github.com/ava-labs/hypersdk/issues/913

x/programs/rust/wasmlanche-sdk/src/program.rs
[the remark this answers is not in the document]
richardpringle: https://github.com/ava-labs/hypersdk/issues/914

PR #908, opened 2024-05-15
A bunch of fixes to the macro

richardpringle:
Enables the usage of null-pointer as a bit of a hack to generically handle no-arg (aside from context) functions. Will work on removing this by serializing args and context into a single struct once this work in merged in main.

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/memory.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/memory.rs
@@ -26,7 +26,13 @@ pub fn from_host_ptr<V>(ptr: *const u8) -> Result<V, StateError>
 where
     V: BorshDeserialize,
 {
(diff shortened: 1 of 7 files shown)


x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -26,7 +26,13 @@ pub fn from_host_ptr<V>(ptr: *const u8) -> Result<V, StateError>
 where
     V: BorshDeserialize,
 {
-    match into_bytes(ptr) {
+    let bytes = if ptr.is_null() {
+        Some(vec![])
+    } else {
+        into_bytes(ptr)
+    };
```
richardpringle, line 33: I don't love this, but we can fix it later

Issue #899, opened 2024-05-14
[x/programs] Write a recursive program test


iFrostizz:
... And make sure that it gets halted with the units metering .ref https://github.com/ava-labs/hypersdk/pull/896#discussion_r1599682399

Issue #865, opened 2024-04-26
[x/programs] Assert that a plan step failed


iFrostizz:
It would be useful to be able to pass a signed integer to the `require` field in the steps specifying it, this way we could assert that a step failed as expected.
Additionally, the error message when an assertion has failed could be improved with the expected and actual value much like the `assert_eq!` macro in Rust. The error message is currently invariably  `result assertion failed`.

najeal:
What the signed integer is supposed to represent? Hypersdk does not give us information on the Error triggered rust program side, we get a `ErrTrapUnreachableCodeReached` error golang host side.

iFrostizz:
@najeal The -1 number is currently returned on an unsuccessful call, see https://github.com/ava-labs/hypersdk/blob/f002156f55e422c79631904582f77ed0811d7419/x/programs/examples/imports/program/program.go#L85
We are moving the assertions out of the simulator though, so this issue will be closed as part of this rewrite.

iFrostizz:
So we now can do that because the error is caught in the `error` field of the response, but I would like to close this issue once we have faillible responses from external calls for a really complete feature.

PR #910, opened 2024-05-15
[x/programs] program recursion test

iFrostizz:
By curiosity, I wanted to write a recursion test just to make sure that unwinding works as expected and that the execution will run out of units.
This is still a WIP because although this property is verified, there is no way to assert that an execution failed. Additionally, there is an error log message from the simulator saying that the execution ran out of units which is annoying to have on the terminal.
I think that it is blocked by https://github.com/ava-labs/hypersdk/issues/865, https://github.com/ava-labs/hypersdk/issues/595, and the fact that the logging is printed on stderr because of the communication pattern.

Closes https://github.com/ava-labs/hypersdk/issues/899

edit: there is also this deserialization error which was not caught. I'm wondering if we could tune the static analysis to detect those blunders.

### DIFF
--- a/x/programs/cmd/simulator/cmd/config.go
+++ b/x/programs/cmd/simulator/cmd/config.go
@@ -164,7 +164,10 @@ func validateAssertion(bytes []byte, require *Require) (bool, error) {
 	}
 
 	actual := int64(0)
-	borsh.Deserialize(&actual, bytes)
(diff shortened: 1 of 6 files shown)


iFrostizz:
This should probably be written in Go or in Rust when we have better unit tests.

Issue #638, opened 2023-12-06
[x/programs] Pass arguments as one serialized tuple


samliok:
As @richardpringle  and I discussed offline, it may be more effective to serialize all arguments in one tuple instead of serializing each argument individually.

Currently, when calling `#[public]` functions from the `host` or with `call_program` we do the following
-  we serialize every individual parameter using borsh and then prepending it with the length of its bytes. 
- deserialize each individual parameter in `rust` by grabbing the byte length and then calling `borsh`

By serializing all arguments into a singular tuple(which `borsh` supports), we can reduce serialization and a good amount of memory allocations. Take the `transfer` function for example
`pub  fn  transfer(program:  Program, sender:  Address, recipient:  Address, amount:  i64) ->  bool`

Currently we must serialize and then deserialize every argument we want to pass in. Instead on the `host` side we could create a tuple like `(program, sender, recipient, amount)` and then perform only one serialization as described above. On the receiving end, we adjust the `#[public]` proc macro to do the following
-  deserialization the entire tuple at once
- split the tuple up and sequentially pass the arguments to the desired function

PR #915, opened 2024-05-16
Combine context and user defined parameters

richardpringle:
While there still has to be special treatment for the `Context` parameter, this is a more generic way of generating code for user-defined functions


x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -54,9 +48,7 @@ pub fn into_bytes(ptr: *const u8) -> Option<Vec<u8>> {
 /// Panics if the pointer exceeds the maximum size of an isize or that the allocated memory is null.
 #[no_mangle]
 pub extern "C" fn alloc(len: usize) -> *mut u8 {
-    if len == 0 {
-        return std::ptr::null_mut();
-    }
+    assert!(len > 0, "cannot allocate 0 sized data");
```
iFrostizz, line 51: Much nicer!

iFrostizz (approved):

x/programs/rust/wasmlanche-sdk/src/memory.rs
richardpringle: 😂  I just put the file back to what it was (I think you wrote that code)

richardpringle:
closes #638

PR #924, opened 2024-05-22
[vryx] Remove Fee Refund

patrick-ogrady:
To prepare for Vryx (https://hackmd.io/@patrickogrady/rys8mdl5p), all fee refunds need to be removed. Whatever units a transaction requests will be charged.

### DIFF
--- a/chain/result.go
+++ b/chain/result.go
@@ -15,6 +15,8 @@ type Result struct {
 
 	Outputs [][][]byte
 
+	// Computing [Units] requires access to [StateManager], so it is returned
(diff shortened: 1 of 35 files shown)


chain/result.go
```diff
@@ -15,8 +15,8 @@ type Result struct {
 
 	Outputs [][][]byte
 
-	Consumed fees.Dimensions
```
patrick-ogrady, line 18: This can be computed trivially from the transaction...probably want to remove?

PR #928, opened 2024-05-23
Fix typo in sdk-macros dependency definition

richardpringle:
I still don't understand how this was working before my change. It somehow was falling back to the workspace, I guess?


iFrostizz (approved):
Probably, it's doing the same if we write `borsh = { }` for instance

richardpringle:
> Probably, it's doing the same if we write `borsh = { }` for instance

@iFrostizz, the question is, why would that work!? Does it just grab the latest version?

PR #929, opened 2024-05-23
Improve config table test in simulator

richardpringle:
This testing definitely is not covering all cases (outside, inside, and on the boundary). However, this was more of an exercise for me as we shouldn't be doing assertions inside the library anyway.


iFrostizz (approved):
👌

PR #934, opened 2024-05-24
Add Test for state imports

dboehm-avalabs:
Adding some tests to make sure the host functions for interacting with state are functional.  Cleaned up some structs that weren't needed.


x/programs/runtime/import_state_test.go
```diff
@@ -0,0 +1,50 @@
...
+	require.Equal(valueBytes, result)
+
+	result, err = runtime.CallProgram(ctx, &CallInfo{ProgramID: programID, State: state, FunctionName: "remove", Params: nil, Fuel: 10000000})
+	require.NoError(err)
+	require.Equal(append([]byte{1}, valueBytes...), result)
+
+	result, err = runtime.CallProgram(ctx, &CallInfo{ProgramID: programID, State: state, FunctionName: "read", Params: nil, Fuel: 10000000})
+	require.NoError(err)
+	require.Equal(valueBytes, result)
+}
```
richardpringle, line 50: These should all be separate tests. Can use `t.Run("Name", func(*testing.T) { ... })`

richardpringle (approved):

Issue #931, opened 2024-05-23
[x/programs] Don't do assertions in the simulator


iFrostizz:
Assertions are currently written in the simulator with the `Require` struct. It is pretty limiting because all types have to be written beforehand as well as operators. We need to remove that feature from the simulator and do the assertions out of it.

PR #935, opened 2024-05-24
[x/programs] Remove assertions from the simulator

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/931


x/programs/cmd/simulator/src/lib.rs
```diff
@@ -17,7 +17,7 @@ mod id;
 pub use id::Id;
 
 /// The endpoint to call for a [Step].
-#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+#[derive(Debug, Serialize, PartialEq, Clone)]
```
iFrostizz, line 20: I have removed `Deserialize` for types which only made sense implementing `Serialize`, should I back this off ?

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -233,46 +206,47 @@ impl Client {
...
+        let reader = stdout.expect("stdout has been captured");
 
         let responses = BufReader::new(reader)
             .lines()
             .map(|line| serde_json::from_str(&line?).map_err(Into::into));
 
-        let mut child = SimulatorChild { writer, responses };
+        let child = SimulatorChild { writer, responses };
+
+        Client { child }
```
richardpringle, line 228: This shouldn't be nested like this, the `Client` doesn't provide any functionality, it only delegates to the `SimulatorChild`. In other words, this type is more complex than it needs to be. What I would do is have a struct called a `ClientBuilder` with a `ClientBuilder::new` that takes in a path and a `ClientBuilder::try_build` method that returns a `Client<impl Write, impl Iterator<Item = Result<PlanResponse, ClientError>>>` (which is a rename on the `SimulatorChild`). 

That would fix the type issues you're having above.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -233,46 +206,47 @@ impl Client {
...
+            .map(|step| self.child.run_step(plan.caller_key, step))
             .collect()
     }
-}
 
-impl Default for Client {
-    fn default() -> Self {
-        Self::new()
+    pub fn run_step(&mut self, caller_key: &str, step: Step) -> Result<PlanResponse, ClientError> {
+        self.child.run_step(caller_key, &step)
```
richardpringle, line 249: (context for this comment in the previous comment)

You can move all this stuff to the `Client` (renamed `SimulatorChild`)

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -127,60 +123,44 @@ mod tests {
```
richardpringle, line 123: Does it still make sense for the plan responses possibly error? Shouldn't the whole plan either pass or fail since we can be as granular as we want? 

It seems like making sure all the plan responses aren't errors should just be part of the `run_plan` method. 

If you create a new issue (if you use `create issue from comment`, please add extra context), link the new issue here, then resolve this comment. No need to make changes in this PR.

richardpringle: Actually, I think the whole `run_plan` method should probably removed since we only run one step at a time synchronously anyway.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -284,5 +237,66 @@ mod tests {
...
+                )
+                .unwrap()
+                .result
+                .response,
+        );
+        assert_eq!(value, 10);
+    }
+
+    fn bytes_to_u64(bytes: Vec<u8>) -> u64 {
+        u64::from_le_bytes(bytes.try_into().unwrap())
```
richardpringle, line 300: Is this a borsh-serialized `u64`? If so, you shouldn't be deserializing it like this, you should be using `borsh::from_slice`

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -284,5 +237,66 @@ mod tests {
                 .filter_map(|resp| resp.error.as_ref())
                 .next()
         );
+
+        let value = bytes_to_u64(
+            simulator
+                .run_step(
```
richardpringle, line 243: Can we not make `run_step`'s return value generic? Something is super off about the way this works. I think it's that we're dealing with Go responses that don't have `enums`, but you can just write a custom deserializer such that you don't need `serde_as` (dependency dropped is a win), and ignore whatever is in the `response` field if the `error` field is populated.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -198,7 +170,8 @@ pub struct PlanResult {
     /// The timestamp of the function call response.
     pub timestamp: u64,
     /// The result of the function call.
-    pub response: Option<String>,
+    #[serde_as(as = "Base64")]
+    pub response: Vec<u8>,
```
richardpringle, line 174: Yeah, after looking at the examples, I think these types are super weird now. `PlanResult` should be `StepResult<T>` (rename and make it generic) where the response is `Result<T, String>`.

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -154,27 +154,24 @@ pub fn get_balance(context: Context<StateKeys>, recipient: Address) -> i64 {
...
-
     use crate::INITIAL_SUPPLY;
+    use simulator::{Endpoint, Key, Param, Plan, Step};
 
     const PROGRAM_PATH: &str = env!("PROGRAM_PATH");
 
     #[test]
     fn create_program() {
-        let simulator = simulator::Client::new();
+        let mut simulator = simulator::Client::new_stdin();
```
richardpringle, line 164: FYI, this comment isn't relevant, given my other comments, but in general, you shouldn't leak an abstraction in a function name like this. You're creating the client here; you shouldn't need to change the name if you change the internals (to use grpc instead of stdin for instance).

richardpringle (changes requested):
It's an excellent start, but there are some goofy artifacts with the types being designed for a `simulator` that was meant to be used in a completely different way. Further, I think the client types should be simplified and we should definitely try to avoid panicking on the Rust side of the simulator code.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -233,46 +206,47 @@ impl Client {
...
+        let Child { stdin, stdout, .. } = Command::new(path)
             .arg("interpreter")
             .arg("--cleanup")
             .arg("--log-level")
             .arg("error")
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
-            .spawn()?;
+            .spawn()
+            .expect("failed to run the simulator in interpreter mode");
```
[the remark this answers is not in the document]
iFrostizz: I was referring to an old comment of yours, saying to have the instantiation of the `Client` struct panicking on error: https://github.com/ava-labs/hypersdk/pull/884#discussion_r1591514932
I will make it faillible as per your review comments.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Good idea!

x/programs/rust/examples/token/src/lib.rs
iFrostizz: I wanted to avoid the implementor to have to specify generics upfront, since we should only have a handful of transport layers, we probably don't need to be to allow that much customization. For instance, it probably doesn't make sense to have the writer being stdin and the responses being sent over IPC.

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: True, as per your above comment I will just make the `PlanResponse` accept a generic!

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -233,46 +206,47 @@ impl Client {
...
 
-        let writer = stdin.as_mut().ok_or(ClientError::StdIo)?;
-        let reader = stdout.as_mut().ok_or(ClientError::StdIo)?;
+        let writer = stdin.expect("stdin has been captured");
+        let reader = stdout.expect("stdout has been captured");
```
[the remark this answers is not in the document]
iFrostizz: Added a new `try_build` method that returns the ~`Simulator`~ `Client` as a faillible function.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -214,13 +187,13 @@ pub enum ClientError {
...
-    /// Path to the simulator binary
-    path: &'static str,
+pub struct Client<W, R> {
+    child: SimulatorChild<W, R>,
 }
 
-impl Client {
-    pub fn new() -> Self {
+impl Client<ChildStdin, Box<dyn Iterator<Item = Result<PlanResponse, ClientError>>>> {
+    pub fn new_stdin() -> Client<ChildStdin, impl Iterator<Item = Result<PlanResponse, ClientError>>>
```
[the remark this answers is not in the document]
iFrostizz: Now that we've moved to a `try_build` method, this might not be relevant anymore (since it's called from within the `ClientBuilder`

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Thanks a lot for this suggestion, I did all of your suggested stuff. The `SimulatorChild` has been flattened inside the `Client` structure and it's now much easier.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Thanks! I got a bit of difficulties regarding typing for `PlanResponse` because they are encoded as base64 from Go and we then need to deserialize it but not in the case of a plan, for legacy reasons. This is to keep the data at the state before deserialization.
Does it still makes sense to do the plan removal in a followup PR ?

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: Not using `serde_as` for the deserializing of data from Go won't help dropping the dependency because it's also used in the `Param` enum. it looks more compact that way but would you prefer writing a custom deserializer method to do that by hand? I also made the `run_step` return data generic.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: thanks

x/programs/rust/examples/token/src/lib.rs
iFrostizz: This is the only communication layer available for now, we will abstract more once we write another one, see https://github.com/ava-labs/hypersdk/issues/939

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: I will not create the issue then, since we will remove this idea of plan in a followup PR

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -198,7 +169,72 @@ pub struct PlanResult {
...
+{
+    /// The ID created from the program execution.
+    pub id: Option<String>,
+    /// An optional message.
+    pub msg: Option<String>,
+    /// The timestamp of the function call response.
+    pub timestamp: u64,
+    /// The result of the function call.
+    pub response: T,
+}
```
iFrostizz, line 189: I was actually not trivial to use a serde `tag` here to define an enum with a typed and an untyped variant. The reason is that we will have to deal with the variants in the `TryInto` function.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -207,19 +243,26 @@ pub enum ClientError {
     Read(#[from] std::io::Error),
     #[error("Deserialization error: {0}")]
     Deserialization(#[from] serde_json::Error),
+    #[error("Borsh deserialization error")]
+    BorshDeserialization,
```
iFrostizz, line 247: `borsh::io::Error` is a subset of `std::io::Error` so I couldn't add the `#[from]` variant or it would overlap with the `Read` error.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -254,39 +294,34 @@ impl Client {
...
-            .iter()
-            .map(|step| child.run_step(&plan.caller_key, step))
-            .collect()
-    }
-}
-
-impl Default for Client {
-    fn default() -> Self {
-        Self::new()
+        Ok(Client { writer, responses })
```
richardpringle, line 304: This is a lot nicer, 🥇

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -296,9 +331,20 @@ where
...
+        &mut self,
+        caller_key: &str,
+        step: &Step,
+    ) -> Result<PlanResponseTyped<T>, ClientError>
+    where
+        T: BorshDeserialize,
+    {
+        self._run_step(caller_key, step)?
+            .try_into()
+            .map_err(|_| ClientError::BorshDeserialization)
```
richardpringle, line 347: We can do it in a follow-up PR, but we should really just let the end-user do the batching since it's all synchronous anyway. 

I think it's a bad practice to do a bunch of stuff and then do your assertions on the result. I suspect that more people (at least the ones following idiomatic testing practices) will be asserting on the results at each step.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -59,201 +59,207 @@ pub fn get_value_external(_: Context, target: Program, max_units: i64, of: Addre
...
-            }),
-        });
-
-        // run plan
         let plan_responses = simulator.run_plan(plan).unwrap();
 
-        // ensure no errors
         assert!(
-            plan_responses.iter().all(|resp| resp.error.is_none()),
+            plan_responses.iter().all(|resp| resp.base.error.is_none()),
```
richardpringle, line 138: Yeah, I don't like this. Using a custom serializer will be better. I think I said it in my other comment, but you can do this as a follow-up.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -296,9 +331,20 @@ where
...
-        let resp = self.responses.next().ok_or(ClientError::Eof)??;
+        self.responses.next().ok_or(ClientError::Eof)?
+    }
 
-        Ok(resp)
+    pub fn run_step<T>(
+        &mut self,
+        caller_key: &str,
+        step: &Step,
+    ) -> Result<PlanResponseTyped<T>, ClientError>
```
richardpringle, line 341: I think I would add a new error type here called a `StepError`:
```rust
#[derive(this_error::Error)]
enum StepError {
    #[error("{0}"]
    Client(#[from] ClientError),
    #[error("{0}"]
    Deserialization(#[from]borsh::io::Error),
}
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -198,7 +169,72 @@ pub struct PlanResult {
...
+impl<T> TryInto<PlanResponseTyped<T>> for PlanResponse
+where
+    T: BorshDeserialize,
+{
+    type Error = borsh::io::Error;
+
+    fn try_into(self) -> Result<PlanResponseTyped<T>, Self::Error> {
+        let PlanResponse {
+            base: BaseResponse { id: resp_id, error },
+            result:
```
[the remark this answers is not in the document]
richardpringle: Commented below, can do this in a follow up

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -198,7 +169,72 @@ pub struct PlanResult {
...
+where
+    T: BorshDeserialize,
+{
+    #[serde(flatten)]
+    pub base: BaseResponse,
+    /// The result of the plan.
+    pub result: PlanResultTyped<T>,
+}
+
+impl<T> TryInto<PlanResponseTyped<T>> for PlanResponse
```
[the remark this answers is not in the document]
richardpringle: (please do this before I approve)

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -236,16 +279,13 @@ impl Client {
...
-            mut stdin,
-            mut stdout,
-            ..
-        } = Command::new(self.path)
+    pub fn try_build(
+        self,
+    ) -> Result<
+        Client<impl Write, impl Iterator<Item = Result<PlanResponse, ClientError>>>,
+        ClientError,
+    > {
```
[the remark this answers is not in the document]
richardpringle: (please do this before I approve)

x/programs/cmd/simulator/src/lib.rs
richardpringle: (please do this before I approve)

richardpringle (changes requested):
Added comments on all the changes I want you to make before we merge this. Can you also go through all the follow-up comments, create issues, and assign yourself to them, please? Then, just post the link to the issue and resolve the comment.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Thanks, I added a new `Serialization` error as well since the step serialization is faillible in the new `StepError` enum.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Absolutely, the plan is not adapted to our use-case anymore, this will be done when closing https://github.com/ava-labs/hypersdk/issues/981

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Thanks for the nice suggestions!

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: Because of the last changes, I just opted for a simpler alternative of creating a new type over `Result`
```rust
type StepResult = Result<PlanResponse, StepError>;
```

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: Answered, will be part of the removal of plans

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: I don't like it either, and I don't want this to get merged so I will work on the custom serializer

iFrostizz: Me neither, but this is a temporary solution.
Once we drop `Plan`, all of the `PlanResponse` will be typed and the intermediary types will be removed, directly inlined in the `PlanResponse` struct, with a custom `Serializer`.

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: Thanks. I have implemented the `TryFrom` instead of the `TryInto`.

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: @richardpringle are you okay with that choice ? I'm wondering if you made the iteration response generic to allow the transport layer to be generic, but I think that we should do that later

dboehm-avalabs (dismissed):

x/programs/cmd/simulator/build.rs
```diff
@@ -36,7 +36,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let go_build_output = Command::new("go")
         .arg("build")
         .arg("-tags")
-        .arg("debug")
```
dboehm-avalabs, line 39: don't we want this to be debug so that the host will log?

iFrostizz: Good catch! This was a debugging leftover, I probably thought to have edited the `simulator` command invokation.

richardpringle (approved):

PR #936, opened 2024-05-24
State::get should return an option

richardpringle:
It should always be valid to check to see if anything has been stored at a specific key without erroring.

Scope increased, I cleaned up the whole of state accesses and inlined a bunch of code to simplify things.

### DIFF
--- a/x/programs/runtime/import_state.go
+++ b/x/programs/runtime/import_state.go
@@ -57,14 +57,14 @@ func NewStateAccessModule() *ImportModule {
 				}
 				return val, nil
 			})},
-			"put": {FuelCost: writeCost, Function: FunctionWithOutput(func(callInfo *CallInfo, input []byte) ([]byte, error) {
+			"put": {FuelCost: writeCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
 				parsedInput := &keyValueInput{}
 				if err := borsh.Deserialize(parsedInput, input); err != nil {
-					return nil, err
+					return err
 				}
 				ctx, cancel := context.WithCancel(context.Background())
 				defer cancel()
-				return nil, callInfo.State.Insert(ctx, prependAccountToKey(callInfo.Account, parsedInput.Key), parsedInput.Value)
+				return callInfo.State.Insert(ctx, prependAccountToKey(callInfo.Account, parsedInput.Key), parsedInput.Value)
 			})},
 			"delete": {FuelCost: deleteCost, Function: FunctionWithOutput(func(callInfo *CallInfo, input []byte) ([]byte, error) {
 				parsedInput := &keyInput{}
--- a/x/programs/rust/examples/counter/src/lib.rs
+++ b/x/programs/rust/examples/counter/src/lib.rs
@@ -6,27 +6,6 @@ pub enum StateKeys {
     Counter(Address),
 }
 
-/// Initializes the program address a count of 0.
-#[public]
-pub fn initialize_address(context: Context<StateKeys>, address: Address) -> bool {
-    let Context { program, .. } = context;
-
-    if program
-        .state()
-        .get::<i64>(StateKeys::Counter(address))
-        .is_ok()
-    {
-        panic!("counter already initialized for address")
-    }
-
-    program
-        .state()
-        .store(StateKeys::Counter(address), &0_i64)
-        .expect("failed to store counter");
-
-    true
-}
-
 /// Increments the count at the address by the amount.
 #[public]
 pub fn inc(context: Context<StateKeys>, to: Address, amount: i64) -> bool {
@@ -60,7 +39,8 @@ fn get_value_internal(context: &Context<StateKeys>, of: Address) -> i64 {
         .program
         .state()
         .get(StateKeys::Counter(of))
-        .expect("failed to get counter")
+        .expect("state corrupt")
+        .unwrap_or_default()
 }
 
 /// Gets the count at the address for an external program.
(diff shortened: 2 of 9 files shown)


x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -6,27 +6,6 @@ pub enum StateKeys {
...
-        panic!("counter already initialized for address")
-    }
-
-    program
-        .state()
-        .store(StateKeys::Counter(address), &0_i64)
-        .expect("failed to store counter");
-
-    true
-}
```
richardpringle, line 28: Fallible initialize never really made sense. If we wanted a counter with explicit initialization, we should have been able to set the number, otherwise it's wasteful to store the 0 case.

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -104,13 +107,15 @@ pub fn transfer(
     let sender_balance = program
         .state()
         .get::<i64>(StateKeys::Balance(sender))
-        .expect("failed to update balance");
+        .expect("failed to deserialize balance")
```
richardpringle, line 110:
```suggestion
        .expect("state corrupt")
```

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: Totally agree, otherwise we could have had a function called `initialized` that returns

```rust
    if program
        .state()
        .get::<i64>(StateKeys::Counter(address))
        .is_ok()
  ```

that would be called after the first call to `inc*` in tests, but that is effectively the same sanity check as the next calls to `get` so I'm fine just removing that function

File: x/programs/rust/examples/counter/src/lib.rs [hunk 3]
```diff
@@ -151,13 +131,6 @@ mod tests {
...
-        plan.add_step(Step {
-            endpoint: Endpoint::Execute,
-            method: "initialize_address".into(),
-            max_units: 1000000,
-            params: vec![counter_id.into(), bob_key.clone()],
-            require: None,
-        });
 
         plan.add_step(Step {
             endpoint: Endpoint::Execute,
```
iFrostizz, line 136: Could you please also remove the call to `initialize_address` in the `external_call` test ? It's not ran for now but may save us 1.5 seconds once we fix it 😄

dboehm-avalabs (approved):

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -95,22 +98,24 @@ pub fn transfer(
...
     // ensure the sender has adequate balance
     let sender_balance = program
         .state()
-        .get::<i64>(StateKeys::Balance(sender))
-        .expect("failed to update balance");
+        .get::<u64>(StateKeys::Balance(sender))
+        .expect("state corrupt")
+        .unwrap_or_default();
 
-    assert!(amount >= 0 && sender_balance >= amount, "invalid input");
```
iFrostizz, line 109: 👍

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -207,20 +210,16 @@ mod host {
...
         }
 
         let result = unsafe { ffi(bytes.as_ptr(), bytes.len()) };
 
-        if result.is_null() {
-            Err(Error::Read)
-        } else {
-            Ok(result)
-        }
+        NonNull::new(result)
```
iFrostizz, line 222: 👏

iFrostizz (dismissed):

x/programs/runtime/import_state.go
```diff
@@ -57,14 +57,14 @@ func NewStateAccessModule() *ImportModule {
 				}
 				return val, nil
 			})},
-			"put": {FuelCost: writeCost, Function: FunctionWithOutput(func(callInfo *CallInfo, input []byte) ([]byte, error) {
+			"put": {FuelCost: writeCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
```
richardpringle, line 60: @dboehm-avalabs, this makes sense, right? 

As I ask the question, I'm realizing that we should probably be consistent between `delete` and `put`. Either both should return the old value like the Rust `HashMap` API, or both should return nothing.

As much as I like the Rust API, I'm in favour of the latter since it could be expensive to copy all the bytes into the wasm module

x/programs/rust/examples/counter/src/lib.rs
richardpringle: I'll be the one re-enabling that test so I deal with it then

x/programs/runtime/import_state.go
dboehm-avalabs: I'd probably lean return nothing.

x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -156,10 +156,9 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
             }
 
             #[no_mangle]
-            unsafe extern "C" fn #name(args: *const u8) {
+            unsafe extern "C" fn #name(args: wasmlanche_sdk::HostPtr) {
```
dboehm-avalabs, line 159: I like this

x/programs/runtime/import_state.go
richardpringle: I'll change that in a follow up

dboehm-avalabs (approved):

PR #942, opened 2024-05-26
[WIP] ERC20 Token Program

dboehm-avalabs:


x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// The total supply of the token. Key prefix 0x0.
+    TotalSupply,
+    /// The name of the token. Key prefix 0x1.
+    Name,
+    /// The symbol of the token. Key prefix 0x2.
+    Symbol,
+    /// The balance of the token by address. Key prefix 0x3 + address.
+    Balance(Address),
+
+    Allowance([Address; 2]),
```
richardpringle, line 18:
```suggestion
    Allowance(Address, Address),
```
with the comment above about `PrefixedBytes`. 

We should be careful here... Using an enum for state-keys might not actually make sense; the size of `StateKeys::Name` is necessarily the same size as `StateKeys::Allowance` (the type has a size, not each variant). I'm not suggesting we change this now, but we should probably benchmark if it makes a difference.

x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+        .get(StateKey::Owner)
+        .expect("failed to get total supply")
+}
+
+pub fn owner_check(program: &Program<StateKey>, actor: Address) {
+    assert!(
+        get_owner(program) == actor,
+        "caller is required to be owner"
+    )
+}
```
richardpringle, line 35: You could actually write an extension trait for the `Context` to call `context.is_owner()` directly.

x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+
+/// Returns the total supply of the token.
+#[public]
+pub fn total_supply(context: Context<StateKey>) -> i64 {
+    let Context { program, .. } = context;
+    program
+        .state()
+        .get(StateKey::TotalSupply)
+        .expect("failed to get total supply")
```
richardpringle, line 73:
```suggestion
        .unwrap_or_defaul()
```

x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    // set token symbol
+    program
+        .state()
+        .store(StateKey::Symbol, &symbol)
+        .expect("failed to store symbol");
+}
+
+/// Returns the total supply of the token.
+#[public]
+pub fn total_supply(context: Context<StateKey>) -> i64 {
```
richardpringle, line 68: no negatives
```suggestion
pub fn total_supply(context: Context<StateKey>) -> u64 {
```

x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,174 @@
+use wasmlanche_sdk::Context;
+use wasmlanche_sdk::{public, state_keys, types::Address, Program};
+
+const INITIAL_SUPPLY: i64 = 123456789;
```
richardpringle, line 4:
```suggestion
const INITIAL_SUPPLY: u64 = 123456789;
```

x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    let Context { program, .. } = context;
+    program
+        .state()
+        .get(StateKey::TotalSupply)
+        .expect("failed to get total supply")
+}
+
+/// Transfers balance from the token owner to the recipient.
+#[public]
+pub fn mint(context: Context<StateKey>, recipient: Address, amount: i64) -> bool {
```
richardpringle, line 78:
```suggestion
pub fn mint(context: Context<StateKey>, recipient: Address, amount: u64) -> bool {
```

x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+    let Context { program, actor } = context;
+    owner_check(&program, actor);
+    let balance = program
+        .state()
+        .get::<i64>(StateKey::Balance(recipient))
```
richardpringle, line 83:
```suggestion
        .get::<u64>(StateKey::Balance(recipient))
```

I would probably use a type-alias at the top of the file `type BalanceType = u64;` and use it for everything. Not 100% sure name, maybe `QuantityType`, 🤷 .

x/programs/test/programs/simple_token/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+wasmlanche-sdk = { workspace = true }
+borsh = { workspace = true }
+
+[build-dependencies]
+wasmlanche-sdk = { workspace = true, features = ["build"] }
```
richardpringle, line 14: Any reason not to extend the current token example? If you think it makes more sense to keep it as is, we can rename this to `token` and call the other one `simple-token` (as it is simpler).

x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+#[public]
+pub fn transfer_from(
+    context: Context<StateKey>,
+    sender: Address,
+    recipient: Address,
+    amount: i64,
+) -> bool {
+    false
```
richardpringle, line 173: 😂

x/programs/test/programs/simple_token/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+        .expect("failed to store allowance");
+    true
+}
+
+/// Transfers balance from the sender to the the recipient.
+#[public]
+pub fn transfer(context: Context<StateKey>, recipient: Address, amount: u64) -> bool {
+    let Context { program, actor } = context;
+    let sender = actor;
+    assert_ne!(sender, recipient, "sender and recipient must be different");
```
dboehm-avalabs, line 141: should this have an error or return false?

x/programs/test/programs/token/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+#[public]
+pub fn transfer_from(
+    context: Context<StateKey>,
+    sender: Address,
+    recipient: Address,
+    amount: u64,
+) -> bool {
+    let Context {program, actor } = context;
```
dboehm-avalabs, line 190: This is currently broken due to reuse of the context across the various functions.  I could create "inner" functions that do the actual allowance and transfer logic that instead take in the deconstructed pieces of the context, but that seems clunky.  Is there a cleaner way to handle this?

x/programs/test/programs/token/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    assert_ne!(actor, recipient, "sender and recipient must be different");
+
+    // ensure the sender has adequate balance
+    let sender_balance = program
+        .state()
+        .get::<u64>(StateKey::Balance(actor))
+        .expect("failed to get sender balance")
+        .unwrap_or_default();
+
+    assert!(sender_balance >= amount, "invalid input");
```
dboehm-avalabs, line 161: Is this the mechanism we want to be using for checking conditions?

iFrostizz: I think that we need to figure out https://github.com/ava-labs/hypersdk/issues/595

EricForgy:
Is this PR doing what I think it is doing? If you are adding ERC20 to HyperSDK, that is such a sad thing for me to see. We have an opportunity to do something much better and ERC20 is a huge step backwards.

dboehm-avalabs:
> Is this PR doing what I think it is doing? If you are adding ERC20 to HyperSDK, that is such a sad thing for me to see. We have an opportunity to do something much better and ERC20 is a huge step backwards.

It isn't the intention to actually have such a thing built into the HyperSDK.  I'm simply running through some common use cases and seeing what issues I run into.  I probably won't even merge this.

Issue #943, opened 2024-05-27
[x/programs] add a macros for generating public methods


iFrostizz:
We find ourselves writing a lot of boilerplate code to write public getters of state keys.
This is the case of many methods of the upcoming ERC20 token example. https://github.com/ava-labs/hypersdk/blob/b646347d25954e93bf19368862a47ae985a3a100/x/programs/test/programs/simple_token/src/lib.rs
It would be great to have a macro attribute to be placed on the relevant `StateKeys` variants to generate this code for us.

richardpringle:
Yeah, I think this is a good idea for now. I think there might be better ways to do this at the VM level, but until we've plugged programs into an actual VM, this is probably a good stop-gap

richardpringle:
Sorry, let me rephrase: I think *using* the `state_keys` macro to create getters is a good idea. There are multiple ways we can accomplish this and I'm not sure that annotating every variant is the best way to accomplish this.

PR #944, opened 2024-05-27
[x/programs] Add an attribute macro to create public getters

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/943

### DIFF
--- a/x/programs/rust/examples/token/src/lib.rs
+++ b/x/programs/rust/examples/token/src/lib.rs
@@ -42,16 +42,6 @@ pub fn init(context: Context<StateKeys>) {
         .expect("failed to store symbol");
 }
 
-/// Returns the total supply of the token.
-#[public]
-pub fn get_total_supply(context: Context<StateKeys>) -> i64 {
-    let Context { program, .. } = context;
-    program
-        .state()
-        .get(StateKeys::TotalSupply)
-        .expect("failed to get total supply")
-}
-
 /// Transfers balance from the token owner to the recipient.
 #[public]
 pub fn mint_to(context: Context<StateKeys>, recipient: Address, amount: i64) -> bool {
@@ -214,7 +204,7 @@ mod tests {
 
         plan.add_step(Step {
             endpoint: Endpoint::ReadOnly,
-            method: "get_total_supply".into(),
+            method: "total_supply".into(),
             max_units: 0,
             params: vec![program_id.into()],
             require: Some(Require {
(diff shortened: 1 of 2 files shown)


x/programs/rust/examples/token/src/lib.rs
```diff
@@ -8,6 +8,7 @@ const INITIAL_SUPPLY: i64 = 123456789;
 #[state_keys]
 pub enum StateKeys {
     /// The total supply of the token. Key prefix 0x0.
+    #[state_keys(getter = i64)]
```
richardpringle, line 11: It's not immediately clear what `getter = i64` means. I think something like `value_type` would make a lot more sense. With that said, I'm not sure that's the approach that we want to take as we would have to annotate every single field. 

There's probably a better option using a declarative macro to define everything. Something like:
```rust
state_schema! {
    MyUnitVariant => i64,
    MyTupleVariant(Address) => String,
}
```
That can construct a `StateKeys` enum with all the state-keys and some other functionality for compile-time guarantees for getting and setting values. 

Before we implement this, I would like to see a proposal for what this looks like for the end-user. Once we agree on something, we can try and tackle an implementation.

richardpringle (changes requested):

x/programs/rust/examples/token/src/lib.rs
iFrostizz: I don't think that there would be any much better way to avoid annotating every field with the expected return type, for instance this is not an information that we could get in a reliable way.
Renaming the attribute from `getter` to `value_type` would be great, this is a change I will do.
About the `state_schema` declarative macro, I think that my proposition is a bit more idiomatic. This is for instance what the `serde` or `serde_as` proc macros are doing.

iFrostizz: Also if we want to be able to also set values, we will need to be able to specify keys that should be a getter or setter, or both. In order to stay secure, we should also be able to be able to apply access control, and bound checks on the variable. I think that the setter is much harder to use and could be a pretty bad footgun,

iFrostizz: Also, this is something that *may* be unrelated, but the token example features a `Balance(Address)` state key which does not have the default implementation. If the value is not present for an address, then it won't panic but return the default value. Should we allow the users of this macro to specify if they want to have a default value ? It seems that we are starting to get into macros that get in the way of the users because they are way too complex for what they may do.

PR #968, opened 2024-05-28
Smart Host Pointer

richardpringle:
So... I'm getting a litte fancy here allowing what is effectively a pointer to deref as `&[u8]`. I don't think this accomplishes much from an efficiency standpoint except that we don't carry around the extra `capacity` `usize` for a `Vec<u8>` when all we really need is a borrow. 

If you are okay with this change, I should probably add some documentation before we merge. The code is safe because it will panic if the `ALLOCATIONS` global detects mutliple borrows trying to convert to an owned value. 

I wanted to do this because the majority of the time we were turning a pointer returned by the host into a `Vec<u8>` we were just borrowing it right after. I'm not sure if the compiler was smart enough to drop capacity tracking, but now it's explicit. We now don't have to worry about any extra code that generated creating a `Vec<u8>`, borrowing, then dropping.

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/memory.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/memory.rs
@@ -4,34 +4,70 @@
 //! the program. These methods are unsafe as should be used
 //! with caution.
 
-use std::{alloc::Layout, cell::RefCell, collections::HashMap};
+use std::{alloc::Layout, cell::RefCell, collections::HashMap, mem::ManuallyDrop, ops::Deref};
 
 thread_local! {
     /// Map of pointer to the length of its content on the heap
-    static GLOBAL_STORE: RefCell<HashMap<*const u8, usize>> = RefCell::new(HashMap::new());
+    static ALLOCATIONS: RefCell<HashMap<*const u8, usize>> = RefCell::new(HashMap::new());
 }
 
-/// Reconstructs the vec that was created using the module's `alloc` function.
-/// # Panics
-/// If you are trying to dereference data that has already been dereferenced
-/// or the pointer is invalid (wasn't created with the module's `alloc` function).
-#[must_use]
-pub fn deref_bytes(ptr: *const u8) -> Vec<u8> {
-    GLOBAL_STORE
-        .with_borrow_mut(|s| s.remove(&ptr))
-        // SAFETY:
-        // the space is reserved by the `alloc` function
-        .map(|len| unsafe { std::vec::Vec::from_raw_parts(ptr.cast_mut(), len, len) })
-        .expect("value missing or taken")
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct HostPtr(*const u8);
+
+impl Deref for HostPtr {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        let len = ALLOCATIONS
+            .with_borrow(|allocations| allocations.get(&self.0).copied())
+            .expect("attempted to deref invalid host pointer");
+
+        unsafe { std::slice::from_raw_parts(self.0, len) }
+    }
 }
 
-/* memory functions ------------------------------------------- */
+impl Drop for HostPtr {
+    fn drop(&mut self) {
+        if self.is_null() {
+            return;
+        }
+
+        let len = ALLOCATIONS
+            .with_borrow_mut(|allocations| allocations.remove(&self.0))
+            .expect("attempted to drop invalid host pointer");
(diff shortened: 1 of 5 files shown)


dboehm-avalabs (approved):

x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -4,34 +4,70 @@
...
+}
+
+impl From<HostPtr> for Vec<u8> {
+    fn from(host_ptr: HostPtr) -> Self {
+        // drop will dealloc the bytes
+        let host_ptr = ManuallyDrop::new(host_ptr);
+
+        let len = ALLOCATIONS
+            .with_borrow_mut(|allocations| allocations.remove(&host_ptr.0))
+            .expect("attempted to drop invalid host pointer");
```
richardpringle, line 52:
```suggestion
            .expect("attempted to convert invalid host pointer");
```

dboehm-avalabs (approved):

x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -42,9 +78,9 @@ pub extern "C" fn alloc(len: usize) -> *mut u8 {
         std::alloc::handle_alloc_error(layout);
     }
 
-    GLOBAL_STORE.with_borrow_mut(|s| s.insert(ptr, len));
+    ALLOCATIONS.with_borrow_mut(|s| s.insert(ptr, len));
 
-    ptr
+    HostPtr(ptr.cast())
```
iFrostizz, line 83: we could be slightly more explicit with a `cast_const()` here
```suggestion
    HostPtr(ptr.cast_const())
```

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -107,17 +107,13 @@ impl<'a, K: Key> State<'a, K> {
...
-            let bytes = {
-                let ptr = unsafe { get_bytes(args_bytes.as_ptr(), args_bytes.len()) };
+            let ptr = unsafe { get_bytes(args_bytes.as_ptr(), args_bytes.len()) };
 
-                if ptr.is_null() {
-                    return Ok(None);
-                }
+            if ptr.is_null() {
+                return Ok(None);
+            }
```
iFrostizz, line 114: This is a leftover of https://github.com/ava-labs/hypersdk/pull/936, but it's probably not okay to return `Ok(None)` on any `err` from the host call to `get` (because we make the assumption that when this happens, the returned pointer is null). The reason is that `ErrNotFound` is not the only error that can be returned along the null pointer, there could also be `ErrClosed` and we should probably propagate that error rather than making the assumption that the value is not in the state.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -147,9 +143,8 @@ impl<'a, K: Key> State<'a, K> {
 
         let args_bytes = borsh::to_vec(&args).map_err(|_| StateError::Serialization)?;
 
-        let ptr = unsafe { delete(args_bytes.as_ptr(), args_bytes.len()) };
+        let bytes = unsafe { delete(args_bytes.as_ptr(), args_bytes.len()) };
```
iFrostizz, line 146: nit: any strong opinion on calling this `bytes` ?

iFrostizz (changes requested):

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle: Posting here for posterity.

As discussed offline, we can't handle any transient errors from the host, as that would be a source of indeterminism. We use the null-pointer to signify no value and cast to `None`. Previously, we were serializing the `None` value as `[0]` and the `Some(<value>)` as `[1, <value-bytes>]`, but it's more efficient to skip serialization here. 

@iFrostizz, can you open an issue if we're returning `nil` for any case other than the value not being in the database?

x/programs/rust/wasmlanche-sdk/src/memory.rs
richardpringle: Good catch

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle: I guess I could call it `bytes_ref` since that's how you're supposed to treat it, but that kind of goes against the convention.

dboehm-avalabs (approved):

PR #970, opened 2024-05-28
[x\programs] Add host function to return remaining fuel

dboehm-avalabs:


iFrostizz (changes requested):

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -74,6 +74,25 @@ impl<K> Program<K> {
...
+            fn get_remaining_fuel() -> *const u8;
+        }
+
+        let bytes = {
+            let ptr = unsafe { get_remaining_fuel() };
+
+            deref_bytes(ptr)
+        };
+
+        borsh::from_slice::<u64>(&bytes).map_err(|_| StateError::Deserialization)
```
richardpringle, line 94: Should we just use the native return instead of serializing deserializing? This method should be infallible. 

In any case, if we do want to do the whole serialization thing, we can just panic if deserialization fails.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -72,6 +72,21 @@ impl<K> Program<K> {
...
+        #[link(wasm_import_module = "program")]
+        extern "C" {
+            #[link_name = "remaining_fuel"]
+            fn get_remaining_fuel() -> HostPtr;
+        }
+
+        let bytes = unsafe { get_remaining_fuel() };
+
+        borsh::from_slice::<u64>(&bytes).expect("failed to deserialize the remaining fuel")
+    }
```
richardpringle, line 89: 👍

richardpringle (approved):

richardpringle (approved):

iFrostizz (approved):

Issue #868, opened 2024-04-26
[x/programs] Change the state's `put` and `delete` methods to handle batches on the host


richardpringle:
Where we can, we should avoid going in and out of the wasm-context. There should always be the option to batch calls. 

Further, I think a single method for updates and deletes would be beneficial. Ideally, we could use something like 
```rust
#[derive(BorshSerialize)]
struct CacheEntry<K> {
    key: K, 
    value: CacheValue,
}

#[derive(BorshSerialize)]
enum CacheValue {
    Get(Vec<u8>),
    Update(Vec<u8>),
    Delete,
}
```
I also think we would further benefit from 
```rust
[derive(BorshSerialize)]
enum Update<K> {
    Put(K, Box<dyn BorshSerialize>),
    Delete(K),
}
```
where we could defer the actual serialization. It might be trickier to downcast the `Box<dyn BorshSerialize>` to a `Box<T>` when fetching from the cache though...

PR #976, opened 2024-05-29
[GH-868] Add batch support to state write

dboehm-avalabs:
GH-868


x/programs/runtime/import_state.go
```diff
@@ -62,6 +64,20 @@ func NewStateAccessModule() *ImportModule {
 				defer cancel()
 				return callInfo.State.Insert(ctx, prependAccountToKey(callInfo.Account, parsedInput.Key), parsedInput.Value)
 			})},
+			"put_many": {FuelCost: writeManyCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
```
iFrostizz, line 67: We should think about how to support dynamically metered host function calls

iFrostizz (approved):
LGTM 🚀

x/programs/runtime/import_state.go
dboehm-avalabs: you can consume additional fuel within the host function via the callinfo.

Issue #927, opened 2024-05-22
[x/programs] Serialize params on the Rust side


iFrostizz:
https://github.com/ava-labs/hypersdk/pull/918#discussion_r1610684389

PR #979, opened 2024-05-29
[x/programs] Serialize params on the Rust side

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/927

### DIFF
--- a/x/programs/cmd/simulator/Cargo.toml
+++ b/x/programs/cmd/simulator/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 include = ["**/*.go"]
 
 [dependencies]
+base64 = "0.22.1"
 serde = { version = "1.0.202", features = ["derive"] }
 serde_json = "1.0.117"
 serde_with = { version = "3.8.1", features = ["base64"] }
--- a/x/programs/cmd/simulator/Cargo.toml
+++ b/x/programs/cmd/simulator/Cargo.toml
@@ -8,6 +8,5 @@ include = ["**/*.go"]
 base64 = "0.22.1"
 serde = { version = "1.0.202", features = ["derive"] }
 serde_json = "1.0.117"
-serde_with = { version = "3.8.1", features = ["base64"] }
 thiserror = { workspace = true }
 borsh = { workspace = true }
--- a/x/programs/cmd/simulator/cmd/plan.go
+++ b/x/programs/cmd/simulator/cmd/plan.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -39,10 +40,10 @@ type runCmd struct {
 	reader *bufio.Reader
 
 	// tracks program IDs created during this simulation
-	programIDStrMap map[string]ids.ID
+	programIDStrMap map[int]ids.ID
 }
 
-func (c *runCmd) New(parser *argparse.Parser, programIDStrMap map[string]ids.ID, lastStep *int, reader *bufio.Reader) {
+func (c *runCmd) New(parser *argparse.Parser, programIDStrMap map[int]ids.ID, lastStep *int, reader *bufio.Reader) {
 	c.programIDStrMap = programIDStrMap
 	c.cmd = parser.NewCommand("run", "Run a HyperSDK program simulation plan")
 	c.file = c.cmd.String("", "file", &argparse.Options{
@@ -179,7 +180,7 @@ func (c *runCmd) RunStep(ctx context.Context, db *state.SimpleMutable) (*Respons
 		if err != nil {
 			return nil, err
 		}
-		c.programIDStrMap[fmt.Sprintf("step_%d", index)] = id
+		c.programIDStrMap[index] = id
 	}
 
 	lastStep := index + 1
@@ -276,8 +277,13 @@ func (c *runCmd) createCallParams(ctx context.Context, db state.Immutable, param
 		switch param.Type {
 		case String, ID:
 			stepIDStr := string(param.Value)
-			if strings.HasPrefix(stepIDStr, "step_") {
-				programID, ok := c.programIDStrMap[stepIDStr]
+			id_string, found := strings.CutPrefix(stepIDStr, "step_")
+			if found {
+				id, err := strconv.ParseInt(id_string, 10, 32)
+				if err != nil {
+					return nil, err
+				}
+				programID, ok := c.programIDStrMap[int(id)]
 				if !ok {
 					return nil, fmt.Errorf("failed to map to id: %s", stepIDStr)
 				}
--- a/x/programs/cmd/simulator/cmd/plan.go
+++ b/x/programs/cmd/simulator/cmd/plan.go
@@ -40,10 +40,10 @@ type runCmd struct {
 	reader *bufio.Reader
 
 	// tracks program IDs created during this simulation
-	programIDStrMap map[string]string
+	programIDStrMap map[string]ids.ID
 }
 
-func (c *runCmd) New(parser *argparse.Parser, programIDStrMap map[string]string, lastStep *int, reader *bufio.Reader) {
+func (c *runCmd) New(parser *argparse.Parser, programIDStrMap map[string]ids.ID, lastStep *int, reader *bufio.Reader) {
 	c.programIDStrMap = programIDStrMap
 	c.cmd = parser.NewCommand("run", "Run a HyperSDK program simulation plan")
 	c.file = c.cmd.String("", "file", &argparse.Options{
@@ -176,7 +176,11 @@ func (c *runCmd) RunStep(ctx context.Context, db *state.SimpleMutable) (*Respons
 	// map all transactions to their step_N identifier
 	txID, found := resp.getTxID()
 	if found {
-		c.programIDStrMap[fmt.Sprintf("step_%d", index)] = txID
+		id, err := ids.FromString(txID)
+		if err != nil {
+			return nil, err
+		}
+		c.programIDStrMap[fmt.Sprintf("step_%d", index)] = id
 	}
 
 	lastStep := index + 1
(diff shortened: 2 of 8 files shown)


x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -346,26 +346,9 @@ func (c *runCmd) createCallParams(ctx context.Context, db state.Immutable, param
...
-				if v < 0 {
-					return nil, fmt.Errorf("%w: %s", errors.New("negative value"), param.Type)
-				}
-				cp = append(cp, Parameter{Value: uint64(v), Type: param.Type})
-			case string:
-				number, err := strconv.ParseUint(v, 10, 64)
-				if err != nil {
-					return nil, fmt.Errorf("%w: %s", ErrFailedParamTypeCast, param.Type)
-				}
-				cp = append(cp, Parameter{Value: number, Type: param.Type})
```
iFrostizz, line 365: we were not using this

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -326,10 +359,6 @@ mod tests {
         let output_json = serde_json::to_value(&param).unwrap();
 
         assert_eq!(output_json, expected_json);
-
-        let output_param: Param = serde_json::from_value(expected_json).unwrap();
-
-        assert_eq!(output_param, expected_param);
```
iFrostizz, line 332: We never needed to deserialize a `Param`, so I removed the derive attribute as well as the test

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -3,7 +3,8 @@
 //! Alternatively the `Plan` can be written in JSON and passed to the
 //! Simulator binary directly.
 
-use serde::{Deserialize, Serialize};
+use base64::{engine::general_purpose::STANDARD as b64, Engine};
+use serde::{ser::SerializeStruct, Deserialize, Serialize};
 use serde_with::{serde_as, DisplayFromStr};
```
iFrostizz, line 8: The `serde_with` dependency will be dropped once we merge https://github.com/ava-labs/hypersdk/pull/935

dboehm-avalabs (dismissed):

x/programs/cmd/simulator/src/lib.rs
iFrostizz: edit: probably after manually implementing `Serialize` on the steps once we get rid of plans.

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -202,7 +201,11 @@ func (c *runCmd) RunStep(ctx context.Context, db *state.SimpleMutable) (*Respons
 	// map all transactions to their step_N identifier
 	txID, found := resp.getTxID()
 	if found {
-		c.programIDStrMap[fmt.Sprintf("step_%d", index)] = txID
+		id, err := ids.FromString(txID)
+		if err != nil {
+			return nil, err
+		}
+		c.programIDStrMap[fmt.Sprintf("step_%d", index)] = id
```
richardpringle, line 208: Why not just use the index instead of the string, the `step_` is redundant here. You could probably also use a sparse `[]id.ID` and fill in a bunch `nil` for all the other outputs. None of this needs to be performant.

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -248,7 +251,11 @@ func runStepFunc(
 
 			return nil
 		}
-		id, response, balance, err := programExecuteFunc(ctx, log, db, params[0].Value.(ids.ID), params[1:], method, maxUnits)
+		id, err := ids.ToID(params[0].Value)
+		if err != nil {
+			return err
+		}
+		id, response, balance, err := programExecuteFunc(ctx, log, db, id, params[1:], method, maxUnits)
```
richardpringle, line 258: remind me again why it's `param[1:]`?

x/programs/cmd/simulator/cmd/plan.go
richardpringle: I'm sure there's more unused code lurking too

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -95,17 +96,48 @@ pub enum Key {
...
-    #[serde(untagged)]
     Key(Key),
 }
 
+impl Serialize for Param {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        let mut val = serializer.serialize_struct("param", 2)?;
```
richardpringle, line 112: Haha this is definitely a big *nit*, but I would do this:
```suggestion
        let mut serializer = serializer.serialize_struct("param", 2)?;
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -3,7 +3,8 @@
 //! Alternatively the `Plan` can be written in JSON and passed to the
 //! Simulator binary directly.
 
-use serde::{Deserialize, Serialize};
+use base64::{engine::general_purpose::STANDARD as b64, Engine};
```
richardpringle, line 6: Is there any real benefit to base64 encoding everything?

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -95,17 +96,48 @@ pub enum Key {
...
+                    Key::Ed25519(key) => (key, "ed25519"),
+                    Key::Secp256r1(key) => (key, "secp256r1"),
+                };
+                val.serialize_field("type", algo)?;
+                val.serialize_field("value", &b64.encode(key))?;
+            }
+        }
+        val.end()
+    }
+}
```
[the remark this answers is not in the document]
richardpringle: I don't quite understand why everything needs to be base64 encoded. If I had to guess, it's because it's easier to decode everything on the Go side, but you should just be able to *switch* on the `type` tag and only base64 decode where necessary. But IMO, it's an extra step that's unnecessary (and adds a little complexity). 

Also, I think it would just be beneficial to use an intermediate type here:
```rust
#[derive(Serialize)]
#[serde(rename_all = "lowercase", tag = "type", content = "value")]
enum StringParam {
    U64(String),
    String(String),
    Id(String),
    #[serde(untagged)]
    Key(Key),
}
```
If you want, you can define the type inside the `Serialize` implementation, do a `From` conversion, then just serialize the intermediate type. This should all get optimized away by the compiler, especially when the type is only used in one place. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=755c35cea6729ce0da44df255dceeaed

richardpringle (commented):
Leaving this as a comment review since I'll need to re-approve anyway when the merge conflicts are resolved

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: It's Go making choices for us when dealing with the `[]byte` type, see https://pkg.go.dev/encoding/json#Marshal
> Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.
I found online that people were forking the package to strip out the base64 part but I don't think that we should do that.

x/programs/cmd/simulator/cmd/plan.go
iFrostizz: Replaced the type of the key of the map to `int`. Using a list would also have worked but would have required a bit more of complexity.

x/programs/cmd/simulator/cmd/plan.go
iFrostizz: That's because within all params, the id of the program that is called is passed as the first of them.
The first parameter always has a special treatment and it is defined in this function that you tagged

x/programs/cmd/simulator/src/lib.rs
iFrostizz: 😨 
(replaced)

x/programs/cmd/simulator/src/lib.rs
iFrostizz: https://github.com/ava-labs/hypersdk/issues/994

x/programs/cmd/simulator/Cargo.toml
```diff
@@ -5,8 +5,8 @@ edition = "2021"
 include = ["**/*.go"]
 
 [dependencies]
+base64 = "0.22.1"
 serde = { version = "1.0.202", features = ["derive"] }
 serde_json = "1.0.117"
-serde_with = { version = "3.8.1", features = ["base64"] }
```
iFrostizz, line 10: I promised!

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -85,23 +85,52 @@ impl Step {
...
+    {
+        match self {
+            Param::U64(num) => {
+                Serialize::serialize(&StringParam::U64(b64.encode(num.to_le_bytes())), serializer)
+            }
+            Param::String(text) => {
+                Serialize::serialize(&StringParam::String(b64.encode(text)), serializer)
+            }
+            Param::Id(id) => {
+                let id = serde_json::to_vec(&id).map_err(serde::ser::Error::custom)?;
```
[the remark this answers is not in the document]
richardpringle: Why not just use `usize::to_string`. It's not fallible

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -169,10 +197,27 @@ pub struct PlanResult {
     /// The timestamp of the function call response.
     pub timestamp: u64,
     /// The result of the function call.
-    #[serde_as(as = "Base64")]
+    #[serde(deserialize_with = "b64_decode")]
     pub response: Vec<u8>,
 }
 
+fn b64_encode<S>(text: &String, serializer: S) -> Result<S::Ok, S::Error>
```
richardpringle, line 204:
```suggestion
fn b64_encode<S>(text: &str, serializer: S) -> Result<S::Ok, S::Error>
```

richardpringle: And while you're at it... I really would prefer:
```suggestion
fn base64_encode<S>(text: &str, serializer: S) -> Result<S::Ok, S::Error>
```
But this is a nit

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -169,10 +197,27 @@ pub struct PlanResult {
...
+    S: Serializer,
+{
+    serializer.serialize_str(&b64.encode(text))
+}
+
+fn b64_decode<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    <std::string::String as Deserialize>::deserialize(deserializer).and_then(|string| {
```
richardpringle, line 215:
```suggestion
    <String as Deserialize>::deserialize(deserializer).and_then(|string| {
```

richardpringle (changes requested):

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: Because it still has the `step_` prefix, but it probably makes sense to drop it here too!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Fixed!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: This is very nice! thanks

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: I have implemented `From<&usize>` instead and applied your suggestions. That's better!

x/programs/cmd/simulator/src/id.rs
```diff
@@ -9,6 +9,12 @@ impl From<usize> for Id {
     }
 }
 
+impl<'a> From<&'a Id> for &'a usize {
+    fn from(val: &'a Id) -> Self {
+        &val.0
+    }
+}
```
richardpringle, line 16: 👍

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -85,23 +85,52 @@ impl Step {
...
+                Serialize::serialize(&StringParam::String(b64.encode(text)), serializer)
+            }
+            Param::Id(id) => {
+                let num: &usize = id.into();
+                let id = format!("step_{}", num);
+                Serialize::serialize(&StringParam::Id(b64.encode(id)), serializer)
+            }
+            Param::Key(key) => Serialize::serialize(key, serializer),
+        }
+    }
```
richardpringle, line 131: This is a little more complex than what I had suggested. Also, you really don't need the `step_`. In json, the `"type"` field will be `"id"`, right? that means that `step_` is redundant.

richardpringle (approved):
All the base-64 stuff is a little convoluted. This is a good step into cleaning things up, but I think there's still more to do.

Issue #969, opened 2024-05-28
[x/programs] Handle mutli-value tuple variants in `state-keys`. 


richardpringle:
Need to get rid of `PrefixedBytes<'a>` and do the serialization directly in the macro expansion. 

Should probably use bytemuck directly since the types are all `Copy` anyway.

PR #980, opened 2024-05-29
Fix multi-value state-keys

richardpringle:
Fix multi-value state-keys by `BorshSerialize`-ing directly inside the macro


dboehm-avalabs (approved):

richardpringle:
closes #969

Issue #977, opened 2024-05-29
[testing] Switch e2e tests from gomega to testify


marun:
The use of the gomega assertion library for e2e tests and testify for unit tests increases the cost of maintaining the hypersdk tests since developers have to be familiar with both libraries. Switching the e2e tests to testify avoids that problem. 

Note that avalanchego and subnet-evm have already been updated to use testify as their only assertion library.

patrick-ogrady:
As mentioned offline, we should attempt to do this once we get further into the "Developer Ready" milestone (as the changes that will occur there may significantly change the form of the tests).

PR #984, opened 2024-05-30
[testing] Migrate from gomega to testify 

marun:
See: #977

### DIFF
--- a/examples/morpheusvm/tests/integration/integration_test.go
+++ b/examples/morpheusvm/tests/integration/integration_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/fatih/color"
 	"github.com/onsi/gomega"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 
 	"github.com/ava-labs/hypersdk/chain"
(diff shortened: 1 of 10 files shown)


examples/morpheusvm/tests/integration/integration_test.go
patrick-ogrady, line 1: You can remove `gomega.RegisterFailHandler(ginkgo.Fail)` from all of these tests 👍

patrick-ogrady:
Can we remove this flag from the script as well to resolve this warning:
```
You're using deprecated Ginkgo functionality:
=============================================
  --ginkgo.slow-spec-threshold is deprecated --slow-spec-threshold has been deprecated and will be removed in a future version of Ginkgo.  This feature has proved to be more noisy than useful.  You can use --poll-progress-after, instead, to get more actionable feedback about potentially slow specs and understand where they might be getting stuck.
```

patrick-ogrady:
> > Can we remove this flag from the script as well to resolve this warning:
> > ```
> > You're using deprecated Ginkgo functionality:
> > =============================================
> >   --ginkgo.slow-spec-threshold is deprecated --slow-spec-threshold has been deprecated and will be removed in a future version of Ginkgo.  This feature has proved to be more noisy than useful.  You can use --poll-progress-after, instead, to get more actionable feedback about potentially slow specs and understand where they might be getting stuck.
> > ```
> 
> I don't see anywhere that we're using this flag. I haven't been able to reproduce locally, but I did notice that the cli version installed by the scripts wasn't consistent with the go.mod version. Maybe this fixes your problem? I've added a commit to sync that. Might be worth parsing the version out of go.mod at some point.

this fixed the issue for me 👍

patrick-ogrady (approved):

Issue #963, opened 2024-05-28
Use `Varint` in the Codec


patrick-ogrady:
This includes:
* Replacing the "Int" used to encode `Bytes`/`String`
* Exposing a `PackVarInt`/`PackUvarInt` for numbers: https://github.com/ava-labs/hypersdk/blob/49527653edb1953550dce3fb1f15eafe394c441a/codec/packer.go#L106-L128

This should use `encoding/binary`: https://pkg.go.dev/encoding/binary#PutUvarint

aaronbuchwald:
Closing in favor of #1906

PR #989, opened 2024-05-30
Add `Varint` methods in the codec

iFrostizz:
Using a `VarInt` (or a `UvarInt`) in the codec will sometimes lead to a smaller impact on storage, considering that the stored number is small enough. Otherwise there may be a penalty of a maximum of 2 bytes in the case of 64 bits numbers.
- Adding packing utils to do that in the codec
- Write tests

Also closes https://github.com/ava-labs/hypersdk/issues/963

### DIFF
--- a/codec/packer.go
+++ b/codec/packer.go
@@ -141,46 +141,29 @@ func (p *Packer) UnpackInt(required bool) int {
 }
 
 func (p *Packer) PackUvarInt(v uint64) {
-	buf := make([]byte, binary.MaxVarintLen64)
-	binary.PutUvarint(buf, v)
-	buf = ZeroOut(buf)
-	p.PackBytes(buf)
+	p.p.Bytes = binary.AppendUvarint(p.p.Bytes, v)
 }
 
 func (p *Packer) UnpackUvarInt(required bool) uint64 {
-	bytes := p.p.UnpackBytes()
-	val, n := binary.Uvarint(bytes)
-	if required && n <= 0 {
+	val, bytesRead := binary.Uvarint(p.Bytes())
+	if required && bytesRead <= 0 {
 		p.addErr(fmt.Errorf("%w: UvarInt field is not populated", ErrFieldNotPopulated))
 	}
+	p.p.Bytes = p.p.Bytes[bytesRead:]
 	return val
 }
 
 func (p *Packer) PackVarInt(v int64) {
-	buf := make([]byte, binary.MaxVarintLen64)
-	binary.PutVarint(buf, v)
-	buf = ZeroOut(buf)
-	p.PackBytes(buf)
+	p.p.Bytes = binary.AppendVarint(p.p.Bytes, v)
 }
 
 func (p *Packer) UnpackVarInt(required bool) int64 {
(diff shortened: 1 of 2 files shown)


codec/packer.go
```diff
@@ -139,6 +140,49 @@ func (p *Packer) UnpackInt(required bool) int {
 	return int(v)
 }
 
+func (p *Packer) PackUvarInt(v uint64) {
+	buf := make([]byte, binary.MaxVarintLen64)
+	binary.PutUvarint(buf, v)
+	buf = ZeroOut(buf)
+	p.PackBytes(buf)
```
patrick-ogrady, line 147: PackBytes will wrap this buf in an integer length. That will be much larger in total than just using the uint64.

patrick-ogrady: There is good logic for how to properly do this inside of `merkledb`: https://github.com/ava-labs/avalanchego/blob/8d18b618be290b0e0fedbbedc76398afba68e6ec/x/merkledb/codec.go#L255-L272

patrick-ogrady: You may need to access the packer raw bytes to do this properly JFYI

iFrostizz: Updated and wrote a fuzz test to verify that encoding a number and decoding it back yields the same number, and that the bytes representation is never exceeding 10 bytes for any random 64 bit. I don't think that we need to add this check https://github.com/ava-labs/avalanchego/blob/8d18b618be290b0e0fedbbedc76398afba68e6ec/x/merkledb/codec.go#L261-L268 since it really looks like that empty bytes should be ignored here but I might be wrong https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/encoding/binary/varint.go;l=78-82

codec/packer.go
```diff
@@ -139,6 +140,32 @@ func (p *Packer) UnpackInt(required bool) int {
 	return int(v)
 }
 
+func (p *Packer) PackUvarInt(v uint64) {
+	p.p.Bytes = binary.AppendUvarint(p.p.Bytes, v)
```
patrick-ogrady, line 144: This does not properly update `p.Offset`: https://github.com/ava-labs/avalanchego/blob/ddf66eaed1659c2caa8531bbda83b360ca85900e/utils/wrappers/packing.go#L50

I think you should make a PR to AvalancheGo that adds support for packing `Uvarint` and `Varint`. We can then use that downstream here (doing it outside of the upstream library seems wrong given we are modifying its internal data structures).

codec/packer.go
```diff
@@ -139,6 +140,32 @@ func (p *Packer) UnpackInt(required bool) int {
 	return int(v)
 }
```
patrick-ogrady, line 142: We will also need some utility functions to determine the size of encoding an int/uint to properly determine bandwidth usage of transaction.

iFrostizz: In this PR ?

codec/packer.go
iFrostizz: I have opened https://github.com/ava-labs/avalanchego/pull/3081 in order to do that downstream first, after that we will be able to call these methods directly !

PR #990, opened 2024-05-30
Prevent too many variants

richardpringle:
Thought I would add this

### DIFF
--- a/x/programs/rust/sdk-macros/tests/ui/too-many-state-key-variants.rs
+++ b/x/programs/rust/sdk-macros/tests/ui/too-many-state-key-variants.rs
@@ -1,0 +1,14 @@
+use seq_macro::seq;
+use wasmlanche_sdk::state_keys;
+
+seq!(N in 0..=255  {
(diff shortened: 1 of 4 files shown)


x/programs/rust/sdk-macros/tests/ui/too-many-state-key-variants.rs
```diff
@@ -0,0 +1,14 @@
...
+
+seq!(N in 0..255  {
+    #[state_keys]
+    pub enum StateKeysFail {
+        #(
+            Variant~N,
+        )*
+        Variant256,
+    }
+});
```
iFrostizz, line 12: tiny nit: this will actually not generate the `Variant255` since the range is not inclusive of this number
Screenshot 2024-05-30 at 23 25 47

iFrostizz (dismissed):
🚀 🌚

dboehm-avalabs (approved):

iFrostizz (approved):

Issue #981, opened 2024-05-29
[x/programs] Remove any reference to plans


iFrostizz:
Since the merging of the interpreter mode in https://github.com/ava-labs/hypersdk/pull/884, we don't need a `Plan` anymore. We can just run single steps and have stronger guarantees on the value of the return data.
ref: https://github.com/ava-labs/hypersdk/pull/935#discussion_r1619319709

PR #1002, opened 2024-06-03
[x/programs] Remove plans

iFrostizz:
Remove plans and any reference to them, we are only leveraging the interpreter mode now and running one step at a time for more corectness.

Closes https://github.com/ava-labs/hypersdk/issues/981

### DIFF
--- a/x/programs/cmd/simulator/src/lib.rs
+++ b/x/programs/cmd/simulator/src/lib.rs
@@ -359,9 +359,11 @@ where
             .next()
             .ok_or(StepError::Client(ClientError::Eof))?
             .and_then(|step| {
-                step.error
-                    .clone()
-                    .map_or(Ok(step), |err| Err(StepError::Program(err)))
+                if let Some(err) = step.error {
+                    Err(StepError::Program(err))
+                } else {
+                    Ok(step)
+                }
             })?
             .try_into()
             .map_err(StepError::BorshDeserialization)
--- a/x/programs/cmd/simulator/src/lib.rs
+++ b/x/programs/cmd/simulator/src/lib.rs
@@ -8,7 +8,6 @@ use borsh::BorshDeserialize;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::{
     io::{BufRead, BufReader, Write},
-    num::TryFromIntError,
     path::Path,
     process::{Child, Command, Stdio},
 };
@@ -97,7 +96,6 @@ pub enum Key {
 #[derive(Clone, Debug, PartialEq)]
 pub enum Param {
     U64(u64),
-    I64(i64),
     String(String),
     Id(Id),
     Key(Key),
@@ -107,7 +105,6 @@ pub enum Param {
 #[serde(rename_all = "lowercase", tag = "type", content = "value")]
 enum StringParam {
     U64(String),
-    I64(String),
     String(String),
     Id(String),
 }
@@ -116,7 +113,6 @@ impl From<&Param> for StringParam {
     fn from(value: &Param) -> Self {
         match value {
             Param::U64(num) => StringParam::U64(b64.encode(num.to_le_bytes())),
-            Param::I64(num) => StringParam::I64(b64.encode(num.to_le_bytes())),
             Param::String(text) => StringParam::String(b64.encode(text.clone())),
             Param::Id(id) => {
                 let num: &usize = id.into();
@@ -143,12 +139,6 @@ impl Serialize for Param {
 impl From<u64> for Param {
     fn from(val: u64) -> Self {
         Param::U64(val)
-    }
-}
-
-impl From<i64> for Param {
-    fn from(val: i64) -> Self {
-        Param::I64(val)
     }
 }
 
@@ -246,8 +236,6 @@ pub enum StepError {
     BorshDeserialization(#[from] borsh::io::Error),
     #[error("Program error: {0}")]
     Program(String),
-    #[error("Int conversion error: {0}")]
-    TryFromInt(#[from] TryFromIntError),
 }
 
 /// A [Client] is required to pass [Step]s to the simulator by calling [run](Self::run_step).
@@ -313,7 +301,7 @@ where
         self.writer.write_all(run_command)?;
 
         let step = SimulatorStep { caller_key, step };
-        let input = serde_json::to_vec(&dbg!(step)).map_err(StepError::Serde)?;
+        let input = serde_json::to_vec(&step).map_err(StepError::Serde)?;
         self.writer.write_all(&input)?;
         self.writer.write_all(b"'\n")?;
         self.writer.flush()?;
--- a/x/programs/cmd/simulator/src/lib.rs
+++ b/x/programs/cmd/simulator/src/lib.rs
@@ -1,6 +1,6 @@
 //! A client and types for the VM simulator. This crate allows for Rust
 //! developers to construct tests for their programs completely in Rust.
-//! Alternatively the `Plan` can be written in JSON and passed to the
+//! Alternatively the `Step`s can be written in JSON and passed to the
 //! Simulator binary directly.
 
 use base64::{engine::general_purpose::STANDARD as b64, Engine};
@@ -31,7 +31,7 @@ pub enum Endpoint {
     Execute,
 }
 
-/// A [Plan] is made up of [Step]s. Each step is a call to the API and can include verification.
+/// A [Step] is a call to the simulator
 #[derive(Debug, Serialize, PartialEq, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct Step {
@@ -95,6 +95,7 @@ pub enum Key {
 // add `Cow` types for borrowing
 #[derive(Clone, Debug, PartialEq)]
 pub enum Param {
+    U32(u32),
     U64(u64),
     String(String),
     Id(Id),
@@ -104,6 +105,7 @@ pub enum Param {
 #[derive(Serialize)]
 #[serde(rename_all = "lowercase", tag = "type", content = "value")]
 enum StringParam {
+    U32(String),
     U64(String),
     String(String),
     Id(String),
@@ -115,6 +117,9 @@ impl Serialize for Param {
         S: serde::Serializer,
     {
         match self {
+            Param::U32(num) => {
+                Serialize::serialize(&StringParam::U32(b64.encode(num.to_le_bytes())), serializer)
+            }
             Param::U64(num) => {
                 Serialize::serialize(&StringParam::U64(b64.encode(num.to_le_bytes())), serializer)
             }
@@ -128,6 +133,12 @@ impl Serialize for Param {
             }
             Param::Key(key) => Serialize::serialize(key, serializer),
         }
+    }
+}
+
+impl From<u32> for Param {
+    fn from(val: u32) -> Self {
+        Param::U32(val)
     }
 }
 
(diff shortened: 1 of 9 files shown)


x/programs/cmd/simulator/src/lib.rs
```diff
@@ -174,12 +148,12 @@ pub struct PlanResult {
 }
 
 #[derive(Debug, Deserialize)]
-pub struct PlanResultTyped<T>
+pub struct StepResultTyped<T>
```
iFrostizz, line 151: I'd have loved to, but it's not easy to avoid having an intermediary type (if possible?).
The issue is that the return type has to somehow be known beforehand.
- `impl BorshDeserialize` would be great, but it's not allowed in fields. Otherwise I would have put it in `StepReponse`
- `Box<dyn BorshDeseialize>` to do the dispatching at runtime, but this type doesn't implement `Deserialize` which makes sense

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -339,17 +309,11 @@ where
...
-        caller_key: &str,
-        step: &Step,
-    ) -> Result<PlanResponseTyped<T>, StepError>
-    where
-        T: BorshDeserialize,
-    {
-        self._run_step(caller_key, step)?
+            .and_then(|step| {
+                step.error
+                    .clone()
```
iFrostizz, line 314: Replacing this with a `if let ...` pattern to only move the error when there is one

x/programs/cmd/simulator/src/lib.rs
richardpringle: FYI

Neither `impl BorshDeserialize` nor `Box<dyn BorshDeserialize>` wouldn't make any sense here. `impl Trait` syntax is meant for opaque types that are known at compile-time OR they are equivalent to a generic with a trait bound. `Box<dyn Trait>` requires object-safety ([refresher](https://doc.rust-lang.org/reference/items/traits.html#object-safety)). `BorshDeserialize` is not object-safe.

richardpringle: Why don't you just remove the `pub` visibility from `StepResult::response` and have a getter method. 
```rust
impl StepResult {
    fn response<T: BorshDeserialize>(&self) -> Result<T, borsh::io::Error> {
        borsh::from_slice(&self.response)
    } 
}
```
Then you can get rid of all the `*Typed<T>` types

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -302,21 +277,26 @@ pub enum ClientError {
...
+    #[error("Client error: {0}")]
     Client(#[from] ClientError),
     #[error("Serialization / Deserialization error: {0}")]
     Serde(#[from] serde_json::Error),
     #[error("Borsh deserialization error: {0}")]
     BorshDeserialization(#[from] borsh::io::Error),
+    #[error("Program error: {0}")]
+    Program(String),
+    #[error("Int conversion error: {0}")]
+    TryFromInt(#[from] TryFromIntError),
```
richardpringle, line 289: Where's this being used? This error does not look like the others...

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -302,21 +277,26 @@ pub enum ClientError {
...
 }
 
-/// A [Client] is required to pass a [Plan] to the simulator, then to [run](Self::run_plan) the actual simulation.
+/// A [Client] is required to run [Step]s to the simulator, by calling [run](Self::run_step).
```
richardpringle, line 292:
```suggestion
/// A [Client] is required to pass [Step]s to the simulator by calling [run](Self::run_step).
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -155,51 +169,10 @@ impl From<Key> for Param {
...
-pub struct PlanError(String);
-impl core::fmt::Debug for PlanError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
-    }
-}
-
 #[derive(Debug, Deserialize)]
-pub struct PlanResult {
+pub struct StepResult {
```
richardpringle, line 173: Can you rename `Step` to `Action`? A "step" only makes sense in the context of a plan (a plan has many steps). A simulator doesn't really have steps; it executes actions.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -392,19 +364,14 @@ where
...
-    where
-        T: BorshDeserialize,
-    {
-        self._run_step(caller_key, step)?
+            .and_then(|step| {
+                if let Some(err) = step.error {
+                    Err(StepError::Program(err))
+                } else {
+                    Ok(step)
+                }
```
richardpringle, line 372: You should do this as part of the deserialization so that we can't represent what is effectively an invalid response. We should never have both a response and an error, it should be one or the other.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -70,34 +71,36 @@ mod tests {
...
+            .run_step::<()>(
+                &owner_key,
+                &Step {
+                    endpoint: Endpoint::Execute,
+                    method: "program_create".into(),
+                    max_units: 1000000,
+                    params: vec![Param::String(PROGRAM_PATH.into())],
+                },
+            )
+            .unwrap();
```
richardpringle, line 103: If you do what I suggest with the `repsonse` method, you wouldn't need to pass in the type argument to `run_step` and that would clean up this code. 

Also, a test without a single assertion isn't a great test (the test before wasn't great, but it's more apparent now). 

Please make the test function return a `Result<(), Error>` so that we actually have good error messages.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -107,41 +110,49 @@ mod tests {
...
-                .iter()
-                .filter_map(|resp| resp.base.error.as_ref())
-                .next()
-        );
+        simulator
+            .run_step::<()>(
+                &owner_key,
+                &Step::create_key(Key::Ed25519(owner_key.clone())),
+            )
+            .unwrap();
```
richardpringle, line 118: These unwraps are also going aren't going to give the best message for failing tests either. You should make this test return a `Result` as well.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -167,47 +178,61 @@ mod tests {
...
+            .run_step::<u32>(
+                &owner_key,
+                &Step {
+                    endpoint: Endpoint::Execute,
+                    method: "initialize_address".into(),
+                    max_units: 1000000,
+                    params: vec![counter2_id.into(), bob_key.clone()],
+                },
+            )
+            .unwrap();
```
richardpringle, line 235: Same comment for all the unwraps

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -243,63 +239,73 @@ mod tests {
...
+                    method: "mint_to".into(),
+                    params: vec![
+                        program_id.into(),
+                        alice_key.clone(),
+                        Param::U64(alice_initial_balance),
+                    ],
+                    max_units: 1000000,
+                },
+            )
+            .unwrap();
```
richardpringle, line 304: Same comment on all these `unwrap`s

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -327,75 +333,98 @@ mod tests {
...
+                    params: vec![
+                        program_id.into(),
+                        alice_key.clone(),
+                        bob_key.clone(),
+                        Param::U64(transfer_amount),
+                    ],
+                    max_units: 1000000,
+                },
+            )
+            .unwrap();
```
richardpringle, line 423: Same comment on the unwraps

richardpringle (changes requested):

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Thanks for the refresher, and for the very nice solution. It solves it all...

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -1,17 +1,19 @@
 //! A client and types for the VM simulator. This crate allows for Rust
 //! developers to construct tests for their programs completely in Rust.
-//! Alternatively the `Plan` can be written in JSON and passed to the
+//! Alternatively the `Step`s can be written in JSON and passed to the
```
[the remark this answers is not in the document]
iFrostizz: You're right, we now have a link to the `Step` type after building the docs, and it seems to be a good practice to use these doc comments for contained items ! From the [book section](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#commenting-contained-items) `We typically use these doc comments inside the crate root file (src/lib.rs by convention) or inside a module to document the crate or the module as a whole.`

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -1,17 +1,19 @@
...
 use borsh::BorshDeserialize;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::{
     io::{BufRead, BufReader, Write},
+    num::TryFromIntError,
     path::Path,
     process::{Child, Command, Stdio},
 };
 use thiserror::Error;
+use wasmlanche_sdk::Gas;
```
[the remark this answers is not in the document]
iFrostizz: I removed the import to this crate

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -105,6 +108,8 @@ pub enum Param {
 #[serde(rename_all = "lowercase", tag = "type", content = "value")]
 enum StringParam {
     U64(String),
+    #[serde(rename = "i64")]
+    Gas(String),
```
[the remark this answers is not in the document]
iFrostizz: True... I'm removing it

x/programs/cmd/simulator/src/lib.rs
iFrostizz: In the `TryFrom` implementation to convert a `StepResponse` into a `StepResponseTyped<T>`, that's because there is no implementation of `From<i32>` for `Id`

x/programs/cmd/simulator/src/lib.rs
iFrostizz: fixed!

x/programs/cmd/simulator/src/lib.rs
iFrostizz: I'm scared that this is going to need a lot of changes, it's also called a step on the go side. Should these be equivalent to actions triggered by transactions ? If not, then we may need to find another name to avoid any confusion

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -137,6 +145,12 @@ impl From<u64> for Param {
     }
 }
 
+impl From<Gas> for Param {
+    fn from(val: Gas) -> Self {
+        Param::Gas(val)
+    }
+}
```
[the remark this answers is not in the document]
iFrostizz: It's tricky that way... Because then we need to implement `From<&Param> for &StringParam` which leads to return a value referenced by the local function

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: I miss Python's `cached_property` so much...

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: ... Figured that you meant `From<&Param> for StringParam`, I implemented it and I'm not actually sure if it's really clearer that way. We need to opt-out the `Key` variant because it's not at the same level as other params

x/programs/cmd/simulator/src/lib.rs
iFrostizz: I think that we need to discuss more about renaming plans

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: I'm going to leave the `unwrap()` since they actually provide a better error message than short-circuiting by the use of `?` and returning a `Result` from the test, because it gives us the backtrace with the line where it panicked in case it happens. It is going to give us an easier time to debug if any of the tests fail since we will exactly know where is failed.

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: but I implemented your suggestion, good catch thanks!

richardpringle (dismissed):

richardpringle:
@dboehm-avalabs, feel free to fix the conflicts approve and merge, you have my blessing

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/981

dboehm-avalabs (approved):

PR #1005, opened 2024-06-03
Centralize gas-type

richardpringle:
This makes it easier to change the type to a u64 if desired (or any other type)


iFrostizz (approved):
LGTM!

iFrostizz:
This might be a leftover https://github.com/ava-labs/hypersdk/blob/ff05db9eb56935fc964c0a88397269312e9dce8e/x/programs/cmd/simulator/src/lib.rs#L43 and I also don't know why we would have this type being a `u64`. I guess that it would still work if you replace by `Gas` here since the bitwise representation of positive numbers *should* almost be the same (except when it grows large)

richardpringle:
> hypersdk/x/programs/cmd/simulator/src/lib.rs

I think this is fine for now... I don't want to add another dependency just for the `Gas` type.

PR #1019, opened 2024-06-05
[x/programs] Switch program ids to address

dboehm-avalabs:
We want the programs to be associated with addresses rather than ids since that maps to where the state will be stored.


x/programs/runtime/runtime.go
```diff
@@ -41,7 +40,7 @@ func NewRuntime(
 		cfg:           cfg,
 		engine:        wasmtime.NewEngineWithConfig(cfg.wasmConfig),
 		hostImports:   NewImports(),
-		programs:      map[ids.ID]*Program{},
+		programs:      map[codec.Address]*Program{},
```
richardpringle, line 43: 👍

richardpringle (approved):

PR #1021, opened 2024-06-05
[x/programs] Create State Loader

dboehm-avalabs:
Allow the user to determine how the path to the state of the programs is generated.

### DIFF
--- a/x/programs/test/programs/call_program/src/lib.rs
+++ b/x/programs/test/programs/call_program/src/lib.rs
@@ -1,4 +1,4 @@
-use wasmlanche_sdk::{public, Context, Gas, Program};
+use wasmlanche_sdk::{public, types::Address, Context, Gas, Program};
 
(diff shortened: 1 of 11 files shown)


x/programs/test/programs/call_program/src/lib.rs
```diff
@@ -10,6 +10,17 @@ pub fn simple_call_external(_: Context, target: Program, max_units: Gas) -> i64
...
+pub fn actor_check(context: Context) -> Address {
+    let Context { actor, .. } = context;
+    actor
+}
+
+#[public]
+pub fn actor_check_external(_: Context, target: Program, max_units: Gas) -> Address {
+    target.call_function("actor_check", (), max_units).unwrap()
+}
+
```
richardpringle, line 23: 👍

richardpringle (approved):

Issue #867, opened 2024-04-26
[x/programs] Cache deletes until wasmtime module exit.


richardpringle:
There are two scenarios where an exit occurs: either the program has successfully completed, or another program has been invoked. #840 handles puts, but not deletes.

PR #1029, opened 2024-06-07
Implement cache deletes

iFrostizz:
Cache deletes in the program state so that we now avoid cache misses when it comes to data deletion.
Also added a cache insertion when calling `get`.
The new cache now tells the caller wether the data is here `Data(Vec<u8>)`, has been deleted `Deleted` or `ToDelete` if it's scheduled to be deleted when the cache is flushed.
There is a caveat though which is that ZST Vecs are completely disabled by borsh-rs. That's because they want to avoid a potential DOS since huge vecs can easily be crafted using ZST (they are free memory-wise!) with `vec![(), !0]`.
See https://github.com/near/borsh-rs/issues/52
We need to make sure that we are not exposed to that, or that fuel is correctly metering to avoid abuses.

Closes https://github.com/ava-labs/hypersdk/issues/867
Closes https://github.com/ava-labs/hypersdk/issues/868

### DIFF
--- a/x/programs/runtime/import_state.go
+++ b/x/programs/runtime/import_state.go
@@ -26,6 +26,10 @@ type keyValueInput struct {
 	Value []byte
 }
 
+type keyInput struct {
+	Key []byte
+}
+
 func NewStateAccessModule() *ImportModule {
 	return &ImportModule{
 		Name: "state",
@@ -98,9 +102,7 @@ func NewStateAccessModule() *ImportModule {
 				return bytes, nil
 			})},
 			"delete_many": {FuelCost: putManyCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
-				// fmt.Fprintln(os.Stderr, "OK")
-				// return nil
-				var parsedInput [][]byte
+				var parsedInput []keyInput
 				if err := borsh.Deserialize(&parsedInput, input); err != nil {
 					fmt.Fprintln(os.Stderr, err)
 					return err
@@ -111,7 +113,7 @@ func NewStateAccessModule() *ImportModule {
 				defer cancel()
 
 				for _, entry := range parsedInput {
-					if err := callInfo.State.GetProgramState(callInfo.Program).Remove(ctx, entry); err != nil {
+					if err := callInfo.State.GetProgramState(callInfo.Program).Remove(ctx, entry.Key); err != nil {
 						fmt.Fprintln(os.Stderr, err)
 						return err
 					}
--- a/x/programs/runtime/import_state.go
+++ b/x/programs/runtime/import_state.go
@@ -33,8 +33,10 @@ func NewStateAccessModule() *ImportModule {
 				if err := borsh.Deserialize(&parsedInput, input); err != nil {
 					return nil, err
 				}
+
 				ctx, cancel := context.WithCancel(context.Background())
 				defer cancel()
+
 				val, err := callInfo.State.GetProgramState(callInfo.Program).GetValue(ctx, parsedInput)
 				if err != nil {
 					if errors.Is(err, database.ErrNotFound) {
@@ -42,6 +44,7 @@ func NewStateAccessModule() *ImportModule {
 					}
 					return nil, err
 				}
+
 				return val, nil
 			})},
 			"put": {FuelCost: putCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
@@ -49,8 +52,10 @@ func NewStateAccessModule() *ImportModule {
 				if err := borsh.Deserialize(parsedInput, input); err != nil {
 					return err
 				}
+
 				ctx, cancel := context.WithCancel(context.Background())
 				defer cancel()
+
 				return callInfo.State.GetProgramState(callInfo.Program).Insert(ctx, parsedInput.Key, parsedInput.Value)
 			})},
 			"put_many": {FuelCost: putManyCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
@@ -58,13 +63,16 @@ func NewStateAccessModule() *ImportModule {
 				if err := borsh.Deserialize(&parsedInput, input); err != nil {
 					return err
 				}
+
 				ctx, cancel := context.WithCancel(context.Background())
 				defer cancel()
+
 				for _, entry := range parsedInput {
 					if err := callInfo.State.GetProgramState(callInfo.Program).Insert(ctx, entry.Key, entry.Value); err != nil {
 						return err
 					}
 				}
+
 				return nil
 			})},
 			"delete": {FuelCost: deleteCost, Function: Function(func(callInfo *CallInfo, input []byte) ([]byte, error) {
@@ -75,6 +83,7 @@ func NewStateAccessModule() *ImportModule {
 
 				ctx, cancel := context.WithCancel(context.Background())
 				defer cancel()
+
 				programState := callInfo.State.GetProgramState(callInfo.Program)
 				bytes, err := programState.GetValue(ctx, parsedInput)
 				if err != nil {
@@ -94,6 +103,22 @@ func NewStateAccessModule() *ImportModule {
 				}
 
 				return bytes, nil
+			})},
+			"delete_many": {FuelCost: putManyCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
+				var parsedInput [][]byte
+				if err := borsh.Deserialize(&parsedInput, input); err != nil {
+					return err
+				}
+
+				ctx, cancel := context.WithCancel(context.Background())
+				defer cancel()
+
+				for _, entry := range parsedInput {
+					if err := callInfo.State.GetProgramState(callInfo.Program).Remove(ctx, entry); err != nil {
+						return err
+					}
+				}
+				return nil
 			})},
 		},
 	}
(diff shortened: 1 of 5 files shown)


x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -257,7 +257,7 @@ pub fn state_keys(_attr: TokenStream, item: TokenStream) -> TokenStream {
 
     // add default attributes
     item_enum.attrs.push(parse_quote! {
-         #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
```
iFrostizz, line 260: a tab was off 😄

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -38,8 +42,15 @@ pub enum Error {
     Delete,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum CachedData {
+    Data(Vec<u8>),
+    Deleted,
+    ToDelete,
+}
```
iFrostizz, line 50: Here, `Deleted` means "we know that there is no value here", it could be a bit unclear since it may mean that some value was here before, while it just means that we checked in the actual storage and there is no value.

iFrostizz: Also, having a value at a key `K` in the cache means "we know the state of the value here". If it's not here that means that it's unknown.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -84,60 +97,81 @@ impl<'a, K: Key> State<'a, K> {
...
+                    let value = from_slice::<V>(&bytes)
+                        .map_err(|_| StateError::Deserialization)
+                        .map(Some);
+                    entry.insert(CachedData::Data(bytes));
+                    value
+                } else {
+                    entry.insert(CachedData::Deleted);
+                    Ok(None)
+                }
+            }
```
iFrostizz, line 122: I'm not sure how to avoid the repetition here, in one branch the data is borrowed because it comes from the storage. On the other, it has been copied from host so it's owned.

iFrostizz: It's working quite well though because we need an owned value to insert in the cache.

x/programs/runtime/import_state.go
```diff
@@ -95,6 +99,22 @@ func NewStateAccessModule() *ImportModule {
 
 				return bytes, nil
 			})},
+			"delete_many": {FuelCost: putManyCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
+				var parsedInput []keyInput
```
dboehm-avalabs, line 103: This should be able to just be [][]byte.

iFrostizz: true, no reason to define a struct

iFrostizz: thank you!

dboehm-avalabs (approved):

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle: Okay, this actually compiles
```rust
    pub fn get<V>(self, key: K) -> Result<Option<V>, Error>
    where
        V: BorshDeserialize,
    {
        #[link(wasm_import_module = "state")]
        extern "C" {
            #[link_name = "get_bytes"]
            fn get_bytes(ptr: *const u8, len: usize) -> HostPtr;
        }

        let mut cache = self.cache.borrow_mut();

        let cached_data = match cache.get(&key) {
            Some(cached_data) => cached_data,
            None => {
                let args_bytes = borsh::to_vec(&key).map_err(|_| StateError::Serialization)?;

                let ptr = unsafe { get_bytes(args_bytes.as_ptr(), args_bytes.len()) };

                if ptr.is_null() {
                    cache.insert(key, CachedData::Deleted);
                    return Ok(None);
                }

                cache.entry(key).or_insert(CachedData::Data(ptr.into()))
            }
        };

        match cached_data {
            CachedData::Data(data) => from_slice::<V>(data)
                .map_err(|_| StateError::Deserialization)
                .map(Some),
            CachedData::ToDelete | CachedData::Deleted => Ok(None),
        }
    }
```

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -84,60 +97,81 @@ impl<'a, K: Key> State<'a, K> {
...
+            Entry::Vacant(entry) => {
+                if let Some(data) = Self::get_from_host(key)? {
+                    entry.insert(CachedData::ToDelete);
+                    from_slice::<V>(&data)
+                        .map_err(|_| StateError::Deserialization)
+                        .map(Some)
+                } else {
+                    entry.insert(CachedData::Deleted);
+                    Ok(None)
+                }
```
richardpringle, line 154: Same as the comment above, I don't think you need all the repetition. There's a lot of surface area for bugs here.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -84,60 +97,81 @@ impl<'a, K: Key> State<'a, K> {
...
-            .map(Some)
+        }
     }
 
-    /// Delete a value from the hosts's storage.
-    /// # Errors
-    /// Returns an [Error] if the key cannot be serialized
-    /// or if the host fails to delete the key and the associated value
-    pub fn delete<T: BorshDeserialize>(self, key: K) -> Result<Option<T>, Error> {
+    fn get_from_host(key: K) -> Result<Option<Vec<u8>>, Error> {
```
richardpringle, line 159: I don't think you need this static method.

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -154,13 +191,24 @@ impl<'a, K: Key> State<'a, K> {
...
+        self.cache
+            .borrow_mut()
             .drain()
-            .map(|(key, value)| PutArgs { key, value })
-            .collect();
-        let serialized_args = borsh::to_vec(&args).expect("failed to serialize");
-        unsafe { put_many_bytes(serialized_args.as_ptr(), serialized_args.len()) };
+            .for_each(|(key, value)| match value {
+                CachedData::Data(value) => puts.push(PutArgs { key, value }),
+                CachedData::ToDelete => deletes.push(key),
```
richardpringle, line 200: Yeah, this tells me that a single API here would be better. I can't remember where I said it (could have been in this review or elsewhere), but I think that putting an empty `Vec<u8>` isn't valid. So you can actually eliminate all invalid state (minimizing the runtime checks) by making empty `Vec` puts the equivalent of `delete`. 

@dboehm-avalabs, thoughts?

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -154,13 +191,24 @@ impl<'a, K: Key> State<'a, K> {
...
+
+        if !puts.is_empty() {
+            let serialized_args = borsh::to_vec(&puts).expect("failed to serialize");
+            unsafe { put_many_bytes(serialized_args.as_ptr(), serialized_args.len()) };
+        }
+
+        if !deletes.is_empty() {
+            let serialized_args = borsh::to_vec(&deletes.as_slice()).expect("failed to serialize");
+            unsafe { delete_many_bytes(serialized_args.as_ptr(), serialized_args.len()) };
+        }
```
richardpringle, line 212: Haha, I just keep seeing good reason to centralize the API into a single call. If there's only one call, you can do a simple `if cache.is_empty() { return }` for `Drop`

richardpringle (changes requested):
Okay, sorry for taking so long with this review. 

I want to check with @dboehm-avalabs to see if he also things it would make sense before you launch into it, but what I would like you to do is centralize the API first. What I mean by that is a single 
```rust
       #[link(wasm_import_module = "state")]
        extern "C" {
            #[link_name = "put_bytes"]
            fn put_bytes(ptr: *const u8, len: usize);
        }
```
That actually takes a something like `Vec<(Vec<u8>, Vec<u8>)>` as the serialized data. The first value in the tuple is the key, the second is the value. If the value is empty, that should signify a delete. I'm not 100% sure what the best way is to represent this in Go, maybe `[][2][]byte` or something along those lines?

You should be able to do that with only minor changes to the Rust (nothing to the actual cache). 

Once we merge that code, we can come back and do the `CachedData` enum.

Off the top of my head, I think the cache would probably be `HashMap<K, Option<Vec<u8>>>`.

x/programs/rust/sdk-macros/src/lib.rs
iFrostizz: It's fine, it's only catching the eye when writing another aligned stuff just below

x/programs/runtime/import_state.go
```diff
@@ -95,6 +95,22 @@ func NewStateAccessModule() *ImportModule {
...
+				ctx, cancel := context.WithCancel(context.Background())
+				defer cancel()
+
+				for _, entry := range parsedInput {
+					if err := callInfo.State.GetProgramState(callInfo.Program).Remove(ctx, entry); err != nil {
+						return err
+					}
+				}
+				return nil
+			})},
```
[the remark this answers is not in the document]
iFrostizz: Opened https://github.com/ava-labs/hypersdk/pull/1057 in order to address https://github.com/ava-labs/hypersdk/pull/1029#discussion_r1633309815

iFrostizz:
Closing in favor of https://github.com/ava-labs/hypersdk/pull/1059

PR #1039, opened 2024-06-10
[x/programs] Reuse Linker to improve execution time

dboehm-avalabs:
By reusing the linker we get about a 10X improvement in runtime based on the added benchmark.


x/programs/runtime/import_program.go
```diff
@@ -63,7 +64,8 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
 				return result, nil
 			})},
 			"set_call_result": {FuelCost: setResultCost, Function: FunctionNoOutput(func(callInfo *CallInfo, input []byte) error {
-				callInfo.inst.result = input
+				// needs to clone because this points into the current store's linear memory which may be gone when this is read
+				callInfo.inst.result = slices.Clone(input)
```
iFrostizz, line 68: Isn't `slices.Clone` doing a shallow copy ? So maybe that the pointer is not cloned

iFrostizz (approved):

richardpringle (approved):

x/programs/runtime/import_program.go
dboehm-avalabs: Elements are bytes in this case, so shallow copy of the bytes is fine.

Issue #1034, opened 2024-06-10
[x/programs] External program calls should return a `Result<T, ExternalCallError>`


richardpringle:
External program calls should return a `Result<T, ExternalCallError>` where `T: BorshDeserialize` (which could include a user-defined `Result` type). 

`ExternalCallError` should be defined as follows:
```rust
#[derive(thiserror::Error)]
#[repr(u8)] // the host is effectively returning an error-code
#[non_exhaustive] // assume that we will make this error more granular as we see fit
enum ExternalCallError {
    ExecutionFailure = 0, // all encompassing error
    CallPanicked = 1, // explicit panic in the external program 
    OutOfFuel = 2,
}
```

Program authors writing programs with external calls should be able to recover from certain scenarios or at the very least, deliver a meaningful message to their callers.

PR #1043, opened 2024-06-11
Return `Result` from public function calls

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1034

We now return a `Result` as part of the public API when calling programs externally or internally.
This is an improvement as we are now able to react to a call failing.
You can see the effect on example tests, which now won't fail if the external call failed, but the assertion will do.
This also lets us pattern match on the error returned, if it's a panic, an out of fuel, or something else.

edit:
also closes https://github.com/ava-labs/hypersdk/issues/865


x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -217,10 +217,7 @@ func (c *runCmd) runStepFunc(
 		if err != nil {
 			return err
 		}
-		response, balance, err := programExecuteFunc(ctx, c.log, db, programAddress, params[1:], method, maxUnits)
-		if err != nil {
-			return err
-		}
```
iFrostizz, line 223: Removed errors that were coming from the vm space from the simulator

x/programs/cmd/simulator/cmd/program.go
```diff
@@ -83,8 +82,6 @@ func programCreateFunc(ctx context.Context, db *state.SimpleMutable, path string
 
 	address, err := SetProgram(ctx, db, programID, programBytes)
 	if err != nil {
-		response := multilineOutput([][]byte{utils.ErrBytes(err)})
-		fmt.Println(response)
```
iFrostizz, line 87: This didn't made sense

x/programs/runtime/import_program.go
```diff
@@ -49,18 +58,16 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
 
 				result, err := r.CallProgram(
 					context.Background(),
-					&newInfo)
```
iFrostizz, line 52: I think that the preference on the go side is the parenthesis sticked at the end of the last line, should I back this off ?

x/programs/runtime/import_program.go
```diff
@@ -49,18 +58,16 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
...
-				if err != nil {
-					return nil, err
-				}
+					&newInfo,
+				)
 
 				// subtract the fuel used during this call from the calling program
 				remainingFuel := newInfo.RemainingFuel()
-				if err := callInfo.ConsumeFuel(parsedInput.Fuel - remainingFuel); err != nil {
-					return nil, err
```
iFrostizz, line 60: @dboehm-avalabs is it okay to treat this error the same as the `OutOfFuel` trap error that is in the runtime ?

x/programs/runtime/runtime.go
```diff
@@ -65,6 +65,23 @@ func (r *WasmRuntime) AddProgram(program codec.Address, bytes []byte) (*Program,
...
+		// https://github.com/bytecodealliance/wasmtime-go/blob/v14.0.0/trap.go#L30
+		switch *trap.Code() {
+		case 9:
+			return append(Err, CallPanicked), err
+		case 11:
+			return append(Err, OutOfFuel), err
+		default:
+			return append(Err, ExecutionFailure), err
+		}
+	}
```
richardpringle, line 80: Don't we already have a single source of truth for converting trap-errors into our own set of errors are whatever? 

There should be an abstraction here and there definitely shouldn't be random numbers just floating in the code. At the very least, these numbers should be defined as constants.

x/programs/rust/examples/counter-external/src/lib.rs
```diff
@@ -75,6 +75,6 @@ mod tests {
             .response::<u64>()
             .unwrap();
 
-        assert_eq!(response, 1u64);
+        assert_eq!(response, Ok(1));
```
richardpringle, line 78: You'll need to put this back when you remove the nested `Result`s

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -278,6 +279,28 @@ mod tests {
             .result
             .response::<u64>()
             .unwrap();
-        assert_eq!(value, 10);
+        assert_eq!(value, Ok(10));
```
richardpringle, line 282: All these will have to go back too

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -278,6 +279,28 @@ mod tests {
...
+                        bob_key.clone(),
+                        10.into(),
+                    ],
+                },
+            )
+            .unwrap()
+            .result
+            .response::<u64>()
+            .unwrap();
+        assert_eq!(res, Err(ExternalCallError::OutOfFuel));
```
richardpringle, line 304: Can you do this in a separate test

x/programs/rust/examples/token/src/lib.rs
```diff
@@ -505,6 +505,6 @@ mod tests {
             .result
             .response::<u64>()
             .unwrap();
-        assert_eq!(balance, 0);
+        assert_eq!(balance, Ok(0));
```
richardpringle, line 508: All these will have to go back too

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -44,22 +58,22 @@ impl<K> Program<K> {
     /// Attempts to call a function `name` with `args` on the given program. This method
     /// is used to call functions on external programs.
     /// # Errors
-    /// Returns a [`StateError`] if the call fails.
+    /// Returns a [`ExternalCallError`] if the call fails.
```
richardpringle, line 61: I don't think you need the "`"s 
```suggestion
    /// Returns a [ExternalCallError] if the call fails.
```
Not 100% sure though

x/programs/test/util.go
```diff
@@ -84,7 +84,7 @@ func prependAccountToKey(account codec.Address, key []byte) []byte {
 
 func Into[T any](data []byte) T {
 	result := new(T)
-	if err := borsh.Deserialize(result, data); err != nil {
+	if err := borsh.Deserialize(result, data[1:]); err != nil {
```
richardpringle, line 87: Wait... what happened here?

richardpringle (changes requested):

x/programs/runtime/import_program.go
dboehm-avalabs: Realistically this should never actually error.  We check that there is enough fuel before calling this and we will fail much earlier if fuel isn't enabled.

x/programs/runtime/import_program.go
```diff
@@ -33,12 +42,12 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
 				newInfo := *callInfo
 				parsedInput := &callProgramInput{}
 				if err := borsh.Deserialize(parsedInput, input); err != nil {
-					return nil, err
+					return append(Err, ExecutionFailure), err
```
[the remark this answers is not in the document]
dboehm-avalabs: Yeah, I hate this.  Lets change the host function signature and let it deal with the serialization somewhere else.

x/programs/runtime/import_program.go
iFrostizz: Thanks, so I will leave it that way

x/programs/cmd/simulator/cmd/plan.go
iFrostizz: It's going to be written in the output, see https://github.com/ava-labs/hypersdk/blob/a40ad2eff0b9622ff588fe33f2e58008b9642519/x/programs/runtime/runtime.go#L69-L78

x/programs/cmd/simulator/cmd/program.go
```diff
@@ -126,20 +119,11 @@ func programExecuteFunc(
 		Params:       bytes,
 	}
 	programOutput, err := rt.CallProgram(ctx, callInfo)
-	output := [][]byte{programOutput}
 	if err != nil {
-		response := multilineOutput(output)
-		return nil, 0, fmt.Errorf("program execution failed: %s, err: %w", response, err)
+		return [][]byte{programOutput}, 0
```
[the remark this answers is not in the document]
iFrostizz: See https://github.com/ava-labs/hypersdk/pull/1043#discussion_r1635165817, the error is actually at the vm level now

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -173,7 +174,7 @@ pub struct StepResult {
 }
 
 impl StepResult {
-    pub fn response<T>(&self) -> Result<T, borsh::io::Error>
+    pub fn response<T>(&self) -> Result<Result<T, ExternalCallError>, borsh::io::Error>
```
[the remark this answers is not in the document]
iFrostizz: Much better, yeah, thanks!

x/programs/runtime/runtime.go
iFrostizz: Thanks, yeah I just used the constants at this link in wasmtime, it's not going to break if we update the dependency that way.

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: It makes it a bit tedious to check on program errors though:

```rust
        let res = simulator
            .run_step(
                &owner_key,
                &Step {
                    endpoint: Endpoint::Execute,
                    method: "inc_external".into(),
                    max_units: 100_000_000,
                    params: vec![
                        counter1_id.into(),
                        counter2_id.into(),
                        100.into(),
                        bob_key.clone(),
                        10.into(),
                    ],
                },
            )
            .unwrap()
            .result
            .response::<u64>()
            .unwrap_err();
        let simulator::StepResponseError::ExternalCall(err) = res else {
            panic!();
        };
        assert_eq!(err, ExternalCallError::OutOfFuel);
```

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: Done!

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Yep you're right!
Screenshot 2024-06-11 at 19 15 11

x/programs/test/util.go
iFrostizz: This is because the data is used from a function call, which contains this 1 or 0 prepended byte for `Ok` or `Err`. But it definitely should not be here

x/programs/runtime/import_program.go
```diff
@@ -18,6 +18,15 @@ const (
 	getFuelCost     = 10000
 )
 
+var (
+	Err = []byte{0}
+	Ok  = []byte{1}
```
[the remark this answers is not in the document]
iFrostizz: I have done what David's told me to do instead, which is to create some helper functions to help define `Ok(T)`, `Err(T)`, `Some(T)`, or `None` as in Rust.

x/programs/runtime/import_program.go
```diff
@@ -49,18 +58,16 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
 
 				result, err := r.CallProgram(
 					context.Background(),
-					&newInfo)
-				if err != nil {
-					return nil, err
-				}
```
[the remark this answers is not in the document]
iFrostizz: Yes that's the reason, I'm not really sure how to rewrite it though since we need to be careful not to reassign the `err` variable.

x/programs/test/util.go
iFrostizz: I almost want to write an `unwrap` function...

x/programs/runtime/import_program.go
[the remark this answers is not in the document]
iFrostizz: Done, I have introduced a new `InfaillbleFunction`, lmk if that works

x/programs/test/util.go
iFrostizz: That's what I've done, let's see if you are happy with it 😄

x/programs/cmd/simulator/cmd/plan.go
iFrostizz: I just ended up removing the err at all, it all makes much more sense now

x/programs/runtime/import_program.go
iFrostizz: yeah I just backed this off.

x/programs/runtime/import_program.go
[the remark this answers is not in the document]
richardpringle: I think it just deserves a comment saying "need to handle fuel before dealing with the error" or whatever

x/programs/test/util.go
richardpringle: But how do you know what you're deserializing into? I'm not suggesting that this code is wrong, it's just not clear _why_ it's right

iFrostizz: I removed the access to `[1:]` and wrote this `Unwrap` function instead, to be used after the function call to the program. Is that what you are talking about ?

x/programs/runtime/import_program.go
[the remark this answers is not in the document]
iFrostizz: As I removed the `err` from the above `r.CallProgram`, I just replace the `fuelErr` by the usual `err`. I think that it doesn't need an err anymore since the flow is the same as previous

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Sooo actually, clippy pedantic wants us to put the backticks
```sh
error: item in documentation is missing backticks
  --> x/programs/rust/wasmlanche-sdk/src/program.rs:61:20
   |
61 |     /// Returns a [ExternalCallError] if the call fails.
   |                    ^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown
note: the lint level is defined here
  --> x/programs/rust/wasmlanche-sdk/src/lib.rs:1:9
   |
1  | #![deny(clippy::pedantic)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[deny(clippy::doc_markdown)]` implied by `#[deny(clippy::pedantic)]`
help: try
   |
61 |     /// Returns a [`ExternalCallError`] if the call fails.
   |                    ~~~~~~~~~~~~~~~~~~~
   ```
   
   I just added them to other parts of the code where they were missing.

Comment (richardpringle) [reply x/programs/rust/wasmlanche-sdk/src/program.rs:61]: >I just added them to other parts of the code where they were missing.

Do we not have pedantic turned on everywhere?

File: x/programs/cmd/simulator/src/lib.rs (hunk 14 above)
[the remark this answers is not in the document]
Comment (richardpringle) [reply x/programs/cmd/simulator/src/lib.rs]: That's fine

File: x/programs/rust/wasmlanche-sdk/src/program.rs (hunk 10 above)
Comment (iFrostizz) [reply x/programs/rust/wasmlanche-sdk/src/program.rs:61]: only on the wasmlanche-sdk

File: x/programs/cmd/simulator/cmd/plan.go (hunk 1 above)
Comment (richardpringle) [reply x/programs/cmd/simulator/cmd/plan.go:223]: Isn't this function used to invoke a program function regardless of the source? What if it's just the runtime?

File: x/programs/cmd/simulator/src/lib.rs [hunk 17]
```diff
@@ -172,12 +173,21 @@ pub struct StepResult {
     response: Vec<u8>,
 }
 
+#[derive(Error, Debug)]
+pub enum StepResponseError {
+    #[error(transparent)]
+    Serialization(#[from] borsh::io::Error),
+    #[error(transparent)]
+    ExternalCall(#[from] ExternalCallError),
+}
```
richardpringle, line 182: 👍

x/programs/runtime/import_program.go
```diff
@@ -48,20 +53,17 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
 				newInfo.Params = parsedInput.Params
 				newInfo.Fuel = parsedInput.Fuel
 
-				result, err := r.CallProgram(
+				result := r.CallProgram(
```
richardpringle, line 56: The logic to handle the trap error should exist here; or at least the decision to use said logic. Can you put `CallProgram` back to the way it was? 

If you want `func (r *WasmTime) ConcertErrorToResponse(err error) Response` or whatever, you can do that so that the logic is re-usable, but this is the "presentation" layer where the code should be making decisions about how the response should specifically be presented to the wasm-module.

x/programs/runtime/import_program.go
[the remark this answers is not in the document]
richardpringle: Please see my other comment that starts with 
>The logic to handle the trap error should exist here

I can't link until I finish the review

x/programs/runtime/utils.go
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package runtime
+
+var None = []byte{0}
```
richardpringle, line 6: None is a prefix, it should not be a `[]byte`. You should only be constructing the `[]byte` where it's needed, not here.

x/programs/runtime/import_program.go
```diff
@@ -19,6 +18,12 @@ const (
 	getFuelCost     = 10000
 )
 
+var (
+	ExecutionFailure = []byte{0}
+	CallPanicked     = []byte{1}
+	OutOfFuel        = []byte{2}
```
richardpringle, line 24: These shouldn't be `[]byte`, they should just be `u8` that are inserted into a `[]byte` when necessary.

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -280,4 +281,82 @@ mod tests {
...
+                },
+            )
+            .unwrap()
+            .result
+            .response::<u64>()
+            .unwrap_err();
+        let simulator::StepResponseError::ExternalCall(err) = res else {
+            panic!();
+        };
+        assert_eq!(err, ExternalCallError::OutOfFuel);
```
richardpringle, line 360: Instead of panicking here without an error message, you can just use `matches!`
```rust
        assert!(matches!(
            err,
            StepResponseError::ExternalCall(ExternalCallError::OutOfFuel)
        ));
```

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -1,11 +1,25 @@
...
-    state::{Error as StateError, Key, State},
+    state::{Key, State},
     types::Address,
     Gas,
 };
 use borsh::{BorshDeserialize, BorshSerialize};
 use std::{cell::RefCell, collections::HashMap};
+use thiserror::Error;
+
+#[derive(Error, Debug, BorshDeserialize, PartialEq)]
```
richardpringle, line 11:
```suggestion
#[derive(Error, Debug, BorshDeserialize)]
```

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -68,11 +82,14 @@ impl<K> Program<K> {
...
-        let args_bytes = borsh::to_vec(&args).map_err(|_| StateError::Serialization)?;
+        let args_bytes = borsh::to_vec(&args).map_err(|_| ExternalCallError::ExecutionFailure)?;
 
         let bytes = unsafe { call_program(args_bytes.as_ptr(), args_bytes.len()) };
 
-        borsh::from_slice(&bytes).map_err(|_| StateError::Deserialization)
+        match borsh::from_slice(&bytes) {
+            Ok(res) => res,
+            Err(_) => Err(ExternalCallError::ExecutionFailure),
+        }
```
richardpringle, line 92:
```suggestion
        borsh::from_slice(&bytes).map_err(|_| ExternalCallError::ExecutionFailure)?
```

richardpringle (changes requested):

x/programs/runtime/import_program.go
iFrostizz: I have renamed the `callProgram` function to a more explicit `CallProgramToResponse` and kept the `CallProgram` function as-is.

x/programs/runtime/utils.go
iFrostizz: I replaced the value of `None` by 0. Also replaced `var` to a `const`.

x/programs/runtime/import_program.go
iFrostizz: Replaced all of them by constants of ints, the `[]byte` value is constructed at runtime now.

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: good one, [`assert_matches!`](https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html) would have been great but it's on nightly.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Removed because we now uses `matches!`

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Thanks!

x/programs/runtime/imports.go
```diff
@@ -100,6 +100,21 @@ func (f Function[T, U]) call(callInfo *CallInfo, caller *wasmtime.Caller, vals [
 	return writeOutputToMemory(callInfo, results, err)
 }
 
+type InfaillibleFunction[T any] func(*CallInfo, T) RawBytes
```
dboehm-avalabs, line 103: Don't need this, especially because it can still fail due to input desserialization.

x/programs/runtime/utils.go
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package runtime
+
+const None = byte(0)
+
+func Ok(data []byte) []byte {
```
dboehm-avalabs, line 8: Use the type system instead of []byte.  Doing it this way makes it unclear when bytes are actually bytes or if they are an option or a result.

x/programs/runtime/import_program.go
```diff
@@ -15,6 +14,10 @@ const (
 	callProgramCost = 10000
 	setResultCost   = 10000
 	getFuelCost     = 10000
+
+	ExecutionFailure = 0
```
dboehm-avalabs, line 18: Create an error type for these.

type ErrorCode uint8
ExecutionFailure ErrorCode = 0

x/programs/runtime/runtime.go
```diff
@@ -72,6 +72,22 @@ func (r *WasmRuntime) AddProgram(program codec.Address, bytes []byte) (*Program,
 	return programModule, nil
 }
 
+func (r *WasmRuntime) CallProgramToResponse(ctx context.Context, callInfo *CallInfo) []byte {
+	result, err := r.CallProgram(ctx, callInfo)
+	if err != nil {
+		trap := convertToTrap(err)
+		switch *trap.Code() {
```
dboehm-avalabs, line 79: I'm not sure this is safe, I don't think you are guaranteed to have a trap code

x/programs/runtime/imports.go
iFrostizz: Isn't it still going to happen when we call `writeOutputToMemory` ?
I wanted to highlight that this function couldn't fail, but I guess that I just return `nil` everywhere too

x/programs/runtime/utils.go
iFrostizz: I will probably just base my branch again on your `AddResult` instead of duplicating efforts

x/programs/runtime/import_program.go
iFrostizz: I have done it, let me know if that looks good. I think that the `[]byte{byte(ExecutionFailure)}` is looking pretty bad but your `AddProgram` branch might solve this.

x/programs/runtime/runtime.go
iFrostizz: And since the new once won't be matched with either of the unreachable or out of fuel, it would just default to `ExecutionFailure`

iFrostizz: I have done a conversion trial with `errors.As`. If that doesn't match as a trap error, then we just serialize to a `ExecutionFailure`.

x/programs/runtime/imports.go
iFrostizz: Removed this `InfaillibleFunction`, just using the `Functions` and returning `nil` for each error since the `InfaillibleFunction` was not bringing much benefits.

x/programs/runtime/import_program.go
```diff
@@ -49,17 +51,21 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
...
 				newInfo.Fuel = input.Fuel
 
-				result := r.CallProgramToResponse(
+				result, err := r.CallProgram(
 					context.Background(),
 					&newInfo)
+				if err != nil {
+
+					return Err[RawBytes, ErrorCode](ExecutionFailure), nil
+				}
```
iFrostizz: Don't we want the behaviour to match with program2program calls ? Here it's returning an `ExecutionFailure` in every case.

x/programs/runtime/import_program.go
```diff
@@ -35,12 +37,12 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
...
 		Name: "program",
 		HostFunctions: map[string]HostFunction{
-			"call_program": {FuelCost: callProgramCost, Function: Function[callProgramInput, Result[RawBytes, uint8]](func(callInfo *CallInfo, input callProgramInput) (Result[RawBytes, uint8], error) {
+			"call_program": {FuelCost: callProgramCost, Function: Function[callProgramInput, Result[RawBytes, ErrorCode]](func(callInfo *CallInfo, input callProgramInput) (Result[RawBytes, ErrorCode], error) {
 				newInfo := *callInfo
 
 				// make sure there is enough fuel in current store to give to the new call
 				if callInfo.RemainingFuel() < input.Fuel {
-					return Err([]byte{byte(ExecutionFailure)}), nil
+					return Err[RawBytes, ErrorCode](OutOfFuel), nil
```
iFrostizz, line 45: In this case we wanted to just return an `ExecutionFailure`, since it should be the client responsibility to make sure that they have enough gas to complete the call.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -81,11 +81,12 @@ impl<K> Program<K> {
...
         };
 
-        let args_bytes = borsh::to_vec(&args).map_err(|_| ExternalCallError::ExecutionFailure)?;
+
+        let args_bytes = borsh::to_vec(&args).expect("failed to serialize args");
 
         let bytes = unsafe { call_program(args_bytes.as_ptr(), args_bytes.len()) };
 
-        borsh::from_slice(&bytes).map_err(|_| ExternalCallError::ExecutionFailure)?
+        borsh::from_slice(&bytes).expect("failed to deserialize")
```
iFrostizz: I would revert this change, this way we could more easily bubble up serialization errors if needed.

x/programs/runtime/import_program.go
iFrostizz: edit: nice you've done it at the next commit

x/programs/runtime/import_program.go
iFrostizz: solved

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: done

x/programs/cmd/simulator/cmd/program.go
```diff
@@ -125,21 +118,16 @@ func programExecuteFunc(
...
-	output := [][]byte{programOutput}
+	result, err := rt.CallProgram(ctx, callInfo)
+	remainingFuel := callInfo.RemainingFuel()
 	if err != nil {
-		response := multilineOutput(output)
-		return nil, 0, fmt.Errorf("program execution failed: %s, err: %w", response, err)
+		if code, ok := runtime.ExtractErrorCode(err); ok {
+			return runtime.Err[runtime.RawBytes, runtime.ErrorCode](code), remainingFuel, nil
+		}
+		return runtime.Err[runtime.RawBytes, runtime.ErrorCode](runtime.ExecutionFailure), remainingFuel, err
```
iFrostizz, line 127: We used to waste all the allocated gas if it was returning an error, now the rest is remaining.

dboehm-avalabs (approved):

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -221,21 +222,19 @@ func (c *runCmd) runStepFunc(
...
 			return err
 		}
-
-		if len(response) > 1 {
-			return errors.New("multi response not supported")
-		}
-		res := response[0]
-		if res != nil {
-			resp.setResponse(res)
+		response, err := runtime.Serialize(output)
```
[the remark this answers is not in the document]
iFrostizz: That's because the `programExecuteFunc` function is now returning a `Result[T any, E any]`, it has to be turned into bytes to fit in the response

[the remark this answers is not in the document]
iFrostizz: Returned `[]byte` as you advised, it's much cleaner now and the `Serialize` functions are renamed to `serialize` which makes diffs smaller.

x/programs/runtime/import_program.go
```diff
@@ -42,6 +42,19 @@ func extractErrorCode(err error) (ErrorCode, bool) {
 	return 0, false
 }
 
+func (e ErrorCode) Error() string {
+	switch e {
+	case CallPanicked:
+		return "The call panicked"
+	case ExecutionFailure:
+		return "An unexpected error happened"
```
richardpringle, line 50: Shouldn't this be the catch-all instead of panicking? I don't think we want the runtime to panic like this.

x/programs/cmd/simulator/cmd/plan.go
```diff
@@ -245,21 +241,16 @@ func (c *runCmd) runStepFunc(
 			return err
 		}
 		// TODO: implement readonly for now just don't charge for gas
-		response, _, err := programExecuteFunc(ctx, c.log, db, programAddress, params[1:], method, math.MaxUint64)
-		if err != nil {
+		output, _ := programExecuteFunc(ctx, c.log, db, programAddress, params[1:], method, math.MaxUint64)
```
richardpringle, line 244: It's here that you should deal with an error return by `programExecuteFunc`.

x/programs/cmd/simulator/Cargo.toml
richardpringle, line 1: 👍

x/programs/cmd/simulator/src/lib.rs
richardpringle, line 1: 👍

x/programs/runtime/import_program_test.go
richardpringle, line 1: 👍

x/programs/runtime/import_state.go
richardpringle, line 1: 👍

x/programs/runtime/import_state_test.go
richardpringle, line 1: 👍

x/programs/runtime/imports.go
richardpringle, line 1: 👍

x/programs/runtime/program.go
richardpringle, line 1: 👍

x/programs/runtime/serialization.go
richardpringle, line 1: 👍

x/programs/runtime/serialization_test.go
richardpringle, line 1: 👍

x/programs/runtime/util_test.go
richardpringle, line 1: 👍

x/programs/rust/examples/counter-external/src/lib.rs
richardpringle, line 1: 👍

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle, line 1: 👍

richardpringle (changes requested):
Please leave all the 👍 on the files unresolved. I'm going to use them in my next round of review so I don't have to go over all the files again.

x/programs/cmd/simulator/cmd/program.go
```diff
@@ -125,21 +118,16 @@ func programExecuteFunc(
...
 
-	return output, callInfo.RemainingFuel(), err
-}
-
-func multilineOutput(resp [][]byte) (response string) {
-	for _, res := range resp {
-		response += string(res) + "\n"
-	}
-	return response
+	return runtime.Ok[runtime.RawBytes, runtime.ErrorCode](result), remainingFuel
```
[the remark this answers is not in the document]
iFrostizz: created a new `resultToOutput` function that does this

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: I forgot ! Thanks for that catch

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -224,4 +225,82 @@ mod tests {
             .unwrap();
         assert_eq!(value, 10);
     }
+
+    #[test]
+    fn insufficient_gas() {
```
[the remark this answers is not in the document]
iFrostizz: Completely removed that tests and added a test called `TestImportOutOfFuel` at `x/programs/runtime/import_program_test.go` which does this independent of the gas rules (it's calling another program with 0 gas).

x/programs/runtime/import_program.go
iFrostizz: I just removed this `Error() string` function as it was supposed to be used because there was some back and forth conversion between the `ErrorCode` and `error` but it got solved.

x/programs/cmd/simulator/cmd/plan.go
iFrostizz: Done by the add of the `resultToOutput` function which gets passed the output as well as the error.

richardpringle (changes requested):
One comment that requires addressing

x/programs/cmd/simulator/cmd/program.go
```diff
@@ -133,14 +129,14 @@ func programExecuteFunc(
 		FunctionName: function,
 		Params:       bytes,
 	}
-	programOutput, err := rt.CallProgram(ctx, callInfo)
-	output := [][]byte{programOutput}
+	result, err := rt.CallProgram(ctx, callInfo)
+	remainingFuel := callInfo.RemainingFuel()
```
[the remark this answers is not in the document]
iFrostizz: Right, I wanted to stick with the original changes but I was returning the remaining fuel in both case. It's now only returning it in the happy path.

dboehm-avalabs (approved):

x/programs/cmd/simulator/cmd/plan.go
richardpringle, line 1: 👍

x/programs/cmd/simulator/cmd/program.go
richardpringle, line 1: 👍

x/programs/runtime/import_program_test.go
richardpringle, line 1: 👍

x/programs/runtime/import_state.go
richardpringle, line 1: 👍

x/programs/runtime/import_state_test.go
richardpringle, line 1: 👍

x/programs/runtime/program.go
richardpringle, line 1: 👍

x/programs/runtime/serialization.go
richardpringle, line 1: 👍

x/programs/runtime/serialization_test.go
richardpringle, line 1: 👍

x/programs/runtime/util_test.go
richardpringle, line 1: 👍

x/programs/rust/wasmlanche-sdk/src/state.rs
richardpringle, line 1: 👍

richardpringle (approved):

PR #1047, opened 2024-06-12
Multisig example

iFrostizz:
TODO


iFrostizz:
Blocked by raw bytes return from `call_function`

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -85,9 +85,8 @@ pub enum Key {
 
 #[derive(Clone, Debug, PartialEq, Default, Serialize)]
 #[serde(rename_all = "camelCase")]
-#[non_exhaustive]
 pub struct TestContext {
-    program_id: Id,
+    pub program_id: Id,
```
iFrostizz, line 89: Without these changes, it was not possible to "autofill" a `TestContext` from another one. Pretty useful in the the case of the `voter_context`, see the multisig test `cast_votes` for more information.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -12,6 +12,7 @@ use thiserror::Error;
 /// Defer deserialization from bytes
 /// <div class="warning">It is possible that this type performs multiple allocations during deserialization. It should be used sparingly.</div>
 #[cfg_attr(feature = "debug", derive(Debug))]
+#[derive(BorshSerialize)]
```
iFrostizz, line 15: Necessary to be able to return a `DeferDeserialize` from the `execute` function. Let me know if it's unneeded, we could just have returned a `Vec<u8>` and casted to a `DeferDeserialize` in the test directly.

x/programs/rust/wasmlanche-sdk/src/types.rs
```diff
@@ -9,7 +9,7 @@ pub type Gas = u64;
 
 /// A struct that enforces a fixed length of 32 bytes which represents an address.
 #[cfg_attr(feature = "debug", derive(Debug))]
-#[derive(Clone, Copy, PartialEq, Eq, BorshSerialize, BorshDeserialize, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, BorshSerialize, BorshDeserialize, Hash, PartialOrd, Ord)]
```
iFrostizz, line 12: This is to be able to write an `Address` in an `HashSet` just like it is for the `voters` field.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: I will remove it because it really looks unnecessary

iFrostizz: I was wondering if I should remove it, it should probably be left like this at the end because casting the return bytes to a `DeferDeserialize` is necessary, otherwise the deserialization expects a length prefix if we wanted bytes such as `Vec<u8>`. Further, the `BorshSerialize` saves us from a manual `into_inner` function that would just be
```rust
impl DeferDeserialize {
    ...

    pub fn into_inner(self) -> Vec<u8> {
        self.0
    }
}
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+    proposal_at(context.program(), proposal_id)
+}
+
+#[public]
+pub fn pub_proposal_id(context: Context<StateKeys>) -> u32 {
+    proposal_id(context.program())
+}
+
+#[public]
+pub fn pub_quorum_reached(
```
richardpringle, line 142:
```suggestion
pub fn proposal_at(context: Context<StateKeys>, proposal_id: u32) -> Option<Proposal> {
    proposal_at(context.program(), proposal_id)
}

#[public]
pub fn proposal_id(context: Context<StateKeys>) -> u32 {
    proposal_id(context.program())
}

#[public]
pub fn quorum_reached(
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+}
+
+pub fn proposal_at(program: &Program<StateKeys>, proposal_id: u32) -> Option<Proposal> {
+    program
+        .state()
+        .get(StateKeys::Proposals(proposal_id))
+        .expect("state corrupt")
+}
+
+pub fn proposal_id(program: &Program<StateKeys>) -> u32 {
```
richardpringle, line 158:
```suggestion
pub fn last_proposal_id(program: &Program<StateKeys>) -> u32 {
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+#[cfg(test)]
+mod tests {
+    use super::ProposalError;
+    use simulator::{Endpoint, Key, Step, TestContext};
+    use wasmlanche_sdk::DeferDeserialize;
+
+    const PROGRAM_PATH: &str = env!("PROGRAM_PATH");
+
+    #[test]
+    fn can_propose() {
```
richardpringle, line 179: You're testing the IDs functionality here... "can propose" is storing state, you don't need to test whether or not you can store state.

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            .unwrap()
+            .result
+            .response()
+            .unwrap();
+
+        assert_eq!(new_pid, 1);
+    }
+
+    #[test]
+    fn cast_votes() {
```
richardpringle, line 225:
```suggestion
    fn cannot_double_vote() {
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+    let last_id = proposal_id(program);
+
+    program
+        .state()
+        .store_by_key(
+            StateKeys::Proposals(last_id),
+            &Proposal {
+                to,
+                method,
+                data,
```
richardpringle, line 55: `data` is args? Maybe name it `args` or `arguments`?

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+                value,
```
richardpringle, line 56: What is `value`?

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            .unwrap()
+            .result
+            .response::<Result<(), ProposalError>>()
+            .unwrap();
+
+        assert!(matches!(res, Err(ProposalError::AlreadyVoted)));
+    }
+
+    #[test]
+    fn vote_execution() {
```
richardpringle, line 294:
```suggestion
    fn execute() {
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            })
+            .unwrap()
+            .result
+            .response::<Result<DeferDeserialize, ProposalError>>()
+            .unwrap()
+            .unwrap();
+
+        let bytes: Vec<u8> = res.deserialize().unwrap();
+
+        assert_eq!(borsh::from_slice::<u32>(&bytes).unwrap(), 1);
```
richardpringle, line 390: This is wrong, why do you need the extra deserialize?

richardpringle (changes requested):
A "multisig" shouldn't just allow anyone and everyone to vote. The voters should be specified in advance

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -114,6 +113,7 @@ pub(crate) struct SimulatorTestContext {
 #[derive(Clone, Debug, PartialEq)]
 pub enum Param {
     U64(u64),
+    U32(u32),
```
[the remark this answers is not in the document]
iFrostizz: This is true, but we still need strong typing on the Rust side. Otherwise we could as well borsh serialize everything but that gives fewer guarantees about the type and is harder to debug.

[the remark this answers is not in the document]
iFrostizz: To clarify, this PR https://github.com/ava-labs/hypersdk/pull/1099 makes it now unnecessary to add the types on the Go side as well, since any unregistered type which doesn't need special treatment is just passed as "bytes".

x/programs/rust/wasmlanche-sdk/src/types.rs
iFrostizz: Oh right, thanks. This is to be able to have the `voters` field in the `Proposer` struct which implements `BorshSerialize`, It's a requirement because maps are sorted by key when serialized.
I think that I will just store the `voters` in a separate state key.

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+        proposal.yea += 1;
+    } else {
+        proposal.nay += 1;
+    }
+    proposal.voters.insert(actor);
+
+    program
+        .state()
+        .store_by_key(StateKeys::Proposals(id), &proposal)
+        .expect("state corrupt");
```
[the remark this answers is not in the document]
iFrostizz: Good point. Yays/Nays are now stored in a separate state key.

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: That won't work because these function names are conflicting with the non-public functions.

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: Renamed!

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: It's the amount of native tokens sent along with the execution.
I wanted to keep it simple, having some kind of value contribution will make it more complex since we need to handle the value amount (with a floor to ensure the account always has enough funds), and refunds if the execution fails.

iFrostizz: So the value is just passed by the actor that executes the proposal. It's probably a bit unfair, but it was a bit better than just passing 0.

iFrostizz:
Currently blocked by address passing https://github.com/ava-labs/hypersdk/pull/1121

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+                params: vec![voter_context.clone().into(), pid.into(), 1000000u64.into()],
+            })
+            .unwrap()
+            .result
+            .response::<Result<DeferDeserialize, ProposalError>>()
+            .unwrap()
+            .unwrap();
+
+        let id_serialized: Vec<u8> = res.deserialize().unwrap();
+        let id: u32 = borsh::from_slice(&id_serialized).unwrap();
```
iFrostizz, line 540: I know that it doesn't look the best but that's because of the `BorshSerialize` that is annotating the `DeferDeserialize`. It should probably also serialize as a white-box.

iFrostizz:
```rust
impl BorshSerialize for DeferDeserialize {
    fn serialize<W: std::io::prelude::Write>(&self, writer: &mut W) -> std::io::Result<()> {
        writer.write_all(&self.0)?;
        Ok(())
    }
}
```

iFrostizz: Opened https://github.com/ava-labs/hypersdk/issues/1123

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+fn _last_proposal_id(context: &Context<StateKeys>) -> u64 {
+    context
+        .get(StateKeys::LastProposalId)
+        .expect("state corrupt")
+        .unwrap_or_default()
+}
+
+fn _quorum_reached(max_votes: u64, yeas: u64, nays: u64) -> bool {
+    yeas + nays >= MIN_VOTES && yeas > nays && yeas >= (max_votes + 1) / 2
+}
```
richardpringle, line 214: Please put all of these (and any others) in a `mod internal` (same file) and remove the prefixes.

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+
+#[public]
+pub fn quorum_reached(
+    context: Context<StateKeys>,
+    proposal_id: u64,
+) -> Result<bool, ProposalError> {
+    let proposal = _proposal_at(&context, proposal_id).ok_or(ProposalError::InexistentProposal)?;
+    let yeas = context.get(StateKeys::Yeas(proposal_id)).unwrap().unwrap();
+    let nays = context.get(StateKeys::Nays(proposal_id)).unwrap().unwrap();
```
richardpringle, line 194: You should be using an `expect` with a `unwrap_or_default` here (and elsewhere)

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+    let mut proposal =
+        _proposal_at(&context, proposal_id).ok_or(ProposalError::InexistentProposal)?;
+
+    if proposal.executed {
+        return Err(ProposalError::AlreadyExecuted);
+    }
+
+    let yeas = context.get(StateKeys::Yeas(proposal_id)).unwrap().unwrap();
+    let nays = context.get(StateKeys::Nays(proposal_id)).unwrap().unwrap();
+    if !_quorum_reached(proposal.voters_len, yeas, nays) {
```
richardpringle, line 165:
```suggestion
    let nays = context.get(StateKeys::Nays(proposal_id)).unwrap().unwrap();

    if !_quorum_reached(proposal.voters_len, yeas, nays) {
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+}
+
+fn _quorum_reached(max_votes: u64, yeas: u64, nays: u64) -> bool {
+    yeas + nays >= MIN_VOTES && yeas > nays && yeas >= (max_votes + 1) / 2
+}
+
+fn _is_not_duplicate(voters: &[Address]) -> bool {
+    if voters.is_empty() {
+        panic!("there should be at least one voter");
+    }
```
richardpringle, line 219: You're coupling logic here. This check should happen somewhere else

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+            )
+            .unwrap();
+    } else {
+        context
+            .store_by_key(
+                StateKeys::Nays(id),
+                &(context.get::<u64>(StateKeys::Nays(id)).unwrap().unwrap() + 1),
+            )
+            .unwrap();
+    }
```
richardpringle, line 135: This isn't very readable. Please store the result from `get` in a variable and use that. 

Please don't use `unwrap` either

x/programs/cmd/simulator/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: I removed the use of this type since it was not *necessary* but it might make sense to introduce it back later.

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+    voters_addr: Vec<Address>,
+    to: Program,
+    method: String,
+    args: Vec<u8>,
+    value: u64,
+) -> u64 {
+    let voters_len = voters_addr.len().try_into().expect("too much voters");
+    assert!(voters_len >= MIN_VOTES);
+    assert!(_is_not_duplicate(&voters_addr));
+    let mut voters: Vec<_> = voters_addr
```
richardpringle, line 55:
```suggestion
    assert!(_is_not_duplicate(&voters_addr));

    let mut voters: Vec<_> = voters_addr
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+        .iter()
+        .map(|address| Voter {
+            address: *address,
+            voted: false,
+        })
```
richardpringle, line 60: nit
```suggestion
        .iter()
        .copied()
        .map(|address| Voter {
            address,
            voted: false,
        })
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+        .collect();
+    let i = _ensure_voter(&context.actor(), &voters).unwrap();
```
richardpringle, line 62:
```suggestion
        .collect();

    let i = _ensure_voter(&context.actor(), &voters).unwrap();
```

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+
+    let actor = context.actor();
+    let voters: Vec<_> = (0..proposal.voters_len)
+        .map(|i| {
+            context
+                .get(StateKeys::Voter(id, i as usize))
+                .unwrap()
+                .unwrap()
+        })
+        .collect();
```
richardpringle, line 117: This is a very inefficient way pull out all the voters, why do you need to do this? 

If you have a key `Vote(Address)`, you check if the data is there? If it is, the person can't vote again, you return an error. If it's not there, increase the yea or nay count, store the vote, done.

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+                        String::new().into(),
+                        Vec::new().into(),
+                        0u64.into(),
+                    ],
+                })
+                .unwrap()
+                .result
+                .response::<()>(),
+            Err(StepResponseError::ExternalCall(_))
+        ));
```
richardpringle, line 294: Please use variables

richardpringle (changes requested):

samliok (commented):
overall this program is a really cool example 🚀! 

One note: important to have thorough commenting on examples, since this is prob the first place devs will look. I would add helpful comments to things like public functions, statekeys, struct fields, etc...

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+}
+
+#[state_keys]
+pub enum StateKeys {
+    LastProposalId,
+    Proposals(u64),
+    Voter(u64, usize),
+    Yeas(u64),
+    Nays(u64),
+}
```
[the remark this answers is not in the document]
iFrostizz: Done !

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+
+#[public]
+pub fn propose(
+    context: Context<StateKeys>,
+    voters_addr: Vec<Address>,
+    to: Program,
+    method: String,
+    args: Vec<u8>,
+    value: u64,
+) -> u64 {
```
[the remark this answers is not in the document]
iFrostizz: That's fine I will do it here

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: Thanks, namespaced these functions in the `internal` module.

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: Removed that check as it was already done in the only function calling this one. `propose`.

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: Thanks, created a new variable and provided some better panic messages with `expect`.

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: Thanks, The voter now passes a `voter_id` where they expects to be in the set and the program will check if it is here and then if it has already voted.

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: Thanks, Created a variable that stored the response and now matches on this one.

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+    let voters: Vec<_> = (0..proposal.voters_len)
+        .map(|i| {
+            context
+                .get(StateKeys::Voter(id, i as usize))
+                .unwrap()
+                .unwrap()
+        })
+        .collect();
+
+    let i = _ensure_voter(&actor, &voters)?;
```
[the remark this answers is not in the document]
iFrostizz: Renamed to `index`.

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+                return false;
+            }
+        }
+    }
+
+    true
+}
+
+/// Check if the passed actor is part of the voters and return its index on success
+fn _ensure_voter(actor: &Address, voters: &[Voter]) -> Result<u64, ProposalError> {
```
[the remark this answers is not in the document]
iFrostizz: Put all of the internal functions in the `internal` module instead. This `ensure_voter` function was removed.

x/programs/rust/examples/multisig/src/lib.rs
iFrostizz: I disagree with the use of a `unwrap_or_default` here, because if the "default" case is returned then an important invariant is broken. I have instead used two `expect` since this function should be called after a proposal has been created. The state should be correctly initialized after the call to `propose`.

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
...
+    AlreadyVoted,
+    InexistentProposal,
+    QuorumNotReached,
+    NotVoter,
+    ExecutionFailed(ExternalCallError),
+}
+
+#[state_keys]
+pub enum StateKeys {
+    LastProposalId,
```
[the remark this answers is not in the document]
iFrostizz: (this key was renamed to `NextProposalId` instead)
I'm not sure if we could have such a host function, the data may not be stored in a contiguous space of memory and it may be hard to have such an information. What we may want though is a `Vec` library program (if we are able to write stateful libraries in the future) that stores this information. This is a good remark !

x/programs/rust/examples/multisig/src/lib.rs
```diff
@@ -0,0 +1,495 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use wasmlanche_sdk::{
+    public, state_keys, Address, Context, DeferDeserialize, ExternalCallError, Program,
+};
+
+pub const MIN_VOTES: u64 = 2;
+
+#[derive(Debug, BorshSerialize, BorshDeserialize)]
+pub struct Proposal {
```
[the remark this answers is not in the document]
iFrostizz: The results are stored in the `StatelessBlock` but we should not rely on this information being stored long term, so I will just apply your suggestion. Thanks !

x/programs/rust/examples/multisig/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: The reason we might want this `Vec` library is that storing a `Vec` as a state key is going to make it costly to pull it from storage because we can only get the full vec.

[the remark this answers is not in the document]
samliok: yea, a wayyyys back the initial thought was to have host functions that would reduce the amount of data pulled into WASM(stuff like getting values in a map, vec, etc...). we don't have host functions for maps anymore because of `StateKeys` but vecs are a little different. 

under the hood a `Vecs` could be stored/tracked differently, and come with additional get functions like `get_index`. Could also do something similar to [solidity](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays), and just handle them i differently under the hood in get/set methods.

iFrostizz:
`DeferDeserialize` cannot be used in the storage https://github.com/ava-labs/hypersdk/pull/1047/files#diff-a262134f908ac01d356baed6d18b02ea08af43560006cd3255dff48394e6cdebR12 because the deserialization implementation by nature exhausts all bytes https://github.com/ava-labs/hypersdk/blob/c348296b6324ecbc998edc3a058f94ccdc33027a/x/programs/rust/wasmlanche-sdk/src/program.rs#L31 meaning that when it is deserialized (after calling `get` let's say), then it's going to read all bytes expected from the `DeferDeserialize` and will read all bytes of the subsequent fields too. Solutions:
1. Add a `AsRef<[u8]` implementation
2. Write the `DeferDeserialize` at the end of all fields

It's a security issue if this `DeferDeserialize` cannot be deserialized in a struct and may create some DOS vectors.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -33,6 +33,12 @@ impl BorshDeserialize for DeferDeserialize {
     }
 }
 
+impl Into<Vec<u8>> for DeferDeserialize {
```
iFrostizz, line 36: The `Into` was deliberately implemented because we only want the `Vec<u8>` -> `DeferDeserialize` to happen with the help of `borsh`.

richardpringle: This doesn't make sense. You should be implementing `From<DeferDeserialize>`

iFrostizz: irrelevant because it has been removed.

x/programs/rust/examples/multisig/src/lib.rs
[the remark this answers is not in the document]
iFrostizz: That's very interesting! I don't know if we really want this at the vm level though, it feels more like a program-level thing, the vm may not need to be aware of that, but I'm happy to try both !!

iFrostizz:
superseded by https://github.com/ava-labs/hypersdk/pull/1581

PR #1048, opened 2024-06-12
Make `call_function` take serialized arguments

richardpringle:
This makes the generic use of `call_function` simpler in that you can call it with pre-serialized arguments. Since we now have the `bindings` feature, external calls can effectively be made manually with `call_function` where the author needs to check that the calling conventions work as expected, or automatically with the bindings feature.

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/program.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/program.rs
@@ -21,7 +21,8 @@ impl<K> BorshSerialize for Program<K> {
             account,
             state_cache: _,
         } = self;
-        BorshSerialize::serialize(account, writer)
+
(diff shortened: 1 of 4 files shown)


x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -47,10 +48,10 @@ impl<K> Program<K> {
     /// Returns a [`StateError`] if the call fails.
     /// # Safety
     /// The caller must ensure that `function_name` + `args` point to valid memory locations.
```
richardpringle, line 50: 🤔 this is out of date. Will create an issue to go and make sure all the docs are up to date and maybe add example usage everywhere.

iFrostizz (approved):

Issue #1045, opened 2024-06-12
Allow any borsh-serializable as state key unnamed field values


iFrostizz:
Currently, the unnamed state key fields are expected to implement `AsRef` as well as having the `len()` function.
This removes a whole class of types such as primitives.
We could instead require the type to be `BorshSerialize`

```diff
diff --git a/x/programs/rust/sdk-macros/src/lib.rs b/x/programs/rust/sdk-macros/src/lib.rs
index eb73cbf0..b53aa9f0 100644
--- a/x/programs/rust/sdk-macros/src/lib.rs
+++ b/x/programs/rust/sdk-macros/src/lib.rs
@@ -293,17 +293,15 @@ pub fn state_keys(_attr: TokenStream, item: TokenStream) -> TokenStream {
                         .enumerate()
                         .map(|(i, field)| Ident::new(&format!("field_{i}"), field.span()));
                     let fields_2 = fields.clone();
-                    let fields_3 = fields.clone();

                     Ok(quote! {
                         Self::#variant_ident(#(#fields),*) => {
-                            let len = 1 + #(#fields_2.as_ref().len())+*;
+                            let as_bytes = borsh::to_vec(&(#(#fields_2),*))?;
+                            let len = 1 + as_bytes.len();
                             let len = u32::try_from(len).map_err(|_| std::io::ErrorKind::InvalidData)?;
                             writer.write_all(&len.to_le_bytes())?;
                             writer.write_all(&[#idx])?;
-                            #(
-                                writer.write_all(#fields_3.as_ref())?;
-                            )*
+                            writer.write_all(&as_bytes)?;
                         }
                     })
                 }
```

richardpringle:
Yeah, this works. It would probably be better to put the `idx` in the tuple though and not worry about length. To be honest, this is a one-way serialization so we actually don't care about length at all

PR #1050, opened 2024-06-12
Allow for any borsh-serializable key for unnamed field

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1045

We used to require the field that is in the state key to implement `AsRef` as well as having a `len()` function. The only requirement now is to implement `BorshSerialize`.

### DIFF
--- a/x/programs/rust/sdk-macros/src/lib.rs
+++ b/x/programs/rust/sdk-macros/src/lib.rs
@@ -297,12 +297,7 @@ pub fn state_keys(_attr: TokenStream, item: TokenStream) -> TokenStream {
 
                     Ok(quote! {
                         Self::#variant_ident(#(#fields),*) => {
-                            let as_bytes = borsh::to_vec(&(#(#fields_2),*))?;
-                            let len = 1 + as_bytes.len();
-                            let len = u32::try_from(len).map_err(|_| std::io::ErrorKind::InvalidData)?;
-                            writer.write_all(&len.to_le_bytes())?;
-                            writer.write_all(&[#idx])?;
-                            writer.write_all(&as_bytes)?;
+                            borsh::to_vec(&(#idx, #(#fields_2),*))?.serialize(writer);
                         }
                     })
                 }
--- a/x/programs/rust/sdk-macros/src/lib.rs
+++ b/x/programs/rust/sdk-macros/src/lib.rs
@@ -294,17 +294,15 @@ pub fn state_keys(_attr: TokenStream, item: TokenStream) -> TokenStream {
                         .enumerate()
                         .map(|(i, field)| Ident::new(&format!("field_{i}"), field.span()));
                     let fields_2 = fields.clone();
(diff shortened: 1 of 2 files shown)


x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -293,17 +293,15 @@ pub fn state_keys(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
-                            let len = 1 + #(#fields_2.as_ref().len())+*;
+                            let as_bytes = borsh::to_vec(&(#(#fields_2),*))?;
+                            let len = 1 + as_bytes.len();
                             let len = u32::try_from(len).map_err(|_| std::io::ErrorKind::InvalidData)?;
                             writer.write_all(&len.to_le_bytes())?;
                             writer.write_all(&[#idx])?;
-                            #(
-                                writer.write_all(#fields_3.as_ref())?;
-                            )*
+                            writer.write_all(&as_bytes)?;
```
richardpringle, line 304: You can simplify this whole thing down to a single line:
```rust
borsh::to_vec(&(#idx, #(#fields_2),*))?.serialize(writer);
```

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -149,8 +149,8 @@ impl<'a, K: Key> State<'a, K> {
         }
 
         #[derive(BorshSerialize)]
-        struct PutArgs<Key> {
-            key: Key,
+        struct PutArgs<K> {
+            key: K,
```
richardpringle, line 153: This change isn't necessary

richardpringle (changes requested):

x/programs/rust/sdk-macros/src/lib.rs
iFrostizz: I don't think that works because we need the length in order to borsh deserialize it on the host side.
If we keep it, then we are going to borsh serialize the bytes twice which is a loss of performance and borsh serialize the `#idx` which is not really needed because it's a one-way conversion

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: I found that it was more correct than repeating the trait, since we probably don't want to have a different key in the `PutArgs` and in the actual state, I reverted it though

x/programs/rust/sdk-macros/src/lib.rs
richardpringle: You have the length: https://docs.rs/borsh/latest/src/borsh/ser/mod.rs.html#254-265

iFrostizz: Thanks, yeah that makes sense!

dboehm-avalabs (approved):

richardpringle (approved):

PR #1052, opened 2024-06-12
[x/programs] Add result type

dboehm-avalabs:


x/programs/runtime/import_program.go
```diff
@@ -46,16 +46,16 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
...
 
 				// subtract the fuel used during this call from the calling program
 				remainingFuel := newInfo.RemainingFuel()
 				if err := callInfo.ConsumeFuel(input.Fuel - remainingFuel); err != nil {
-					return nil, err
+					return Err[[]byte, uint8](0), err
 				}
 
-				return result, nil
+				return Ok[[]byte, uint8](result), nil
```
richardpringle, line 58: Looks like Rust haha

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -52,14 +52,14 @@ impl<K> Program<K> {
         function_name: &str,
         args: ArgType,
         max_units: Gas,
-    ) -> Result<T, StateError> {
+    ) -> Result<T, u8> {
```
richardpringle, line 55: This is where @iFrostizz's `ExternalCallResult` should go instead of the u8.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -52,14 +52,14 @@ impl<K> Program<K> {
...
         #[link(wasm_import_module = "program")]
         extern "C" {
             #[link_name = "call_program"]
             fn call_program(ptr: *const u8, len: usize) -> HostPtr;
         }
 
-        let args_ptr = borsh::to_vec(&args).map_err(|_| StateError::Serialization)?;
+        let args_ptr = borsh::to_vec(&args).expect("failed to serialize args");
```
richardpringle, line 62: serialization no longer happens here on `main`

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -68,11 +68,16 @@ impl<K> Program<K> {
             max_units,
         };
 
-        let args_bytes = borsh::to_vec(&args).map_err(|_| StateError::Serialization)?;
+        let args_bytes = borsh::to_vec(&args).expect("failed to serialize args");
```
richardpringle, line 71: Args will always serialize so this is fine.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -68,11 +68,16 @@ impl<K> Program<K> {
...
 
         let bytes = unsafe { call_program(args_bytes.as_ptr(), args_bytes.len()) };
 
-        borsh::from_slice(&bytes).map_err(|_| StateError::Deserialization)
+        crate::dbg!({
+            let bytes: &[u8] = &bytes;
+            bytes
+        });
+
```
richardpringle, line 79: Who put this here?
```suggestion
```
😜

x/programs/runtime/import_program.go
dboehm-avalabs: That's the goal, right?

x/programs/rust/wasmlanche-sdk/src/program.rs
dboehm-avalabs: fixed

x/programs/runtime/import_program.go
```diff
@@ -46,16 +70,16 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
...
+					return Err[RawBytes, ErrorCode](ExecutionFailure), err
 				}
 
-				// subtract the fuel used during this call from the calling program
-				remainingFuel := newInfo.RemainingFuel()
-				if err := callInfo.ConsumeFuel(input.Fuel - remainingFuel); err != nil {
-					return nil, err
-				}
+				// return any remaining fuel to the calling program
+				callInfo.AddFuel(newInfo.RemainingFuel())
```
iFrostizz, line 80: I see that `RemainingFuel` is calculated from the amount of consumed fuel of the instance.
Since the reuse of instance, are we resetting the fuel metering ?

x/programs/runtime/import_program.go
```diff
@@ -28,12 +53,11 @@ func NewProgramModule(r *WasmRuntime) *ImportModule {
...
 		HostFunctions: map[string]HostFunction{
-			"call_program": {FuelCost: callProgramCost, Function: Function[callProgramInput, RawBytes](func(callInfo *CallInfo, input callProgramInput) (RawBytes, error) {
+			"call_program": {FuelCost: callProgramCost, Function: Function[callProgramInput, Result[RawBytes, ErrorCode]](func(callInfo *CallInfo, input callProgramInput) (Result[RawBytes, ErrorCode], error) {
 				newInfo := *callInfo
 
-				// make sure there is enough fuel in current store to give to the new call
-				if callInfo.RemainingFuel() < input.Fuel {
-					return nil, errors.New("remaining fuel is less than requested fuel")
+				if err := callInfo.ConsumeFuel(input.Fuel); err != nil {
+					return Err[RawBytes, ErrorCode](OutOfFuel), nil
```
iFrostizz, line 60: It's so nice that we now only have one place where the fuel consumption is calculated

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: They could be user-defined, I would tend to think that it would be better to return an `ExecutionFailure` here, a serialization error could be added in the future if needed

iFrostizz (approved):

x/programs/runtime/import_program.go
dboehm-avalabs: we aren't currently reusing instances

Issue #685, opened 2024-01-17
[chain] refactor `Base` into `Transaction`


wlawt:
For example, both `Base` and `Transaction` have an `Execute` function. However, `Base.Execute` is called within `Transaction.PreExecute`. Refactoring would make this relationship clearer

PR #1054, opened 2024-06-13
[chain] Refactor `Base` into `Transaction`

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/685

Making the `Base` part of the `Transaction` so that their relationship is clearer.


chain/base.go
```diff
@@ -1,68 +0,0 @@
...
-	case b.ChainID != chainID:
-		return ErrInvalidChainID
-	default:
-		return nil
-	}
-}
-
-func (*Base) Size() int {
-	return BaseSize
-}
```
iFrostizz, line 50: I noticed that `Size()`, even if defined, didn't seemed to be used. So I just deleted the whole file because all of the other methods have been inlined into `Transaction`.

chain/transaction.go
```diff
@@ -251,8 +259,16 @@ func (t *Transaction) PreExecute(
 	im state.Immutable,
 	timestamp int64,
 ) error {
-	if err := t.Base.Execute(r.ChainID(), r, timestamp); err != nil {
-		return err
+	switch {
+	case t.Timestamp%consts.MillisecondsPerSecond != 0:
+		// TODO: make this modulus configurable
```
iFrostizz, line 264: I opened an issue to keep track of this

cli/spam.go
```diff
@@ -471,7 +471,7 @@ type timeModifier struct {
 	Timestamp int64
 }
 
-func (t *timeModifier) Base(b *chain.Base) {
+func (t *timeModifier) Base(b *chain.Transaction) {
 	b.Timestamp = t.Timestamp
 }
```
iFrostizz, line 476: Do we really want the modifier to have access to the full `Transaction` ?

rpc/jsonrpc_client.go
```diff
@@ -157,20 +157,16 @@ func (cli *JSONRPCClient) GenerateTransactionManual(
 	// Construct transaction
 	now := time.Now().UnixMilli()
 	rules := parser.Rules(now)
-	base := &chain.Base{
-		Timestamp: utils.UnixRMilli(now, rules.GetValidityWindow()),
-		ChainID:   rules.ChainID(),
-		MaxFee:    maxFee,
-	}
+	timestamp := utils.UnixRMilli(now, rules.GetValidityWindow())
+	tx := chain.NewTx(timestamp, rules.ChainID(), maxFee, actions)
```
aaronbuchwald, line 161: Constructing the transaction here and then modifying it in the loop below may incorrectly mutate the transaction.

Previously, each modifier ran on `base` before it was passed into the transaction, so that it was constructed correctly with the final value.

Now that the modifiers are run after the transaction has been constructed, it will mutate the field without updating the private fields that depend on it.

See `Digest()` as an example: https://github.com/ava-labs/hypersdk/blob/main/chain/transaction.go#L49

It looks like the required order of operations is to construct the transaction and then call Digest/Sign.

iFrostizz: Nice catch, thank you !
I'm wondering if the fix would be to give a reference to every fields that used to belong to base instead of the full transaction to avoid footguns.

iFrostizz: Fixed in https://github.com/ava-labs/hypersdk/pull/1054/commits/92b1b776a1de50c38d82a1150ca2bd3ee0744dc8
We got rid of the Base but the 3 fields are always hanging together loosely, not sure if it's desirable

cli/spam.go
```diff
@@ -471,8 +471,8 @@ type timeModifier struct {
 	Timestamp int64
 }
 
-func (t *timeModifier) Base(b *chain.Base) {
-	b.Timestamp = t.Timestamp
+func (t *timeModifier) Base(timestamp int64, _ ids.ID, _ uint64) {
+	t.Timestamp = timestamp
```
aaronbuchwald, line 475: It looks like this will modify the timestamp of `t` instead of the value that's being passed in. This seems to be opposite the intent of the original code.

iFrostizz: Oops.... Thank you for this !

iFrostizz: Fixed!

aaronbuchwald:
I don't think we want to use the modifier as it's used in this PR long term. Closing in favor of a different tx/action refactor later.

PR #1057, opened 2024-06-13
Remove host `delete` and `put` functions

iFrostizz:
Rename the `put_many` function to `put` and make it handle deletes if the value is empty
Addresses https://github.com/ava-labs/hypersdk/pull/1029#discussion_r1633309815

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/state.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/state.rs
@@ -158,7 +158,7 @@ impl<'a, K: Key> State<'a, K> {
             key,
             value: Vec::new(),
         };
-        let args_bytes = borsh::to_vec(&[args]).map_err(|_| StateError::Serialization)?;
+        let args_bytes = borsh::to_vec(&vec![args]).map_err(|_| StateError::Serialization)?;
 
         unsafe { put(args_bytes.as_ptr(), args_bytes.len()) };
 
--- a/x/programs/rust/wasmlanche-sdk/src/state.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/state.rs
@@ -158,7 +158,8 @@ impl<'a, K: Key> State<'a, K> {
             key,
             value: Vec::new(),
         };
-        let args_bytes = borsh::to_vec(&vec![args]).map_err(|_| StateError::Serialization)?;
+        let args_bytes =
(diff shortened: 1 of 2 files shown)


x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -99,6 +99,10 @@ impl<'a, K: Key> State<'a, K> {
         let mut cache = self.cache.borrow_mut();
 
         let val_bytes = if let Some(val) = cache.get(&key) {
+            if val.is_empty() {
+                return Ok(None);
+            }
```
richardpringle, line 104: 🤔 we should probably make sure empty `Vec`s don't get inserted. Can you create an issue for that after you merge this?

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -124,34 +128,41 @@ impl<'a, K: Key> State<'a, K> {
...
         // we should actually cache deletes as well
         // to avoid cache misses after delete
-        self.cache.borrow_mut().remove(&key);
-
-        let args_bytes = borsh::to_vec(&key).map_err(|_| StateError::Serialization)?;
+        let args = PutArgs {
+            key,
+            value: Vec::new(),
+        };
+        let args_bytes = borsh::to_vec(&vec![args]).map_err(|_| StateError::Serialization)?;
```
richardpringle, line 153:
```suggestion
        let args_bytes = borsh::to_vec(&[args]).map_err(|_| StateError::Serialization)?;
```

richardpringle (changes requested):
tiny change

richardpringle (approved):

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: added `as_slice` so it wasn't treated as a fixed length array by Borsh

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: I updated the code so that it checks in the `store` function if the serialized data is empty, thanks for that suggestion!

richardpringle (approved):

PR #1058, opened 2024-06-13
[x/programs] Add CallContext to programs to allow setting of default values

dboehm-avalabs:
Add the ability to construct callinfo using a builder style to make it easier to make a sequence of calls that share a lot of the same call info.

### DIFF
--- a/x/programs/cmd/simulator/Cargo.toml
+++ b/x/programs/cmd/simulator/Cargo.toml
@@ -10,3 +10,4 @@ serde = { version = "1.0.203", features = ["derive"] }
 serde_json = "1.0.117"
 thiserror = { workspace = true }
 borsh = { workspace = true }
+wasmlanche-sdk = { workspace = true }
--- a/x/programs/cmd/simulator/cmd/plan.go
+++ b/x/programs/cmd/simulator/cmd/plan.go
@@ -23,6 +23,7 @@ import (
 	"github.com/ava-labs/hypersdk/codec"
 	"github.com/ava-labs/hypersdk/crypto/ed25519"
 	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/x/programs/runtime"
 )
 
 var _ Cmd = (*runCmd)(nil)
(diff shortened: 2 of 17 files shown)


iFrostizz (approved):

x/programs/runtime/call_context.go
```diff
@@ -0,0 +1,88 @@
...
+	defaults := reflect.ValueOf(c.defaultCallInfo)
+	for i := 0; i < defaults.NumField(); i++ {
+		defaultField := defaults.Field(i)
+		if !defaultField.IsZero() {
+			resultField := resultInfo.Elem().Field(i)
+			if !resultField.IsZero() {
+				return nil, fmt.Errorf("%w %s", errCannotOverwrite, callInfoTypeInfo.Field(i).Name)
+			}
+			resultField.Set(defaultField)
+		}
```
iFrostizz, line 37: Should we not set a "default" value if the default has not been set ?
For instance, that would allow us to set a `CallContext` with a default `FunctionName` and `Fuel` with an `Actor` that *may* be overridden.
If it's not set, I believe that it would be undefined behaviour because then the non-default fields are going to be `nil`.

iFrostizz: I guess I would need to see how it would integrate into the `programExecuteFunc` function

x/programs/runtime/call_context.go
```diff
@@ -0,0 +1,88 @@
...
+type CallContext struct {
+	r               *WasmRuntime
+	defaultCallInfo CallInfo
+}
+
+func (c CallContext) createCallInfo(callInfo *CallInfo) (*CallInfo, error) {
+	newCallInfo := *callInfo
+	resultInfo := reflect.ValueOf(&newCallInfo)
+	callInfoTypeInfo := reflect.TypeOf(CallInfo{})
+	defaults := reflect.ValueOf(c.defaultCallInfo)
```
iFrostizz, line 28: What is the reason why you create a new `callInfoTypeInfo` here, can't we use `defaults` ?

x/programs/runtime/runtime.go
```diff
@@ -58,6 +58,10 @@ func NewRuntime(
 	return runtime
 }
 
+func (r *WasmRuntime) WithDefaults(callInfo *CallInfo) CallContext {
```
iFrostizz, line 61: Wondering if it would be a good practice to add a `nil` check here for the pointer

iFrostizz (approved):

iFrostizz (approved):

PR #1059, opened 2024-06-13
Implement cache deletes

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/867, Closes https://github.com/ava-labs/hypersdk/issues/868

Rewrite of https://github.com/ava-labs/hypersdk/pull/1029
Opening this in favor of https://github.com/ava-labs/hypersdk/pull/1029 to avoid to force push and rewrite all the history. Richards comments are great too so we also avoid to make the commits have no meaning.

Cache deletes were missing, they were only working for "puts". This PR also adds cache for "get".
This PR is also based on https://github.com/ava-labs/hypersdk/pull/1057 so that we have only one centralized `put` API (which used to be `put_many`) that handles deletes by writing an empty vec. Meaning that deletes are now purely managed on the guest.

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/state.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/state.rs
@@ -130,9 +130,9 @@ impl<'a, K: Key> State<'a, K> {
         let ptr = unsafe { get_bytes(args_bytes.as_ptr(), args_bytes.len()) };
 
         if ptr.is_null() {
-            return Ok(None);
+            Ok(None)
         } else {
-            return Ok(Some(ptr.into()));
+            Ok(Some(ptr.into()))
         }
     }
 
@@ -147,9 +147,9 @@ impl<'a, K: Key> State<'a, K> {
             Some(Some(val)) => {
                 if val.is_empty() {
                     return Ok(None);
-                } else {
-                    val
                 }
+
+                val
             }
             Some(None) => return Ok(None),
             None => {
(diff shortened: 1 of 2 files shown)


x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -39,7 +39,7 @@ pub enum Error {
 }
 
 pub struct State<'a, K: Key> {
-    cache: &'a RefCell<HashMap<K, Vec<u8>>>,
+    cache: &'a RefCell<HashMap<K, Option<Vec<u8>>>>,
```
iFrostizz, line 42: The reason we have an `Option<Vec<u8>>` is to be able to know which is the data we know the state of (in the cache). As a guideline:
- `None`: We don't know the state of the data
- `Some(vec![])`: The data is has been deleted. The `get_host` is crucial to avoid deleting already deleted data (it would make the host error)
- `Some(data)`: There is some data

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -98,72 +98,73 @@ impl<'a, K: Key> State<'a, K> {
...
+                }
+            }
+            Some(None) => return Ok(None),
+            None => {
+                &if let Some(val) = Self::get_host(key)? {
+                    cache.entry(key).or_insert(Some(Vec::new()));
+                    val
+                } else {
+                    return Ok(None);
+                }
```
iFrostizz, line 161: This is only possible thanks to 1.79.0!

dboehm-avalabs (approved):

PR #1067, opened 2024-06-20
Unify `StatelessBlock` and `StatefulBlock`

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/342

Create a single `Block` interface, with `StatelessBlock` and `StatefulBlock` which implements it.
New names for `StatelessBlock` and `StatefulBlock` are TBD.


iFrostizz:
We should probably rename those to `Block` for stateless and `WrappedBlock` or `VMBlock` for stateful.

Issue #819, opened 2024-04-05
[x/programs] Deterministic program-ids 


richardpringle:
The Readme currently states[^1] that we will "likely use the transaction-id" as the program-id. But making program-ids deterministic is preferred as it enables programs to exist across multiple chains with the same ID. 

While transaction-ids are actually also deterministic, what I'm referring to here is determinism independent of nonce/replay protection, such that a program may be deployed with the same ID no matter how many transactions/actions have been sent by the deployer. 

How accomplish this is _to be determined_. 

[^1]: https://github.com/ava-labs/hypersdk/blob/main/README.md?plain=1#L136-L139

PR #1072, opened 2024-06-21
[x/programs] Enable Program based program deployment 

dboehm-avalabs:
Adds a host function that will deploy a program to a new account.

### DIFF
--- a/x/programs/cmd/simulator/cmd/program.go
+++ b/x/programs/cmd/simulator/cmd/program.go
@@ -167,6 +167,7 @@ func (s *ProgramStore) GetAccountProgram(ctx context.Context, account codec.Addr
 }
 
 func (s *ProgramStore) GetProgramBytes(ctx context.Context, programID ids.ID) ([]byte, error) {
+	// TODO: take fee out of balance?
 	programBytes, exists, err := getProgram(ctx, s, programID)
 	if err != nil {
 		return []byte{}, err
--- a/x/programs/cmd/simulator/cmd/program.go
+++ b/x/programs/cmd/simulator/cmd/program.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/akamensky/argparse"
+	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"go.uber.org/zap"
 
@@ -81,7 +82,14 @@ func programCreateFunc(ctx context.Context, db *state.SimpleMutable, path string
 		return codec.EmptyAddress, err
 	}
 
-	address, err := SetProgram(ctx, db, programID, programBytes)
+	err = setProgram(ctx, db, programID, programBytes)
+	if err != nil {
+		response := multilineOutput([][]byte{utils.ErrBytes(err)})
+		fmt.Println(response)
+		return codec.EmptyAddress, fmt.Errorf("program creation failed: %w", err)
+	}
+
+	account, err := deployProgram(ctx, db, programID, []byte{})
 	if err != nil {
 		response := multilineOutput([][]byte{utils.ErrBytes(err)})
 		fmt.Println(response)
@@ -94,7 +102,7 @@ func programCreateFunc(ctx context.Context, db *state.SimpleMutable, path string
 		return codec.EmptyAddress, err
 	}
 
-	return address, nil
+	return account, nil
 }
 
 func programExecuteFunc(
@@ -146,9 +154,20 @@ type ProgramStore struct {
 	state.Mutable
 }
 
-func (s *ProgramStore) GetProgramBytes(ctx context.Context, program codec.Address) ([]byte, error) {
-	// TODO: take fee out of balance?
-	programBytes, exists, err := GetProgram(ctx, s, program)
+func (s *ProgramStore) GetAccountProgram(ctx context.Context, account codec.Address) (ids.ID, error) {
+	programID, exists, err := getAccountProgram(ctx, s, account)
+	if err != nil {
+		return ids.Empty, err
+	}
+	if !exists {
+		return ids.Empty, errors.New("unknown account")
+	}
+
+	return programID, nil
+}
+
+func (s *ProgramStore) GetProgramBytes(ctx context.Context, programID ids.ID) ([]byte, error) {
+	programBytes, exists, err := getProgram(ctx, s, programID)
 	if err != nil {
 		return []byte{}, err
 	}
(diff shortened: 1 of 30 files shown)


x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -21,6 +21,15 @@ pub enum ExternalCallError {
...
 }
 
+#[derive(Error, Debug, BorshDeserialize)]
+#[repr(u8)]
+#[non_exhaustive]
+#[borsh(use_discriminant = true)]
+pub enum DeployError {
+    #[error("an error happened during the deploy")]
+    DeployFailure = 0,
+}
```
iFrostizz, line 31: I see little benefit in having a separate `DeployError`, it might be better to merge the `ExternalCallError` and `DeployError` into a `ProgramError`

x/programs/cmd/simulator/cmd/program.go
```diff
@@ -146,9 +154,20 @@ type ProgramStore struct {
...
+	}
+	if !exists {
+		return ids.Empty, errors.New("unknown account")
+	}
+
+	return programID, nil
+}
+
+func (s *ProgramStore) GetProgramBytes(ctx context.Context, programID ids.ID) ([]byte, error) {
+	programBytes, exists, err := getProgram(ctx, s, programID)
```
iFrostizz, line 170: Should we back off the TODO ? It doesn't seem to be resolved yet:
```suggestion
	// TODO: take fee out of balance?
	programBytes, exists, err := getProgram(ctx, s, programID)
```

x/programs/cmd/simulator/cmd/storage.go
```diff
@@ -54,31 +59,76 @@ func (s *prefixedStateMutable) Remove(ctx context.Context, key []byte) error {
...
 //
 // Program
 //
 
-func programStateKey(key []byte) (k []byte) {
+func accountStateKey(key []byte) (k []byte) {
+	k = make([]byte, 2+len(key))
+	k[0] = accountPrefix
+	copy(k[1:], key)
+	k[len(key)-1] = accountStatePrefix
```
iFrostizz, line 73: This function rewrites the 3rd byte from the end, is your intention to write the last one ? You could maybe write
```suggestion
	k[len(key)+1] = accountStatePrefix
```
You could add a unit test in `storage_test.go`

```go
package cmd

import (
	"testing"

	"github.com/stretchr/testify/require"
)

func TestPrefix(t *testing.T) {
	require := require.New(t)
	key := []byte{0, 1, 2, 3}
	stateKey := accountStateKey(key)
	require.Equal(len(stateKey), len(key) + 2)
	expectedStateKey := append([]byte{accountPrefix}, key...)
	expectedStateKey = append(expectedStateKey, accountStatePrefix)
	require.Equal(stateKey, expectedStateKey)
}
```

x/programs/cmd/simulator/cmd/storage.go
```diff
@@ -54,31 +59,76 @@ func (s *prefixedStateMutable) Remove(ctx context.Context, key []byte) error {
...
+	return
+}
+
+func accountDataKey(key []byte) (k []byte) {
+	k = make([]byte, 2+len(key))
+	k[0] = accountPrefix
+	copy(k[1:], key)
+	k[len(key)-1] = accountDataPrefix
```
iFrostizz, line 81:
```suggestion
	k[len(key)+1] = accountDataPrefix
```

iFrostizz (changes requested):

x/programs/rust/wasmlanche-sdk/src/program.rs
dboehm-avalabs: I'd rather have more specific errors for the calls if possible.  It gives an easy way to see what errors are applicable.

iFrostizz: what I meant is that it's a little bit weird to have a separate error enum while we use `ExternalCallError` for all other faillible program operations because then we would need to import two different error enums, just thought about having
```rust
#[derive(Error, Debug, BorshDeserialize)]
#[repr(u8)]
#[non_exhaustive]
#[borsh(use_discriminant = true)]
pub enum ExternalCallError {
    #[error("an error happened during the execution")]
    ExecutionFailure = 0,
    #[error("the call panicked")]
    CallPanicked = 1,
    #[error("not enough fuel to cover the execution")]
    OutOfFuel = 2,
    #[error("an error happened during the deploy")]
    DeployFailure = 3,
}
```

But since `ExternalCallError` isn't really a good name anymore because of the `DeployFailure`, I suggested `ProgramError`.

richardpringle: The reason this error is called `ExternalCallError` is because it should only be returned on external calls. I'm not sure that a `DeployFailure` deserves its own category when making an external call. What are the circumstances in which a deployment can fail? You could run out of fuel, certainly... but that should be an `OutOfFuel` error. 

I don't think a deployment failure should be recoverable. Either that, or we need to give a better reason as to why the deployment failed.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -22,7 +23,16 @@ pub enum ExternalCallError {
...
 
-/// Represents the current Program in the context of the caller, or an external
+#[derive(Error, Debug, BorshDeserialize)]
+#[repr(u8)]
+#[non_exhaustive]
+#[borsh(use_discriminant = true)]
+pub enum DeployError {
+    #[error("an error happened during the deploy")]
+    DeployFailure = 0,
+}
```
richardpringle, line 33: When would we want to return this? Can talk about it in this thread and if we agree that *we do* want to return it, we should add the docs. 

We should probably either just panic here or give details that one would want to act on.

x/programs/cmd/simulator/cmd/storage.go
```diff
@@ -89,14 +139,24 @@ func GetProgram(
...
+
+func deployProgram(
+	ctx context.Context,
+	mu state.Mutable,
+	programID ids.ID,
+	accountCreationData []byte,
 ) (codec.Address, error) {
-	address := codec.CreateAddress(programPrefix, programID)
-	return address, mu.Insert(ctx, address[:], program)
+	newID := sha256.Sum256(append(programID[:], accountCreationData...))
```
richardpringle, line 157: 👍

x/programs/rust/wasmlanche-sdk/src/program.rs
dboehm-avalabs: yeah, lets worry about this later.

richardpringle (approved):

dboehm-avalabs:
Helps with #819 the account_creation_data is meant to be used to allow the program to pass in the appropriate data to make the account deterministic.

Issue #1062, opened 2024-06-19
Allow Context injection in the simulator


iFrostizz:
The simulator should allow context injection for top-level calls. This will allow us to advance time, block, change actor, ...
https://github.com/ava-labs/hypersdk/pull/1058 already implements a builder to pass the call info more easily, so we should extend it to allow the default `CallInfo` to be altered from the outside.

PR #1076, opened 2024-06-24
Add support for context overrides

iFrostizz:
A `TestContext` can now be passed to the simulator in order to override any value of the `callInfo`.
This is firstly used to call a program (alter the `ProgramId`), but also to spoof the timestamp, block height, actor, and so on.

Closes https://github.com/ava-labs/hypersdk/issues/1062


x/programs/cmd/simulator/src/lib.rs
```diff
@@ -91,6 +93,27 @@ pub enum Key {
...
+pub struct TestContext {
+    pub program_id: Id,
+}
+
+#[derive(Clone, Debug, PartialEq, Serialize)]
+#[serde(tag = "type", rename = "testContext")]
+pub(crate) struct SimulatorTestContext {
+    #[serde(serialize_with = "base64_struct_encode")]
+    value: TestContext,
+}
```
iFrostizz, line 108: It's clearly a hack. There is no way to externally tag a struct to have this `type` `value` structure like we have for enums.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -24,22 +25,14 @@ pub enum ExternalCallError {
 /// Represents the current Program in the context of the caller. Or an external
 /// program that is being invoked.
 #[cfg_attr(feature = "debug", derive(Debug))]
+#[derive(Clone, BorshSerialize, Serialize)]
```
iFrostizz, line 28: I will back off this `Serialize` add. I wanted to be able to pass the `Context` but it was not really easy to construct it in a nice way because of traits such as `Default` for `Program<K>`. Now that we have the `TestContext` we have a clean builder API.

x/programs/runtime/util_test.go
```diff
@@ -20,12 +20,14 @@ type testRuntime struct {
 	DefaultGas uint64
 }
 
-func (t *testRuntime) CallProgram(program codec.Address, actor codec.Address, function string, params ...interface{}) ([]byte, error) {
+func (t *testRuntime) CallProgram(program codec.Address, actor codec.Address, height uint64, timestamp uint64, function string, params ...interface{}) ([]byte, error) {
```
dboehm-avalabs, line 23: Let's not add these and then use the CallContext instead.

x/programs/cmd/simulator/src/id.rs
```diff
@@ -1,6 +1,6 @@
 use serde::Serialize;
 
-#[derive(Clone, Copy, Debug, PartialEq, Serialize)]
+#[derive(Clone, Copy, Debug, PartialEq, Serialize, Default)]
```
richardpringle, line 3: 🤔, why do you need a default here now but you didn't before?

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -91,6 +91,49 @@ pub enum Key {
...
 }
 
+#[derive(Clone, Debug, PartialEq, Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
+pub struct TestContext {
+    program_id: Id,
+    actor_key: Option<Key>,
+    height: u64,
+    timestamp: u64,
```
richardpringle, line 101: Why not make all of these fields `pub`, this is only meant to be used in tests, and there are no properties that you need to be private

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -91,6 +91,49 @@ pub enum Key {
...
+
+    pub fn with_timestamp(mut self, timestamp: u64) -> Self {
+        self.timestamp = timestamp;
+        self
+    }
+
+    pub fn clear_actor_key(&mut self) {
+        self.actor_key = None;
+    }
+}
```
richardpringle, line 135: This is overkill when you can set every property:
```suggestion
```

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -99,6 +142,8 @@ pub enum Param {
     String(String),
     Id(Id),
     Key(Key),
+    #[allow(private_interfaces)]
+    TestContext(SimulatorTestContext),
```
richardpringle, line 146: Why do we have to do this?

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -188,6 +240,14 @@ where
     serializer.serialize_str(&b64.encode(text))
 }
 
+fn base64_struct_encode<S>(struc: &TestContext, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let bytes = serde_json::to_vec(struc).unwrap();
+    serializer.serialize_str(&b64.encode(bytes))
```
richardpringle, line 248: What's the benefit of doing this?

x/programs/runtime/runtime_test.go
```diff
@@ -57,3 +58,27 @@ func TestRuntimeCallProgramComplexReturn(t *testing.T) {
...
+
+	_, err = program.Call("can_change_height")
+	require.NoError(err)
+	_, err = program.CallWithHeight(1, "can_change_height")
+	require.NoError(err)
+
+	_, err = program.Call("can_change_actor")
+	require.NoError(err)
+	_, err = program.CallWithActor(codec.CreateAddress(0, ids.GenerateTestID()), "can_change_actor")
+	require.NoError(err)
```
richardpringle, line 83: I think each of these should be in their own test. You should be able to instantiate the program once and then use it in each proceeding test. I've done it before but don't remember what I did.

x/programs/runtime/util_test.go
richardpringle: +1

x/programs/rust/examples/counter-external/src/lib.rs
```diff
@@ -38,15 +38,20 @@ mod tests {
...
         let counter = simulator
             .run_step(&owner, &Step::create_program(counter_path))
             .expect("should be able to create the counter")
             .id;
         let counter = Param::Id(counter);
 
-        let params = vec![counter_external, counter.clone(), owner_key.clone()];
+        let params = vec![
+            TestContext::default()
+                .with_program_id(counter_external)
```
richardpringle, line 50: It would be better to do `impl From<Id> for TextContext` where all the other fields get default values.

x/programs/rust/examples/counter/src/lib.rs
richardpringle, line 1: If you always have to enter the program-id, then you shouldn't be able to set a default value. This is an anti-pattern

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -25,22 +25,13 @@ pub enum ExternalCallError {
...
-    fn serialize<W: std::io::prelude::Write>(&self, writer: &mut W) -> std::io::Result<()> {
-        let Self {
-            account,
-            state_cache: _,
-        } = self;
-
-        account.serialize(writer)
-    }
-}
-
```
richardpringle, line 43: Why did you remove this?

x/programs/rust/wasmlanche-sdk/src/lib.rs
```diff
@@ -34,24 +35,6 @@ pub struct Context<K = ()> {
...
-            action_id,
-        } = self;
-        BorshSerialize::serialize(program, writer)?;
-        BorshSerialize::serialize(actor, writer)?;
-        BorshSerialize::serialize(height, writer)?;
-        BorshSerialize::serialize(timestamp, writer)?;
-        BorshSerialize::serialize(action_id, writer)?;
-        Ok(())
-    }
-}
```
richardpringle, line 53: Why did you remove this?

x/programs/test/programs/context_injection/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+            assert!(new_timestamp > timestamp)
+        }
+        None => {
+            context
+                .program
+                .state()
+                .store(StateKeys::LastTimestamp, &new_timestamp)
+                .unwrap();
+        }
+    }
```
richardpringle, line 29: Why don't you just return the timestamp then validate that it was the one that was injected. This is overcomplicated.

x/programs/test/programs/context_injection/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+        Some(actor) => assert!(new_actor != actor),
+        None => {
+            context
+                .program
+                .state()
+                .store(StateKeys::LastActor, &new_actor)
+                .unwrap();
+        }
+    }
+}
```
richardpringle, line 60: Same goes for these.

x/programs/test/programs/context_injection/Cargo.toml
richardpringle, line 1: If I'm not mistaken, all the tests that are in `x/programs/tests` are meant to be tested from Go. This seems more like an e2e test that belongs in the `x/programs/cmd/simulator` directory somewhere

richardpringle (changes requested):

x/programs/runtime/util_test.go
iFrostizz: Replaced fields by a `Context` instead.

x/programs/cmd/simulator/src/id.rs
iFrostizz: That's because `TestContext` has `program_id` as a field which is of type `Id`.

x/programs/cmd/simulator/src/lib.rs
iFrostizz: Because we want the user space to only have access to `TestContext`, `SimulatorTestContext` is private and only meant for the simulator, because struct cannot be externally tagged. See https://github.com/ava-labs/hypersdk/pull/1076#discussion_r1651672553

x/programs/cmd/simulator/src/lib.rs
iFrostizz: The Go simulator reads the value from the `testContext` parameter (Type field) as a `[]byte`. Thus we need to encode this inner one as base64 because that's what `encoding/json` does on the go side. I hate it.

x/programs/runtime/runtime_test.go
iFrostizz: Do you mean that the test should be instantiated globally and that each test should call only one function ? That sounds like fixture-based testing

x/programs/rust/examples/counter-external/src/lib.rs
iFrostizz: Agree with this design decision. Updated!

x/programs/rust/examples/counter/src/lib.rs
iFrostizz: Makes sense, we should never need to call the empty program id.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: We achieve the same result with `#[borsh(skip)]`

x/programs/rust/wasmlanche-sdk/src/lib.rs
iFrostizz: Because the `BorshSerialize` attribute does the same thing

x/programs/test/programs/context_injection/src/lib.rs
iFrostizz: true... Your idea is more "atomic"

x/programs/test/programs/context_injection/src/lib.rs
iFrostizz: Changed them as well.

x/programs/test/programs/context_injection/Cargo.toml
iFrostizz: Done, I moved this test in `x/programs/cmd/simulator/test/`

iFrostizz:
All of the CI is ran because it thinks that not only programs have been ran, we might want to extend the grep range

x/programs/rust/wasmlanche-sdk/src/lib.rs
richardpringle: Can you put this back? This evaluates to 
```rust
impl<K: BorshSerialize> for Context<K> { // ...
```
But we don't need the trait bound on `K`.

x/programs/runtime/runtime_test.go
```diff
@@ -57,3 +58,24 @@ func TestRuntimeCallProgramComplexReturn(t *testing.T) {
...
+	require.Equal(uint64(1), into[uint64](result))
+
+	result, err = program.CallWithHeight(1, "get_height")
+	require.NoError(err)
+	require.Equal(uint64(1), into[uint64](result))
+
+	result, err = program.CallWithActor(codec.CreateAddress(0, ids.GenerateTestID()), "get_actor")
+	require.NoError(err)
+	require.NotEqual(ids.Empty, into[ids.ID](result))
```
richardpringle, line 80: These should be three different tests. You can use `t.Run` like I did in this PR to separate these into three different tests.

The issue is that if `CallWithTimeStamp` fails, you should still run `CallWithHeight` and `CallWithActor`. 

When anyone is debugging this, it's a lot more telling where the issue is if three different assertions fail rather than adding a patch only to get the first assertion to pass. Do you see what I mean? Unit tests are supposed to help with debugging as much as they're meant to prove that something works as expected.

x/programs/cmd/simulator/test/context_injection/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+
+#[public]
+pub fn get_height(context: Context) -> u64 {
+    context.height()
+}
+
+#[public]
+pub fn get_actor(context: Context) -> Address {
+    context.actor()
+}
```
richardpringle, line 16: 👍

x/programs/cmd/simulator/test/context_injection/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+                },
+            )
+            .unwrap()
+            .result
+            .response::<Address>()
+            .unwrap();
+
+        assert_ne!(response, Address::default());
+    }
+}
```
richardpringle, line 140: Way better, 🥇

richardpringle (changes requested):

x/programs/rust/wasmlanche-sdk/src/lib.rs
iFrostizz: Sorry ! Indeed I expanded the code and it constraint `K` to be `BorshSerialize`, this is not the case on the automatic serialization for `Program` though. I'm backing this off.

x/programs/runtime/runtime_test.go
iFrostizz: Thanks, yeah you are right that isolating unit tests makes them much easier to debug. I have used `t.Run` but it feels like the test is repeating. David's PR on default struct does not solve that because the function would panic if the default value was already set.
I have probably found a happy solution, passed a function pointer in one of the fields of the test property and executed the changing logic here.

richardpringle: 😅 
>like I did in this PR

I forgot to provide the link

richardpringle (approved):

PR #1079, opened 2024-06-26
Add the ability to return raw bytes from `call_function`

iFrostizz:
There is currently no way to call an external function if we don't know the type of the return value ahead of time.
Adding the ability to return a `DeferDeserialize` struct (thanks @richardpringle ) from `call_function` that contains these raw bytes and has a method `deserialize` that deserialize these to the desired type.

### DIFF
--- a/x/programs/runtime/import_program_test.go
+++ b/x/programs/runtime/import_program_test.go
@@ -152,7 +152,6 @@ func TestImportGetRemainingFuel(t *testing.T) {
 	require.NoError(err)
 	require.LessOrEqual(into[uint64](result), program.Runtime.DefaultGas)
 }
-<<<<<<< HEAD
 
 func TestImportOutOfFuel(t *testing.T) {
 	require := require.New(t)
@@ -165,5 +164,3 @@ func TestImportOutOfFuel(t *testing.T) {
 	require.NoError(err)
 	require.Equal([]byte{byte(OutOfFuel)}, result)
 }
-=======
->>>>>>> ef69c9a9 (write unit tests, write docs, faillible `deserialize`)
--- a/x/programs/runtime/import_program_test.go
+++ b/x/programs/runtime/import_program_test.go
@@ -152,6 +152,7 @@ func TestImportGetRemainingFuel(t *testing.T) {
 	require.NoError(err)
 	require.LessOrEqual(into[uint64](result), program.Runtime.DefaultGas)
 }
+<<<<<<< HEAD
 
 func TestImportOutOfFuel(t *testing.T) {
 	require := require.New(t)
@@ -164,3 +165,5 @@ func TestImportOutOfFuel(t *testing.T) {
 	require.NoError(err)
 	require.Equal([]byte{byte(OutOfFuel)}, result)
 }
+=======
+>>>>>>> ef69c9a9 (write unit tests, write docs, faillible `deserialize`)
--- a/x/programs/rust/wasmlanche-sdk/src/program.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/program.rs
@@ -24,7 +24,6 @@ impl DeferDeserialize {
 }
 
 impl BorshDeserialize for DeferDeserialize {
-    /// <div class="warning">This function does not provide any size-hint.</div>
     fn deserialize_reader<R: Read>(reader: &mut R) -> std::io::Result<Self> {
         let mut inner = Vec::new();
         reader.read_to_end(&mut inner)?;
--- a/x/programs/rust/wasmlanche-sdk/src/program.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/program.rs
@@ -9,6 +9,16 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use std::{cell::RefCell, collections::HashMap};
 use thiserror::Error;
 
+#[cfg_attr(feature = "debug", derive(Debug))]
+pub struct DeferDeserialize(Vec<u8>);
+
+impl DeferDeserialize {
+    pub fn deserialize<T: BorshDeserialize>(self) -> Result<T, std::io::Error> {
+        let Self(bytes) = self;
+        borsh::from_slice(&bytes)
+    }
+}
+
 /// An error that is returned from call to public functions.
 #[derive(Error, Debug, BorshSerialize, BorshDeserialize)]
 #[repr(u8)]
@@ -71,20 +81,19 @@ impl<K> Program<K> {
     /// # use wasmlanche_sdk::{types::Address, Program};
     /// #
     /// # let program_id = [0; Address::LEN];
-    /// # let target: Program<()> = borsh::from_slice(&program_id).expect("the program should deserialize");
+    /// # let target: Program<()> = borsh::from_slice(&program_id)?;
     /// let increment = 10;
-    /// let params = borsh::to_vec(&increment).expect("failed to borsh serialize params");
+    /// let params = borsh::to_vec(&increment)?;
     /// let max_units = 1000000;
-    /// target
-    ///     .call_function("call_with_param", &params, max_units)
-    ///     .unwrap()
+    /// let bytes = target.call_function("increment", &params, max_units)?;
+    /// let incremented = bytes.deserialize()?;
     /// ```
-    pub fn call_function<T: BorshDeserialize>(
+    pub fn call_function(
         &self,
         function_name: &str,
         args: &[u8],
         max_units: Gas,
-    ) -> Result<T, ExternalCallError> {
+    ) -> Result<DeferDeserialize, ExternalCallError> {
         #[link(wasm_import_module = "program")]
         extern "C" {
             #[link_name = "call_program"]
@@ -102,7 +111,22 @@ impl<K> Program<K> {
 
         let bytes = unsafe { call_program(args_bytes.as_ptr(), args_bytes.len()) };
 
-        borsh::from_slice(&bytes).expect("failed to deserialize")
+        let result_bytes = match bytes.first().expect("missing result prefix byte") {
+            0 => {
+                if bytes.len() != 2 {
+                    panic!("wrong encoding for result");
+                } else {
+                    let error_code = bytes.get(1).unwrap();
+                    let error_res: ExternalCallError = borsh::from_slice(&[*error_code][..])
+                        .expect("failed to deserialize error code");
+                    return Err(error_res);
+                }
+            }
+            1 => bytes.get(1..).unwrap(),
+            _ => panic!("wrong error code"),
+        };
+
+        Ok(DeferDeserialize(result_bytes.to_vec()))
     }
 
     /// Gets the remaining fuel available to this program
--- a/x/programs/rust/wasmlanche-sdk/src/program.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/program.rs
@@ -10,13 +10,13 @@ use std::{cell::RefCell, collections::HashMap, io::Read};
 use thiserror::Error;
 
 /// Defer deserialization from bytes
+/// <div class="warning">It is possible that this type performs multiple allocations during deserialization. It should be used sparingly.</div>
 #[cfg_attr(feature = "debug", derive(Debug))]
 pub struct DeferDeserialize(Vec<u8>);
 
 impl DeferDeserialize {
     /// # Errors
     /// Returns a [`std::io::Error`] if there was an issue deserializing the value
-    #[allow(clippy::must_use_candidate)]
     pub fn deserialize<T: BorshDeserialize>(self) -> Result<T, std::io::Error> {
         let Self(bytes) = self;
         borsh::from_slice(&bytes)
@@ -24,8 +24,7 @@ impl DeferDeserialize {
 }
 
 impl BorshDeserialize for DeferDeserialize {
-    /// <div class="warning">This function does not provide any size-hint.
-    /// It is possible that this type performs multiple allocations during deserialization. It should be used sparingly.</div>
+    /// <div class="warning">This function does not provide any size-hint.</div>
     fn deserialize_reader<R: Read>(reader: &mut R) -> std::io::Result<Self> {
         let mut inner = Vec::new();
         reader.read_to_end(&mut inner)?;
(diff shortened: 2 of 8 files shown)


x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -71,13 +92,13 @@ impl<K> Program<K> {
...
     /// let increment = 10;
-    /// let params = borsh::to_vec(&increment).expect("failed to borsh serialize params");
+    /// let params = borsh::to_vec(&increment).expect("serialization error");
     /// let max_units = 1000000;
-    /// target
-    ///     .call_function("call_with_param", &params, max_units)
-    ///     .unwrap()
+    /// let has_incremented: bool = target.call_function("increment", &params, max_units)?;
+    /// assert!(has_incremented);
+    /// # Ok::<(), wasmlanche_sdk::ExternalCallError>(())
```
iFrostizz, line 101: My reasoning for these is that the borsh deserialization and serialization above are infaillible. So it was a nice opportunity to return an error from the example.

x/programs/test/programs/call_program/src/lib.rs
```diff
@@ -51,12 +52,16 @@ pub fn call_with_two_params_external(
...
         .unwrap()
 }
+
+#[public]
+pub fn call_deferred(_: Context, target: Program, max_units: Gas) -> i64 {
+    let bytes = target
+        .call_function::<DeferDeserialize>("simple_call", &[], max_units)
+        .unwrap();
+    bytes.deserialize()
+}
```
richardpringle, line 67: You don't need an integration test to test this feature, it can be done with a unit test.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -6,9 +6,30 @@ use crate::{
...
+pub struct DeferDeserialize(Vec<u8>);
+
+impl DeferDeserialize {
+    /// # Panics
+    /// Panics if there was an issue deserializing the value
+    #[allow(clippy::must_use_candidate)]
+    pub fn deserialize<T: BorshDeserialize>(self) -> T {
+        let Self(bytes) = self;
+        borsh::from_slice(&bytes).expect("failed to deserialize")
+    }
```
richardpringle, line 22:
```suggestion
    pub fn deserialize<T: BorshDeserialize>(self) -> Result<T, borsh::io::Error> {
        borsh::from_slice(&self.0)
    }
```
Don't have to change this to one line, but you should let the user decide what to do in the case of an error.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -6,9 +6,30 @@ use crate::{
     Gas,
 };
 use borsh::{BorshDeserialize, BorshSerialize};
-use std::{cell::RefCell, collections::HashMap};
+use std::{cell::RefCell, collections::HashMap, io::Read};
 use thiserror::Error;
 
+#[cfg_attr(feature = "debug", derive(Debug))]
+pub struct DeferDeserialize(Vec<u8>);
```
richardpringle, line 13: Can you add a warning doc-comment stating that "since the `Reader` in the `BorshDeserialize::deserialize_reader` function doesn't provide any size-hint, it's possible that this type performs multiple allocations during deserialization"? And also add that "it should be used sparingly."

x/programs/runtime/import_program_test.go
```diff
@@ -151,3 +151,20 @@ func TestImportGetRemainingFuel(t *testing.T) {
...
+	program := newTestProgram(ctx, "call_program")
+
+	expected, err := serialize(int64(0))
+	require.NoError(err)
+
+	result, err := program.Call("call_deferred",
+		program.Address, uint64(1000000))
+	require.NoError(err)
+	require.Equal(expected, result)
+}
```
richardpringle, line 170: This is simply a convenience type that *should* make it easier to write the multi-sig. It should have a unit test, but adding integration tests like this is overkill.

richardpringle (changes requested):

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Done, the `deserialize` function now returns a `Result`. Thanks!

x/programs/runtime/import_program_test.go
iFrostizz: Got it. I added 3 unit tests that shows the usage of `DeferDeserialize` compared to a borsh deserialization.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Added a doc comment on the `deserialize_reader` function directly. I couldn't get a working hyperlink to the function in a warning div.

dboehm-avalabs (dismissed):

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -6,11 +6,35 @@ use crate::{
...
+    /// Returns a [`std::io::Error`] if there was an issue deserializing the value
+    #[allow(clippy::must_use_candidate)]
+    pub fn deserialize<T: BorshDeserialize>(self) -> Result<T, std::io::Error> {
+        let Self(bytes) = self;
+        borsh::from_slice(&bytes)
+    }
+}
+
+impl BorshDeserialize for DeferDeserialize {
+    /// <div class="warning">This function does not provide any size-hint.
```
richardpringle, line 27:
```
<div class="warning">
```
🥇

x/programs/rust/wasmlanche-sdk/src/program.rs
richardpringle: Can you move those docs here? The type itself should be used sparingly. But the docs you wrote below are perfect (especially that warning)

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -6,11 +6,35 @@ use crate::{
...
 use thiserror::Error;
 
+/// Defer deserialization from bytes
+#[cfg_attr(feature = "debug", derive(Debug))]
+pub struct DeferDeserialize(Vec<u8>);
+
+impl DeferDeserialize {
+    /// # Errors
+    /// Returns a [`std::io::Error`] if there was an issue deserializing the value
+    #[allow(clippy::must_use_candidate)]
```
richardpringle, line 19: Do you need to do this? `Result` is already `must_use`.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -6,11 +6,35 @@ use crate::{
...
+    fn deserialize_reader<R: Read>(reader: &mut R) -> std::io::Result<Self> {
+        let mut inner = Vec::new();
+        reader.read_to_end(&mut inner)?;
+        Ok(Self(inner))
+    }
+}
+
 /// An error that is returned from call to public functions.
-#[derive(Error, Debug, BorshSerialize, BorshDeserialize)]
+#[derive(Error, Debug, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
```
richardpringle, line 37:
```suggestion
#[derive(Error, Debug, BorshSerialize, BorshDeserialize)]
```

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -171,3 +195,38 @@ impl<K> BorshSerialize for CallProgramArgs<'_, K> {
...
+
+#[cfg(test)]
+mod tests {
+    use super::{DeferDeserialize, ExternalCallError};
+
+    #[test]
+    fn defer_ok_variant() {
+        let bytes = DeferDeserialize(vec![1, 0]);
+        assert_eq!(bytes.deserialize::<Result<_, ()>>().unwrap(), Ok(false));
+    }
```
richardpringle, line 207: This is interesting. What are you actually testing here? The implementation of the `derserialize` function is trivial, so I think you're actually testing the `BorshDeserialize` implementation of `Result<bool, ()>`.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -171,3 +195,38 @@ impl<K> BorshSerialize for CallProgramArgs<'_, K> {
...
+    #[test]
+    fn defer_err_variant() {
+        let bytes = DeferDeserialize(vec![0, 1]);
+        assert_eq!(
+            bytes
+                .deserialize::<Result<(), ExternalCallError>>()
+                .unwrap(),
+            Err(ExternalCallError::CallPanicked)
+        );
+    }
```
richardpringle, line 218: Same comment here

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -171,3 +195,38 @@ impl<K> BorshSerialize for CallProgramArgs<'_, K> {
...
+
+    #[test]
+    fn defer_no_len_prefix() {
+        let inner_bytes = vec![1, 4, 3, 2, 1];
+        let defer_res = borsh::from_slice::<Result<DeferDeserialize, ()>>(&inner_bytes);
+        assert!(defer_res.is_ok());
+        let Ok(defer) = defer_res else {
+            unreachable!();
+        };
```
richardpringle, line 227: Just use `unwrap` instead of doing this

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -171,3 +195,38 @@ impl<K> BorshSerialize for CallProgramArgs<'_, K> {
...
+        assert!(defer.is_ok_and(|bytes| bytes.0 == vec![4, 3, 2, 1]));
+        let defer = borsh::from_slice::<Result<Vec<u8>, ExternalCallError>>(&inner_bytes);
+        assert!(defer.is_err());
+    }
```
richardpringle, line 231: The test here is overcomplicated. Why don't you do something more like this:
```rust
type ExptectedType = u64;

let expected: ExpectedType = 42;
let serialized = borsh::to_vec(expected).unwrap();
let deferred: DeferDeserialize = borsh::from_slice(&serialized);
assert_eq!(deferred.0 == serialized);
let actual = deferred.deserialize::<ExpectedType>().unwrap();
assert!(actual, expected);
```
(I didn't actual compile and run this, but it should be something this simple).

x/programs/test/programs/call_program/src/lib.rs
richardpringle, line 1: Why are these changes even in this PR?

richardpringle (changes requested):

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: 😄

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: I have split the warning text and have framed those two with the warning.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Good catch, looks like I don't because this function used to be infaillible.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: I had to do that for the test `defer_err_variant`. It would not only be useful for the test but also for pattern-matching the error in programs directly, don't you think that we should leave it ?

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: What was I thinking about ...

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: True, well I just removed other tests and replaced them by yours because they were not properly testing the `DeferDeserialize` struct but felt duplicated.

x/programs/test/programs/call_program/src/lib.rs
iFrostizz: I thought that they were small enough, they don't really follow the aim of this PR though so I could open a new PR that fixes these if you prefer.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Removed this test in favour of yours.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Removed it.

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Now the test is gone, but the usage of testing if this error matches is still useful in programs context.

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -6,11 +6,34 @@ use crate::{
...
+    /// # Errors
+    /// Returns a [`std::io::Error`] if there was an issue deserializing the value
+    pub fn deserialize<T: BorshDeserialize>(self) -> Result<T, std::io::Error> {
+        let Self(bytes) = self;
+        borsh::from_slice(&bytes)
+    }
+}
+
+impl BorshDeserialize for DeferDeserialize {
+    /// <div class="warning">This function does not provide any size-hint.</div>
```
richardpringle, line 27: Haha, the size hint is related to the allocations. Maybe it's confusing and you can just delete this. Do it in a follow up if this is the only comment.

x/programs/rust/wasmlanche-sdk/src/program.rs
richardpringle: No one should be comparing the internals of this type unless it's a test here which can do the comparison with `deferred.0`.

richardpringle (changes requested):
Two tiny changes

x/programs/rust/wasmlanche-sdk/src/program.rs
richardpringle: Sorry, above comment doesn't make sense

x/programs/rust/wasmlanche-sdk/src/program.rs
iFrostizz: Nah that's fine I will do it now while I'm here 😄

richardpringle (approved):

Issue #1096, opened 2024-07-05
Update `State.store` to be able to store multiple key-values


iFrostizz:
As raised in https://github.com/ava-labs/hypersdk/pull/1095#discussion_r1666055805
The current way of putting stuff in storage is by calling `store` on the state directly which takes as an input a `K: Key`, `V: BorshSerialize`. This is problematic if there is a lot of data to store because we need to call the function independently multiple times which takes a lot of code space.
We should create a function that allows storing a list of tuple (K, V) in storage.

PR #1097, opened 2024-07-05
Make `store` take a list of (key, value) to "store many"

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1096

Update `store` so that it takes a list of tuples, and create a new `store_single` to match the old behavior.
The cache is not passed by mutable reference in order to avoid the runtime cost of borrowing it mutably many times.

The only pitfall is that because the function signature of `store` is now `fn store<V: BorshSerialize, Pairs: IntoIterator<Item = (K, &V)>`, the `V` has to be the same consistent type across the list.

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/state.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/state.rs
@@ -1,5 +1,6 @@
 use crate::{memory::HostPtr, state::Error as StateError};
 use borsh::{from_slice, to_vec, BorshDeserialize, BorshSerialize};
+use std::cell::RefMut;
 use std::{cell::RefCell, collections::HashMap, hash::Hash};
 
 #[derive(Clone, thiserror::Error, Debug)]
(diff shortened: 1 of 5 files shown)


x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -38,15 +42,37 @@ impl<'a, K: Key> State<'a, K> {
...
+        self,
+        pairs: Pairs,
+    ) -> Result<(), Error> {
+        let cache = &mut self.cache.borrow_mut();
+        for (key, value) in pairs {
+            Self::store_internal(cache, key, value)?;
+        }
+
+        Ok(())
+    }
```
richardpringle, line 59:
```suggestion
    pub fn store<'b, V: BorshSerialize + 'b, Pairs: IntoIterator<Item = (K, &'b V)>>(
        self,
        pairs: Pairs,
    ) -> Result<(), Error> {
        let cache = &mut self.cache.borrow_mut();

        pairs
            .into_iter()
            .map(|(k, v)| borsh::to_vec(&v).map(|bytes| (k, Some(bytes))))
            .try_for_each(|result| {
                result.map(|(k, v)| {
                    cache.insert(k, v);
                })
            })
            .map_err(|_| StateError::Serialization)?;

        Ok(())
    }
```

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -38,15 +42,37 @@ impl<'a, K: Key> State<'a, K> {
...
     /// Returns an [`Error`] if the key or value cannot be
     /// serialized or if the host fails to handle the operation.
-    pub fn store<V>(self, key: K, value: &V) -> Result<(), Error>
-    where
-        V: BorshSerialize,
-    {
+    pub fn store_single<V: BorshSerialize>(self, key: K, value: &V) -> Result<(), Error> {
+        let cache = &mut self.cache.borrow_mut();
+        Self::store_internal(cache, key, value)
+    }
```
richardpringle, line 69:
```suggestion
    pub fn store_by_key<V: BorshSerialize>(self, key: K, value: &V) -> Result<(), Error> {
        self.store([(key, value)])
    }
```

x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -38,15 +42,37 @@ impl<'a, K: Key> State<'a, K> {
...
+
+    fn store_internal<V: BorshSerialize>(
+        cache: &mut RefMut<HashMap<K, Option<Vec<u8>>>>,
+        key: K,
+        value: &V,
+    ) -> Result<(), Error> {
```
richardpringle, line 75: This isn't a good pattern to follow if there's another way to do it.

richardpringle (changes requested):

richardpringle (approved):

PR #1105, opened 2024-07-09
Simplify auth in examples

containerman17:
Currently, each example app folder has its own auth folder. I suggest moving all auth schemes from the `examples` folder to the `hypersdk/chain/auth` package.

If there's a better package for this, please let me know. I'm still getting familiar with the repo structure.


aaronbuchwald:
This LGTM, could we change the package structure from `chain/auth/` to just `auth/` ?

containerman17:
1. Moved auth to it's own package
2. Signed commits

aaronbuchwald (approved):

PR #1106, opened 2024-07-09
Fix Rust Build Errors

samliok:
fixes the rust build errors I was receiving when trying to run the rust tests

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/state.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/state.rs
@@ -145,9 +145,9 @@ impl<'a, K: Key> State<'a, K> {
             }
             Some(None) => return Ok(None),
             None => {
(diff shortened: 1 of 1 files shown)


x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -145,9 +145,9 @@ impl<'a, K: Key> State<'a, K> {
             }
             Some(None) => return Ok(None),
             None => {
-                &if let Some(val) = Self::get_host(key)? {
-                    cache.entry(key).or_insert(Some(Vec::new()));
-                    val
+                // if no value in cache, fetch from host
+               if let Some(val) = Self::get_host(key)? {
+                    cache.entry(key).or_insert(Some(val)).as_ref().unwrap()
```
samliok, line 150: why are we even adding to the cache if this value is being deleted?

PR #1109, opened 2024-07-09
Fix fee window roll for < 1s block intervals

aaronbuchwald:
This PR migrates a bug fix from https://github.com/ava-labs/hypersdk/pull/1091 / https://github.com/ava-labs/hypersdk/commit/0a6a97a63807e0bc4b787df1bf12527adaeb0599

This fixes a bug where block timestamps are measured in milliseconds, but the fee manager only rolls the multi-dimensional consumption windows in 1s increments. This means that if blocks are produced < 1s apart, the fee manager observes each of these blocks as occurring 0s apart (rounding down).

This means that a sequence of N blocks each occurring < 1s apart, the fee manager will continue counting the consumption as occurring within the same interval since it uses the parent and current block timestamps as its only points of reference.

### DIFF
--- a/fees/manager.go
+++ b/fees/manager.go
@@ -35,13 +35,20 @@ type (
 
 // Manager is safe for concurrent use
 type Manager struct {
-	l   sync.RWMutex
+	l sync.RWMutex
+
+	// Layout: [timestamp(s)][dimension[0].price][dimension[0].window][dimension[0].lastConsumed]...
+	//
+	// Note, we must store the timestamp because computing the time difference exclusively off
+	// of block parent timestamps may never result in a window roll (difference may always be less
+	// than 1 second). This bug would result in the unit price never changing (or even going up if the
+	// last consumed is larger than the target).
 	raw []byte
 }
 
 func NewManager(raw []byte) *Manager {
 	if len(raw) == 0 {
-		raw = make([]byte, FeeDimensions*dimensionStateLen)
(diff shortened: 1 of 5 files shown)


fees/manager.go
```diff
@@ -78,19 +85,22 @@ func (f *Manager) LastConsumed(d Dimension) uint64 {
...
 	targetUnits := r.GetWindowTargetUnits()
 	unitPriceChangeDenom := r.GetUnitPriceChangeDenominator()
 	minUnitPrice := r.GetMinUnitPrice()
-	since := int((currTime - lastTime) / consts.MillisecondsPerSecond)
-	bytes := make([]byte, dimensionStateLen*FeeDimensions)
+	lastTimeSeconds := int64(binary.BigEndian.Uint64(f.raw[0:consts.Int64Len]))
+	currTimeSeconds := int64(currTime / consts.MillisecondsPerSecond)
+	since := uint64(currTimeSeconds - lastTimeSeconds)
+	bytes := make([]byte, consts.Int64Len+dimensionStateLen*FeeDimensions)
+	binary.BigEndian.PutUint64(bytes[0:consts.Int64Len], uint64(currTimeSeconds))
```
aaronbuchwald, line 103: By storing the timestamp in seconds, as soon as we move 1.9 to 2.1 it will be treated as a 1 second interval elapsing. Previously, this would have been treated as 200ms rounding down to no difference.

aaronbuchwald:
Closing in favor of including in https://github.com/ava-labs/hypersdk/pull/1091

PR #1110, opened 2024-07-09
A little moving and renaming

richardpringle:
As the title suggests


x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -14,6 +14,22 @@ pub enum Error {
     Deserialization,
 }
 
+/// Gets the balance for the specified address
+/// # Panics
+/// Panics if there was an issue deserializing the balance
+#[must_use]
+pub fn get_balance(account: Address) -> u64 {
```
iFrostizz, line 21: The `get_balance` seems to rather belong to the `program.rs` file, is doesn't really deal with state.

iFrostizz (dismissed):
Just a little moving suggestion

iFrostizz (approved):

x/programs/rust/wasmlanche-sdk/src/state.rs
iFrostizz: It wouldn't deal with `program` either in the case of an EOA, so it is left here.

PR #1115, opened 2024-07-10
Fix usage of feature

richardpringle:
Not sure how the simulator was compiling before


iFrostizz:
Do we actually need it ? I don't see any code behind the `debug` feature 🤔

iFrostizz:
> Do we actually need it ? I don't see any code behind the `debug` feature 🤔

Wow my bad, I thought that it was added to the `simulator` package.

iFrostizz (approved):

PR #1118, opened 2024-07-10
Pass actor address directly to simulator

iFrostizz:
In the process of stopping to use keys with the simulator since it is handy for interacting with it from the command line, but we are moving away from that. The actor address is directly passed to the simulator.

### DIFF
--- a/x/programs/cmd/simulator/test/context_injection/src/lib.rs
+++ b/x/programs/cmd/simulator/test/context_injection/src/lib.rs
@@ -112,7 +112,7 @@ mod tests {
             .unwrap();
 
         let mut test_context = TestContext::from(program_id);
-        test_context.actor_key = Some(actor_key);
+        test_context.actor = Address::new([1; Address::LEN]);
 
         let response = simulator
             .run_step(&Step {
--- a/x/programs/rust/wasmlanche-sdk/src/types.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/types.rs
@@ -1,4 +1,4 @@
-use borsh::BorshDeserialize;
+use borsh::{BorshDeserialize, BorshSerialize};
 
 /// Byte length of an action ID.
 pub const ID_LEN: usize = 32;
@@ -9,7 +9,7 @@ pub type Gas = u64;
 
 /// A struct that enforces a fixed length of 32 bytes which represents an address.
 #[cfg_attr(feature = "debug", derive(Debug))]
-#[derive(Clone, Copy, PartialEq, Eq, borsh::BorshSerialize, BorshDeserialize, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, BorshSerialize, BorshDeserialize, Hash)]
 pub struct Address([u8; Self::LEN]);
 
 impl Address {
--- a/x/programs/rust/wasmlanche-sdk/src/types.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/types.rs
@@ -1,4 +1,5 @@
 use borsh::BorshDeserialize;
+#[cfg(feature = "test-utils")]
 use serde::Serialize;
 
 /// Byte length of an action ID.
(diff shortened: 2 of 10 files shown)


x/programs/cmd/simulator/test/context_injection/src/lib.rs
```diff
@@ -112,7 +112,7 @@ mod tests {
             .unwrap();
 
         let mut test_context = TestContext::from(program_id);
-        test_context.actor_key = Some(actor_key);
+        test_context.actor = Address::new([1; Address::LEN]);
```
iFrostizz, line 115: I actually tried to implement an `Address::random` function. That didn't worked, at all...
1. There is no way to have the `random` function being `#[cfg(test)]` and using it across crates (typically example, since they are importing the simulator)
2. No way to have it using a custom `test-utils` flag because then the WASM target will complain that it has the `rand` crate.
It's probably okay this way for now.

iFrostizz: I have instead created an `Address::from_str` that takes in a name and convert it to bytes behind the `test-utils` feature. Coming in the next PR!

x/programs/rust/wasmlanche-sdk/src/types.rs
```diff
@@ -1,4 +1,5 @@
-use borsh::{BorshDeserialize, BorshSerialize};
```
richardpringle, line 1: This change doesn't make sense

x/programs/rust/wasmlanche-sdk/src/types.rs
```diff
@@ -9,8 +10,8 @@ pub type Gas = u64;
 
 /// A struct that enforces a fixed length of 32 bytes which represents an address.
 #[cfg_attr(feature = "debug", derive(Debug))]
-#[derive(Clone, Copy, PartialEq, Eq, BorshSerialize, BorshDeserialize, Hash)]
-pub struct Address([u8; Self::LEN]);
+#[derive(Clone, Copy, PartialEq, Eq, borsh::BorshSerialize, BorshDeserialize, Serialize, Hash)]
+pub struct Address(#[serde(serialize_with = "<[_]>::serialize")] [u8; Self::LEN]);
```
richardpringle, line 14: Should this be limited to `cfg(test)`?

x/programs/cmd/simulator/test/context_injection/src/lib.rs
richardpringle: Why don't you just return an address when you create a key. What you're doing here doesn't really make sense and I don't quite understand your explanation. 

>I have instead created an Address::from_str that takes in a name and convert it to bytes behind the test-utils feature. Coming in the next PR!

I don't quite understand this either.

Wouldn't it be easier if you just passed in an `Address` to create_key?

x/programs/rust/wasmlanche-sdk/src/types.rs
```diff
@@ -1,4 +1,5 @@
-use borsh::{BorshDeserialize, BorshSerialize};
+use borsh::BorshDeserialize;
+use serde::Serialize;
```
richardpringle, line 2: I think you can do what you're looking for with `AsRef<[u8]>` instead of making `Address` `Serialize`.

richardpringle (changes requested):

x/programs/rust/wasmlanche-sdk/src/types.rs
iFrostizz: I haven't found a solution to disambiguate the `<[_]>::serialize` to be from `serde::Serialize` or `borsh::BorshSerialize` so it was my workaround.

iFrostizz: If you have a suggestion I'd be really happy because it's not looking the best.

x/programs/rust/wasmlanche-sdk/src/types.rs
iFrostizz: Sure I can do that. Not for `#[cfg(test)]` though because it doesn't support inter-crate compilation sadly, that could be the `test-utils` feature though

x/programs/cmd/simulator/test/context_injection/src/lib.rs
iFrostizz: The `create_key` endpoint was meant for manual interaction with the simulator. It was just constraining for no reason when interacting with it programmatically, because we had to store named keys.
In the simulator, we don't really need private components, it's much easier if everything can be impersonated by just injecting an address to the `testContext`.

x/programs/rust/wasmlanche-sdk/src/types.rs
iFrostizz: I'm not sure why, `Address` has to be serialized in the `actor` field of the `TestContext`

x/programs/cmd/simulator/test/context_injection/src/lib.rs
iFrostizz: I would leave private components to the vm

x/programs/rust/wasmlanche-sdk/src/types.rs
iFrostizz: Thanks for the merged branch !

x/programs/rust/wasmlanche-sdk/src/types.rs
iFrostizz: You have written a manual implementation of `TestContext`, that's definitely better. Thanks!

richardpringle (approved):

PR #1119, opened 2024-07-10
Remove NetworkID and ChainID from Rules

joshua-kim:
I don't think there's a use-case where a VM defines a set of fork rules that doesn't use the snow context specified `ChainID` or `NetworkID`.  Hypersdk `VM` already knows this information from it's own `snow.Context`, so there isn't a point in having the application define it a second time.


examples/tokenvm/cmd/token-faucet/manager/manager.go
```diff
@@ -129,7 +129,8 @@ func (m *Manager) sendFunds(ctx context.Context, destination codec.Address, amou
 		To:    destination,
 		Asset: ids.Empty,
 		Value: amount,
-	}}, m.factory)
+	}}, m.factory,
```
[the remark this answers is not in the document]
joshua-kim: Will revert

chain/block.go
```diff
@@ -419,7 +419,7 @@ func (b *StatelessBlock) innerVerify(ctx context.Context, vctx VerifyContext) er
 	}
 
 	// Process transactions
-	results, ts, err := b.Execute(ctx, b.vm.Tracer(), parentView, feeManager, r)
+	results, ts, err := b.Execute(ctx, b.vm.ChainID(), b.vm.Tracer(), parentView, feeManager, r)
```
joshua-kim, line 422: Need to use the executing VM's `chainID` here to guard against replay attacks (signer signs the chainID they issue so this will guard against replay attacks across vm instances)

chain/dependencies.go
```diff
@@ -116,11 +118,6 @@ type Mempool interface {
 
 // TODO: add fixed rules as a subset of this interface
 type Rules interface {
-	// Should almost always be constant (unless there is a fork of
-	// a live network)
-	NetworkID() uint32
-	ChainID() ids.ID
-
```
aaronbuchwald, line 123: Slightly arbitrary, but `NetworkID` and `ChainID` fit nicely under the concept of rules imo.

Currently, the chainID is only used for replay protection, but this would likely be exposed directly to the VM in the future as well. For example, in the Warp precompile for Subnet-EVM/Coreth, we expose this directly via `getBlockchainID`: https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/30-avalanche-warp-x-evm#precompile-implementation.

joshua-kim: I agree that the VM needs to know the `NetworkID` and `ChainID`, but I just don't know if `Rules` is the right place to put it. I understand the `Rules` abstraction as a config that exists per fork/time ([ref](https://github.com/ava-labs/hypersdk/blob/bed09991e604939b5dda29ec3e28a2c00d39e717/vm/dependencies.go#L94), this interface signature suggests to me that the point of rules is that it changes over time as time is a parameter). `NetworkID` and `ChainID` are always going to be constant throughout the lifecycle of a VM, so I don't see a reason to bind this to the `Rules` interface since the `Controller` should just be using the `ChainID` and `NetworkID` in `snow.Context` during `Initialize`. Basically I think that this is something that `avalanchego` defines, not something the `Controller` defines.

aaronbuchwald: The comment above explains why. If there's a fork of a live network, you may want to switch the networkID/chainID used from the fork point on, so that transactions sent on one network cannot be replayed on the other.

PR #1121, opened 2024-07-10
Remove `key_create` endpoint and use `Address` everywhere

iFrostizz:
As the title says


iFrostizz:
This PR should be merged after https://github.com/ava-labs/hypersdk/pull/1118

iFrostizz:
TODO: also remove the `Key` enum as well as the `Param` variant

samliok (approved):

x/programs/rust/wasmlanche-sdk/src/types.rs
```diff
@@ -24,6 +24,19 @@ impl Address {
...
+    #[must_use]
+    /// # Panics
+    /// Panics if the length of the bytes representation of the name is longer than [`Address::LEN`]
+    pub fn from_str(name: &str) -> Self {
+        let mut bytes = [0u8; Self::LEN];
+        let name_bytes = name.as_bytes();
+        assert!(name_bytes.len() <= Self::LEN, "passed name is too long");
+        bytes[0..name_bytes.len()].copy_from_slice(name_bytes);
+        Self::new(bytes)
+    }
```
richardpringle, line 39: This isn't a good way to do this generically. It would be fine for a particular program if the author wanted to do this, but it shouldn't be the encouraged way to create addresses. 

Why not just use addresses? One can easily generate random addresses with the `rand` crate. If you need to expose a `From<[u8; 32]>` implementation to make that easier, just do that. 

I don't want to add a whole other feature flag just for this one function

richardpringle (changes requested):

x/programs/rust/wasmlanche-sdk/src/types.rs
iFrostizz: Solved by just using `Address::new` in tests. The `From<[u8; 32]>` would have been unnecessary.

samliok (approved):

samliok (approved):

richardpringle (approved):

Issue #1123, opened 2024-07-11
Make `DeferDeserialize` borsh-serialize as a white box


iFrostizz:
`DeferDeserialize` is used to be able to hold raw bytes and to delay the deserialization of the inner content.
This struct is currently annotated with `BorshSerialize` in https://github.com/ava-labs/hypersdk/pull/1047 which makes it annoying to use when it's returned from a function (which is the primary use of this `BorshSerialize`) because it prepends the returned bytes with their length as a little-endian u32 number. Here is the fix:

```rust
impl BorshSerialize for DeferDeserialize {
    fn serialize<W: std::io::prelude::Write>(&self, writer: &mut W) -> std::io::Result<()> {
        writer.write_all(&self.0)?;
        Ok(())
    }
}
```

PR #1124, opened 2024-07-11
Manually implement `BorshSerialize` for `DeferDeserialize`

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1123

Implement `BorshSerialize` for `DeferDeserialize` such that it doesn't encode the length. This is useful when `DeferDeserialize` is returned from functions such as ones that execute an arbitrary call and return the raw bytes as returned by the called function.

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/program.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/program.rs
@@ -17,7 +17,7 @@ pub struct DeferDeserialize(Vec<u8>);
 impl BorshSerialize for DeferDeserialize {
     /// # Errors
     /// Returns a [`std::io::Error`] if there was an issue writing
-    fn serialize<W: std::io::prelude::Write>(&self, writer: &mut W) -> std::io::Result<()> {
+    fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
         writer.write_all(&self.0)?;
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -14,6 +14,15 @@ use thiserror::Error;
 #[cfg_attr(feature = "debug", derive(Debug))]
 pub struct DeferDeserialize(Vec<u8>);
 
+impl BorshSerialize for DeferDeserialize {
+    /// # Errors
+    /// Returns a [`std::io::Error`] if there was an issue serializing the value
```
richardpringle, line 19:
```suggestion
    /// Returns a [`std::io::Error`] if there was an issue writing
```

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -14,6 +14,15 @@ use thiserror::Error;
...
+    fn serialize<W: std::io::prelude::Write>(&self, writer: &mut W) -> std::io::Result<()> {
```
richardpringle, line 20:
```suggestion
    fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
```

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -14,6 +14,15 @@ use thiserror::Error;
...
+        writer.write_all(&self.0)?;
+        Ok(())
```
richardpringle, line 22:
```suggestion
        writer.write_all(&self.0)
```

richardpringle (changes requested):

richardpringle (approved):

PR #1127, opened 2024-07-11
Refactor block + state databases out of Controller

joshua-kim:
We currently allow the Controller to define it's own databases that merkle state + vm data is stored in. This PR removes this from the implementation and bubbles it down into the hypersdk to reduce `Initialize` complexity in the VM. We can always expose configuration options in the future if this is something people want to override, but I suspect most people will be happy with the default.

This PR changes the structure of chain data. Because the databases for VM data + state are now bubbled down, we give the Controller its own dedicated `ChainDataDir` we override.


vm/vm.go
```diff
@@ -45,6 +48,12 @@ import (
 	avasync "github.com/ava-labs/avalanchego/x/sync"
 )
 
+const (
+	blockDB   = "blockdb"
+	stateDB   = "statedb"
+	vmDataDir = "vm"
```
joshua-kim, line 54: Open to considerations for this directory's name. `Controller` as a name seems weird to me so left it as `vm`.

examples/morpheusvm/controller/controller.go
```diff
@@ -42,7 +43,7 @@ type Controller struct {
 
 	metrics *metrics
 
-	metaDB database.Database
```
joshua-kim, line 45: Renaming this db variable since it's now the only db the VM knows about.

examples/tokenvm/storage/storage.go
```diff
@@ -40,10 +40,10 @@ type ReadState func(context.Context, [][]byte) ([][]byte, []error)
 // 0x5/ (hypersdk-fee)
 
 const (
-	// metaDB
+	// VM state
 	txPrefix = 0x0
 
-	// stateDB
+	// HyperSDK state
```
samliok, line 46: what is the difference between VM state and hyperSDK state?

samliok (approved):

examples/tokenvm/storage/storage.go
joshua-kim: VM state is the database that is used by the implementation VM (hypersdk does not know about this) + HyperSDK state is known by hypersdk and contains data referenced by state keys. The database for the latter is defined by hypersdk but they state keys are defined by the VM

joshua-kim: Updated comment

vm/dependencies.go
```diff
@@ -81,9 +80,6 @@ type Controller interface {
 		builder builder.Builder,
 		gossiper gossiper.Gossiper,
 		// TODO: consider splitting out blockDB for use with more experimental
```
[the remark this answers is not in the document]
joshua-kim: yeah ;-; will remove.

vm/vm.go
```diff
@@ -159,13 +168,46 @@ func (vm *VM) Initialize(
...
+		AVAXAssetID:    vm.snowCtx.AVAXAssetID,
+		Log:            vm.snowCtx.Log,
+		Keystore:       vm.snowCtx.Keystore,
+		SharedMemory:   vm.snowCtx.SharedMemory,
+		BCLookup:       vm.snowCtx.BCLookup,
+		Metrics:        vm.snowCtx.Metrics,
+		WarpSigner:     vm.snowCtx.WarpSigner,
+		ValidatorState: vm.snowCtx.ValidatorState,
+		ChainDataDir:   filepath.Join(vm.snowCtx.ChainDataDir, vmDataDir),
+	}
```
aaronbuchwald, line 204: (not necessary for this PR) thoughts on adding one more layer of nesting here ie.

outerVM
- block
- state

innerVM
- vm

examples/tokenvm/storage/storage.go
```diff
@@ -40,10 +41,10 @@ type ReadState func(context.Context, [][]byte) ([][]byte, []error)
 // 0x5/ (hypersdk-fee)
 
 const (
-	// metaDB
+	// VM state
 	txPrefix = 0x0
 
-	// stateDB
+	// State keys
```
aaronbuchwald, line 47: This naming confuses me a bit. I think the difference here is indexes (currently called meta, which I'd say is also a slight misnomer) and StateDB, which is really just the active state of the chain.

I also don't think we should use state keys here as it overlaps with the state keys specified by actions.

Could we rename these as:
```
// Indexes

// Active state
```

(sorry, can't add a suggestion here)

examples/morpheusvm/storage/storage.go
```diff
@@ -37,10 +37,10 @@ type ReadState func(context.Context, [][]byte) ([][]byte, []error)
 // 0x3/ (hypersdk-fee)
 
 const (
-	// metaDB
+	// VM state
 	txPrefix = 0x0
 
-	// stateDB
+	// State keys
```
aaronbuchwald, line 43: same renaming comment here

examples/tokenvm/storage/storage.go
joshua-kim: Yeah I don't even know what `meta` means. I guess it means `metadata` but it should probably be called `txDB`.

vm/vm.go
joshua-kim: Don't really have thoughts on this... if someone has strong opinions I can be convinced for or against this though...

aaronbuchwald (approved):

PR #1129, opened 2024-07-11
Remove unused mocks

joshua-kim:
Deletes unused mocks. I would prefer if we stayed away from `mockgen` entirely but the remaining mocks actually have usages.


aaronbuchwald (approved):
LGTM once CI passes

PR #1132, opened 2024-07-12
Some cleanup in the workspace

iFrostizz:
Make the `wasmlanche-sdk` the only default member so it's the one that pops up when opening the documentation.
Update some documentation.


samliok (dismissed):

Cargo.toml
```diff
@@ -11,6 +11,9 @@ members = [
     "x/programs/test/programs/*",
     "x/programs/cmd/simulator/test/*",
 ]
+default-members = [
+    "x/programs/rust/wasmlanche-sdk",
+]
```
richardpringle, line 16: From the docs: 
>Packages to operate on when a specific package wasn’t selected.

I'm not sure why we would want this, we want to operate on the entire workspace when a package isn't specified. 

```suggestion
```

x/programs/rust/wasmlanche-sdk/src/lib.rs
```diff
@@ -33,6 +33,7 @@ pub struct Context<K = ()> {
     action_id: Id,
 }
 
+/// Getters of the `Context` components
```
richardpringle, line 36:
```suggestion
```
These aren't necessarily just getters.

richardpringle (changes requested):
>Panics if there was an issue deserializing the account.

Typically, you don't put a period at the end of text if the text isn't a complete sentence. The above is a sentence fragment as it's missing a subject and is therefore dramatically incorrect to put a period.

Cargo.toml
richardpringle: Yeah, If I run `cargo test` it doesn't run for the whole workspace. This isn't desired behaviour.

richardpringle:
I closed this PR as none of the changes really make sense. It was gramatically correct to put a period on at least one of the doc-comments, but I think it's better to be consistent and just leave them off. The comment about getters is correct at the moment, but too narrowly describes what could be added to that `impl` block as well.

PR #1133, opened 2024-07-12
[x\programs] Remove old balance code and add an additional test

dboehm-avalabs:
I had worked on a version of the balance code that refunded unspent value that was sent to another program, but took most of the code out.  This line should have also been removed, but was missed.  Also add a test to confirm that the balance moves from program 1 to program 2 on calls that send value.


x/programs/runtime/import_balance_test.go
```diff
@@ -14,6 +14,38 @@ import (
...
+	defer cancel()
+	program := newTestProgram(ctx, "balance")
+	r := program.Runtime
+	stateManager := r.StateManager.(test.StateManager)
+	stateManager.Balances[program.Address] = 3
+
+	// create a new instance of the balance program
+	newInstanceAddress := codec.CreateAddress(0, ids.GenerateTestID())
+	programID := stateManager.AccountMap[program.Address]
+	r.StateManager.(test.StateManager).AccountMap[newInstanceAddress] = programID
```
iFrostizz, line 30: You could use public methods for more consistency

```suggestion
	programID, err := stateManager.GetAccountProgram(ctx, program.Address)
	require.NoError(err)
	r.StateManager.SetAccountProgram(ctx, newInstanceAddress, programID)
```

iFrostizz (approved):

iFrostizz (approved):

PR #1134, opened 2024-07-12
Auto cleanup

sukantoraymond:


examples/morpheusvm/scripts/deploy.devnet.sh
```diff
@@ -174,7 +174,8 @@ EOF
 # Setup devnet
 CLUSTER="vryx-$(date +%s)"
 function cleanup {
-  echo -e "\n\n${RED}run this command to destroy the devnet:${NC} ${TMPDIR}/avalanche node destroy ${CLUSTER}\n"
+  echo -e "\n\n${RED}destroying the devnet, running:${NC} ${TMPDIR}/avalanche node destroy ${CLUSTER}\n"
+  ${TMPDIR}/avalanche node destroy ${CLUSTER} -y
```
aaronbuchwald, line 178: Won't this terminate the whole network when the script completes as well, whereas it should be left running in that case and cleaned up if the script fails or is cancelled part way through?

Issue #1138, opened 2024-07-14
Implement Action Test Suite - Execute


aaronbuchwald:
Implement a test suite helper for the `chain.Action` type to unit test actions (outside the context of a full HyperVM).

This should look roughly like adding a `test_action.go` file to the chain directory:

```go

type ActionTest struct {
	Rules           Rules
	State           state.Mutable
	Timestamp       uint64
	Actor           codec.Address
	ActionID        ids.ID
	ExpectedOutputs [][]byte
	ExpectedErr     error
}

type ActionTestSuite struct {
	Tests  map[string]ActionTest
	Action chain.Action
}

func (suite *ActionTestSuite) Run(t *testing.T) {
	for name, test := range suite.Tests {
		t.Run(name, func(t *testing.T) {
			require := require.New(t)
			outputs, err := suite.Action.Execute(
				context.TODO(),
				test.Rules,
				test.State,
				test.Timestamp,
				test.Actor,
				test.ActionID,
			)
			require.Equal(test.ExpectedOutputs, outputs)
			require.Equal(test.ExpectedErr, err)
		})
	}
}
```

We should then integrate this first into MorpheusVM and then TokenVM.

PR #1139, opened 2024-07-15
Implement action test suite helper

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1138

Create an action test suite helper that would be integrated into MorpheusVM and TokenVM and could be used for VM authors to simplify the parametrized testing setup.


chain/action_test_helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	ExpectedOutputs   [][]byte
+	ExpectedErr       error
+	ContainsErrString string
+}
+
+type ActionTestSuite struct {
+	Tests map[string]ActionTest
+
+	Setup    func()
+	Teardown func()
```
iFrostizz, line 35: I'm actually not sure if we will need those, I can remove them for this PR and we can add them back later if so.

chain/action_test_helpers.go
```diff
@@ -0,0 +1,78 @@
...
+
+				if test.ExpectedErr != nil {
+					require.Equal(err, test.ExpectedErr)
+				}
+				if len(test.ContainsErrString) > 0 {
+					require.Contains(err.Error(), test.ContainsErrString)
+				}
+			}
+
+			require.Equal(output, test.ExpectedOutputs)
```
iFrostizz, line 71: err and output should be mutually exclusive but it doesn't hurt to still check it at the end, even if `err != nil`

chain/action_test_helpers.go
iFrostizz: Removed for now. These hook functions could be useful if we pass a reference to the storage in order to be able to assert stuff on the written storage.

iFrostizz:
> Maybe include integration with existing test suites in this PR? I don't think its ideal to have to review something without also getting to see how it is used.

Absolutely makes sense ! Thanks :)

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,80 @@
...
+			},
+			State:       ts.NewView(map[string]state.Permissions{}, map[string][]byte{}),
+			ExpectedErr: tstate.ErrInvalidKeyOrPermission,
+		},
+		"NotEnoughBalance": {
+			Action: &Transfer{
+				To:    codec.EmptyAddress,
+				Value: 1,
+			},
+			Actor: codec.EmptyAddress,
```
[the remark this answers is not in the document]
iFrostizz: Good catch! I thought that it would make the tests reach further code paths but the actor just defaults to the empty value so I will just remove them

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,80 @@
...
+			State: func() state.Mutable {
+				keys := make(state.Keys)
+				k := storage.BalanceKey(codec.EmptyAddress)
+				keys.Add(string(k), state.All)
+				stor := map[string][]byte{}
+				tsv := ts.NewView(keys, stor)
+				b := make([]byte, 8)
+				binary.LittleEndian.PutUint64(b, uint64(1))
+				require.NoError(tsv.Insert(context.TODO(), k, b))
```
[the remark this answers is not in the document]
iFrostizz: Yes and no - To solve this we may need to move this part of the code as a function pointer that would be called in the iteration of the test suite, which doesn't seem like a good idea to me at the moment. If there is something easier I missed please let me know !

iFrostizz:
I'm not satisfied with the way we do storage. For instance, if we had a test-only in-memory database that implements `state.Mutable`, we could just use the `SetBalance` helper functions. The one that would be the most appropriate for that is the `TestDB` that is defined in `tstate/tstate_test.go` but it cannot be accessed from `examples/morpheusvm/actions/transfer_test.go`.
Since this struct is not used, I'm just wondering if it would make sense to move it to the test helpers so that we can access it from other tests as well. Would it be shockingly bad if we can access a test-helpers db in the codebase ?

iFrostizz:
Implemnted an `InMemoryStore` in [900cf09](https://github.com/ava-labs/hypersdk/pull/1139/commits/900cf09d7821d9d7c40faf4f189ab0472a78541d) which also contains a public `Storage` field that is passed to the tstate once the initial state has been set.

examples/tokenvm/storage/storage.go
iFrostizz, line 1: Linting has done this, and CI passes. Should we leave this alone ?

examples/tokenvm/tests/integration/integration_test.go
iFrostizz, line 1: Same as https://github.com/ava-labs/hypersdk/pull/1139/files#r1679210787

examples/tokenvm/storage/storage.go
ARR4N: The new version feels idiomatic to me, so I'd accept the change.

chain/action_test_helpers.go
ARR4N: You won't reach the output check because you're using `require`; swapping to `assert` will use `t.Errorf()` under the hood instead of `t.Fatalf()`, but I don't think you should.

> it doesn't hurt to still check it at the end, even if `err != nil`

I think it does actually, even if very minor, because it forces every error path to have a strict return value (presumably `nil`) when in reality the value is undefined.

[More context](https://google.github.io/styleguide/go/decisions#keep-going) should anyone be interested.

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,74 @@
...
+		},
+		"SimpleTransfer": {
+			Action: &Transfer{
+				To:    codec.EmptyAddress,
+				Value: 1,
+			},
+			State: func() state.Mutable {
+				keys := make(state.Keys)
+				store := chain.NewInMemoryStore()
+				require.NoError(storage.SetBalance(context.TODO(), store, codec.EmptyAddress, 1))
```
ARR4N, line 61: `context.Background()` is the [correct value](https://google.github.io/styleguide/go/decisions#keep-going:~:text=In%20tests%3A%20TestXXX%2C%20BenchmarkXXX%2C%20FuzzXXX) to use here.

`TODO()` is generally used for when a `Context` isn't available and a future refactoring is needed to introduce it. I think it's so you can do a global plain-text search for the phrase `TODO`.

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,74 @@
...
+				Value: 0,
+			},
+			ExpectedErr: ErrOutputValueZero,
+		},
+		"InvalidStateKey": {
+			Action: &Transfer{
+				To:    codec.EmptyAddress,
+				Value: 1,
+			},
+			State:       ts.NewView(map[string]state.Permissions{}, map[string][]byte{}),
```
ARR4N, line 37: (nit) It's a little cleaner if you just use `make(State.Keys)` for the first argument.

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,74 @@
...
+			Action: &Transfer{
+				To:    codec.EmptyAddress,
+				Value: 1,
+			},
+			State: func() state.Mutable {
+				keys := make(state.Keys)
+				store := chain.NewInMemoryStore()
+				require.NoError(storage.SetBalance(context.TODO(), store, codec.EmptyAddress, 1))
+				keys.Add(string(emptyBalanceKey), state.All)
+				tsv := ts.NewView(keys, store.Storage)
```
ARR4N, line 63: No need to set the variable here. Just `return ts.NewView(...)`

ARR4N (changes requested):

chain/action_test_helpers.go
```diff
@@ -0,0 +1,77 @@
...
+
+type ActionTestSuite struct {
+	Tests map[string]ActionTest
+}
+
+func (suite *ActionTestSuite) Run(t *testing.T) {
+	for testName := range suite.Tests {
+		t.Run(testName, func(t *testing.T) {
+			require := require.New(t)
+			test := suite.Tests[testName]
```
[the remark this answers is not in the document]
iFrostizz: Thanks!

chain/action_test_helpers.go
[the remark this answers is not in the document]
iFrostizz: I think that it would be nice to use conditional compilation for that later, I will move this to a new `chaintesting` package. Having test-only stuff that could be imported in production is definitely no good.

chain/action_test_helpers.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/state"
+)
+
+var _ state.Mutable = (*InMemoryStore)(nil)
+
+type InMemoryStore struct {
```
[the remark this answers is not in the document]
iFrostizz: Added comments for `InMemoryStore`, `ActionTest`, and `ActionTest.Run`

chain/action_test_helpers.go
```diff
@@ -0,0 +1,77 @@
...
+	ExpectedOutputs [][]byte
+	ExpectedErr     error
+}
+
+type ActionTestSuite struct {
+	Tests map[string]ActionTest
+}
+
+func (suite *ActionTestSuite) Run(t *testing.T) {
+	for testName := range suite.Tests {
```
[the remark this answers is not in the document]
iFrostizz: Thanks, I have applied your for `name, test := range suite.Tests` suggestion which solves this.

examples/morpheusvm/actions/transfer_test.go
iFrostizz: Thanks. Replaced the `context.TODO()` by a `context.Background()`.

examples/morpheusvm/actions/transfer_test.go
iFrostizz: THanks!

examples/morpheusvm/actions/transfer_test.go
iFrostizz: Done, thanks!

chaintesting/action_test_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/state"
+)
+
+var _ state.Mutable = (*InMemoryStore)(nil)
+
+// InMemoryStore is a storage that acts as a wrapper around a map and implements state.Mutable.
+type InMemoryStore struct {
+	Storage map[string][]byte
+}
```
aaronbuchwald, line 24: Could we put `InMemoryStore` in the same package as `state.Mutable` ?

Could we also export the constructor, so that tests can create a new in-memory store, but leave the type un-exported?

chaintesting/action_test_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+	Actor     codec.Address
+	ActionID  ids.ID
+
+	ExpectedOutputs [][]byte
+	ExpectedErr     error
+}
+
+type ActionTestSuite struct {
+	Tests map[string]ActionTest
+}
```
aaronbuchwald, line 67: Could we move this to `chain/` so that it's in the same package as the interface it's used to test?

chaintesting/action_test_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Run execute all tests from the test suite and make sure that assertions passes.
```
aaronbuchwald, line 69:
```suggestion
// Run execute all tests from the test suite and make sure all assertions pass.
```

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,73 @@
...
+		},
+		"InvalidStateKey": {
+			Action: &Transfer{
+				To:    codec.EmptyAddress,
+				Value: 1,
+			},
+			State:       ts.NewView(make(state.Keys), map[string][]byte{}),
+			ExpectedErr: tstate.ErrInvalidKeyOrPermission,
+		},
+		"NotEnoughBalance": {
```
aaronbuchwald, line 40: Could we add a test case for an overflow / `ErrInvalidBalance` as well?

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,73 @@
...
+				Value: 1,
+			},
+			State: func() state.Mutable {
+				keys := make(state.Keys)
+				store := chaintesting.NewInMemoryStore()
+				require.NoError(storage.SetBalance(context.Background(), store, codec.EmptyAddress, 1))
+				keys.Add(string(emptyBalanceKey), state.All)
+				return ts.NewView(keys, store.Storage)
+			}(),
+		},
```
aaronbuchwald, line 65: Could we add an optional assertion to the test helper, so that this test can assert the balance of the receiver is updated correctly?

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,73 @@
...
+	emptyBalanceKey := storage.BalanceKey(codec.EmptyAddress)
+
+	tests := map[string]chaintesting.ActionTest{
+		"ZeroTransfer": {
+			Action: &Transfer{
+				To:    codec.EmptyAddress,
+				Value: 0,
+			},
+			ExpectedErr: ErrOutputValueZero,
+		},
```
aaronbuchwald, line 31: Could we explicitly set the `Actor` field of each `ActionTest` ?

It's currently accessed by these tests and using `codec.EmptyAddress` since it's not populated.

chaintesting/action_test_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+func (s *InMemoryStore) Remove(_ context.Context, key []byte) error {
+	delete(s.Storage, string(key))
+	return nil
+}
+
+// ActionTest is a single parameterized test. It calls Execute on the action with the passed parameters
+// and checks that the assertions passes.
```
aaronbuchwald, line 51:
```suggestion
// and checks that all assertions pass.
```

chaintesting/action_test_helpers.go
iFrostizz: @ARR4N told me to move it from `chain/` to `chaintesting/` in https://github.com/ava-labs/hypersdk/pull/1139#discussion_r1679256896 since these are test-only constructs, to avoid them leaking to production.

examples/morpheusvm/actions/transfer_test.go
iFrostizz: Done!

examples/morpheusvm/actions/transfer_test.go
iFrostizz: Added an `Assertion` field that takes in a function pointer that gets passed the state so that assertions can be done after the test was ran.

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,73 @@
...
+			},
+			State: func() state.Mutable {
+				keys := make(state.Keys)
+				keys.Add(string(emptyBalanceKey), state.Read)
+				tsv := ts.NewView(keys, map[string][]byte{})
+				return tsv
+			}(),
+			ExpectedErr: storage.ErrInvalidBalance,
+		},
+		"SimpleTransfer": {
```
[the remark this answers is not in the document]
iFrostizz: good idea, I added a test that transfers the value to an address with many `1`.

examples/morpheusvm/actions/transfer_test.go
iFrostizz: Done !

chaintesting/action_test_helpers.go
iFrostizz: I don't know if we want it to be in a production package, or maybe moved to the `state` package with some compilation tags ?

chaintesting/action_test_helpers.go
```diff
@@ -0,0 +1,87 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package chaintesting
```
joshua-kim, line 4: nit: rename to `chaintest`

chaintesting/action_test_helpers.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+var _ state.Mutable = (*inMemoryStore)(nil)
+
+// inMemoryStore is a storage that acts as a wrapper around a map and implements state.Mutable.
+type inMemoryStore struct {
+	Storage map[string][]byte
+}
+
+func NewInMemoryStore() *inMemoryStore {
```
joshua-kim, line 26: I think it's return to return a non-exported type from a exported function. We should either return the interface or just export `inMemoryStore`. I recommend the latter so that the caller can decide to use it as a concrete type or as an interface

chaintesting/action_test_helpers.go
```diff
@@ -0,0 +1,87 @@
...
+	return &inMemoryStore{
+		Storage: make(map[string][]byte),
+	}
+}
+
+func (s *inMemoryStore) GetValue(_ context.Context, key []byte) ([]byte, error) {
```
joshua-kim, line 32: nit: make the receiver `i` (to be in-line with the struct name)

chaintesting/action_test_helpers.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (s *inMemoryStore) Remove(_ context.Context, key []byte) error {
+	delete(s.Storage, string(key))
+	return nil
+}
+
+// ActionTest is a single parameterized test. It calls Execute on the action with the passed parameters
+// and checks that all assertions pass.
+type ActionTest struct {
+	Action chain.Action
```
joshua-kim, line 53: I think it's more idiomatic to add a `string` field `Name` here instead of putting the test name as the key in `map[string]ActionTest` in the `Suite` struct below. I think if we do this we can make that map type just a slice.

examples/tokenvm/storage/storage.go
```diff
@@ -19,7 +19,6 @@ import (
 	"github.com/ava-labs/hypersdk/state"
 
 	smath "github.com/ava-labs/avalanchego/utils/math"
-
```
joshua-kim, line 22: nit: undo these diffs that aren't related

chaintesting/action_test_helpers.go
iFrostizz: Done!

chaintesting/action_test_helpers.go
iFrostizz: The point of the `NewInMemoryStore` was to be a constructor for the `inMemoryStore` struct, I don't understand why the value associated with the `inMemoryStore` could not be used as an interface, for instance we can currently call `SetBalance` with the parameter `mu state.Mutable` as type without any issue

chaintesting/action_test_helpers.go
iFrostizz: Replaced!

chaintesting/action_test_helpers.go
iFrostizz: Thanks ! So I have removed this `ActionTestSuite` type and instead passed the `[]ActionTest` as part of the `Run` function.

iFrostizz: (And it's now passing the `Name` as one of the fields of `ActionTest`.

examples/tokenvm/storage/storage.go
iFrostizz: reverted but these are either done automatically or when the `./scripts/test.lint.sh` is ran.

iFrostizz: Confirmed that this is the `./scripts/test.lint.sh` script doing this

chaintesting/action_test_helpers.go
joshua-kim: This compiles fine now because you're just using it as a variable:

```
db := NewInMemoryStore()
db.Put()
db.Get()
...
```

should compile. But because the concrete-type isn't exported outside of the package a caller can't do this for example:

```
type Foo struct {
    db *chaintest.inMemoryStore
}
```

Because that type isn't exported outside of the `chaintest` package so other packages don't know that it type is referring to. If we're going to return a type outside of the package via `NewInMemoryStore`, I think we should allow callers to know what that concrete type is. This go pattern is considered non-idiomatic (go linters will warn you with "Exported function with the unexported return type", my IDE does this automatically).

iFrostizz: I see ! Thanks a lot for the explanation !

joshua-kim (dismissed):

chaintesting/action_test_helpers.go
aaronbuchwald: Ah, will defer so this can be merged and start a separate conversation on what convention we should go with

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,126 @@
...
+				require.NoError(storage.SetBalance(context.Background(), store, codec.EmptyAddress, 1))
+				keys.Add(string(emptyBalanceKey), state.All)
+				keys.Add(string(storage.BalanceKey(oneAddr)), state.All)
+				return ts.NewView(keys, store.Storage)
+			}(),
+			Assertion: func(store state.Mutable) bool {
+				balance, err := storage.GetBalance(context.Background(), store, oneAddr)
+				require.NoError(err)
+				return balance == 1
+			},
```
aaronbuchwald, line 121: Can we add a check that the sender's address has been decreased correctly?

iFrostizz: Done, thanks!

chain/action_test_helpers.go
ARR4N: > I think we should keep using require, otherwise the test author needs to keep thinking about which failures will cause the test to fail vs where the test can continue.

That's actually the point. It avoids whack-a-mole fixes where you get partial information about problems. The test author really should be thinking about that because they're the one person with the greatest insight.

My rule of thumb is that errors use `require` because that almost certainly guarantees that all other returned values are incorrect. Incorrect `actual` or `got` values, however, are typically more suited to `assert`.

> Additionally, require is more concise than if err != nil style checks, which is needed to use t.Fatal.

`require` is equivalent to `t.Fatal` so there's no need for the `err != nil` check. The alternative is `assert`, which is equivalent to `t.Error`.

chaintesting/action_test_helpers.go
darioush: In a way, returning the unexported type is an advantage.
It forces the caller to either only use exported methods on the returned type, or declare an explicit interface on their side (caller side), which is the correct side for the dependency to live.
This makes it very easy to switch in a different callee package or to return a wrapped type later, since it avoids embedding types by name (either struct or interfaces) in the user (caller)'s code.

aaronbuchwald (changes requested):

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,128 @@
...
+				keys.Add(string(emptyBalanceKey), state.All)
+				keys.Add(string(storage.BalanceKey(oneAddr)), state.All)
+				return ts.NewView(keys, store.Storage)
+			}(),
+			Assertion: func(store state.Mutable) bool {
+				receiverBalance, err := storage.GetBalance(context.Background(), store, oneAddr)
+				require.NoError(err)
+				senderBalance, err := storage.GetBalance(context.Background(), store, codec.EmptyAddress)
+				require.NoError(err)
+				return receiverBalance == 1 && senderBalance == 0
```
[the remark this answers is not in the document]
ARR4N: +1 to adding a parameter, but I think it should be a `*testing.T` for flexibility. Users can wrap it if they want, but aren't forced to.

> We should also avoid using the local version of `require` within this function.

:100:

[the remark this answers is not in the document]
iFrostizz: Thanks for this feedback.
I have renamed the `require` that was used on the global test suite to `req` since it is less used and it is more natural to type `require`. Also removed the `bool` return from the `Assertion` function.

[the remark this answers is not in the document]
iFrostizz: Taking @ARR4N suggestion ! Thanks !

chaintest/action_test_helpers.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (i *InMemoryStore) GetValue(_ context.Context, key []byte) ([]byte, error) {
+	val, ok := i.Storage[string(key)]
+	if !ok {
+		return nil, database.ErrNotFound
+	}
+	return val, nil
+}
+
+func (i *InMemoryStore) Insert(_ context.Context, key []byte, value []byte) error {
```
ARR4N, line 40: What is the expected behaviour on a duplicate key? The interface isn't documented. Similarly, but for a missing key, when calling `Remove()`? Should these return errors?

iFrostizz: When looking at `Database`, `TStateView`, `DB`, all of these seem to have the same behaviour. In the case of stores that are using a database, they are returning an error if the inner database was closed. For the tstate, if the key does not have enough permissions.
For `Remove`, if the value is `nil` then no error (nil) is returned, which is reflected by this implementation because `delete` won't panic if the value is `nil`.

chaintest/action_test_helpers.go
```diff
@@ -0,0 +1,85 @@
...
+	Assertion func(state.Mutable) bool
+}
+
+// Run execute all tests from the test suite and make sure all assertions pass.
+func Run(t *testing.T, tests []ActionTest) {
+	for _, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
+			require := require.New(t)
+
+			output, err := test.Action.Execute(context.TODO(), test.Rules, test.State, test.Timestamp, test.Actor, test.ActionID)
```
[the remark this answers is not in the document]
iFrostizz: Thanks! I have replace the `context.TODO` by a `context.Background` and the test now checks that the `Err()` is not `nil` by using `require.NoError(ctx.Err())`. Is that what you intended ?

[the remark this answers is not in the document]
iFrostizz: Though if the context is passed from the `Run`, is that really okay to check for the done signal for each of the subtest ? Wouldn't it be better to send a fatal error so that it completely stops the test suite ? because the context is going to be passed to the subsequent call to `Execute` which would already be populated in the done channel ?
I'm wondering if the context shouldn't be created for each subtest again here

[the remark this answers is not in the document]
ARR4N: > Thanks for the litterature!

No problem

> I have replaced the context.TODO by a context.Background

The important thing is that this happens as far "out" as possible, i.e. the `TestXXX()` function (as you've done). The idea being that the users of the suite can supply their own.

> and the test now checks that the Err() is not nil by using require.NoError(ctx.Err()). Is that what you intended ?

No, you can completely ignore `ctx.Err()` and probably should as it's feasible that a test may want to test for this error (i.e. an error is expected).

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,131 @@
...
+			Name:  "SelfTransfer",
+			Actor: codec.EmptyAddress,
+			Action: &Transfer{
+				To:    codec.EmptyAddress,
+				Value: 1,
+			},
+			State: func() state.Mutable {
+				keys := make(state.Keys)
+				store := chaintest.NewInMemoryStore()
+				req.NoError(storage.SetBalance(context.Background(), store, codec.EmptyAddress, 1))
```
ARR4N, line 76: These should also be plumbed in from outside. Rule of thumb: if you're using a `context.Context` inside any function other than `main()`, `init()` or `{Test,Bench,Fuzz}XXX()` then that function should receive the context as its first argument.

If changing the function signature to accept the `Context` isn't feasible in that PR then use `context.TODO()` to record the fact that a refactoring is necessary.

chaintest/action_test_helpers.go
```diff
@@ -0,0 +1,83 @@
...
+// InMemoryStore is a storage that acts as a wrapper around a map and implements state.Mutable.
+type InMemoryStore struct {
+	Storage map[string][]byte
+}
+
+func NewInMemoryStore() *InMemoryStore {
+	return &InMemoryStore{
+		Storage: make(map[string][]byte),
+	}
+}
```
iFrostizz, line 30: We could have modified the API such that these methods never return an error for instance because all of them cannot really fail, in the case of `GetValue` then the check of `nil` could have just been used for the value but it's a bad idea for an API that tries to mimic other databases so it "simulates" errors.

examples/morpheusvm/actions/transfer_test.go
iFrostizz: Thank you so much for https://github.com/ava-labs/hypersdk/pull/1170 !

chaintest/action_test_helpers.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/state"
+)
+
+var _ state.Mutable = (*InMemoryStore)(nil)
+
+// InMemoryStore is a storage that acts as a wrapper around a map and implements state.Mutable.
```
aaronbuchwald, line 21:
```suggestion
// InMemoryStore is an in-memory implementation of `state.Mutable`
```

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,137 @@
...
+				return ts.NewView(keys, store.Storage)
+			}(),
+			Assertion: func(ctx context.Context, t *testing.T, store state.Mutable) {
+				require := require.New(t)
+				receiverBalance, err := storage.GetBalance(ctx, store, oneAddr)
+				require.NoError(err)
+				senderBalance, err := storage.GetBalance(ctx, store, codec.EmptyAddress)
+				require.NoError(err)
+				require.Equal(receiverBalance, uint64(1))
+				require.Equal(senderBalance, uint64(0))
```
aaronbuchwald, line 121:
```suggestion
				require := require.New(t)
				receiverBalance, err := storage.GetBalance(ctx, store, oneAddr)
				require.NoError(err)
				require.Equal(receiverBalance, uint64(1))
				senderBalance, err := storage.GetBalance(ctx, store, codec.EmptyAddress)
				require.NoError(err)
				require.Equal(senderBalance, uint64(0))
```

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func createAddressWithByte(b byte) (codec.Address, error) {
+	addrSlice := make([]byte, codec.AddressLen)
+	for i := range addrSlice {
+		addrSlice[i] = b
+	}
+	return codec.ToAddress(addrSlice)
+}
```
aaronbuchwald, line 137: Optional for this PR - we should move this utility to a separate shared test package since other VMs will need this as well

aaronbuchwald (changes requested):

examples/morpheusvm/actions/transfer_test.go
iFrostizz: Opened https://github.com/ava-labs/hypersdk/issues/1186

aaronbuchwald (approved):

PR #1140, opened 2024-07-15
Remove rejected from controller interface

aaronbuchwald:
This PR removes the `Rejected` call from the `Controller` interface because it is unused and there's no current use case for the controller to index rejected data. If there's a use case in the future, it's easy to add it back.


scripts/mocks.mockgen.txt
```diff
@@ -1,3 +1,2 @@
-github.com/ava-labs/hypersdk/chain=Auth=chain/mock_auth.go
```
aaronbuchwald, line 1: Removed after running mockgen script because this was unused

joshua-kim (approved):

PR #1141, opened 2024-07-15
Pass Reference of External Call Context

samliok:
Forces `ExternalCallContext` to be passed by reference. This allows programs to reuse the same context without having to create a new one for each external call.


iFrostizz:
I'm just wondering if it would make sense to also borrow the fields inside of the `ExternalCallContext`, for instance you cannot do:

```rust
#[public]
pub fn inc(_: Context, external: Program, address: Address) {
    let ctx = ExternalCallContext::new(external, 1_000_000, 0);
    counter::inc(&ctx, address, 1);
    let ctx = ExternalCallContext::new(external, 1_000_000, 0);
    counter::inc(&ctx, address, 1);
}
```

because `external` is moved since `Program` doesn't implement `Copy` nor we (probably) want

iFrostizz (dismissed):

richardpringle (approved):

PR #1143, opened 2024-07-15
Add Tx Indexer Extension

aaronbuchwald:
This PR refactors the tx indexing from the MorpheusVM/TokenVM controller code into a separate module under `extensions/indexer`

This also fixes an apparent bug where the database created internally to the MorpheusVM/TokenVM for tx indexing would never be closed.


examples/tokenvm/tests/integration/integration_test.go
```diff
@@ -42,7 +43,6 @@ import (
 	"github.com/ava-labs/hypersdk/rpc"
 	"github.com/ava-labs/hypersdk/vm"
 
-	"github.com/ava-labs/hypersdk/auth"
```
joshua-kim, line 45: Not going to block on this but we need to fix this repo's imports. There doesn't seem to be any rationale for how the imports are ordered currently so I keep having to manually undo the diffs my ide auto-organizes.

extensions/indexer/indexer.go
joshua-kim, line 1: nit: rename to `extension/indexer`. I think it's idiomatic in Go to use singular package names (except for when it doesn't, like `errors`, or `strings` which I believe is because it would collide with the existing primitive type).

extensions/indexer/indexer.go
```diff
@@ -0,0 +1,80 @@
...
+
+type AcceptedIndexer interface {
+	Accepted(ctx context.Context, blk *chain.StatelessBlock) error
+}
+
+type SuccessfulTxIndexer interface {
+	Accepted(ctx context.Context, tx *chain.Transaction, result *chain.Result) error
+}
+
+type TxIndexer struct {
```
joshua-kim, line 24: If we're going to make this an extension that other vms will pull in it probably makes sense to add unit tests for this

examples/morpheusvm/controller/action_indexer.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
+	"github.com/ava-labs/hypersdk/extensions/indexer"
+)
+
+var _ indexer.SuccessfulTxIndexer = &successfulTxIndexer{}
+
+type successfulTxIndexer struct {
```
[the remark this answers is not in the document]
aaronbuchwald: We could rename it as well. We could more accurately call it an `AcceptedSubscriber`.

This would be more accurate to exactly what it's doing and an indexer would be one thing you could build as an `AcceptedSubscriber`. Similarly, you could pass in an `acceptedActionHandler` or similar, which would be just this for the MorpheusVM and would also maintain the in-memory order book in the case of TokenVM.

examples/morpheusvm/controller/controller.go
```diff
@@ -96,10 +97,15 @@ func (c *Controller) Initialize(
 	}
 	snowCtx.Log.Info("loaded genesis", zap.Any("genesis", c.genesis))
 
-	c.db, err = hstorage.New(pebble.NewDefaultConfig(), snowCtx.ChainDataDir, "db", gatherer)
+	c.txDB, err = hstorage.New(pebble.NewDefaultConfig(), snowCtx.ChainDataDir, "db", gatherer)
 	if err != nil {
 		return vm.Config{}, nil, nil, nil, nil, nil, nil, nil, err
 	}
+	c.txIndexer = indexer.NewTxIndexer(
```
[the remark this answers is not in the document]
aaronbuchwald: This is just keeping the same pattern that was there before. The for loop currently does two things, it iterates and indexes every transaction and then for each successful transaction it performs some action for each of the multi-actions that were included.

extensions/indexer/indexer.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (indexer *TxIndexer) Accepted(ctx context.Context, blk *chain.StatelessBlock) error {
+	batch := indexer.db.NewBatch()
+	defer batch.Reset()
+
+	timestamp := blk.GetTimestamp()
+	results := blk.Results()
+	for i, tx := range blk.Txs {
+		result := results[i]
+		if indexer.enabled {
```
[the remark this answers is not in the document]
aaronbuchwald: `result.Success` just indicates whether the execution result was successful. Regardless of that status, the transaction was included in the block. Failed transactions can still pay fees and should be indexed.

extensions/indexer/indexer.go
aaronbuchwald: Wasn't sure if that should be part of this PR, but can add it

extensions/indexer/indexer.go
```diff
@@ -0,0 +1,80 @@
...
+	successfulTxIndexer SuccessfulTxIndexer,
+) *TxIndexer {
+	return &TxIndexer{
+		db:                  db,
+		enabled:             enabled,
+		successfulTxIndexer: successfulTxIndexer,
+	}
+}
+
+func (indexer *TxIndexer) Accepted(ctx context.Context, blk *chain.StatelessBlock) error {
```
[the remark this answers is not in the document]
aaronbuchwald: Ya

extensions/indexer/indexer.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/ava-labs/hypersdk/fees"
+)
+
+type AcceptedIndexer interface {
+	Accepted(ctx context.Context, blk *chain.StatelessBlock) error
+}
+
+type SuccessfulTxIndexer interface {
+	Accepted(ctx context.Context, tx *chain.Transaction, result *chain.Result) error
+}
```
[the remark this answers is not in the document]
aaronbuchwald: I'll move the actions to their own accepted subscriber as suggested here https://github.com/ava-labs/hypersdk/pull/1143#discussion_r1679659760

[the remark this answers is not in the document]
aaronbuchwald: Still leaving this extra interface in, so that each controller can pass in its own successful tx handler. The alternative would be for both of them to just implement the full interface. Either is fine imo, lmk if you'd prefer the latter.

extensions/indexer/storage.go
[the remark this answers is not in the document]
aaronbuchwald: Removed the static functions

extensions/indexer/indexer.go
aaronbuchwald: I think we should make adding a unit test a separate PR because you currently need a builder in order to construct a stateless block...

We need to clean a lot of this code up to make writing unit tests more reasonable, so I'd prefer to defer this to a separate PR and keep the scope of this one a little bit smaller.

aaronbuchwald:
https://github.com/ava-labs/hypersdk/pull/1143/commits/707dad3ce39f85a2dd58ff53cb5fc103de8fe495 fixes https://github.com/ava-labs/hypersdk/actions/runs/9963000960/job/27528179448?pr=1143#step:5:433

Screenshot 2024-07-16 at 16 07 11

where I did not not invoke the combined subscriber correctly (still called the tx indexer instead)

aaronbuchwald:
> [707dad3](https://github.com/ava-labs/hypersdk/commit/707dad3ce39f85a2dd58ff53cb5fc103de8fe495) fixes https://github.com/ava-labs/hypersdk/actions/runs/9963000960/job/27528179448?pr=1143#step:5:433
> 
> Screenshot 2024-07-16 at 16 07 11
> where I did not not invoke the combined subscriber correctly (still called the tx indexer instead)

Digging into this a bit more, this happens for the TokenVM tests when the `actionIndexer` is a subscriber registered AFTER the `txIndexer`. If I reverse the order, the tests pass reliably locally.

I can also fix this test by inserting a sleep statement between marking the block as accepted and invoking the `tcli`. It seems that this code is racey because `controller.Accepted` is actually called from a goroutine processes accepted blocks asynchronously: https://github.com/ava-labs/hypersdk/blob/a061629d14a57898514a1d3549906b8d91f4276a/vm/resolutions.go#L213.

This race condition is not introduced by this change, but it does appear to happen reliably with this change iff the `actionIndexer` comes second.

I think it's reasonable to swap the order, so that this passes CI and add a TODO calling out the race condition in the integration test.

examples/morpheusvm/controller/action_indexer.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
+	"github.com/ava-labs/hypersdk/extension/indexer"
+)
+
+var _ indexer.SuccessfulTxSubscriber = (*actionHandler)(nil)
+
+type actionHandler struct {
```
joshua-kim, line 16: Should we call this something else...? Something related metrics or something?

extension/indexer/tx_indexer.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return false, 0, false, fees.Dimensions{}, 0, err
+	}
+	fee := binary.BigEndian.Uint64(v[consts.Uint64Len+1+fees.DimensionsLen:])
+	return true, t, success, d, fee, nil
+}
+
+type noopTxIndexer struct{}
+
+func NewNoopTxIndexer() *noopTxIndexer {
```
joshua-kim, line 108: nit: `NoOp`

extension/indexer/tx_indexer.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	d, err := fees.UnpackDimensions(v[consts.Uint64Len+1 : consts.Uint64Len+1+fees.DimensionsLen])
+	if err != nil {
+		return false, 0, false, fees.Dimensions{}, 0, err
+	}
+	fee := binary.BigEndian.Uint64(v[consts.Uint64Len+1+fees.DimensionsLen:])
+	return true, t, success, d, fee, nil
+}
+
+type noopTxIndexer struct{}
```
joshua-kim, line 106: nit: export type so caller can use this type if it wants to. Applicable to rest of concrete-types in this package

extension/indexer/accepted_subscriber.go
```diff
@@ -0,0 +1,31 @@
...
+
+type AcceptedSubscriber interface {
+	Accepted(ctx context.Context, blk *chain.StatelessBlock) error
+}
+
+type CombinedAcceptedSubscriber struct {
+	subscribers []AcceptedSubscriber
+}
+
+func NewCombinedAcceptedSubscriber(subscribers ...AcceptedSubscriber) *CombinedAcceptedSubscriber {
```
joshua-kim, line 20: Can we call this function + this type `func NewAcceptedSubscribers(...) *AcceptedSubscribers`?

extension/indexer/tx_indexer.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/ava-labs/hypersdk/consts"
+	"github.com/ava-labs/hypersdk/fees"
+)
+
+var (
+	_ AcceptedSubscriber = (*txIndexer)(nil)
+	_ AcceptedSubscriber = (*noopTxIndexer)(nil)
+)
+
+var (
```
joshua-kim, line 24: nit: combine the two `var` blocks

examples/morpheusvm/controller/action_indexer.go
joshua-kim, line 1: nit: I think you meant to name this file whatever we end up naming this struct

examples/tokenvm/controller/action_indexer.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/examples/tokenvm/actions"
+	"github.com/ava-labs/hypersdk/extension/indexer"
+)
+
+var _ indexer.SuccessfulTxSubscriber = (*actionHandler)(nil)
+
+type actionHandler struct {
```
joshua-kim, line 16: Similar comment as the corresponding comment in morpheusvm

examples/tokenvm/controller/action_indexer.go
```diff
@@ -0,0 +1,56 @@
...
+		case *actions.BurnAsset:
+			a.c.metrics.burnAsset.Inc()
+		case *actions.Transfer:
+			a.c.metrics.transfer.Inc()
+		case *actions.CreateOrder:
+			a.c.metrics.createOrder.Inc()
+			a.c.orderBook.Add(chain.CreateActionID(tx.ID(), uint8(i)), tx.Auth.Actor(), action)
+		case *actions.FillOrder:
+			a.c.metrics.fillOrder.Inc()
+			outputs := result.Outputs[i]
```
joshua-kim, line 36: Wondering if the order-book updates should happen in their own subscriber since they are logically decoupled from the metrics

extension/indexer/tx_indexer.go
```diff
@@ -0,0 +1,118 @@
...
+	failureByte = byte(0x0)
+	successByte = byte(0x1)
+)
+
+type TxIndexer interface {
+	Accepted(ctx context.Context, blk *chain.StatelessBlock) error
+	GetTransaction(txID ids.ID) (bool, int64, bool, fees.Dimensions, uint64, error)
+}
+
+type txIndexer struct {
```
joshua-kim, line 34: Could probably just call this `DiskIndexer`. Feels weird to have the concrete-type just be the name of the interface.

extension/indexer/tx_indexer.go
```diff
@@ -0,0 +1,118 @@
...
+	_ AcceptedSubscriber = (*noopTxIndexer)(nil)
+)
+
+var (
+	failureByte = byte(0x0)
+	successByte = byte(0x1)
+)
+
+type TxIndexer interface {
+	Accepted(ctx context.Context, blk *chain.StatelessBlock) error
```
joshua-kim, line 30: Probably should just replace this line w/ an embed for the `AcceptedSubscriber` interface it since that's the interface we're using to get access to accepted blocks

extension/indexer/tx_indexer.go
```diff
@@ -0,0 +1,118 @@
...
+	GetTransaction(txID ids.ID) (bool, int64, bool, fees.Dimensions, uint64, error)
```
joshua-kim, line 31: Should we return a struct here instead? We can define an `indexer.Tx` struct type or something that the indexer defines for indexed tx data.

extension/indexer/tx_indexer.go
aaronbuchwald: I disagree with this. This is intentionally hiding that detail assuming callers choose to create this type and don't need to know the details or change anything about it after the fact.

Open to changing my mind on this though

extension/indexer/tx_indexer.go
aaronbuchwald: I generally prefer a separate `var` block for interface assertions, but don't see that used elsewhere in this repo and do see a combined `var` block with another variable used in one place.

Will update.

examples/tokenvm/controller/action_indexer.go
aaronbuchwald: I think it's fine to do more than one thing here.

Doing this async (as it's handled prior to this PR as well) does introduce a race condition because the in-memory order book is not updated during block accept. We may want to support a `SyncAcceptedSubscriber` and `AsyncAcceptedSubscriber` to fix that.

examples/morpheusvm/controller/action_indexer.go
aaronbuchwald: acceptedActionMetricsHandler? This is a bit too long imo and I slightly prefer actionHandler in this case unless you have an alternative suggestion.

examples/morpheusvm/controller/action_indexer.go
aaronbuchwald: Renaming

examples/tokenvm/controller/action_indexer.go
aaronbuchwald: Renaming

extension/indexer/tx_indexer.go
aaronbuchwald: Renamed to `txDBIndexer` and renaming the constructor as well

extension/indexer/tx_indexer.go
joshua-kim: 1. If you don't want to let the caller know about the type for some reason, then you can just return the interface. Returning an unexported struct makes the type awkward to use. You can't embed, wrap, return a non-exported type from a calling package.
2. At the same time what's the disadvantage of exporting the type? Returning a concrete type gives more flexibility to the caller to use it either as the interface or as the concrete type. I think we should let consumers of the package decide how they want to use our api (this way they can use our indexer interface, wrap it with their own, or use the type).

extension/indexer/tx_indexer.go
aaronbuchwald: This is an interface, so we can swap a noop version in if tx indexing is disabled as per https://github.com/ava-labs/hypersdk/pull/1143#discussion_r1679556902

extension/indexer/tx_indexer.go
aaronbuchwald: Done

extension/indexer/tx_indexer.go
aaronbuchwald: Done

examples/morpheusvm/controller/action_indexer.go
aaronbuchwald: Done

examples/tokenvm/controller/action_indexer.go
aaronbuchwald: Done

extension/indexer/tx_indexer.go
aaronbuchwald: Done

examples/tokenvm/controller/action_indexer.go
joshua-kim: Yeah I think long term we're gonna try to get rid of the metrics in here and push it up into hypersdk anyways so... not too concerned

extension/indexer/accepted_subscriber.go
```diff
@@ -0,0 +1,31 @@
...
+
+type AcceptedSubscribers struct {
+	subscribers []AcceptedSubscriber
+}
+
+func NewAcceptedSubscribers(subscribers ...AcceptedSubscriber) *AcceptedSubscribers {
+	return &AcceptedSubscribers{subscribers: subscribers}
+}
+
+func (c *AcceptedSubscribers) Accepted(ctx context.Context, blk *chain.StatelessBlock) error {
```
joshua-kim, line 24: nit: update to `a` since we updated this struct name

extension/indexer/successful_tx_subscriber.go
```diff
@@ -0,0 +1,39 @@
...
+
+func NewSuccessfulTxSubscriber(subscriber SuccessfulTxSubscriber) AcceptedSubscriber {
+	return &successfulTxSubscriber{subscriber}
+}
+
+func (s *successfulTxSubscriber) Accepted(ctx context.Context, blk *chain.StatelessBlock) error {
+	results := blk.Results()
+	for i, tx := range blk.Txs {
+		result := results[i]
+		if result.Success {
```
joshua-kim, line 31: I think this is how the current code behaves... but just wondering if we should perform this check or leave it to the implementation. We don't have to worry about it for this PR though.

extension/indexer/tx_indexer.go
```diff
@@ -0,0 +1,116 @@
...
+
+type txDBIndexer struct {
+	db database.Database
+}
+
+func NewTxDBIndexer(db database.Database) *txDBIndexer {
+	return &txDBIndexer{db: db}
+}
+
+func (i *txDBIndexer) Accepted(_ context.Context, blk *chain.StatelessBlock) error {
```
joshua-kim, line 40: nit: export the struct + rename the receiver

extension/indexer/accepted_subscriber.go
```diff
@@ -0,0 +1,31 @@
...
+package indexer
+
+import (
+	"context"
+
+	"github.com/ava-labs/hypersdk/chain"
+)
+
+type AcceptedSubscriber interface {
+	Accepted(ctx context.Context, blk *chain.StatelessBlock) error
```
joshua-kim, line 13: I don't think we need this context + in the other tx subscriber interface currently... we could leave it in since I think it's reasonable that you would want a context though (say if you were archiving into s3 or something).

extension/indexer/tx_indexer.go
aaronbuchwald: Exported

extension/indexer/accepted_subscriber.go
aaronbuchwald: Updated

extension/indexer/successful_tx_subscriber.go
aaronbuchwald: Discussed here https://github.com/ava-labs/hypersdk/pull/1143#discussion_r1679916304 and I opted to keep this, so that each VM only handles accepted actions

extension/indexer/accepted_subscriber.go
aaronbuchwald: Ya I think this is reasonable to leave in as part of the interface

extension/indexer/accepted_subscriber.go
aaronbuchwald: Done

extension/indexer/tx_indexer.go
aaronbuchwald: Done

joshua-kim (approved):

PR #1146, opened 2024-07-16
Refactor `vm.Config` out of `vm.Controller`

joshua-kim:
`Controller` implementations are supposed to parse the provided config bytes upon initialization into a hypersdk config. This PR parses the config in the hypersdk, and supplies the remaining bytes to the controller to reduce boilerplate code in the implementation.

### DIFF
--- a/examples/morpheusvm/tests/integration/integration_test.go
+++ b/examples/morpheusvm/tests/integration/integration_test.go
@@ -221,7 +221,13 @@ var _ = ginkgo.BeforeSuite(func() {
 			genesisBytes,
 			nil,
 			[]byte(
-				`{"parallelism":3, "testMode":true, "logLevel":"debug"}`,
+				`{
+				  "config": {
+				    "parallelism":3,
+				    "testMode":true,
+				    "logLevel":"debug"
+				  }
+				}`,
 			),
 			toEngine,
 			nil,
--- a/examples/tokenvm/config/config.go
+++ b/examples/tokenvm/config/config.go
@@ -5,59 +5,19 @@ package config
 
 import (
 	"encoding/json"
-	"fmt"
-	"strings"
-	"time"
 
-	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/logging"
-	"github.com/ava-labs/avalanchego/utils/profiler"
 
-	"github.com/ava-labs/hypersdk/codec"
-	"github.com/ava-labs/hypersdk/examples/tokenvm/consts"
-	"github.com/ava-labs/hypersdk/examples/tokenvm/version"
 	"github.com/ava-labs/hypersdk/gossiper"
-	"github.com/ava-labs/hypersdk/trace"
-	"github.com/ava-labs/hypersdk/vm"
-)
-
-const (
-	defaultContinuousProfilerFrequency = 1 * time.Minute
-	defaultContinuousProfilerMaxFiles  = 10
-	defaultStoreTransactions           = true
-	defaultMaxOrdersPerPair            = 1024
 )
 
 type Config struct {
-	vm.Config
-
-	// Concurrency
-	AuthVerificationCores     int `json:"authVerificationCores"`
-	RootGenerationCores       int `json:"rootGenerationCores"`
-	TransactionExecutionCores int `json:"transactionExecutionCores"`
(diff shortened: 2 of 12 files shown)


examples/tokenvm/config/config.go
```diff
@@ -5,159 +5,44 @@ package config
...
-		Enabled:         c.TraceEnabled,
-		TraceSampleRate: c.TraceSampleRate,
-		AppName:         consts.Name,
-		Agent:           c.nodeID.String(),
-		Version:         version.Version.String(),
-	}
-}
-func (c *Config) GetStateSyncServerDelay() time.Duration { return c.StateSyncServerDelay }
-func (c *Config) GetStreamingBacklogSize() int           { return c.StreamingBacklogSize }
-func (c *Config) GetContinuousProfilerConfig() *profiler.Config {
```
joshua-kim, line 147: Some of these config values weren't even used... the profiler is interesting though so we could keep this if we wanted to provide it out-of-the-box (although it isn't used right now).

mempool/mempool.go
```diff
@@ -50,20 +50,16 @@ type Mempool[T Item] struct {
 	streamedItems     set.Set[ids.ID]
 	nextStream        []T
 	nextStreamFetched bool
-
-	// sponsors that are exempt from [maxSponsorSize]
-	exemptSponsors set.Set[codec.Address]
```
joshua-kim, line 55: This PR deletes this config as per offline discusison.

vm/dependencies.go
```diff
@@ -27,33 +26,34 @@ import (
...
+	ParsedBlockCacheSize             int             `json:"parsedBlockCacheSize"`
+	AcceptedBlockWindow              int             `json:"acceptedBlockWindow"`
+	AcceptedBlockWindowCache         int             `json:"acceptedBlockWindowCache"`
+	ContinuousProfilerConfig         profiler.Config `json:"continuousProfilerConfig"`
+	TargetBuildDuration              time.Duration   `json:"targetBuildDuration"`
+	ProcessingBuildSkip              int             `json:"processingBuildSkip"`
+	TargetGossipDuration             time.Duration   `json:"targetGossipDuration"`
+	BlockCompactionFrequency         int             `json:"blockCompactionFrequency"`
+	// Config is defined by the Controller
+	Config map[string]any `json:"config"`
```
joshua-kim, line 56: Decided to not make this `[]byte` because it makes the UX weird... the downside is that we have to re-marshal this to give `[]byte` to the implementing VM but I think this is fine since it's only done once upon initialization.

iFrostizz (dismissed):
It's clear that most of these functions were not used, but couldn't they be part of a public API that all VM implementers would expect these to comply with ?

joshua-kim:
Chatted offline. I think it's fine for the VM to not know about the hypersdk config since it holds values that only hypersdk will ever care about. If it holds something the VM needs we would expect that to be in the VM config.

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -221,7 +221,13 @@ var _ = ginkgo.BeforeSuite(func() {
 			genesisBytes,
 			nil,
 			[]byte(
-				`{"parallelism":3, "testMode":true, "logLevel":"debug"}`,
+				`{
+				  "config": {
+				    "parallelism":3,
```
aaronbuchwald, line 226: Noticed while reviewing this PR, it looks like `parallelism` field is unused in both this and TokenVM. Could we remove it in either this or a follow up?

aaronbuchwald (approved):

examples/morpheusvm/tests/integration/integration_test.go
joshua-kim: Yeah looks stale. Will create a follow-up.

vm/dependencies.go
aaronbuchwald: Not sure how this will go later since what we're doing is non-standard, but I think this is alright for now. We can find a better approach if it causes issues.

examples/morpheusvm/tests/integration/integration_test.go
joshua-kim: Followup PR: https://github.com/ava-labs/hypersdk/pull/1158

Issue #1145, opened 2024-07-15
Fix Accepted/Rejected reliability


aaronbuchwald:
Trying to refactor the shared tx / action indexer out of both VMs and into a shared module as part of https://github.com/ava-labs/hypersdk/pull/1143, I found a bug where the HyperSDK could fail to notify the controller interface of blocks that have been marked as accepted/rejected, which could result in an incomplete transaction index.

The engine marks a block as accepted from the following chain of calls, which adds a block to an accepted queue to notify the controller on an async thread for performance reasons: https://github.com/ava-labs/hypersdk/blob/a061629d14a57898514a1d3549906b8d91f4276a/chain/block.go#L567
https://github.com/ava-labs/hypersdk/blob/a061629d14a57898514a1d3549906b8d91f4276a/chain/block.go#L567
https://github.com/ava-labs/hypersdk/blob/a061629d14a57898514a1d3549906b8d91f4276a/chain/block.go#L580
https://github.com/ava-labs/hypersdk/blob/a061629d14a57898514a1d3549906b8d91f4276a/vm/resolutions.go#L230
https://github.com/ava-labs/hypersdk/blob/a061629d14a57898514a1d3549906b8d91f4276a/vm/resolutions.go#L281
https://github.com/ava-labs/hypersdk/blob/a061629d14a57898514a1d3549906b8d91f4276a/vm/resolutions.go#L213

The final call is where the HyperSDK will actually call Accepted on the controller and trigger indexing the transactions in the block. However, we do not track what was the last block where we actually succeeded in calling `Accepted` on the controller interface.

HyperSDK uses its own database, so AvalancheGo can't close the DB out from under it, but it will kill the process if it takes too long, so it seems like this could cause us to drop some transactions from the index.

To fix this, we need to correctly handle writes across multiple DBs. The HyperSDK outer VM only has access to its own DB and other writes are performed when the HyperSDK sends a notification to the controller for example: https://github.com/ava-labs/hypersdk/blob/a061629d14a57898514a1d3549906b8d91f4276a/vm/resolutions.go#L172-L191.

To handle this correctly, we'll need to store an index of the last block we finished processing at the end of `processAcceptedBlock`. This must be committed after sending all of the notifications without error. On restart, we will also need to read that index and call `processAcceptedBlock` for all of the blocks between the last accepted block that was processed and the actual last accepted block (or we could just add those blocks to the queue to be processed).

[...]

PR #1150, opened 2024-07-16
Keep track of the last accepted block at the end of the processing

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1145

The `processAcceptedBlocks` didn't kept track of the last accepted block. This was a bug because if the node stopped while it was processing some blocks, the operation couldn't be resumed from the last point because the last accepted block loaded from disk was never written and so the `vm.lastAcceptedBlock` was always written to genesis.

Tests that shows that an issue while the node is processing blocks write to disk and then resume from that last point are TBD.


vm/vm.go
```diff
@@ -433,6 +433,34 @@ func (vm *VM) Initialize(
...
+				blk, err = vm.GetDiskBlock(ctx, height)
+				if err != nil {
+					vm.snowCtx.Log.Info("could not find block on-disk", zap.Uint64("height", height))
+					return err
+				}
+			}
+			vm.processAcceptedBlock(blk)
+		}
+	}
+
```
aaronbuchwald, line 463: This `Initialize` function is already huge. Could we move this into its own function since it is pretty well-defined?

vm/vm.go
```diff
@@ -433,6 +433,34 @@ func (vm *VM) Initialize(
 	webSocketServer, pubsubServer := rpc.NewWebSocketServer(vm, vm.config.StreamingBacklogSize)
 	vm.webSocketServer = webSocketServer
 	vm.handlers[rpc.WebSocketEndpoint] = pubsubServer
+
+	has, err = vm.HasLastProcessed()
+	if err != nil {
+		snowCtx.Log.Error("could not determine if have last processed")
+		return err
+	}
+	if has {
```
aaronbuchwald, line 442: Once this is moved to a separate function, we can also add an early return here since there's nothing else happening after it.

Then we can make this:

```golang
if !has {
    return nil
}
```

and then the remainder can go outside of the if block, which is a bit cleaner

vm/vm.go
```diff
@@ -433,6 +433,34 @@ func (vm *VM) Initialize(
...
+			} else {
+				blk, err = vm.GetDiskBlock(ctx, height)
+				if err != nil {
+					vm.snowCtx.Log.Info("could not find block on-disk", zap.Uint64("height", height))
+					return err
+				}
+			}
+			vm.processAcceptedBlock(blk)
+		}
+	}
```
aaronbuchwald, line 462: Instead of processing them synchronously within `Initialize`, could we add them to the accepted queue?

We should add them to the queue synchronously to guarantee that accepted blocks are processed by any listeners in order.

vm/vm.go
iFrostizz: Moved it to the new `restoreLastAcceptedBlock` that is called at the end of the initialization of the `VM`.

vm/vm.go
iFrostizz: It's cleaner! Thanks!

vm/vm.go
iFrostizz: Done, the blocks are now added to the queue so that they are not blocking the initialization of the `VM` with `vm.acceptedQueue <- blk`

aaronbuchwald (changes requested):
LGTM after addressing these comments

vm/vm.go
```diff
@@ -1128,6 +1129,40 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+		return nil
+	}
+
+	lastProcessedHeight, err := vm.GetLastProcessedHeight()
+	if err != nil {
+		vm.snowCtx.Log.Error("could not get last processed height", zap.Error(err))
+		return err
+	}
+
+	for height := lastProcessedHeight; height <= vm.lastAccepted.Height(); height++ {
```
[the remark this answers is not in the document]
iFrostizz: Wow nice catch! This one is already processed so starting from `lastProcessedHeight + 1`

vm/vm.go
```diff
@@ -1128,6 +1129,40 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
 	return nil
 }
 
+func (vm *VM) restoreLastAcceptedBlock(ctx context.Context) error {
+	has, err := vm.HasLastProcessed()
+	if err != nil {
+		vm.snowCtx.Log.Error("could not determine if have last processed")
```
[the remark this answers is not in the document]
iFrostizz: Done by returning an `error` type that then gets formatted with an error message that is specific to this restoring error so we know where it's coming from.

vm/vm.go
```diff
@@ -1128,6 +1129,40 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+		return err
+	}
+	if !has {
+		return nil
+	}
+
+	lastProcessedHeight, err := vm.GetLastProcessedHeight()
+	if err != nil {
+		vm.snowCtx.Log.Error("could not get last processed height", zap.Error(err))
```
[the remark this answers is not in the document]
iFrostizz: done here too!

vm/vm.go
```diff
@@ -1128,6 +1129,40 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+		return err
+	}
+
+	for height := lastProcessedHeight; height <= vm.lastAccepted.Height(); height++ {
+		var blk *chain.StatelessBlock
+		if height == vm.lastAccepted.Height() {
+			blk = vm.lastAccepted
+		} else {
+			blk, err = vm.GetDiskBlock(ctx, height)
```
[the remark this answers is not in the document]
iFrostizz: Thanks, opened https://github.com/ava-labs/hypersdk/issues/1165 !

vm/vm.go
```diff
@@ -1128,6 +1129,40 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+			if err != nil {
+				vm.snowCtx.Log.Info("could not find block on-disk", zap.Uint64("height", height))
+				return err
+			}
+		}
+
+		vm.acceptedQueue <- blk
```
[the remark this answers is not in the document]
iFrostizz: Added the two info logs!

vm/vm.go
```diff
@@ -1128,6 +1129,40 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+
+	for height := lastProcessedHeight; height <= vm.lastAccepted.Height(); height++ {
+		var blk *chain.StatelessBlock
+		if height == vm.lastAccepted.Height() {
+			blk = vm.lastAccepted
+		} else {
+			blk, err = vm.GetDiskBlock(ctx, height)
+			if err != nil {
+				vm.snowCtx.Log.Info("could not find block on-disk", zap.Uint64("height", height))
+				return err
```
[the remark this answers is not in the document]
iFrostizz: Added bounds to `AcceptorSize` and `AcceptedBlockWindow`

vm/vm.go
```diff
@@ -1128,6 +1141,46 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
 	return nil
 }
 
+func (vm *VM) restoreAcceptedQueue(ctx context.Context) error {
+	has, err := vm.HasLastProcessed()
+	if err != nil {
+		return errors.New("could not determine if have last processed")
+	}
```
aaronbuchwald, line 1148: nit: this should wrap the original error with `fmt.Errorf(...)`

vm/vm.go
```diff
@@ -1128,6 +1141,46 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+	vm.snowCtx.Log.Info("accepted blocks have to be restored", zap.Uint64("n", acceptedToRestore))
+
+	for height := start; height <= end; height++ {
+		var blk *chain.StatelessBlock
+		if height == vm.lastAccepted.Height() {
+			blk = vm.lastAccepted
+		} else {
+			blk, err = vm.GetDiskBlock(ctx, height)
+			if err != nil {
+				return fmt.Errorf("could not find block at height %d on disk", height)
```
aaronbuchwald, line 1173:
```suggestion
				return fmt.Errorf("could not find accepted block at height %d on disk", height)
```

vm/vm.go
```diff
@@ -1128,6 +1141,46 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+			}
+		}
+
+		vm.acceptedQueue <- blk
+	}
+	vm.snowCtx.Log.Info("accepted blocks have been added to the queue")
```
aaronbuchwald, line 1179:
```suggestion
	vm.snowCtx.Log.Info("finished restoring accepted queue")
```

vm/vm.go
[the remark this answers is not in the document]
aaronbuchwald: We should also remove the check to see if it's the last accepted block from this function in the follow on.

There should just be a helper function that will fetch an accepted block by height and the caller shouldn't need to worry about a special case.

vm/vm.go
```diff
@@ -1128,6 +1141,46 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+		return fmt.Errorf("could not get last processed height: %w", err)
+	}
+
+	start := lastProcessedHeight + 1
+	end := vm.lastAccepted.Height()
+	if end >= start {
+		return nil
+	}
+	acceptedToRestore := end - start
+	vm.snowCtx.Log.Info("accepted blocks have to be restored", zap.Uint64("n", acceptedToRestore))
```
aaronbuchwald, line 1164:
```suggestion
	vm.snowCtx.Log.Info("restoring accepted blocks to the accepted queue", zap.Uint64("blocks", acceptedToRestore))
```

vm/vm.go
```diff
@@ -1128,6 +1141,46 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
...
+	}
+
+	lastProcessedHeight, err := vm.GetLastProcessedHeight()
+	if err != nil {
+		return fmt.Errorf("could not get last processed height: %w", err)
+	}
+
+	start := lastProcessedHeight + 1
+	end := vm.lastAccepted.Height()
+	if end >= start {
```
aaronbuchwald, line 1160: Why do we terminate here if `end >= start` but we do one more iteration within the for loop when `height == end` ?

aaronbuchwald (changes requested):

vm/vm.go
iFrostizz: Wow that's a good point, a bug that I introduced. Replaced the `>=` by a `>`. Thanks!

vm/vm.go
iFrostizz: Done thanks!

vm/vm.go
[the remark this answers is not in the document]
iFrostizz: Thanks, noted on the issue.

vm/vm.go
iFrostizz: Actually the order was reversed!

vm/vm.go
```diff
@@ -283,6 +284,12 @@ func (vm *VM) Initialize(
 	if err != nil {
 		return err
 	}
+	if vm.config.AcceptorSize > 256 {
+		return errors.New("AcceptorSize should be less than 256")
+	}
+	if vm.config.AcceptedBlockWindow < 1024 {
```
aaronbuchwald, line 290: Let's set 256 and 1024 as constants at the top of the file

aaronbuchwald: Can we change both of these to formatted errors that include the actual provided value, so that it's clearer to the user?

We should also fix the error messages ie. "should be less than or equal to 256" and "should be greater than or equal to 1024". The values 256 and 1024 should be formatted to include the constants set at the top of the file, so that we do not need to update the error string if we change these defaults in the future.

vm/vm.go
```diff
@@ -442,6 +449,12 @@ func (vm *VM) Initialize(
 	webSocketServer, pubsubServer := rpc.NewWebSocketServer(vm, vm.config.StreamingBacklogSize)
 	vm.webSocketServer = webSocketServer
 	vm.handlers[rpc.WebSocketEndpoint] = pubsubServer
+
+	err = vm.restoreAcceptedQueue(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to restore the pending accepted blocks: %w", err)
```
aaronbuchwald, line 455:
```suggestion
		return fmt.Errorf("failed to restore accepted blocks to the queue: %w", err)
```

vm/vm.go
```diff
@@ -1128,6 +1141,46 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
 	return nil
 }
 
+func (vm *VM) restoreAcceptedQueue(ctx context.Context) error {
+	has, err := vm.HasLastProcessed()
+	if err != nil {
+		return fmt.Errorf("could not determine if have last processed: %w", err)
```
aaronbuchwald, line 1147: if have last processed is a bit hard to read.

```suggestion
		return fmt.Errorf("could not load last processed block: %w", err)
```

aaronbuchwald (changes requested):

vm/vm.go
iFrostizz: Thanks ! Added constants and updated the error messages.

aaronbuchwald: Could you include the actual provided value in the error message?

iFrostizz: Sure, done!

aaronbuchwald (approved):

PR #1153, opened 2024-07-16
Expose state functions from context

richardpringle:
@dboehm-avalabs suggested that we just put the state functions on the `Context` directly and I agreed

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/context.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/context.rs
@@ -1,0 +1,129 @@
+use crate::{
+    state::{Error, Key, State},
+    types::Address,
+    Gas, Id, Program,
+};
+use borsh::{BorshDeserialize, BorshSerialize};
+use std::{cell::RefCell, collections::HashMap};
+
+/// Representation of the context that is passed to programs at runtime.
+#[cfg_attr(feature = "debug", derive(Debug))]
+pub struct Context<K = ()> {
+    program: Program,
+    actor: Address,
+    height: u64,
+    timestamp: u64,
+    action_id: Id,
+    state_cache: RefCell<HashMap<K, Option<Vec<u8>>>>,
+}
+
+impl<K> BorshDeserialize for Context<K> {
+    fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
+        let program = Program::deserialize_reader(reader)?;
+        let actor = Address::deserialize_reader(reader)?;
+        let height = u64::deserialize_reader(reader)?;
+        let timestamp = u64::deserialize_reader(reader)?;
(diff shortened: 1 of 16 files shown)


x/programs/rust/wasmlanche-sdk/src/context.rs
```diff
@@ -0,0 +1,129 @@
...
+    types::Address,
+    Gas, Id, Program,
+};
+use borsh::{BorshDeserialize, BorshSerialize};
+use std::{cell::RefCell, collections::HashMap};
+
+/// Representation of the context that is passed to programs at runtime.
+#[cfg_attr(feature = "debug", derive(Debug))]
+pub struct Context<K = ()> {
+    program: Program,
```
samliok, line 12: could change the `program` field to just be an address?

x/programs/rust/wasmlanche-sdk/src/context.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /// See [`State::store_by_key`].
+    /// # Errors
+    /// See [`State::store_by_key`].
+    pub fn store_by_key<V>(&self, key: K, value: &V) -> Result<(), crate::state::Error>
+    where
+        V: BorshSerialize,
+    {
+        State::new(&self.state_cache).store_by_key(key, value)
```
samliok, line 76: since `flush` will get called once this `state` is dropped, we could move the cache flushing to be a method of context

https://github.com/ava-labs/hypersdk/blob/97a86924856b9dec1210c4a7faf247dae8730b3a/x/programs/rust/wasmlanche-sdk/src/state.rs#L145

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -76,24 +70,12 @@ pub fn send(to: Address, amount: u64) -> Result<(), ExternalCallError> {
 /// Represents the current Program in the context of the caller, or an external
 /// program that is being invoked.
 #[cfg_attr(feature = "debug", derive(Debug))]
-#[derive(BorshSerialize)]
-pub struct Program<K = ()> {
+#[derive(Clone, Copy, BorshSerialize, BorshDeserialize)]
```
samliok, line 73: niceee

samliok (approved):

x/programs/rust/wasmlanche-sdk/src/context.rs
richardpringle: Yeah, this is my next change

x/programs/rust/wasmlanche-sdk/src/context.rs
richardpringle: Will do in a follow up

iFrostizz (approved):

PR #1166, opened 2024-07-18
[x/Programs] Refactor Simulator

samliok:
# Refactor Simulator 

This PR aims to make the simulator more useable & functional. The previous design of the simulator was intended to be used both via a CLI and rust tests. Since we are moving away from a CLI, the simulator was in need of a refactor.

### Main changes
- Changes simulator nomenclature from `Plan`, `Step`, `StepError` etc.. to request/response  verbiage. `Plan` and `Step` were named for passing in files/using the CLI but caused confusion on the rust side. Now, we query the simulator with a `SimulatorRequest` and get a `SimulatorResponse`.
- Simulator code was in one large `lib` file. This PR decomposes the code into more readable files. 
- Requests are sent to the simulator by calling either 
	-  `read(&mut  self, method:  String, params:  Vec<Param>)`
	- `pub  fn  execute(&mut  self,method:  String,params:  Vec<Param>,max_units:  u64)`
	-  `create_program<P:  AsRef<Path>>(&mut  self, path:  P)`
	- The `read` method allows for `state` writes. Disallowing/panicking on state writes in `readonly` program calls will be done in a follow up pr.

- Decouples the `Execute` endpoint into `CreateProgram` and `Execute`
-  Removes the ability to pass in a `Plan` aka a series of program actions.

### Usage

```rust
let mut simulator = build_simulator().unwrap();

 let program_id = simulator.create_program(PROGRAM_PATH).unwrap().id;
 let test_context = TestContext::from(program_id);
 
 simulator
     .execute(
         "init".into(),
         vec![
             test_context.clone().into(),
             Param::String("Test".into()),
             Param::String("TST".into()),
         ],
         MAX_UNITS,
     )
     .unwrap();

 let supply = simulator
     .read("total_supply".into(), vec![test_context.clone().into()])
     .unwrap()
     .result
     .response::<u64>()
     .unwrap();
```

### Previous usage

```rust
 let mut simulator = simulator::ClientBuilder::new().try_build().unwrap();

 let program_id = simulator
     .run_step(&Step::create_program(PROGRAM_PATH))
     .unwrap()
     .id;

 let test_context = TestContext::from(program_id);

 simulator
     .run_step(&Step {
         endpoint: Endpoint::Execute,
         method: "init".into(),
         params: vec![
             test_context.clone().into(),
             Param::String("Test".into()),
             Param::String("TST".into()),
         ],
         max_units: 1000000,
     })
     .unwrap();

 let supply = simulator
     .run_step(&Step {
         endpoint: Endpoint::ReadOnly,
         method: "total_supply".into(),
         max_units: 0,
         params: vec![test_context.clone().into()],
     })
     .unwrap()
     .result
     .response::<u64>()
     .unwrap();
```

While this PR is not perfect, I think it is a step in the right direction to make the simulator more useable and flexible.


x/programs/cmd/simulator/src/util.rs
```diff
@@ -0,0 +1,16 @@
+use std::path::Path;
```
samliok, line 1: not sure if this should be in a seperate `util` file

richardpringle: This is only used in one place, you should just define it at the bottom of that file.

x/programs/cmd/simulator/src/util.rs
```diff
@@ -0,0 +1,16 @@
...
+pub(crate) fn get_path() -> &'static str {
+    let path = env!("SIMULATOR_PATH");
+
+    if !Path::new(path).exists() {
+        eprintln!();
+        eprintln!("Simulator binary not found at path: {path}");
+        eprintln!();
+        eprintln!("Please run `cargo clean -p simulator` and rebuild your dependent crate.");
+        eprintln!();
```
richardpringle, line 10:
```suggestion
        eprint!(r#"
Simulator binary not found at path: {path}

Please run `cargo clean -p simulator` and rebuild your dependent crate.

"#);
```
Or whatever the equivalent is

x/programs/rust/examples/counter/src/lib.rs
```diff
@@ -38,53 +38,43 @@ fn get_value_internal(context: &Context<StateKeys>, of: Address) -> Count {
...
-        simulator
-            .run_step(&Step::create_program(PROGRAM_PATH))
-            .unwrap();
+        simulator.create_program(PROGRAM_PATH).unwrap();
     }
 
     #[test]
     fn increment() {
-        let mut simulator = simulator::ClientBuilder::new().try_build().unwrap();
+        let mut simulator = build_simulator().unwrap();
```
richardpringle, line 55: You could probably change this to be `Simulator::new()` and use `expect` *inside* the `new` function. I can't imagine how anyone would try to recover from the simulator instantiation erroring out

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+    path::Path,
+    process::{Child, Command, Stdio},
+};
+use thiserror::Error;
+use wasmlanche_sdk::ExternalCallError;
+
+/// The endpoint to call for a [`SimulatorRequest`].
+#[derive(Debug, Serialize, PartialEq, Clone)]
+#[serde(rename_all = "lowercase")]
+pub enum Endpoint {
```
richardpringle, line 17: Is this still being used anywhere outside of this crate? If not, don't leave the `pub` there.

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
+use crate::codec::{base64_decode, id_from_usize};
+use crate::util::get_path;
+use crate::{param::Param, Id};
```
richardpringle, line 3: Please merge these

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+    #[error("Missing handle")]
+    StdIo,
+}
+
+pub struct Simulator<W, R> {
+    writer: W,
+    responses: R,
+}
+
+pub fn build_simulator(
```
richardpringle, line 43: This should be a constructor, see my other comment about not returning the error.

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+) -> Result<Simulator<impl Write, impl Iterator<Item = SimulatorResponseItem>>, ClientError> {
+    let path = get_path();
```
richardpringle, line 45: You can bring this into this file, and I would rename it to `get_binary_path` or something along those lines.

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+        let input = serde_json::to_vec(&SimulatorRequest::new_create_program(path.into()))?;
+
+        self.writer.write_all(&input)?;
+        self.writer.write_all(b"'\n")?;
+        self.writer.flush()?;
+
+        self.responses
+            .next()
+            .ok_or(SimulatorError::Client(ClientError::Eof))?
+            .and_then(|resp: SimulatorResponse| {
```
richardpringle, line 86:
```suggestion
            .and_then(|resp| {
```

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+impl<W, R> Simulator<W, R>
+where
+    W: Write,
+    R: Iterator<Item = SimulatorResponseItem>,
+{
+    const RUN_COMMAND: &'static [u8] = b"run --message '";
+
+    pub fn create_program<P: AsRef<Path>>(&mut self, path: P) -> SimulatorResponseItem {
+        let path = path.as_ref().to_string_lossy();
+        self.writer.write_all(Self::RUN_COMMAND)?;
```
richardpringle, line 75:
```suggestion
        self.writer.write_all(Self::RUN_COMMAND)?;
        let path = path.as_ref().to_string_lossy();
```

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+        self.writer.flush()?;
+
+        self.responses
+            .next()
+            .ok_or(SimulatorError::Client(ClientError::Eof))?
+            .and_then(|request| {
+                if let Some(err) = request.error {
+                    Err(SimulatorError::Program(err))
+                } else {
+                    Ok(request)
```
richardpringle, line 111: This should be consistent, the variable is called `resp` above.

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
```
richardpringle, line 137: Same comment as above

richardpringle: Also, the third time I'm seeing the exact same logic. Duplicating twice is fine, but three times deserves an abstraction. 

```rust
impl From<SimulatorResponse> for Result<SimulatorResponse, SimulatorError> {
// ...
}
```

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+pub struct SimulatorRequest {
+    /// The API endpoint to call.
+    pub endpoint: Endpoint,
+    /// The method to call on the endpoint.
+    pub method: String,
+    /// The maximum number of units the Request can consume.
+    pub max_units: u64,
+    /// The parameters to pass to the method.
+    pub params: Vec<Param>,
+}
```
richardpringle, line 155: Is this used outside of this crate? If not, should remove all the `pub`s

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub fn new_execute(method: String, params: Vec<Param>, max_units: u64) -> Self {
+        Self {
+            endpoint: Endpoint::Execute,
+            method,
+            max_units,
+            params,
+        }
+    }
+
+    pub fn new_create_program(path: String) -> Self {
```
richardpringle, line 176: The constructor is already implied by the fact that it's a static method. 
```suggestion
    pub fn read(method: String, params: Vec<Param>) -> Self {
        Self {
            endpoint: Endpoint::ReadOnly,
            method: method.to_string(),
            max_units: 0,
            params,
        }
    }

    pub fn execute(method: String, params: Vec<Param>, max_units: u64) -> Self {
        Self {
            endpoint: Endpoint::Execute,
            method,
            max_units,
            params,
        }
    }

    pub fn create_program(path: String) -> Self {
```

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+        Self {
+            endpoint: Endpoint::CreateProgram,
+            // TODO: this does not need to be passed in
```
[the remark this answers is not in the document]
richardpringle: What does not need to be passed in? The path?

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+pub enum SimulatorError {
+    #[error("Client error: {0}")]
+    Client(#[from] ClientError),
+    #[error("Serialization / Deserialization error: {0}")]
+    Serde(#[from] serde_json::Error),
+    #[error("Borsh deserialization error: {0}")]
+    BorshDeserialization(#[from] borsh::io::Error),
+    #[error("Program error: {0}")]
+    Program(String),
+}
```
richardpringle, line 197: Can you move this to the top of the file?

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+pub(crate) type SimulatorResponseItem = Result<SimulatorResponse, SimulatorError>;
+
+#[derive(Debug, Deserialize)]
+pub struct SimulatorResult {
+    /// The ID created from the program execution.
+    pub action_id: Option<String>,
```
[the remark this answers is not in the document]
richardpringle: @iFrostizz, did you put this here or @dboehm-avalabs? Why is it here? It's not used if I'm not mistaken.

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+    response: Vec<u8>,
+}
+
+#[derive(Error, Debug)]
+pub enum SimulatorResponseError {
+    #[error(transparent)]
+    Serialization(#[from] borsh::io::Error),
+    #[error(transparent)]
+    ExternalCall(#[from] ExternalCallError),
+}
```
richardpringle, line 217: If we can put all `pub` errors at the top of the file, that would be ideal.

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct SimulatorResponse {
+    /// The numeric id of the response.
+    #[serde(deserialize_with = "id_from_usize")]
+    pub id: Id, // TODO override of the Id Deserialize before removing the prefix
+    /// An optional error message.
+    pub error: Option<String>,
+    pub result: SimulatorResult,
```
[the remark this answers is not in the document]
richardpringle: I would prefer it if there weren't public types representing an invalid state. There should be a private (to the `Deserialize<'de>` implementation) type that deals with deserializing from json because Go doesn't have enums, but that shouldn't leak to the user.

Can fix this now or in a follow up.

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,367 @@
...
+
+        let param = Param::from(value);
+        let expected_param = Param::Address(value);
+
+        assert_eq!(param, expected_param);
+
+        let output_json = serde_json::to_value(&param).unwrap();
+
+        assert_eq!(output_json, expected_json);
+    }
```
richardpringle, line 366: All these tests (should we decide to keep them) belong in `param.rs`, not here. 

However, I question their value. If we change the way we serialize, we have to change all the tests. The tests will only fail if we make an explicit change as it stands. All of the logic lives in `serde_derive`, we're only testing how we tag our structs. 

I think this is more of a maintenance burden than adding any value. Tests like these only really make sense when we are making promises about the crate-API. Right now it's just a maintenance burden.

x/programs/cmd/simulator/cmd/config.go
```diff
@@ -8,17 +8,12 @@ import (
 	"fmt"
 )
 
-const (
-	ProgramCreate  = "program_create"
-	ProgramExecute = "execute"
-)
```
richardpringle, line 14: Were these not being used? Don't we have linter warnings for this type of stuff?

x/programs/cmd/simulator/cmd/request.go
```diff
@@ -32,28 +32,28 @@ var _ Cmd = (*runCmd)(nil)
 type runCmd struct {
 	cmd *argparse.Command
 
-	lastStep *int
-	file     *string
-	planStep *string
+	// counts the number of requests made
+	numRequests *int
```
richardpringle, line 36: huge nit (you can just resolve this if you disagree)

`requestCount`? `num` isn't descriptive, that's what the type-system is for....

richardpringle (changes requested):
Overall looks good! Love the deletions compared to the additions! 

I'm nitpicking a little here, sure... but the point of the PR makes things clearer and I think my comments add to that.

x/programs/cmd/simulator/src/simulator.rs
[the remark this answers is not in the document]
samliok: the method name

x/programs/cmd/simulator/src/simulator.rs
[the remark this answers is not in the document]
iFrostizz: git blamed @dboehm-avalabs ! https://github.com/ava-labs/hypersdk/pull/1056
I can't find of any really good use-case besides some funny PoW stuff

x/programs/cmd/simulator/cmd/config.go
samliok: hmm, I believe they might have been used just not anymore

x/programs/cmd/simulator/cmd/request.go
samliok: reason i had `numRequests` over `requests` was because there's already a `request` field with a different type. And `requests` in our rust side has a different type.

x/programs/cmd/simulator/src/simulator.rs
samliok: if it's in the constructor wouldn't we'd need to specify the type when calling `Simulator::new()`?

so it'd be `Simulator::<W, R>::new()`

I could be missing some rust syntax knowledge but I've ran into some rust generic compiler errors trying to do this

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[serde(deserialize_with = "id_from_usize")]
+    pub id: Id, // TODO override of the Id Deserialize before removing the prefix
+    /// An optional error message.
+    pub error: Option<String>,
+    pub result: SimulatorResult,
+}
+
+/// The endpoint to call for a [`SimulatorRequest`].
+#[derive(Debug, Serialize, PartialEq, Clone)]
+#[serde(rename_all = "lowercase")]
```
iFrostizz, line 208:
```suggestion
#[serde(rename_all = "snake_case")]
```

x/programs/cmd/simulator/cmd/config.go
```diff
@@ -33,6 +28,8 @@ const (
 	/// function call. A program's function can internally optionally call other
 	/// functions including program to program.
 	EndpointExecute Endpoint = "execute"
+	/// Create a new program on-chain
+	EndpointCreateProgram Endpoint = "createprogram"
```
iFrostizz, line 32:
```suggestion
	EndpointCreateProgram Endpoint = "create_program"
```

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,238 @@
...
+pub struct Simulator<W, R> {
+    writer: W,
+    responses: R,
+}
+
+pub fn build_simulator() -> Simulator<impl Write, impl Iterator<Item = SimulatorResponseItem>> {
+    let path = get_binary_path();
+
+    let Child { stdin, stdout, .. } = Command::new(path)
+        .arg("interpreter")
```
iFrostizz, line 54: Have you considered making the interpreter mode a default ? Since we don't use the subcommands natively at all it may simplify some stuff with these subcommands but if that is a good choice that could probably be better as a followup of this pr

x/programs/cmd/simulator/src/simulator.rs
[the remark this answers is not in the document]
iFrostizz: Oh wow that's totally unrelated ^ this is for the context !
No it was a legacy code, but we don't seem to need that. And it could simplified to just have the `result`, so that the `response` is now just a method of the `SimulatorResponse`.

x/programs/cmd/simulator/src/simulator.rs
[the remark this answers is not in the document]
iFrostizz: Since the fields other than `response` inside of `SimulatorResult` are not used it might be nice to remove the `SimulatorResult` and move the `result` inside of `SimulatorResponse` yeah

x/programs/cmd/simulator/src/util.rs
iFrostizz, line 1: Looks like you moved the `get_binary_path` from this file, please delete it

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// Create a transaction on-chain from a possible state changing program
+    /// function call. A program's function can internally optionally call other
+    /// functions including program to program.
+    Execute,
+    /// Create a new program on-chain
+    CreateProgram,
+}
+
+use std::path::Path;
+pub(crate) fn get_binary_path() -> &'static str {
```
iFrostizz, line 221:
```suggestion
pub fn get_binary_path() -> &'static str {
```

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,238 @@
...
+    ReadOnly,
+    /// Create a transaction on-chain from a possible state changing program
+    /// function call. A program's function can internally optionally call other
+    /// functions including program to program.
+    Execute,
+    /// Create a new program on-chain
+    CreateProgram,
+}
+
+use std::path::Path;
```
iFrostizz, line 220: Please import this at the top of the file

x/programs/cmd/simulator/src/simulator.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    pub fn execute(method: String, params: Vec<Param>, max_units: u64) -> Self {
+        Self {
+            endpoint: Endpoint::Execute,
+            method,
+            max_units,
+            params,
+        }
+    }
```
iFrostizz, line 160: Not sure why we used to pass in a String, a &str would work just fine !

```suggestion
impl<'a> SimulatorRequest<'a> {
    pub fn read(method: &'a str, params: Vec<Param>) -> Self {
        Self {
            endpoint: Endpoint::ReadOnly,
            method,
            max_units: 0,
            params,
        }
    }

    pub fn execute(method: &'a str, params: Vec<Param>, max_units: u64) -> Self {
        Self {
            endpoint: Endpoint::Execute,
            method,
            max_units,
            params,
        }
    }
```

iFrostizz: Then you could change the signature of `Simulator::read` and `Simulator::write` so that they take the a lifetime as `Simulator::read<'a>` and `Simulator::write<'a>` and pass it to the `method` parameter. This suggestion seems a bit out of scope of this PR though so that could be done as a separate one.

iFrostizz (changes requested):
The use of endpoints is so much cleaner !

x/programs/cmd/simulator/cmd/config.go
iFrostizz: the `ProgramCreate` was used since we used to check if the request was a program creation from the method name, which was really odd.

x/programs/cmd/simulator/cmd/config.go
iFrostizz, line 32:
```suggestion
	CreateProgram Endpoint = "createprogram"
```

These names are a bit repetitive, their purpose is already written in the type

x/programs/cmd/simulator/cmd/request.go
```diff
@@ -32,28 +32,28 @@ var _ Cmd = (*runCmd)(nil)
...
+	// the message send to the simulator
```
iFrostizz, line 37:
```suggestion
	// the message sent to the simulator
```

x/programs/cmd/simulator/cmd/request.go
```diff
@@ -78,25 +78,13 @@ func (c *runCmd) Happened() bool {
...
-			planStep, err = os.ReadFile(*c.file)
-			if err != nil {
-				return err
-			}
-		}
-	default:
-		return errors.New("please specify either a --plan or a --file flag")
+	var marshaledBytes []byte
+	if c.requestMessage == nil || *c.requestMessage == "" {
+		return errors.New("please specify either a valid run command")
```
iFrostizz, line 83:
```suggestion
		return errors.New("please specify a valid request")
```

x/programs/cmd/simulator/cmd/request.go
```diff
@@ -66,7 +66,7 @@ func (c *runCmd) Run(ctx context.Context, log logging.Logger, db *state.SimpleMu
 	if err = c.Verify(); err != nil {
 		return newResponse(0), err
```
iFrostizz, line 67: Should we not pass the `c.numRequests` here instead of 0 ? It's always safe to pass in the deref of the poniter to the value since it's initialized to 0 and only incremented in `RunRequest` (never set to nil) if I'm not mistaken.

x/programs/cmd/simulator/src/simulator.rs
samliok: moved it down b/c @richardpringle comment
https://github.com/ava-labs/hypersdk/pull/1166#discussion_r1683354975

x/programs/cmd/simulator/cmd/config.go
samliok: don't think we should use snake_case here since its not `go` convention + enum types are UpperCamelCase anyways

x/programs/cmd/simulator/src/simulator.rs
samliok: same comment as below

x/programs/cmd/simulator/cmd/request.go
samliok: i dont think it matters since we are erroring here anyways and this is an empty response

x/programs/cmd/simulator/src/simulator.rs
samliok: will do in a follow up, made an issue
https://github.com/ava-labs/hypersdk/issues/1169

x/programs/cmd/simulator/src/simulator.rs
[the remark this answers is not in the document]
samliok: noted these comments in an issue, will change in a follow up
https://github.com/ava-labs/hypersdk/issues/1169

x/programs/cmd/simulator/src/simulator.rs
samliok: yea I agree, need to look into it more was out of the scope of this PR since its getting big already. noted for a follow up
https://github.com/ava-labs/hypersdk/issues/1169

x/programs/rust/examples/counter/src/lib.rs
samliok: If we have a `Simulator::new()` function the caller would need to explicitly specify the generic type arguments, I think it's cleaner to just have a `build_simulator` function and have the caller worry about this

something like `let simulator = Simulator::<Childstdin, ...>::new()`

x/programs/cmd/simulator/src/simulator.rs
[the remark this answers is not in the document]
richardpringle: haha whoops... obvious now that I look at it

x/programs/cmd/simulator/src/simulator.rs
iFrostizz: I see, you might as well scope it inside of the `get_binary_path` function since it's only used here.

x/programs/cmd/simulator/cmd/request.go
```diff
@@ -249,6 +222,11 @@ func (c *runCmd) runStepFunc(
 		// TODO: implement readonly for now just don't charge for gas
 		result, _, err := programExecuteFunc(ctx, c.log, db, testContext, params[1:], method, math.MaxUint64)
 		output := resultToOutput(result, err)
+		// check if the db has any new changes
+		if db.HasChanges() {
+			changes := db.Changes()
+			return fmt.Errorf("%w: %s, %s", ErrInvalidRequest, "readonly request should not have any changes", changes)
+		}
```
iFrostizz, line 229: I think that the `ReadOnly` endpoint should also not be able to deploy, send value, and call with value. Maybe it would make more sense to build these restrictions as part of the VM as compared to the simulator (in host calls). You could call `programExecuteFunc` function with an additional `readOnly bool` flag possibly

iFrostizz, line 229: I think that the `ReadOnly` endpoint should also not be able to deploy, send value, and call with value. Maybe it would make more sense to build these restrictions as part of the VM as compared to the simulator (in host calls). You could call `programExecuteFunc` function with an additional `readOnly bool` flag possibly

iFrostizz (commented):
Just need CI to pass, but it looks great !

samliok:
closing this pr, was part of a much larger update in #1205

PR #1168, opened 2024-07-19
Write some tests for Permissions.Has

iFrostizz:
The implementation of `Has` is very concise and  may look confusing. There were no tests written for that function.


state/keys_test.go
```diff
@@ -84,3 +84,50 @@ func TestMalformedKey(t *testing.T) {
 	keys := make(Keys)
 	require.False(keys.Add("", Read))
 }
+
+func TestHasPermissions(t *testing.T) {
```
aaronbuchwald, line 88: if we're going to add unit tests here, let's add them for all of the relevant cases

aaronbuchwald (changes requested):

state/keys_test.go
iFrostizz: Done. I added at least 2 unit tests for each variant

state/keys_test.go
```diff
@@ -84,3 +84,116 @@ func TestMalformedKey(t *testing.T) {
...
+			perm:     Read,
+			contains: All,
+			failing:  true,
+		},
+		{
+			name:     "read not write",
+			perm:     Read,
+			contains: Write,
+			failing:  true,
+		},
```
aaronbuchwald, line 112: Let's add read does not have allocate as a test case

state/keys_test.go
```diff
@@ -84,3 +84,116 @@ func TestMalformedKey(t *testing.T) {
...
+
+func TestHasPermissions(t *testing.T) {
+	tests := []struct {
+		name     string
+		perm     Permissions
+		contains Permissions
+		failing  bool
+	}{
+		{
+			name:     "read is read",
```
aaronbuchwald, line 96: Let's change the names from "X is X" and "X not Y" -> "X has Y" or "X does not have Y"

state/keys_test.go
```diff
@@ -84,3 +84,116 @@ func TestMalformedKey(t *testing.T) {
 	keys := make(Keys)
 	require.False(keys.Add("", Read))
 }
+
+func TestHasPermissions(t *testing.T) {
+	tests := []struct {
+		name     string
+		perm     Permissions
+		contains Permissions
+		failing  bool
```
aaronbuchwald, line 93: nit: rename failing to `has`

state/keys_test.go
```diff
@@ -84,3 +84,116 @@ func TestMalformedKey(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require := require.New(t)
+			sample := tt.perm.Has(tt.contains)
+
+			if tt.failing {
+				require.False(sample)
+			} else {
+				require.True(sample)
+			}
```
aaronbuchwald, line 196:
```suggestion
			has := tt.perm.Has(tt.contains)
			require.Equal(tt.has, has)
```

state/keys_test.go
aaronbuchwald: Could we add test cases for each of Read/Write/Allocate/All/None has/has not each of the individual permissions ie. read/write/allocate?

aaronbuchwald (changes requested):

state/keys_test.go
iFrostizz: Done, thanks!

state/keys_test.go
iFrostizz: renamed!

state/keys_test.go
iFrostizz, line 1: Rewrote the tests using sets.
The only thing I'm a bit unhappy with is that all subtests don't have independent names. For instance, we could have named the test `tt.name .. has .. perm.String()` but there is no way to convert a permission to a string.

iFrostizz: Found a solution

state/keys_test.go
iFrostizz: Instead used sets to check if the perm of the current subtest is contained in all the perms. If yes, we check that it is what we expected, the same is true for the opposite.

state/keys_test.go
```diff
@@ -84,3 +86,71 @@ func TestMalformedKey(t *testing.T) {
...
+	case Write:
+		return "write"
+	case None:
+		return "none"
+	case All:
+		return "all"
+	default:
+		return "unknown"
+	}
+}
```
aaronbuchwald, line 156: This should be a `String()` function implemented on the `Permissions` type

state/keys_test.go
```diff
@@ -84,3 +86,71 @@ func TestMalformedKey(t *testing.T) {
...
+			t.Run(name, func(t *testing.T) {
+				require := require.New(t)
+				contains := slices.Contains(tt.has, perm)
+				require.Equal(contains, has)
+			})
+		}
+	}
+}
+
+func (p Permissions) String() string {
```
aaronbuchwald, line 141: let's move this to `keys.go` instead of the test file

iFrostizz: Sure,
My reasoning with only having it in the test file is that these permissions can be added together and it would make this function very limiting in the amount of permissions it could represent. We may want to make it a bit more clever by making the string representation dynamic in the future

iFrostizz: I could do that as a followup

aaronbuchwald: I don't think this should ever be dynamic. I'd just add it into `keys.go`, this is fine to include outside of the test package imo and I don't think we should define a `String()` function on this type only within the test package.

PR #1170, opened 2024-07-19
Context plumbing in `ActionTest` suite

ARR4N:
As you mentioned, the `Context` could be polluted if shared across tests (it's unlikely, but I agree that it's worth avoiding, and it's quite simple). I changed `Run` to a method instead.

Wherever possible, the incoming `Context` is plumbed through to where it's used, if possible/appropriate.


iFrostizz (approved):
Thank you so much!

PR #1173, opened 2024-07-22
[WIP] NFT-VM + CFMM-VM

RodrigoVillar:
NFT-VM Functionality:

- [x] Store NFT Collections + Instances
- [x] Implement associated NFT functions (transfer, mint)
- [x] Create NFT Marketplace with associated functions (list, buy)

CFMM-VM Functionality:

- [x] Store tokens
- [x] Implement associated token functions (transfer, mint, burn)
- [x] Implement logic for creating LPs with given constant function formula
- [x] Implement associated LP functions (deposit liquidity, remove liquidity, swap) 
- [ ] Implement mechanism for fee generation + withdrawal


examples/nftvm/tests/integration/nft_collection_test.go
```diff
@@ -0,0 +1,19 @@
+// Tests core logic of NFT Collections and Instances
+
+package integration_test
+
+import "github.com/ava-labs/hypersdk/codec"
+
+const (
+	CollectionNameOne     = "The Louvre Museum"
```
aaronbuchwald, line 8: Does this need to be a separate file?

examples/nftvm/storage/errors.go
```diff
@@ -0,0 +1,11 @@
...
+// See the file LICENSE for licensing terms.
+
+package storage
+
+import "errors"
+
+var ErrInvalidBalance = errors.New("invalid balance")
+
+var ErrOrderAlreadyExists = errors.New("order already exists")
+var ErrCorruptInstanceMarketplaceIndicator = errors.New("instance marketplace indicator is corrupt")
```
aaronbuchwald, line 11: Could this error be more specific? What makes it corrupt?

examples/nftvm/storage/marketplace_storage.go
```diff
@@ -0,0 +1,94 @@
...
+
+func MarketplaceOrderStateKey(
+	orderID ids.ID,
+) (k []byte) {
+	k = make([]byte, consts.Uint16Len+ids.IDLen+consts.Uint16Len)
+	k[0] = marketplaceOrderPrefix
+	copy(k[1:], orderID[:])
+	binary.BigEndian.PutUint16(k[1+ids.IDLen:], MarketplaceOrderStateChunks)
+	return
+}
```
aaronbuchwald, line 28:
```suggestion
) ([]byte) {
	k := make([]byte, consts.Uint16Len+ids.IDLen+consts.Uint16Len)
	k[0] = marketplaceOrderPrefix
	copy(k[1:], orderID[:])
	binary.BigEndian.PutUint16(k[1+ids.IDLen:], MarketplaceOrderStateChunks)
	return k
}
```

same comment throughout - we more commonly use this style rather than naming the return parameter. This makes it easier to follow imo

examples/nftvm/storage/marketplace_storage.go
```diff
@@ -0,0 +1,94 @@
...
+func GenerateOrderID(
+	parentCollection codec.Address,
+	instanceNum uint32,
+	price uint64,
+) (ids.ID, error) {
+	v := make([]byte, codec.AddressLen+consts.Uint32Len+consts.Uint64Len)
+	// TODO: Check that this is OK
+	id := hashing.ComputeHash256Array(v)
+	return id, nil
+}
```
aaronbuchwald, line 39: It looks like this will always construct a byte array of the same length and compute the same hash. The parameters are not used.

examples/nftvm/storage/marketplace_storage.go
```diff
@@ -0,0 +1,94 @@
...
+
+func CreateMarketplaceOrder(
+	ctx context.Context,
+	mu state.Mutable,
+	parentCollection codec.Address,
+	instanceNum uint32,
+	price uint64,
+) (orderID ids.ID, err error) {
+	// Generate orderID + state key
+	orderID, err = GenerateOrderID(parentCollection, instanceNum, price)
```
aaronbuchwald, line 49: I think this will run into the bug mentioned above

examples/nftvm/storage/marketplace_storage.go
aaronbuchwald: What should the DB layout for the orders be? Should they be clustered together? If the state key is hashed, then they'll be scattered randomly across the database including interspersing them with other data that's stored under the same prefix.

examples/nftvm/storage/marketplace_storage.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func GetMarketplaceOrder(
+	ctx context.Context,
+	f ReadState,
+	orderID ids.ID,
+) (price uint64, err error) {
+	k := MarketplaceOrderStateKey(orderID)
+	values, errs := f(ctx, [][]byte{k})
+	return innerGetMarketplaceOrder(values[0], errs[0])
```
aaronbuchwald, line 73: I think this function is small enough to be inlined in both places

examples/nftvm/storage/marketplace_storage.go
```diff
@@ -0,0 +1,94 @@
...
+	v := make([]byte, consts.Uint64Len)
+	binary.BigEndian.PutUint64(v, price)
+
+	// Order is new, complete execution
+	return orderID, mu.Insert(ctx, orderStateKey, v)
+}
+
+func GetMarketplaceOrder(
+	ctx context.Context,
+	f ReadState,
```
aaronbuchwald, line 68: Can we avoid passing `ReadState` in as a function? This seems a bit hacky. If we can avoid it, we should

examples/nftvm/storage/nft_storage.go
```diff
@@ -0,0 +1,273 @@
...
+
+	mconsts "github.com/ava-labs/hypersdk/examples/nftvm/consts"
+)
+
+// PREFIXES MUST START AFTER 0x3 TO AVOID CONFLICTS
+const (
+	nftCollectionPrefix    = 0x4
+	nftInstancePrefix      = 0x5
+	marketplaceOrderPrefix = 0x6
+)
```
aaronbuchwald, line 23: Reading this creates a mental load because now I need to think about where the other prefixes are being used and if there are any prefixes after this.

I'd recommend defining the DB layout and all of the prefixes in a single file and then if you want to split up the functions, dividing those up.

examples/nftvm/storage/nft_storage.go
```diff
@@ -0,0 +1,273 @@
...
+	metadata = v[codec.AddressLen+consts.Uint16Len : codec.AddressLen+consts.Uint16Len+metadataLen]
+	// Extract marketplace indicator
+	marketplaceIndicator := binary.BigEndian.Uint16(v[codec.AddressLen+consts.Uint16Len+metadataLen:])
+	if marketplaceIndicator == uint16(1) {
+		isListedOnMarketplace = true
+	} else if marketplaceIndicator == uint16(0) {
+		isListedOnMarketplace = false
+	} else {
+		return codec.EmptyAddress, []byte{}, false, ErrCorruptInstanceMarketplaceIndicator
+	}
```
aaronbuchwald, line 269: can we replace this with a switch statement and replace `uint16(1)` and `uint16(0)` with constants? I'd recommend using `iota` and following the style here: https://github.com/ava-labs/avalanchego/blob/f4d8a3c0647b2b96aa0dbf5de6961d609d030b88/snow/state.go#L13

examples/nftvm/storage/nft_storage.go
```diff
@@ -0,0 +1,273 @@
...
+	return innerGetNFTInstance(value, err)
+}
+
+func innerGetNFTInstance(
+	v []byte,
+	err error,
+) (owner codec.Address, metadata []byte, isListedOnMarketplace bool, e error) {
+	if err != nil {
+		return codec.EmptyAddress, []byte{}, false, err
+	}
```
aaronbuchwald, line 255: We should check the error outside of this function rather than passing it in to do the error check here.

aaronbuchwald: we should leave to good design and re-usable code to reduce the amount of code it takes to implement a VM rather than try to eliminate every line possible

examples/nftvm/storage/nft_storage.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Insert metadata
+	binary.BigEndian.PutUint16(v[codec.AddressLen:], uint16(metadataLen))
+	copy(v[codec.AddressLen+consts.Uint16Len:], metadata)
+
+	// Insert marketplace indicator
+	if isListedOnMarketplace {
+		binary.BigEndian.PutUint16(v[codec.AddressLen+consts.Uint16Len+metadataLen:], 1)
+	} else {
+		binary.BigEndian.PutUint16(v[codec.AddressLen+consts.Uint16Len+metadataLen:], 0)
```
aaronbuchwald, line 219: let's use constants instead of 1 and 0 here

examples/nftvm/actions/test_helpers.go
```diff
@@ -0,0 +1,50 @@
...
+	collectionOneStateKey = storage.CollectionStateKey(collectionOneAddress)
+
+	instanceOneStateKey      = storage.InstanceStateKey(collectionOneAddress, 0)
+	instanceOneNum           = uint32(0)
+	instanceOneOrderPrice    = uint64(5)
+	instanceOneOrderID, _    = storage.GenerateOrderID(collectionOneAddress, instanceOneNum, instanceOneOrderPrice)
+	instanceOneOrderStateKey = storage.MarketplaceOrderStateKey(instanceOneOrderID)
+)
+
+func createAddressWithSameDigits(num uint8) (codec.Address, error) {
```
aaronbuchwald, line 38: Could you make a separate PR to the HyperSDK to generate a random `codec.Address` ?

examples/nftvm/actions/outputs.go
```diff
@@ -0,0 +1,39 @@
...
+// See the file LICENSE for licensing terms.
+
+package actions
+
+import "errors"
+
+var ErrOutputValueZero = errors.New("value is zero")
+
+var (
+	ErrOutputCollectionNameEmpty     = errors.New("collection name is empty")
```
aaronbuchwald, line 11: let's put the above error in the same var block

examples/nftvm/actions/outputs.go
```diff
@@ -0,0 +1,39 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package actions
```
aaronbuchwald, line 4: can we rename this file to `errors.go` ?  I think this was copy pasted from a bad example in MorpheusVM (stale code that should have been renamed

examples/nftvm/actions/test_helpers.go
RodrigoVillar: https://github.com/ava-labs/hypersdk/issues/1187

examples/nftvm/storage/marketplace_storage.go
RodrigoVillar: Addressing the first point, I forgot to populate `v` with the function arguments passed through.

Addressing the second point, I choose to hash a portion of order state keys as this way, orders have a negligible probability of colliding with each other. I'm not sure how orders under this model would be organized in the database, but my main thought was that since orders share the same prefix (`marketplaceOrderPrefix`), then they would be together or at least this would allow for iterating across the database for orders. I can definitely change how order IDs are generated though

examples/nftvm/tests/integration/nft_collection_test.go
RodrigoVillar: This was a remanent of the original unit testing setup, I moved it back into `integration_test.go`

RodrigoVillar:
Closing this as will be rewriting CFMM-VM to account for recent changes to the VM interface

PR #1176, opened 2024-07-22
Fix macros, typos, and convenience

iFrostizz:
- Fix some typos
- Fix indentation
- Wrong error message on state keys max length exceeded
- Remove redundant call to `cast_const()`

### DIFF
--- a/x/programs/cmd/simulator/src/lib.rs
+++ b/x/programs/cmd/simulator/src/lib.rs
@@ -340,8 +340,7 @@ impl ClientBuilder<'_> {
         let Child { stdin, stdout, .. } = Command::new(self.path)
             .arg("interpreter")
             .arg("--cleanup")
-            .arg("--log-level")
-            .arg("error")
+            .args(["--log-level", "error"])
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
             .spawn()?;
--- a/x/programs/rust/sdk-macros/src/lib.rs
+++ b/x/programs/rust/sdk-macros/src/lib.rs
@@ -267,9 +267,12 @@ pub fn state_keys(_attr: TokenStream, item: TokenStream) -> TokenStream {
     const MAX_VARIANTS: usize = u8::MAX as usize + 1;
 
     if variants.len() > MAX_VARIANTS {
-        return Error::new(variants[MAX_VARIANTS].span(), "Cannot exceed 256 variants")
-            .into_compile_error()
-            .into();
+        return Error::new(
+            variants[MAX_VARIANTS].span(),
+            format!("Cannot exceed {} variants", MAX_VARIANTS),
+        )
+        .into_compile_error()
+        .into();
     }
 
     let match_arms: Result<Vec<_>, _> = variants
--- a/x/programs/rust/sdk-macros/src/lib.rs
+++ b/x/programs/rust/sdk-macros/src/lib.rs
@@ -56,25 +56,25 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
 
             arg => {
                 let err = match arg {
-                Some(FnArg::Typed(PatType { ty, .. })) => {
-                    Error::new(
-                        ty.span(),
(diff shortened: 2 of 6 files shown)


x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -114,20 +114,4 @@ mod tests {
...
-    //     alloc((isize::MAX / 2) as usize + 1);
-    //     alloc((isize::MAX / 2) as usize + 1);
-    // }
-
-    //     #[test]
-    //     #[should_panic]
-    //     fn null_pointer_allocation() {
-    //         // see https://doc.rust-lang.org/1.77.2/std/alloc/struct.Layout.html#method.array
-    //         alloc(isize::MAX as usize);
-    //     }
```
iFrostizz, line 132: We might have a chance running these in a thread, in which case I will remove that change and write tests in another PR. otherwise will leave as-is.

x/programs/cmd/simulator/src/lib.rs
```diff
@@ -340,8 +340,7 @@ impl ClientBuilder<'_> {
         let Child { stdin, stdout, .. } = Command::new(self.path)
             .arg("interpreter")
             .arg("--cleanup")
-            .arg("--log-level")
-            .arg("error")
+            .args(["--log-level", "error"])
```
iFrostizz, line 343: It just makes our life easier since these two are so tightly coupled, it happens sometimes to write an arg in between and struggle to understand why the call to the simulator panics.

x/programs/rust/wasmlanche-sdk/src/memory.rs
iFrostizz: Given that "abort" means that the program won't do anything else, it seems unlikely to be able to catch these signals. Running a function in a new process is not really a thing either. see https://stackoverflow.com/questions/58242918/is-it-possible-to-run-a-function-as-a-process-in-rust

x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -267,12 +267,9 @@ pub fn state_keys(_attr: TokenStream, item: TokenStream) -> TokenStream {
     const MAX_VARIANTS: usize = u8::MAX as usize + 1;
 
     if variants.len() > MAX_VARIANTS {
-        return Error::new(
-            variants[MAX_VARIANTS].span(),
-            "Cannot exceed `u8::MAX` variants",
-        )
-        .into_compile_error()
-        .into();
+        return Error::new(variants[MAX_VARIANTS].span(), "Cannot exceed 256 variants")
```
samliok, line 270: nit: i'd rather not hardcode 256, and say `u8::MAX + 1` or even better 

`format!("Cannot exceed {} variants", MAX_VARIANTS))`

 but i'm not too opinionated on this.

samliok (dismissed):

x/programs/rust/sdk-macros/src/lib.rs
iFrostizz: Thanks! Applied your suggestion to use `MAX_VARIANTS`, it's better indeed.

iFrostizz:
Waiting for state schema to land

richardpringle (changes requested):
Blocking this for now as there are too many conflicts with my changes and these changes aren't a priority

x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -81,7 +81,7 @@ pub extern "C" fn alloc(len: usize) -> HostPtr {
 
     ALLOCATIONS.with_borrow_mut(|s| s.insert(ptr, len));
 
-    HostPtr(ptr.cast_const())
+    HostPtr(ptr)
```
iFrostizz, line 84: Closing this since the others changes have been incorporated as part of rewrites. The only missing thing is this redundant `cast_const` call

PR #1181, opened 2024-07-23
Remove `DeferDeserialize` as it is not safe to use

richardpringle:
We will keep it in the example it's needed in only


iFrostizz (approved):

iFrostizz:
I'm not sure if we want to encourage program writers to use this type by putting it in example, that may be the effect and people will shoot their feet

iFrostizz:
If there is no other solution, then maybe with enough caution comments

iFrostizz (approved):

PR #1188, opened 2024-07-23
Replace Controller Initialize with Factory initialization

joshua-kim:
Removes usage of `Initialize` to use a factory pattern to create VMs.

### DIFF
--- a/examples/morpheusvm/controller/controller.go
+++ b/examples/morpheusvm/controller/controller.go
@@ -31,28 +31,14 @@ import (
 	hstorage "github.com/ava-labs/hypersdk/storage"
 )
 
-var _ vm.Controller = (*Controller)(nil)
+var (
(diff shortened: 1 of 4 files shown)


examples/morpheusvm/controller/controller.go
```diff
@@ -141,10 +129,29 @@ func (c *Controller) Initialize(
...
+	metrics *metrics
+
+	txDB               database.Database
+	txIndexer          indexer.TxIndexer
+	acceptedSubscriber indexer.AcceptedSubscriber
+}
+
+func New() *vm.VM {
+	return vm.New(&factory{}, version.Version)
 }
```
aaronbuchwald, line 155: Optional - could we move this back to the top?

aaronbuchwald (dismissed):

aaronbuchwald (approved):

PR #1192, opened 2024-07-24
Refactor controller initialize return values into vm dependencies

joshua-kim:


vm/vm.go
```diff
@@ -55,19 +55,47 @@ const (
...
+		options.BlockBuilder = blockBuilder
+	})
+}
+
+func WithTxGossiper(txGossiper gossiper.Gossiper) Option {
+	return OptionFunc(func(_ *VM, options *Options) {
+		options.TxGossiper = txGossiper
+	})
+}
+
```
joshua-kim, line 86: Considering removing these

vm/vm.go
```diff
@@ -55,19 +55,47 @@ const (
...
```
joshua-kim, line 86: Wondering if we should keep this in or not

vm/vm.go
```diff
@@ -130,12 +158,39 @@ type VM struct {
...
-	return &VM{
-		factory: factory,
-		v:       v,
-		config:  NewConfig(),
+func New(
+	factory ControllerFactory,
+	v *version.Semantic,
+	actionRegistry chain.ActionRegistry,
+	authRegistry chain.AuthRegistry,
+	authEngine map[uint8]AuthEngine,
```
joshua-kim, line 166: Another idea is to set default auth to use (bls, ed25519) as `options`.

vm/vm.go
```diff
@@ -133,12 +161,39 @@ type VM struct {
...
+
+	// Set defaults
+	vm.options = &Options{
+		BlockBuilder: builder.NewTime(vm),
+		TxGossiper:   txGossiper,
 	}
+
+	for _, opt := range opts {
+		opt.Apply(vm, vm.options)
+	}
```
aaronbuchwald, line 194: Thought - Could go either way. What do you think of setting factories here and initializing within `Initialize` ?

I think as long as we're not starting any goroutines or allocating much here it's fine to do in `New` instead, but the way I had thought about it was passing in factories to `New` so that we're doing the same thing in `Initialize`.

vm/vm.go
```diff
@@ -58,19 +58,47 @@ const (
 	MinAcceptedBlockWindow = 1024
 )
 
+var _ Option = (*OptionFunc)(nil)
+
+type Options struct {
+	BlockBuilder builder.Builder
+	TxGossiper   gossiper.Gossiper
+}
```
aaronbuchwald, line 66: Could we store these values directly on the VM instead of in an options struct or is this style the standard? Since `Options` has direct access to the `*VM` type it seems like we can.

Going forward, when should we use an option as opposed to an argument to the constructor? I assume the constructor takes in the values every VM needs to provide and options enable configuring anything that is more custom.

aaronbuchwald: Should options be moved to a separate file?

aaronbuchwald (changes requested):

vm/vm.go
joshua-kim: This would require another interface to implement which we don't currently need... technically we could be more consistent by initializing within `Initialize` but I want to remove `Initialize` from the VM in the long-term

examples/morpheusvm/controller/controller.go
```diff
@@ -22,11 +21,11 @@ import (
 	"github.com/ava-labs/hypersdk/examples/morpheusvm/storage"
 	"github.com/ava-labs/hypersdk/examples/morpheusvm/version"
 	"github.com/ava-labs/hypersdk/extension/indexer"
-	"github.com/ava-labs/hypersdk/gossiper"
 	"github.com/ava-labs/hypersdk/pebble"
 	"github.com/ava-labs/hypersdk/vm"
 
 	ametrics "github.com/ava-labs/avalanchego/api/metrics"
+
```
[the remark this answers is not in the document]
joshua-kim: Ugh no... but we need one. This repo's imports are super messed up.

vm/vm.go
joshua-kim: I re-implemented... let me know if you think this looks better. 

> Going forward, when should we use an option as opposed to an argument to the constructor? I assume the constructor takes in the values every VM needs to provide and options enable configuring anything that is more custom.

I agree with this

> Should options be moved to a separate file?

Moved... I think we do want to probably delete the options I added though in the future when we finish merging https://github.com/ava-labs/hypersdk/pull/1195.

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -973,3 +979,15 @@ func (*appSender) SendCrossChainAppResponse(context.Context, ids.ID, uint32, []b
...
+
+func (*testBlockBuilderOption) Apply(vm *vm.VM, options *vm.Options) {
+	options.BlockBuilder = builder.NewManual(vm)
+}
+
+type testTxGossiperOption struct{}
+
+func (*testTxGossiperOption) Apply(vm *vm.VM, options *vm.Options) {
+	options.TxGossiper = gossiper.NewManual(vm)
+}
```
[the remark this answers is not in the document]
joshua-kim: Ended up moving this back into the `vm` package... let me know how you feel. I don't love this but the alternative would require us to export the builder/gossiper on the VM struct which I don't want to do.

vm/vm.go
```diff
@@ -133,12 +133,37 @@ type VM struct {
...
+	}
+
+	txGossiper, err := gossiper.NewProposer(vm, gossiper.DefaultProposerConfig())
+	if err != nil {
+		return nil, err
+	}
+
+	// Set defaults
+	vm.builder = builder.NewTime(vm)
+	vm.gossiper = txGossiper
```
aaronbuchwald, line 160: optional nit to set gossiper right after it's created

```suggestion
	txGossiper, err := gossiper.NewProposer(vm, gossiper.DefaultProposerConfig())
	if err != nil {
		return nil, err
	}
	// Set defaults
	vm.gossiper = txGossiper
	vm.builder = builder.NewTime(vm)
```

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -213,7 +213,8 @@ var _ = ginkgo.BeforeSuite(func() {
 		toEngine := make(chan common.Message, 1)
 		db := memdb.New()
 
-		v := controller.New()
+		v, err := controller.New(vm.WithManualGossiper(), vm.WithManualBuilder())
```
aaronbuchwald, line 216: non-commitable suggestion, but could we make this:

```suggestion
		v, err := controller.New(vm.WithManual())
```

Just a convenience function for testing

aaronbuchwald (approved):

Issue #1183, opened 2024-07-23
Bindings should copy the parameter names


iFrostizz:
When using bindings in programs, params are by default named "param_0", "param_1", ...
It would be nice to name them to have meaningful parameter names in documentation and when the LSP autocompletes the call to the function with prefilled parameter names.

PR #1193, opened 2024-07-24
Rewrite of parts of the `#[public]` macro

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1183

Copy parameters in bindings definition:
The parameter names used to be `param_0`, `param_1`, ...
The original parameter names have been copied to have better docs and LSP support.

You can verify this with `cargo doc -p counter -F bindings --open`

Screenshot 2024-07-24 at 13 42 06

As part of this, much of the code could be simplified further


x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -18,105 +19,89 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
+        Some(FnArg::Typed(pat_type)) if is_context(&pat_type.ty) => {
+            // pass parameter if ignored with underscore
+            if matches!(*pat_type.pat, Pat::Wild(_)) {
+                pat_type.pat = Box::new(Pat::Ident(PatIdent {
+                    attrs: vec![],
+                    by_ref: None,
+                    mutability: None,
+                    ident: Ident::new("_ctx", Span::call_site()),
+                    subpat: None,
+                }))
```
iFrostizz, line 41: This is now needed because the parameters are all user-defined. Meaning that the context that used to be `param_0` as expected could also be a wildcard `_` and this parameter cannot be passed around.
Now the issue is that this is going to fail if any other parameter is named `_ctx` such as with this function signature:
```rust
#[public]
pub fn inc(_: Context, _ctx: Context, external: Program, address: Address)
```

iFrostizz:
Closing as these changes have been incorporated into main.

PR #1198, opened 2024-07-25
Standardize chain.Action.Marshal function

containerman17:
As we aim to reduce the amount of code in VMs by following a "Sane Defaults" approach, I propose removing the need for explicit Marshal/Unmarshal functions in actions. Instead, we should use automatic marshalling by default, with an option for easy overrides.

~~Currently, manually marshalling 100,000 actions takes 40ms. Under this PR, it would take 94ms. An additional 54ms per 100k transactions seems acceptable.~~

- [x] Implement JavaScript counterpart of marshal/unmarshal
- [x] Allow overriding Marshal/Unmarshal
- [x] Integrate it into hypersdk
- [x] ~~Explore a further speed up with go:generate~~
- [x] ~~Check if Fuzz tests are working and required~~


containerman17:
## Benchmark Results
### TL;DR
It is double or even triple slower, but no one cares because marshal-unmarshal takes an insignificant portion of time at 100k TPS.

### 100k Marshal-Unmarshal Cycles
#### Transfer Action
The current manual approach takes 13ms per 100k cycles when single-threaded and 5ms when using 8 threads. The proposed reflection-based solution takes 44ms for single-threaded and 15ms for 8-threaded cycles. 15ms for 100k full cycles! Even on an 8-core machine (and potentially 16-32 cores), it is fast enough.

### A Struct with 100 Inner Structs
For an exaggerated structure with an array of 100 inner structs, the 8-threaded result for the reflection-based approach takes 1067ms for 100k cycles, while the manual approach takes 535ms.

### A Megabyte []byte Array
For an array with a million byte elements over 100k iterations, the reflection approach takes 13084ms on 8 cores, and the manual approach takes 14082ms. I commented out this test as there is no significant difference when comparing a single-field struct.

### Raw results
```
// goos: linux
// goarch: amd64
// pkg: github.com/ava-labs/hypersdk/chain
// cpu: AMD EPYC 7763 64-Core Processor
// BenchmarkMarshalUnmarshal/Transfer-Reflection-1-8                     25          44908275 ns/op        35200248 B/op     500002 allocs/op
// BenchmarkMarshalUnmarshal/Transfer-Reflection-2-8                     39          28402783 ns/op        35200235 B/op     500003 allocs/op
// BenchmarkMarshalUnmarshal/Transfer-Reflection-4-8                     52          19412363 ns/op        35200397 B/op     500006 allocs/op
// BenchmarkMarshalUnmarshal/Transfer-Reflection-8-8                     67          15940310 ns/op        35201158 B/op     500012 allocs/op
// BenchmarkMarshalUnmarshal/Transfer-Manual-1-8                         79          13639643 ns/op        14400048 B/op     200002 allocs/op
// BenchmarkMarshalUnmarshal/Transfer-Manual-2-8                        138           8520965 ns/op        14400173 B/op     200003 allocs/op
// BenchmarkMarshalUnmarshal/Transfer-Manual-4-8                        190           6174020 ns/op        14400186 B/op     200005 allocs/op
// BenchmarkMarshalUnmarshal/Transfer-Manual-8-8                        230           5177846 ns/op        14400366 B/op     200009 allocs/op
// BenchmarkMarshalUnmarshal/Complex-Reflection-1-8                       8         131652394 ns/op        62400150 B/op    1200002 allocs/op
// BenchmarkMarshalUnmarshal/Complex-Reflection-2-8                      15          79716490 ns/op        62400238 B/op    1200003 allocs/op
// BenchmarkMarshalUnmarshal/Complex-Reflection-4-8                      22          53110410 ns/op        62400367 B/op    1200005 allocs/op
// BenchmarkMarshalUnmarshal/Complex-Reflection-8-8                      27          42143834 ns/op        62400900 B/op    1200010 allocs/op
// BenchmarkMarshalUnmarshal/Complex-Manual-1-8                          19          57420578 ns/op        40800108 B/op     900002 allocs/op
// BenchmarkMarshalUnmarshal/Complex-Manual-2-8                          33          33921873 ns/op        40800186 B/op     900003 allocs/op
// BenchmarkMarshalUnmarshal/Complex-Manual-4-8                          51          22917034 ns/op        40800419 B/op     900005 allocs/op
// BenchmarkMarshalUnmarshal/Complex-Manual-8-8                          60          18573280 ns/op        40800467 B/op     900009 allocs/op
// PASS
// ok      github.com/ava-labs/hypersdk/chain      20.763s
```

### Transfer struct setup
```golang
type Transfer struct {
	To    codec.Address `json:"to"`
	Value uint64        `json:"value"`
	Memo  []byte        `json:"memo"`
}

transfer := Transfer{
	To:    codec.Address{1, 2, 3, 4, 5, 6, 7, 8, 9},
	Value: 12876198273671286,
	Memo:  []byte("Hello World"),
}

```


### Complex struct setup
```golang
type InnerStruct struct {
	Field1 int32
	Field2 string
	Field3 bool
	Field5 []byte
}

type TestStruct struct {
	Uint64Field uint64
	StringField string
	BytesField  []byte
	IntField    int
	BoolField   bool
	Uint16Field uint16
	Int8Field   int8
	InnerField  []InnerStruct
}

test := TestStruct{
	Uint64Field: 42,
	StringField: "Hello, World!",
	BytesField:  []byte{1, 2, 3, 4, 5},
	IntField:    -100,
	BoolField:   true,
	Uint16Field: 65535,
	Int8Field:   -128,
	InnerField: func() []InnerStruct {
		inner := make([]InnerStruct, 100)
		for i := 0; i < 100; i++ {
			inner[i] = InnerStruct{
				Field1: int32(i),
				Field2: fmt.Sprintf("Inner string %d", i),
				Field3: i%2 == 0,
				Field5: []byte{byte(i), byte(i + 1), byte(i + 2), byte(i + 3), byte(i + 4)},
			}
		}
		return inner
	}(),
}
	```

**Comment by containerman17** [comment]
```golang
func (p *Packer) PackInt(v int) {
	p.p.PackInt(uint32(v))
}

func (p *Packer) UnpackInt(required bool) int {
	v := p.p.UnpackInt()
	if required && v == 0 {
		p.addErr(fmt.Errorf("%w: Int field is not populated", ErrFieldNotPopulated))
	}
	return int(v)
}
```

Changed to 
```golang
// PackInt now accepts uint32 to ensure full support for 64-bit machines
func (p *Packer) PackInt(v uint32) {
	p.p.PackInt(v)
}

func (p *Packer) UnpackInt(required bool) uint32 {
	v := p.p.UnpackInt()
	if required && v == 0 {
		p.addErr(fmt.Errorf("%w: Int field is not populated", ErrFieldNotPopulated))
	}
	return v
}
```

containerman17:
Another idea closely related to the current issue is to consider making the `Size` method optional for `chain.Action` structs. While preallocating memory speeds up the unpacking process, it's not required for the VM to function. Perhaps we can make the `Size` method optional and create a "Performance Checklist" with a recommendation that any actions called more than 10 times per second should have a `Size` method implemented.

richardpringle:
>I propose removing the need for explicit Marshal/Unmarshal functions in actions. Instead, we should use automatic marshalling by default, with an option for easy overrides.

I agree with this, for what it's worth.

.gitignore
```diff
@@ -78,3 +78,5 @@ compose-dev.yaml
 # Do not ignore specific .txt files for license check
 !license-header.txt
 !license-style.txt
+
+*.prof
```
richardpringle, line 82: I really don't like how we ignore all kinds of files like this. Yesterday, I ran into an issue with some code picking up the wrong `.so` file because I didn't realize that we were git-ignoring all `.so` files. This is a bad habit. Files that should be ignored should be put in specific folders. Can we stick the `.prof` files in a profiles directory where we ignore the contents?

chain/dependencies.go
```diff
@@ -244,6 +246,7 @@ type Action interface {
 
 type Auth interface {
 	Object
+	Marshaler
```
richardpringle, line 249: `Marshaller` should have two `l`s.

richardpringle (commented):
Sorry I don't have anything more valuable to add

chain/dependencies.go
containerman17: Thanks, will use "Marshal" according to Go conventions

containerman17:
Converting to draft to try out `codec.reflectcodec` from avalancehgo

containerman17:
So, @aaronbuchwald introduced me to the `avalanchego/codec/reflectcodec` implementation.

Here are the results of my benchmarking:

First of all, all four methods of marshaling produce exactly the same bytes for transfers, which is a good start.

Both `avalanchego/codec/hierarchycodec` and `avalanchego/codec/linearcodec` demonstrate acceptable performance:

https://github.com/ava-labs/hypersdk/blob/ec9da79a940242ae541fbcb2762454277a9ae0a5/codec/auto_marshal_bench_test.go#L23-L34

Please do not compare this benchmark with previous ones, as the earlier ones performed a full marshal-unmarshal cycle, while this only handles the unmarshal process—we're not concerned with the speed of the marshal function.

Please note the "100k" results. With a manual implementation, unpacking 100k items on 8 cores takes 0.7ms. My implementation takes 5.3ms, `avalanchego/codec/hierarchycodec` takes 7ms, and `avalanchego/codec/linearcodec` takes 7.3ms.

This means that even for VRYX's 100k TPS, the time spent on unpacking is negligibly low.

Yes, the avalanchego implementation is the slowest, but in the context of 100k TPS, a 2ms difference would almost certainly make no real-world impact.

containerman17:
While implementing `reflectcodec`, I encountered a problem with the `codec.Codec` interface. The `Unmarshal` method only takes a byte array, which doesn’t work for our case:

```go
type Codec interface {
	MarshalInto(interface{}, *wrappers.Packer) error
	Unmarshal([]byte, interface{}) error
	Size(value interface{}) (int, error)
}
```
[source](https://github.com/ava-labs/avalanchego/blob/85ab9693d28fa243d1121df59e555693be4cc0ad/codec/codec.go#L23)

We need to use the `unmarshal` method from `genericCodec`, which reads directly from `wrappers.Packer` without exhausting it. However, `genericCodec` is private:

```go
func (c *genericCodec) unmarshal(
	p *wrappers.Packer,
	value reflect.Value,
	typeStack set.Set[reflect.Type],
) error {
```
[source](https://github.com/ava-labs/avalanchego/blob/85ab9693d28fa243d1121df59e555693be4cc0ad/codec/reflectcodec/type_codec.go#L529C1-L533C10)

I’ve created a prototype with a modified `Codec` interface in the [standardize-marshal-function-reflectcodec](https://github.com/ava-labs/hypersdk/tree/standardize-marshal-function-reflectcodec) branch. 

Here are the possible solutions:
1. **Custom Implementation**: Use the [dedicated implementation defined in this PR](https://github.com/ava-labs/hypersdk/blob/containerman/standardize-marshal-function/codec/auto_marshal.go). It's faster and independent but adds another serialization codec to the Avalanche ecosystem.
2. **Use `reflectcodec`**: Fewer implementations mean fewer bugs and less maintenance, but it requires making `genericCodec` public.
3. **Use Protobuf**: Minimizes maintenance but increases complexity and was previously rejected in `avalanchego` due to non-deterministic encoding.

containerman17:
We considered the `near/borsch-go` serializer, but it has the same issue as `reflectcodec`: the method capable of working with a reader instead of `[]byte` is private. Therefore, a fork would be inevitable if we choose `near/borsch-go`.

containerman17:
I redid the PR today and minimized the number of changes.

chain/dependencies.go
```diff
@@ -197,10 +197,14 @@ type Object interface {
...
 	// -1 means no start/end
 	ValidRange(Rules) (start int64, end int64)
+}
 
+type HasMarshal interface {
 	// Marshal encodes an [Action] as bytes.
 	Marshal(p *codec.Packer)
+}
 
+type HasSize interface {
```
aaronbuchwald, line 207: Could we define this as a single `Marshaller` interface? The recommended pattern in golang is to name single function interfaces as "action"-er, so `Marshaller` would be appropriate in this case. I think it makes sense to use just Marshaller because there's no reason to implement one and not the other.

Someone that defines their own `Marshal` function should also define their own size function instead of using the linear codec's size function. Falling back to it won't error, but it doesn't make sense to implement one and not the other.

examples/morpheusvm/actions/transfer.go
```diff
@@ -82,25 +81,33 @@ func (*Transfer) ComputeUnits(chain.Rules) uint64 {
...
+// }
+
+// Register this in registry/registry.go as the second argument to Action.Register
+// func UnmarshalTransfer(p *codec.Packer) (chain.Action, error) {
+// 	var transfer Transfer
+// 	p.UnpackAddress(&transfer.To)
+// 	transfer.Value = p.UnpackUint64(true)
+// 	p.UnpackBytes(MaxMemoSize, false, &transfer.Memo)
+// 	return &transfer, p.Err()
+// }
```
aaronbuchwald, line 113: Let's uncomment these and revert this diff to a comment over the original functions to explain that these are optional to implement - we should avoid keeping commented out code in the codebase

codec/type_parser.go
```diff
@@ -3,37 +3,60 @@
...
 	return &TypeParser[T]{
-		typeToIndex:    map[string]uint8{},
-		indexToDecoder: map[uint8]*decoder[T]{},
+		typeToIndex:     map[string]uint8{},
+		indexToDecoder:  map[uint8]*decoder[T]{},
+		registeredTypes: []HasTypeID{},
 	}
 }
 
+type HasTypeID interface {
```
aaronbuchwald, line 32: let's avoid `Has-` interface names , maybe `Typed`

examples/morpheusvm/registry/registry.go
```diff
@@ -25,12 +25,12 @@ func init() {
 	errs := &wrappers.Errs{}
 	errs.Add(
 		// When registering new actions, ALWAYS make sure to append at the end.
-		Action.Register((&actions.Transfer{}).GetTypeID(), actions.UnmarshalTransfer),
+		Action.Register(&actions.Transfer{}, nil),
```
aaronbuchwald, line 28: as per other comment, let's keep `actions.UnmarshalFrom` so we don't have commented out code

chain/actions.go
```diff
@@ -0,0 +1,21 @@
...
+	return codec.LinearCodec.Size(action)
+}
+
+func marshalActionInto(action Action, p *codec.Packer) error {
+	if actionWithMarshal, ok := action.(HasMarshal); ok {
+		actionWithMarshal.Marshal(p)
+		return nil
+	}
+	return codec.LinearCodec.MarshalInto(action, p.Packer)
+}
```
aaronbuchwald, line 21: Since we are just performing an interface check and then calling the optional interface here, could we switch this to taking in an interface here?

chain/actions_test.go
```diff
@@ -0,0 +1,104 @@
...
+	actionNoSizeFunc := &mockAction{}
+	actionWithSizeFunc := &mockActionWithSizeFunc{}
+
+	size1, err := getActionSize(actionNoSizeFunc)
+	require.NoError(err)
+	require.Equal(2, size1)
+
+	size2, err := getActionSize(actionWithSizeFunc)
+	require.NoError(err)
+	require.Equal(100000, size2)
```
aaronbuchwald, line 86: Could we use a constant here instead of putting `100000` in two places?

chain/actions_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	p1 := codec.NewWriter(0, consts.NetworkSizeLimit)
+	err := marshalActionInto(actionNoMarshal, p1)
+	require.NoError(err)
+	require.Equal([]byte{0, 7}, p1.Bytes())
+
+	p2 := codec.NewWriter(0, consts.NetworkSizeLimit)
+	err = marshalActionInto(actionWithMarshal, p2)
+	require.NoError(err)
+	require.Equal([]byte{1, 2, 3, 4, 5}, p2.Bytes())
```
aaronbuchwald, line 103: same comment on re-using an expected value here, so that the test reader does not have to look between here and where the function above is defined to return `[]byte{1, 2, 3, 4, 5}`

aaronbuchwald (changes requested):

chain/transaction_test.go
```diff
@@ -0,0 +1,155 @@
...
+				Memo:  []byte("world"),
+			},
+			&action2{
+				A: 2,
+				B: 4,
+			},
+		},
+	}
+
+	privBytes, err := hex.DecodeString("323b1d8f4eed5f0da9da93071b034f2dce9d2d22692c172f3cb252a64ddfafd01b057de320297c29ad0c1f589ea216869cf1938d88c9fbd70d6748323dbf2fa7")
```
RodrigoVillar, line 124: Nit: if the private key is random, we could use something like
`ed25519.GeneratePrivateKey()`; otherwise, perhaps we could set this to a
constant

chain/transaction_test.go
```diff
@@ -0,0 +1,155 @@
...
+func (*action2) StateKeysMaxChunks() []uint16 {
+	panic("unimplemented")
+}
+
+func (*action2) ValidRange(chain.Rules) (start int64, end int64) {
+	panic("unimplemented")
+}
+
+func (*mockTransferAction) ComputeUnits(chain.Rules) uint64 {
+	panic("ComputeUnits unimplemented")
```
RodrigoVillar, line 62:
```suggestion
	panic("unimplemented")
```
saying "function-name unimplemented" is redundant as in the error logs, it'll show the function as the one responsible for panicking.

codec/packer.go
```diff
@@ -147,10 +148,12 @@ func (p *Packer) UnpackWindow(w *window.Window) {
 	copy((*w)[:], p.p.UnpackFixedBytes(window.WindowSliceSize))
 }
 
+// Deprecated: Use PackBytes for better performance.
 func (p *Packer) PackString(s string) {
 	p.p.PackStr(s)
 }
 
+// Deprecated: Use UnpackBytes for better performance.
```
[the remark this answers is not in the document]
containerman17: Unpacking strings is very slow, much slower than working with the same data in bytes. It should never be used in on-chain operations. However, it is used in the WebSocket server, where performance is not a critical concern.

chain/actions.go
containerman17: No, we can't, because some actions (probably most of them) are not designed to implement the Marshaler interface.

containerman17: The point of this function is to hide the interface check

containerman17: Got it! Fixed, thanks!

chain/actions_test.go
```diff
@@ -0,0 +1,66 @@
...
+	size1, err := getSize(obj)
+	require.NoError(err)
+	require.Equal(consts.Uint8Len, size1)
+
+	objMarshaler := &mockObjectMarshaler{
+		mockSize: 100000,
+	}
+	size2, err := getSize(objMarshaler)
+	require.NoError(err)
+	require.Equal(100000, size2)
```
aaronbuchwald, line 47: let's use a constant for 100000

chain/actions_test.go
```diff
@@ -0,0 +1,66 @@
...
+	require.NoError(err)
+	require.Equal([]byte{7}, p1.Bytes())
+
+	objMarshaler := &mockObjectMarshaler{
+		mockBytes: []byte{1, 2, 3, 4, 5},
+	}
+	p2 := codec.NewWriter(0, consts.NetworkSizeLimit)
+	err = marshalInto(objMarshaler, p2)
+	require.NoError(err)
+	require.Equal([]byte{1, 2, 3, 4, 5}, p2.Bytes())
```
aaronbuchwald, line 65: Could we use `objMarshaller.mockBytes` instead of copying the same byte array

aaronbuchwald (dismissed):
LGTM with two nits that we should include as code quality improvements in the tests

aaronbuchwald (approved):

PR #1199, opened 2024-07-25
Write auth unit tests

iFrostizz:
Write some unit tests for the currently untested auth package and move all of the auth variants into their own package.


auth/consts.go
```diff
@@ -3,18 +3,10 @@
...
 	ED25519ID   uint8 = 0
 	SECP256R1ID uint8 = 1
 	BLSID       uint8 = 2
 )
-
-func Engines() map[uint8]vm.AuthEngine {
-	return map[uint8]vm.AuthEngine{
-		ED25519ID: &ED25519AuthEngine{},
-	}
-}
```
iFrostizz, line 20: This function was unused and now creates an import cycle. It was also probably not correctly placed.

auth/bls/bls_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/hypersdk/crypto"
+	"github.com/ava-labs/hypersdk/crypto/bls"
+)
+
+func TestBLSSign(t *testing.T) {
```
joshua-kim, line 16: For tests like this it's idiomatic to write a table-driven test ([ref](https://go.dev/wiki/TableDrivenTests)) to make it easy to add/modify tests.

I suspect that we can get a decent table-driven test by using a table like:
```
type test struct {
    name string
    message []byte
    signature []byte
    expected error
}
```

If we do this we should be able to clean up this test - since this test technically covers two different cases (one where the signature is valid, and one where the signature is invalid) and we should be testing for one.

auth/ed25519/ed25519_test.go
```diff
@@ -0,0 +1,81 @@
...
+	for i, signedMessage := range signedMessages {
+		verifyFunc := verifier.Add(signedMessage.Msg, signedMessage.Rauth)
+		if (i+1)%batchSize == 0 {
+			require.NotNil(verifyFunc)
+			err := verifyFunc()
+			require.NoError(err)
+		} else {
+			require.Nil(verifyFunc)
+		}
+	}
```
iFrostizz, line 80: It's not clear how `Done` should be called here. Since we need to know the batch size to know how much signature need to be aggregated, we already know when a batch is full and that the verification function pointer is returned.
This makes `Done` irrelevant.

auth/consts.go
iFrostizz: Oh actually it's called `examples/morpheusvm/controller/controller.go`, let's find a better place

iFrostizz: Moved it to `auth/common`, let me know if it's a good choice.

auth/bls/bls_test.go
iFrostizz: Sure, will do! 
Maybe for this purpose we could write fuzz tests, because it seems hard to think about messages that would find edge cases since they should all verify as expected, also the signature cannot really be guessed so we can only cheat here by checking the output and writing it as the test input.

iFrostizz: But absolutely fine for few cases that checks if verify works/fails as expected ! Thanks!

joshua-kim: Hmm yeah you're right. In that case you can either have a `setup func() ([]byte, []byte)` that generates the message + signature pair, or we can just split the current UT into two UTs (valid/invalid). We can also do fuzz testing like you mentioned... i think these are all reasonable so pick whatever you think looks best.

iFrostizz: So I have changes the bls test suite to have an `expectedSignature` in addition to the `signature`.
The `signature` will overwrite the signature that gets written by blst when calling `Sign` so that we can check that the verification fails as expected. Not sure it it's the best solution.

auth/secp256r1/secp256r1_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/hypersdk/crypto"
+	"github.com/ava-labs/hypersdk/crypto/secp256r1"
+)
+
+func TestSecp256r1Sign(t *testing.T) {
+	require := require.New(t)
+
+	priv, err := secp256r1.GeneratePrivateKey()
```
iFrostizz, line 19: It seems pretty hard to generate a deterministic secp256r1 key.

PR #1200, opened 2024-07-25
Remove reference to `snow.Context` to Controller

joshua-kim:
`snow.Context` exposes too much irrelevant stuff to the implementation (e.g consensus lock, atomic memory, ...). If we need to give the controller more information we can just update the `New` signature with a new parameter.

### DIFF
--- a/examples/morpheusvm/controller/controller.go
+++ b/examples/morpheusvm/controller/controller.go
@@ -9,7 +9,8 @@ import (
 	"net/http"
 
 	"github.com/ava-labs/avalanchego/database"
-	"github.com/ava-labs/avalanchego/snow"
+	"github.com/ava-labs/avalanchego/ids"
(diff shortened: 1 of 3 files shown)


examples/morpheusvm/controller/controller.go
```diff
@@ -78,8 +84,7 @@ func (*factory) New(
 		return nil, nil, nil, nil, nil, nil, nil, nil, err
 	}
 
-	c.snowCtx.Log.SetLevel(c.config.LogLevel)
```
joshua-kim, line 81: I don't know why we do this - I think we should just use whatever log level avalanchego uses.

aaronbuchwald: Makes sense to me, we should remove `LogLevel` from the config in this case

aaronbuchwald (approved):

PR #1201, opened 2024-07-26
Write FIFO cache unit tests

iFrostizz:


cache/fifo_test.go
```diff
@@ -0,0 +1,86 @@
...
+package cache
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFIFOCache(t *testing.T) {
+	limit := 10
```
joshua-kim, line 13: Maybe limit should be a parameter in the test table?

iFrostizz: Good point!

iFrostizz: Resolved

iFrostizz: thanks!

cache/fifo_test.go
```diff
@@ -0,0 +1,99 @@
...
+			limit:  10,
+			maxVal: 10,
+			check: func(t *testing.T, cache *FIFO[int, int]) {
+				require := require.New(t)
+				_, ok := cache.Get(0)
+				require.False(ok)
+				v, ok := cache.Get(1)
+				require.True(ok)
+				require.Equal(1, v)
+			},
```
joshua-kim, line 42: I think these tests are on the right track but still find that these are a little hard to reason about (particularly the limit/maxval) parameter are not obvious on how they work until you read the test body.

I think one way we could make this cleaner (and remove the `check` field) would be to make a struct like so:

```
func TestFIFOCache(t *testing.T) {
...
type put struct {
    i     int
    ok  bool
}

type get struct {
    i     int
    ok  bool
}
```

and if we did this I think we could write tests as:
```
{
    puts:[]put{
        {i: 1, ok: true}
        {i: 1, ok: false}
    }
    gets:[]get{
        {i: 1,	ok: true}
    }
}
```

cache/fifo_test.go
```diff
@@ -0,0 +1,99 @@
...
+		},
+		{
+			name:   "inserting before limit can insert again",
+			limit:  10,
+			maxVal: 8,
+			check: func(t *testing.T, cache *FIFO[int, int]) {
+				require := require.New(t)
+				exists := cache.Put(9, 9)
```
joshua-kim, line 50: Usually it's idiomatic to use `ok` in Go instead of `exists`

cache/fifo_test.go
```diff
@@ -0,0 +1,99 @@
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require := require.New(t)
+
+			cache, err := NewFIFO[int, int](tt.limit)
+			require.Equal(tt.expected, err)
```
joshua-kim, line 81: nit: use `ErrorIs`. This does a match to check if the error was wrapped instead of just the text which is more durable ([ref](https://go.dev/blog/go1.13-errors))

cache/fifo_test.go
```diff
@@ -0,0 +1,99 @@
...
+			if err != nil {
+				return
+			}
```
joshua-kim, line 84: If we want we can make the "check for valid `limit` test" a separate unit test. This way we don't need to have this check - but I don't feel strongly for or against.

aaronbuchwald: Let's make this a separate test, so we don't need this check in the body with the rest of the tests

cache/fifo_test.go
aaronbuchwald: This structure looks like a nice improvement. This should make it a bit more readable and we can just use a size like 2, so it's not excessive.

I think we should test the following cases here:
- no elements removed when the cache is less than full
- no elements removed when cache is exactly full
- elements removed in FIFO order when cache overfills
- elements removed in FIFO order after additional get/put ops for element to be removed FIFO

I think we're currently missing the last case. There's a unit test to make sure if we insert the first element again, then it stays in the cache, but we should additionally test that this does not change its position in the queue for removal to differentiate FIFO from LRU.

cache/fifo_test.go
iFrostizz: Mhh.. `ok` seems to suggest that something can go wrong, like inserting twice in a database that would not accept that. Here it's returning `true` if the value was already in the cache but that doesn't mean that it's not going to insert the changes afterall

iFrostizz: For `Get` it makes more sense though!

iFrostizz: I think that we should only use `ok` instead of `exists` for `Get` because `ok` seems to imply that something can go wrong while it just means that the value is already in the cache so no need to insert it again. It seems like `ok` could be used if the cache does not support overwriting value for instance.

cache/fifo_test.go
iFrostizz: Thanks both !
I have changed the test suite to only have a `name`, `osp` pairs and it's way more compact/readable.
Also added tests cases to cover @aaronbuchwald ones, haven't implemented the last one for now (TODO!)

cache/fifo_test.go
iFrostizz: I can't use `ErrorIs`, the error that says "maxSize must be greater than 0" is private and cannot be imported. It's defined in https://github.com/ava-labs/avalanchego/blob/321e727328a86bcfdb8f5bef164e7d50e698408b/utils/buffer/bounded_nonblocking_queue.go#L11
This error check has been moved to a unit test `TestEmptyCacheFails` that checks that this error occurs when the cache is initialized with a length of 0, so it's better contained.

cache/fifo_test.go
iFrostizz: I'm not sure if there is a real difference between the two last test cases, the case "elements removed in FIFO order when cache overfills" seems to fulfil both, please tell me if that's wrong.

iFrostizz: Actually they are different, see the test case "elements removed in FIFO order and not LRU".

aaronbuchwald (approved):

aaronbuchwald (approved):

PR #1203, opened 2024-07-26
Write builder unit tests

iFrostizz:
Writing parametrized tests for the `nextTime` function of the time builder. More tests for the builder to come!

### DIFF
--- a/builder/time_test.go
+++ b/builder/time_test.go
@@ -7,73 +7,205 @@ import (
 	"context"
 	"testing"
 
-	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/engine/common"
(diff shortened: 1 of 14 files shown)


builder/time_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+			time.lastQueue = tt.lastQueue
+
+			next := time.nextTime(tt.now, tt.preferred)
+			require.Equal(tt.expected, next)
+		})
+	}
+}
+
+func TestNextTimeIncreases(t *testing.T) {
```
iFrostizz, line 140: Have explicitly written this tests separately from the one above because they don't serve the exact same purpose. The first one is only based on one call to `nextTime` while the other one calls it twice.

PR #1204, opened 2024-07-26
Move state layout to options

joshua-kim:


vm/vm.go
```diff
@@ -158,6 +165,9 @@ func New(
 	// Set defaults
 	vm.builder = builder.NewTime(vm)
 	vm.gossiper = txGossiper
+	vm.heightKey = []byte("height")
+	vm.timestampKey = []byte("timestamp")
+	vm.feeKey = []byte("fee")
```
joshua-kim, line 170: todo we might introduce keys that conflict w/ the controller's predefined keys so it would be good to have a way to detect conflicts in the implementation side during initialization

aaronbuchwald: Alternatively, we could make the default to prefix the HyperSDK metadata and have a separate prefix for standard VM implementations portion of state.

joshua-kim: Another idea is to have an entirely separate state db for hypersdk defined state vs for state the controller defines

Issue #1165, opened 2024-07-18
Avoid disk re-read on restoring of accepted blocks processing


iFrostizz:
https://github.com/ava-labs/hypersdk/pull/1150 introduced restoration of the accepted blocks processing which guarantees that block subscribers will have a "at least one" subscription in case the processing fails in the middle. Instead, these blocks are stored on disk to start from the last point in order to broadcast it (or again) to the subscribers.
The issue is that we may read blocks multiple times from disk and should use a cache to avoid this. See https://github.com/ava-labs/hypersdk/pull/1150#discussion_r1683008413

iFrostizz:
Additional point from @aaronbuchwald https://github.com/ava-labs/hypersdk/pull/1150#discussion_r1684480745

> We should also remove the check to see if it's the last accepted block from this function in the follow on.
> 
> There should just be a helper function that will fetch an accepted block by height and the caller shouldn't need to worry about a special case.

PR #1214, opened 2024-07-29
Avoid to read same accepted blocks twice from disk

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1165


aaronbuchwald (changes requested):
Could we just use:

```golang
		blkID, err := vm.GetBlockIDAtHeight(ctx, height)
		if err != nil {
			return fmt.Errorf("could not find accepted block at height %d: %w", height, err)
		}

		blk, err := vm.GetStatelessBlock(ctx, blkID)
		if err != nil {
			return fmt.Errorf("could not find accepted block (%s) at height %d: %w", blkID, height, err)
		}
```

inside of the for loop so we re-use the existing functions that already check the cache

iFrostizz:
Thanks @aaronbuchwald , updated the `GetCachedBlock` function so that it uses these already-written functions. Placed them inside of the `GetCachedBlock` function to avoid repetition.

aaronbuchwald:
> Thanks @aaronbuchwald , updated the `GetCachedBlock` function so that it uses these already-written functions. Placed them inside of the `GetCachedBlock` function to avoid repetition.

I think we can just get rid of the new function completely and use https://github.com/ava-labs/hypersdk/pull/1214#pullrequestreview-2206378640 in the for loop body

vm/vm.go
```diff
@@ -1171,7 +1171,12 @@ func (vm *VM) loadAcceptedBlocks(ctx context.Context) error {
 		start = vm.lastAccepted.Hght - lookback
 	}
 	for i := start; i <= vm.lastAccepted.Hght; i++ {
-		blk, err := vm.GetDiskBlock(ctx, i)
+		blkID, err := vm.GetBlockIDAtHeight(ctx, i)
+		if err != nil {
+			vm.snowCtx.Log.Info("could not find block ID on-disk", zap.Uint64("height", i))
+			continue
+		}
+		blk, err := vm.GetStatelessBlock(ctx, blkID)
```
aaronbuchwald, line 1179: This code should be unchanged, this is called on startup when there are no blocks in the cache, so we need to load all of these from disk and can call `GetDiskBlock` directly

iFrostizz: Thanks, done!

aaronbuchwald (approved):

PR #1211, opened 2024-07-29
Update AvalancheGo to v1.11.10

aaronbuchwald:
This PR updates the AvalancheGo dependency to v1.11.10: https://github.com/ava-labs/avalanchego/releases/tag/v1.11.10

PR #1232, opened 2024-07-31
Remove usage of choices.Status

aaronbuchwald:
This PR replaces the usage of `choices.Status` with an `accepted` bool on the stateful block type.

This keeps the semantics the same for duplicate tx checking and where to return a block's inner view from. We should fully remove this boolean


aaronbuchwald:
Follow up from https://github.com/ava-labs/hypersdk/pull/1211

StephenButtolph (approved):

PR #1234, opened 2024-07-31
Handle wildcard in public function arguments

richardpringle:
Once we have upgrades, this will likely be more relevant. If someone wants to set an API but they don't know what they want to do with a particular parameter yet, they should be able to use wildcards

### DIFF
--- a/x/programs/rust/sdk-macros/src/lib.rs
+++ b/x/programs/rust/sdk-macros/src/lib.rs
@@ -2,16 +2,16 @@ extern crate proc_macro;
 
 use proc_macro::TokenStream;
 use proc_macro2::{Literal, Span};
-use quote::{quote, ToTokens};
+use quote::{format_ident, quote, ToTokens};
 use std::str::FromStr;
 use syn::{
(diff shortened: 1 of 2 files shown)


x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -94,13 +94,38 @@ pub fn public(_: TokenStream, item: TokenStream) -> TokenStream {
...
+                fn_arg.span(),
+                "Functions with the `#[public]` attribute cannot have a `self` parameter.",
+            )),
+            FnArg::Typed(pat_type) => {
+                let pat = match pat_type.pat.as_ref() {
+                    Pat::Wild(PatWild { attrs, .. }) => Pat::Ident(PatIdent {
+                        attrs: attrs.clone(),
+                        by_ref: None,
+                        mutability: None,
+                        ident: format_ident!("arg{}", i),
```
samliok, line 114: basically renaming the param to arg{index}

samliok (approved):

PR #1235, opened 2024-07-31
Refactor APIs out of Controller

joshua-kim:
Refactors the APIs out of the Controller.


examples/morpheusvm/cmd/morpheusvm/main.go
```diff
@@ -43,11 +52,54 @@ func main() {
...
+	if err != nil {
+		return fmt.Errorf("failed to initialize log: %w", err)
+	}
+
+	if err := ulimit.Set(ulimit.DefaultFDLimit, log); err != nil {
 		return fmt.Errorf("%w: failed to set fd limit correctly", err)
 	}
 
-	controller, err := controller.New()
+	txDBIndexer := indexer.NewTxDBIndexer(memdb.New())
```
joshua-kim, line 68: Using a memdb since it's an example vm + didn't want to add more code for cli flags to configure the db.

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -212,7 +214,44 @@ var _ = ginkgo.BeforeSuite(func() {
...
+			vm.WithManualGossiper[*controller.Controller](),
+			vm.WithManualBuilder[*controller.Controller](),
+			vm.WithBlockSubscriptions[*controller.Controller](indexerFactory, acceptBlockSubscription),
+			vm.WithRemoveTxSubscriptions[*controller.Controller](removeTxSubscription),
+			vm.WithVMAPIs[*controller.Controller](
+				rpc.JSONRPCServerFactory{},
+				indexerAPI,
+				webSocketFactory,
+			),
+			vm.WithControllerAPIs[*controller.Controller](&lrpc.JSONRPCServerFactory{}),
```
joshua-kim, line 252: I think this is pretty weird. I don't think we should have a concept of "controller" apis. I think apis should just be registering subscriptions into the VM and the VM doesn't even know what a controller api is. I think we can refactor the existing controller apis which are used to get hypersdk metadata like balances into the VM apis since that's not something the controller defines anyways.

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -226,6 +265,7 @@ var _ = ginkgo.BeforeSuite(func() {
 			app,
 		))
 
+		server.Initialize(v)
```
joshua-kim, line 268: Gross pattern, but we should be able to delete all of this once we get grpc-based apis in.

examples/morpheusvm/cmd/morpheusvm/main.go
```diff
@@ -43,11 +52,54 @@ func main() {
...
+	controller, err := controller.New(
+		vm.WithBlockSubscriptions[*controller.Controller](indexerFactory, blockSubscription),
+		vm.WithTxRemovedSubscriptions[*controller.Controller](txRemovedSubscription),
+		vm.WithVMAPIs[*controller.Controller](
+			rpc.JSONRPCServerFactory{},
+			indexerAPI,
+			webSocketFactory,
+		),
+		vm.WithControllerAPIs[*controller.Controller](&lrpc.JSONRPCServerFactory{}),
+	)
```
aaronbuchwald, line 102: As discussed yesterday, worth looking at an optional interface pattern where all of the options can optionally implement any of the interfaces, so that they'll be registered by the VM.

In this case, it's important to include any changes in the release notes and for users to include an interface check to make sure that they catch any changes from upstream instead of silently no longer registering their dependencies as desired if the optional interface signature changed.

vm/option.go
```diff
@@ -5,19 +5,46 @@ package vm
...
+	return func(vm *VM[T]) {
+		vm.vmAPIHandlerFactories = apiHandlerFactories
+	}
+}
+
+func WithControllerAPIs[T Controller](apiHandlerFactories ...rpc.HandlerFactory[T]) Option[T] {
+	return func(vm *VM[T]) {
+		vm.controllerAPIHandlerFactories = apiHandlerFactories
+	}
+}
```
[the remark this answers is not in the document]
joshua-kim: Yeah I think the existing controller apis should just be merged w/ the hypersdk api. The info being served in the controller api is hypersdk defined information so it makes sense for hypersdk to own it.

api/indexer/client.go
```diff
@@ -0,0 +1,65 @@
...
+	"strings"
+
+	"github.com/ava-labs/avalanchego/ids"
+
+	"github.com/ava-labs/hypersdk/requester"
+	"github.com/ava-labs/hypersdk/rpc"
+)
+
+func NewClient(uri string, name string, endpoint string) *Client {
+	uri = strings.TrimSuffix(uri, "/")
```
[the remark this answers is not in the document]
joshua-kim: I don't like this pattern either but this is to be consistent w/ all the other clients (ref: https://github.com/ava-labs/hypersdk/blob/23ccb7e692d9fe24eeb768cfdea0e89f878fa1c2/rpc/jsonrpc_client.go#L36)

vm/option.go
```diff
@@ -5,19 +5,46 @@ package vm
...
 
-func WithManualGossiper() Option {
-	return func(vm *VM) {
+func WithManualGossiper[T Controller]() Option[T] {
+	return func(vm *VM[T]) {
 		vm.gossiper = gossiper.NewManual(vm)
 	}
 }
+
+func WithBlockSubscriptions[T Controller](subscriptions ...event.SubscriptionFactory[*chain.StatelessBlock]) Option[T] {
```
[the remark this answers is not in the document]
joshua-kim: It matters in the sense that we call `Accept` in the order that the subscriptions are provided in. We could consider doing something like inserting these into an intermediary buffered channel instead of blocking the critical path of the hypersdk code.

api/indexer/api.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+type APIFactory struct {
+	path    string
+	name    string
+	indexer *TxDBIndexer
+}
+
+func (f *APIFactory) New(vm rpc.VM) (rpc.HTTPHandler, error) {
```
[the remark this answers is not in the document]
joshua-kim: So my thoughts are:
1. `New` needs to give us the type to build our api
  - If we want to use generics to get that type, we are effectively going to need to keep the controller vs vm api pattern that we have now to support hypersdk defined abstractions and application defined abstractions.
  - If this package defines its own interface to use (so some interface `indexer.API`), hypersdk can't depend on that package since it would be a circular dependency as the indexer is logically a downstream component from the vm.
2. I think it's reasonable for hypersdk to define a `VM` type (or set of types) that downstream apis can build off of. I think ideally this allows access to current state, the mempool, and subscriptions for consensus events (block acceptance).

vm/vm.go
```diff
@@ -466,32 +471,49 @@ func (vm *VM) Initialize(
...
+		}
+
+		vm.txRemovedSubscriptions = append(vm.txRemovedSubscriptions, subscription)
 	}
-	webSocketServer, pubsubServer := rpc.NewWebSocketServer(vm, vm.config.StreamingBacklogSize)
-	vm.webSocketServer = webSocketServer
-	vm.handlers[rpc.WebSocketEndpoint] = pubsubServer
 
-	// Add optional direct state read handler
-	if vm.config.EnableJSONRPCStateHandler {
```
joshua-kim, line 486: This PR removes all of the references to this new api we just merged in because it's not used in any tests.

api/indexer/tx_indexer.go
```diff
@@ -14,30 +13,58 @@ import (
...
+	}
+	apiFactory := &apiFactory{
+		path:    path,
+		name:    name,
+		indexer: indexer,
+	}
+
+	options := []vm.Option[T]{
+		vm.WithBlockSubscriptions[T](subscriptionFactory),
+		vm.WithVMAPIs[T](apiFactory),
```
joshua-kim, line 45: I feel like this looks weird (i.e you can now re-specify options and they append onto each other) ... but it seems to be better than the alternatives we discussed so I don't want to block on this + it seems good enough for now.

examples/morpheusvm/cmd/morpheusvm/main.go
joshua-kim: Replaced with the indexer-defined option as Aaron suggested offline.

rpc/websocket_server.go
```diff
@@ -186,19 +246,26 @@ func (w *WebSocketServer) MessageCallback(vm VM) pubsub.Callback {
...
 		}
 	}
 }
+
+func (w *WebSocketServer) Initialize(vm VM) {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+
+	w.vm = vm
+}
```
aaronbuchwald, line 271: Is this needed? It seems that it's only called from the integration tests

rpc/jsonrpc_server.go
```diff
@@ -10,12 +10,29 @@ import (
...
+	"github.com/ava-labs/hypersdk/api"
 	"github.com/ava-labs/hypersdk/chain"
 	"github.com/ava-labs/hypersdk/codec"
 	"github.com/ava-labs/hypersdk/consts"
 	"github.com/ava-labs/hypersdk/fees"
 )
 
+var _ api.HandlerFactory[VM] = (*JSONRPCServerFactory)(nil)
+
+type JSONRPCServerFactory struct{}
```
aaronbuchwald, line 22: Thoughts on using this pattern here: https://github.com/ava-labs/avalanchego/pull/2717#discussion_r1486611976 ?

vm/dependencies.go
```diff
@@ -33,7 +29,6 @@ type Config struct {
 	TransactionExecutionCores        int             `json:"transactionExecutionCores"`
 	StateFetchConcurrency            int             `json:"stateFetchConcurrency"`
 	MempoolSponsorSize               int             `json:"mempoolSponsorSize"`
-	StreamingBacklogSize             int             `json:"streamingBacklogSize"`
```
aaronbuchwald, line 36: If we're removing this here, we should remove it from the run script / e2e test fixture

vm/dependencies.go
```diff
@@ -51,7 +46,6 @@ type Config struct {
 	ProcessingBuildSkip              int             `json:"processingBuildSkip"`
 	TargetGossipDuration             time.Duration   `json:"targetGossipDuration"`
 	BlockCompactionFrequency         int             `json:"blockCompactionFrequency"`
-	EnableJSONRPCStateHandler        bool            `json:"enableJSONRPCStateHandler"`
```
aaronbuchwald, line 54: In the future, we should require an e2e test before adding a piece of functionality like this, but for now, should we add a test or GitHub issue to add a test, so that we don't leave this as dead code?

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -212,7 +215,40 @@ var _ = ginkgo.BeforeSuite(func() {
...
+
+		acceptBlockSubscription := &rpc.SubscriptionFuncFactory[*chain.StatelessBlock]{
+			AcceptF: func(event *chain.StatelessBlock) error {
+				return server.AcceptBlock(event)
+			},
+		}
+
+		v, err := controller.New(
+			vm.WithManualGossiper[*controller.Controller](),
+			vm.WithManualBuilder[*controller.Controller](),
```
aaronbuchwald, line 241: We need `WithManualGossiper` and `WithManualBuilder` here as overrides for the integration tests, but can we move the others into `controller.New(...)` itself, so that we don't need to place them here and in the entrypoint to the VM?

examples/morpheusvm/controller/controller.go
```diff
@@ -18,24 +15,17 @@ import (
...
 var (
-	_ vm.Controller        = (*Controller)(nil)
-	_ vm.ControllerFactory = (*factory)(nil)
+	_ vm.Controller                     = (*Controller)(nil)
+	_ vm.ControllerFactory[*Controller] = (*factory)(nil)
 )
 
-func New(options ...vm.Option) (*vm.VM, error) {
-	return vm.New(
+func New(options ...vm.Option[*Controller]) (*vm.VM[*Controller], error) {
```
aaronbuchwald, line 27: The controller should define the default options ie. supported APIs, subscriptions, etc. here or even further down (set as defaults in the VM itself) rather than pushing it to the caller of this function to decide which APIs to enable

examples/morpheusvm/cmd/morpheusvm/main.go
```diff
@@ -43,11 +52,50 @@ func main() {
...
+		},
+	}
+
+	blockSubscription := &rpc.SubscriptionFuncFactory[*chain.StatelessBlock]{
+		AcceptF: func(event *chain.StatelessBlock) error {
+			return server.AcceptBlock(event)
+		},
+	}
+
+	controller, err := controller.New(
```
aaronbuchwald, line 89: Same comment about pushing these into `controller.New`

aaronbuchwald:
As discussed offline, we should figure out how to move configuration options to the VM itself instead of depending on the config bytes passed in through AvalancheGo because we don't receive those until `vm.Initialize`

aaronbuchwald (changes requested):

vm/dependencies.go
joshua-kim: Added an issue: https://github.com/ava-labs/hypersdk/issues/1292

rpc/websocket_server.go
joshua-kim: Was needed previously to fix the circular dependency issue but with the new option pattern we're doing this can be removed.

vm/event/event.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package event
```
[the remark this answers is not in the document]
joshua-kim: We could make it its own directory. I felt like these are events that are defined by the VM so it made sense to me to have this as vm/event, but If we want to make it its own top-level package called `vmevent` or `event` I am open to that as well.

rpc/jsonrpc_server.go
joshua-kim: Seems weird to me since either you hide the concrete type and the caller loses type information or you expose the type and the caller has two ways of doing the same thing. I understand the argument for the aesthetic look of it though... but personally not a huge fan of the pattern

examples/morpheusvm/tests/integration/integration_test.go
joshua-kim: This creates an annoying circular dependency between the morpheusvm `controller` and `rpc` packages (since now the controller consumes the rpc package to initialize the default api, but the rpcs also need to know the controller to expose the api on the concrete type). I'm going to fix this by just merging the two packages together since the controller at this point is logically just controller api + genesis/upgrade and we will have a follow-up to merge the controller-defined apis into the VM apis since the controller apis right now are in a weird spot where they serve hypersdk-defined data like balances.

vm/event/event.go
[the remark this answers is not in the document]
joshua-kim: Actually I think it makes sense in its own package. The VM defines a concrete type of event that it emits so I think your proposal makes sense.

examples/morpheusvm/tests/integration/integration_test.go
aaronbuchwald: Sounds good

vm/event/event.go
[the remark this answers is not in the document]
aaronbuchwald: Yup I only ask since it's generic

aaronbuchwald (changes requested):

vm/vm.go
```diff
@@ -57,26 +58,34 @@ const (
 	MinAcceptedBlockWindow = 1024
 )
 
-type VM struct {
-	factory ControllerFactory
-	c       Controller
+type VM[T Controller] struct {
```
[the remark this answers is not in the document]
joshua-kim: Don't think we can do this until the controller is removed - we need the concrete type since the VM knows about the controller. We also have controller-defined apis now which the vm needs to set up so it needs to know the controller concrete type for this as well.

examples/morpheusvm/cmd/morpheusvm/main.go
```diff
@@ -43,11 +45,20 @@ func main() {
...
+	if err != nil {
+		return fmt.Errorf("failed to initialize log: %w", err)
+	}
+
+	if err := ulimit.Set(ulimit.DefaultFDLimit, log); err != nil {
 		return fmt.Errorf("%w: failed to set fd limit correctly", err)
 	}
 
-	controller, err := controller.New()
+	controller, err := controller.New(log, trace.Noop, memdb.New())
```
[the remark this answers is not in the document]
joshua-kim: I avoided this since this is an example + I didn't wanna creep scope of adding cli flag parsing to determine the db path. Added a TODO.

vm/network_tx_gossip.go
```diff
@@ -12,23 +12,23 @@ import (
...
-type TxGossipHandler struct {
-	vm *VM
+type TxGossipHandler[T Controller] struct {
+	vm *VM[T]
 }
 
-func NewTxGossipHandler(vm *VM) *TxGossipHandler {
-	return &TxGossipHandler{vm}
+func NewTxGossipHandler[T Controller](vm *VM[T]) *TxGossipHandler[T] {
+	return &TxGossipHandler[T]{vm}
```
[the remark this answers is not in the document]
joshua-kim: Added

examples/morpheusvm/controller/controller.go
```diff
@@ -4,38 +4,49 @@
...
-	_ vm.Controller        = (*Controller)(nil)
-	_ vm.ControllerFactory = (*factory)(nil)
+	_ vm.Controller                     = (*Controller)(nil)
+	_ vm.ControllerFactory[*Controller] = (*factory)(nil)
 )
 
-func New(options ...vm.Option) (*vm.VM, error) {
-	return vm.New(
+// New returns a VM with defaults configured with the specified options
+func New(log logging.Logger, tracer trace.Tracer, db database.Database, options ...vm.Option[*Controller]) (*vm.VM[*Controller], error) {
```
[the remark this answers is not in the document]
joshua-kim: Added a comment here.

vm/vm.go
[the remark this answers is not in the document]
joshua-kim: Nvm - I think we have to get rid of this now that the integration test code is common. Will unstage the generics diff.

vm/resolutions.go
```diff
@@ -197,6 +182,13 @@ func (vm *VM) processAcceptedBlock(b *chain.StatelessBlock) {
 	if err := vm.SetLastProcessedHeight(b.Height()); err != nil {
 		vm.Fatal("failed to update the last processed height", zap.Error(err))
 	}
+
+	// Update subscriptions
+	for _, subscription := range vm.blockSubscriptions {
+		if err := subscription.Accept(b); err != nil {
+			vm.Fatal("subscription failed to process block", zap.Error(err))
+		}
+	}
```
aaronbuchwald, line 191: We need to send the subscription updates before setting the last processed height otherwise we will re-open this issue: https://github.com/ava-labs/hypersdk/issues/1145. We should guarantee to each subscription that they receive an accepted block notification at least once.

aaronbuchwald (changes requested):

api/ws/server.go
```diff
@@ -1,26 +1,119 @@
...
+	AcceptF func(t T) error
+}
+
+func (s SubscriptionFunc[T]) Accept(t T) error {
+	return s.AcceptF(t)
+}
+
+func (SubscriptionFunc[_]) Close() error {
+	return nil
+}
```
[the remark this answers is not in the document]
joshua-kim: Going to just un-export this since it's only used within this package

api/state/client.go
```diff
@@ -10,10 +10,10 @@ import (
 	"github.com/ava-labs/hypersdk/requester"
 )
 
-func NewJSONRPCStateClient(uri string) *JSONRPCStateClient {
+func NewJSONRPCStateClient(uri string, name string) *JSONRPCStateClient {
 	uri = strings.TrimSuffix(uri, "/")
 	uri += JSONRPCStateEndpoint
-	req := requester.New(uri, Name)
+	req := requester.New(uri, name)
```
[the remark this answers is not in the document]
joshua-kim: Unstaged

examples/morpheusvm/cmd/morpheusvm/main.go
```diff
@@ -43,14 +44,20 @@ func main() {
...
 	if err != nil {
-		return fmt.Errorf("failed to initialize controller: %w", err)
+		return fmt.Errorf("failed to initialize log: %w", err)
+	}
+
+	if err := ulimit.Set(ulimit.DefaultFDLimit, log); err != nil {
+		return fmt.Errorf("%w: failed to set fd limit correctly", err)
 	}
 
+	controller := controller.New(log, trace.Noop)
```
aaronbuchwald, line 60: Can we get the log and trace level from the VM in the same way as the database?

api/ws/server.go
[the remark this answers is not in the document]
aaronbuchwald: I think this could be useful as a general purpose util, but un-exporting it since it's only used here makes sense for now too.

examples/morpheusvm/cmd/morpheusvm/main.go
```diff
@@ -43,14 +43,18 @@ func main() {
...
-		return fmt.Errorf("failed to initialize controller: %w", err)
+		return fmt.Errorf("failed to initialize log: %w", err)
+	}
+
+	if err := ulimit.Set(ulimit.DefaultFDLimit, log); err != nil {
+		return fmt.Errorf("%w: failed to set fd limit correctly", err)
 	}
 
-	return rpcchainvm.Serve(context.TODO(), controller)
+	return rpcchainvm.Serve(context.TODO(), controller.New())
```
aaronbuchwald, line 59: Could we revert this diff? I think it's fine to pass `NoLog` into `ulimit.Set(...)` rather than create a log to pass in here.

aaronbuchwald (approved):

aaronbuchwald (dismissed):

aaronbuchwald (approved):

PR #1239, opened 2024-08-01
Fix bug in calculating transaction usage

dboehm-avalabs:
The amount being added should be the recorded max chunks rather than the index of the max chunks


aaronbuchwald (approved):
LGTM after this passes CI

PR #1241, opened 2024-08-01
Convert ProgramIDs from ids.ID to []byte to support size suffix for state keys

dboehm-avalabs:
Convert ProgramIDs from ids.ID to []byte to support size suffix for state keys


richardpringle:
@dboehm-avalabs 
>to support size suffix for state keys

Is the size-suffix not a fixed size itself?

x/programs/rust/wasmlanche-sdk/src/program.rs
```diff
@@ -120,7 +120,7 @@ impl Program {
     /// # Panics
     /// Panics if there was an issue deserializing the account
     #[must_use]
-    pub fn deploy(&self, program_id: Id, account_creation_data: &[u8]) -> Address {
+    pub fn deploy(&self, program_id: &[u8], account_creation_data: &[u8]) -> Address {
```
richardpringle, line 123: Can you just change the `Id` type instead? Then you can make this `&Id` (or the type could still be owned).

richardpringle (changes requested):

x/programs/rust/wasmlanche-sdk/src/program.rs
dboehm-avalabs: it would have to be a distinct type since IDs still exist as the result of hashes.

dboehm-avalabs:
> @dboehm-avalabs
> 
> > to support size suffix for state keys
> 
> Is the size-suffix not a fixed size itself?

In this case, yes they are, but thought it might be worthwhile to support program ids that aren't to allow more flexibility to VM creators.  Having them be fixed size does have advantages though performance wise, so something I want to think about more in the future.

Issue #1225, opened 2024-07-30
Support External Indexer via gRPC AcceptedSubscriber + Optional Internal Indexing


aaronbuchwald:
The HyperSDK should enable VMs to index relevant data either in the node (for simple setup and quickly launching a network w/ everything they need) or using an external indexer that subscribes to the node in order to reliably process every block.

https://github.com/ava-labs/hypersdk/pull/1143 introduces a simple interface to subscribe to all blocks accepted by the HyperSDK and https://github.com/ava-labs/hypersdk/issues/1145 fixes a previous bug where the HyperSDK would not guarantee at least once delivery of accepted blocks.

The HyperSDK should make it as easy as possible for a basic data ingestion pipeline to process accepted blocks and spit out the relevant static data to index (and display in an explorer).

## External Subscriber

We should implement a gRPC service that implements the `AcceptedSubscriber` logic. This would consist of a sidecar that runs a server and listens on a given port, so that a user can configure the HyperSDK with a gRPC client that will dial the server and push all accepted blocks to the server.

The sidecar would export a listening address, which could then be passed into the HyperSDK APIs via config:

```json
{
    "exportedBlockSubscribers": "localhost:9001"
}
```

The external subscriber then needs to guarantee that:

1) It sends an acknowledgement back to the HyperSDK after it's processed each block (allow HyperSDK to clean up the block and continue processing the accepted queue)
2) Ensure block operation is idempotent to ensure it handles multiple deliveries correctly

## Standard Indexer to optionally Serve APIs within HyperSDK

This proposes a small change from the original ethos of the HyperSDK to do the absolute minimum inside of the node. Instead, this proposes to support an optional set of APIs to get started as quickly as possible.

This should include at least block and transaction indexing:
- block height -> block bytes index
- blockID -> block height
- blockID -> block bytes lookup
- txID -> txBytes
- results (TODO: this can be implemented in the future if results indexing becomes a clear requirement for getting started)

This can then support the basic APIs - `GetTransaction`, `GetBlockByHeight` and `GetBlock`.

## Transform Block to Static Data

Export a function to convert from a block to the relevant JSON for the block including all of the included transactions.

This can be as simple as:

```golang
var _ AcceptedSubscriber = (*pipeline)(nil)

type pipeline struct {
	blockProcessor BlockProcessor
	txProcessor txProcessor
}

func (p *pipeline) Accepted(ctx context.Context, blk *chain.StatelessBlock) error {
[...]

PR #1252, opened 2024-08-05
MorpheusVM External Subscriber Support

RodrigoVillar:
This PR addresses #1225 by creating the following:

- An `ExternalSubscriber` type which, when called, makes gRPC calls to a pre-specified VM Sidecar
- A standard indexer which allows for the indexing of both transactions and blocks
- A sidecar server for MorpheusVM which processes received blocks from the VM in addition to also being able to receive the parser for MorpheusVM

**Setting Up the External Indexer**
- In the `examples/morpheusvm/externalsubscriber/main` directory, run `go run .` which will start the sidecar server. Keep this instance open.
- In the main `morpheusvm` directory, run `./scripts/run.sh` which will start up a test instance of MorpheusVM. After a period of time, you will see that 1) the sidecar has received the parser information from the VM and 2) the sidecar is processing blocks that it has received from the VM


examples/morpheusvm/externalsubscriber/main/main.go
```diff
@@ -0,0 +1,37 @@
...
+	logger.Println("listening to TCP port 9001")
+
+	s := grpc.NewServer()
+	// pb.RegisterExternalSubscriberServer(s,
+	// &external_subscriber.MorpheusSidecar{})
+	morpheusSidecar := externalsubscriber.NewMorpheusSidecar()
+	mpb.RegisterMorpheusSubscriberServer(s, morpheusSidecar)
+	pb.RegisterExternalSubscriberServer(s, morpheusSidecar)
+
+	logger.Println("Server listening on port 9001")
```
aaronbuchwald, line 32: That's over 9000

RodrigoVillar:
Implemented a REST-API service for the MorpheusVM Sidecar

The following port now serves as the entry point for REST commands: `http://localhost:8080`

Below is the interface for the API
```
/block
Required Parameter: height
Returns: a JSON representation of the requested block

/tx
Require Parameter: txID
Returns: a JSON representation of the requested TX
 ```

**Comment by aaronbuchwald** [comment]
Nice, could you switch to implementing this as a JSON RPC API in the same style as in AvalancheGo? https://github.com/ava-labs/avalanchego/blob/master/api/server/router.go#L25

Instead of implementing our own API handlers, we should directly re-use the exact same http.Handler that is already implemented in the HyperSDK.

File: extension/indexer/standard_indexer.go [hunk 2]
```diff
@@ -0,0 +1,226 @@
...
+		v[consts.Uint64Len] = successByte
+	} else {
+		v[consts.Uint64Len] = failureByte
+	}
+	copy(v[consts.Uint64Len+1:], units.Bytes())
+	binary.BigEndian.PutUint64(v[consts.Uint64Len+1+fees.DimensionsLen:], fee)
+	return batch.Put(txID[:], v)
+}
+
+func (*StandardDBIndexer) storeTransactionStateful(
```
aaronbuchwald, line 185: Ah I see the results are not included in the stateful block (misnomer)

examples/morpheusvm/externalsubscriber/sidecar.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/ava-labs/hypersdk/pebble"
+
+	ametrics "github.com/ava-labs/avalanchego/api/metrics"
+	mpb "github.com/ava-labs/hypersdk/examples/morpheusvm/proto"
+	pb "github.com/ava-labs/hypersdk/proto"
+	hstorage "github.com/ava-labs/hypersdk/storage"
+)
+
+// TODO: clean up directory structure
+var stdDBDir = "./stdDB/"
```
aaronbuchwald, line 30: nit: make a const

examples/morpheusvm/externalsubscriber/sidecar.go
```diff
@@ -0,0 +1,160 @@
...
+
+type ExternalMorpheusAPI interface {
```
aaronbuchwald, line 32: We should make this a JSON RPC API https://github.com/ava-labs/hypersdk/pull/1252/#issuecomment-2274005695

examples/morpheusvm/externalsubscriber/sidecar.go
```diff
@@ -0,0 +1,160 @@
...
+
+type MorpheusSidecar struct {
+	pb.ExternalSubscriberServer
+	mpb.MorpheusSubscriberServer
+	ExternalMorpheusAPI
+	standardIndexer indexer.StandardIndexer
+	parser          *rpc.Parser
+}
+
+func NewMorpheusSidecar() *MorpheusSidecar {
```
aaronbuchwald, line 45: Can we make the directory a parameter instead of a constant?

examples/morpheusvm/externalsubscriber/sidecar.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return &MorpheusSidecar{standardIndexer: indexer.NewStandardDBIndexer(stdDB)}
+}
+
+func (m *MorpheusSidecar) Initialize(_ context.Context, initMsg *mpb.InitMsg) (*mpb.InitMsgAck, error) {
+	if m.parser != nil {
+		// TODO: implement better error handling for case where parser is
+		// already initialized
+		return &mpb.InitMsgAck{Success: true}, nil
+	}
```
aaronbuchwald, line 58: When will this function be called more than once?

examples/morpheusvm/externalsubscriber/sidecar.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Unmarshal chainID, genesis
+	chainID := ids.ID(initMsg.ChainID)
+	var gen genesis.Genesis
+	if err := json.Unmarshal(initMsg.Genesis, &gen); err != nil {
+		logger.Println("Unable to unmarhsal genesis", zap.Any("genesis", initMsg.Genesis))
+		return nil, err
+	}
+	m.parser = rpc.NewParser(initMsg.NetworkID, chainID, &gen)
```
aaronbuchwald, line 67: Is setting the genesis and parser the only part of this code that is specific to MorpheusVM? If so, could we factor it out so that this can be re-used for other VMs by replacing the genesis/parser specific part?

examples/morpheusvm/externalsubscriber/sidecar.go
```diff
@@ -0,0 +1,160 @@
...
+	logger.Println("External Subscriber has initialized the parser associated with MorpheusVM")
```
aaronbuchwald, line 68: Can we add the logger as a param to the constructor and get rid of the use of the global logger variable?

examples/morpheusvm/externalsubscriber/sidecar.go
```diff
@@ -0,0 +1,160 @@
...
+	// Unmarshal block
+	blk, err := chain.UnmarshalBlock(b.BlockData, m.parser)
+	if err != nil {
+		log.Fatalln("Failed to unmarshal block", zap.Any("block", b.BlockData))
+	}
+
+	// Call accept only if block has been indexed
+	if m.standardIndexer.BlockAlreadyIndexed(blk.Hght) {
+		return &pb.BlockAck{Success: true}, nil
+	}
```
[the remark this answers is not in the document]
RodrigoVillar: This function call makes sure that the operation is idempotent but more importantly, we are not repeating identical operations (i.e. writing the same block to DB multiple times)

examples/morpheusvm/externalsubscriber/sidecar.go
RodrigoVillar: Each node at initialization will send a parser init message; we therefore ignore identical parser init messages after the parser has been created

extension/indexer/standard_indexer.go
```diff
@@ -0,0 +1,210 @@
...
+	binary.BigEndian.PutUint64(k, height)
+	exists, _ := s.db.Has(k)
+	return exists
+}
+
+/*
+Helper storage functions
+*/
+
+func (*StandardDBIndexer) storeTransactionStateful(
```
aaronbuchwald, line 169: Can we be consistent with the naming of these functions instead of using both `store` and `map` as the prefixes, could we just use `put...` ?

aaronbuchwald (changes requested):

extension/indexer/external_subscriber_client.go
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package indexer
```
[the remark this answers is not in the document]
RodrigoVillar: > Can we move both the client and server into a sub-directory ie. grpcindexer ?
> 
> We should implement the server in a way that is VM agnostic rather than implementing it specifically for MorpheusVM

I've moved `external_subscriber_client.go` into it's own directory (`grpcindexer`). However, I haven't done the same yet for sidecar because of it's dependency on the MorpheusVM parser:

https://github.com/ava-labs/hypersdk/blob/4a6cafc648284c95fd186f3535638ceb3fd64208/examples/morpheusvm/rpc/jsonrpc_client.go#L130C1-L134C2

I think that for the grpc server to be fully VM agnostic, we should move this parser logic outside of MorpheusVM so that the server can utilize the VM-agnostic parser and thus be fully VM-agnostic

examples/morpheusvm/externalsubscriber/sidecar.go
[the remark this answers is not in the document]
aaronbuchwald: It should be the case that performing the same operation twice is just overwriting an existing index, which results in no change. In other words, processing block A should index block A and if you call it twice, it would just write the exact same key-value pairs resulting in no change. There's no need for an extra check like this.

extension/indexer/external_subscriber_client.go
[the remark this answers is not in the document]
aaronbuchwald: Why can't we pass this in as this interface instead: https://github.com/ava-labs/hypersdk/blob/86d90df4e07287ab2abe8bd4c79b7bc7a9d2a42a/chain/dependencies.go#L30 ?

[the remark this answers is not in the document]
RodrigoVillar: The parser dependency for the external subscriber server is more so because of the parser being sent during initialization. 

https://github.com/RodrigoVillar/hypersdk/blob/fd26c8185c021b391fcbae9efcdb019daf35884a/examples/morpheusvm/externalsubscriber/sidecar.go#L79-L86

In the case above, we are sending the parser details from the VM to the server, who will then call `rpc.NewParser()` with the values it extracted from the gRPC call. If we generalize the parser, then we can get rid of the Morpheus parser dependency and then be able to move the external subscriber server to `grpcindexer`

scripts/proto_codegen.sh
```diff
@@ -0,0 +1,41 @@
...
+
+TARGET=$PWD/proto
+if [ -n "${1:-}" ]; then
+  TARGET="$1"
+fi
+
+cd "$TARGET"
+
+echo "Creating .pb.go files"
+protoc --go_out=. --go-grpc_out=. subscriber.proto
```
aaronbuchwald, line 41: nit: add newline

aaronbuchwald, line 41: Can we copy the same file structure from AvalancheGo https://github.com/ava-labs/avalanchego/tree/master/proto/pb ie. `proto/servicename/servicename.proto` and then putting the generated files in `proto/pb/servicename/...`

We should also copy the same style as used here: https://github.com/ava-labs/avalanchego/blob/master/scripts/protobuf_codegen.sh and make sure that we are using either the same versions as in AvalancheGo or the latest versions of each dependency.

extension/grpcindexer/external_subscriber_client.go
```diff
@@ -0,0 +1,67 @@
...
+	pb "github.com/ava-labs/hypersdk/proto"
+)
+
+var _ indexer.AcceptedSubscriber = (*ExternalSubscriberClient)(nil)
+
+type ExternalSubscriberClient struct {
+	client pb.ExternalSubscriberClient
+}
+
+func NewExternalSubscriberClient(ctx context.Context, server string, networkID uint32, chainID ids.ID, genBytes []byte) (*ExternalSubscriberClient, error) {
```
aaronbuchwald, line 25: Could we rename `server -> serverAddr` and `genBytes -> genesisBytes` ? With this number of arguments, can we put one argument on each line and the return value on the final line as `) (*ExternalSubscriberClient, error) {`

extension/grpcindexer/external_subscriber_server.go
```diff
@@ -0,0 +1,85 @@
...
+	Logger        *log.Logger
+}
+
+func NewExternalSubscriberServer(stdDBDir string, logger *log.Logger, parserFactory ParserFactory) *ExternalSubscriberServer {
+	stdDB, err := hstorage.New(pebble.NewDefaultConfig(), stdDBDir, "db", avametrics.NewLabelGatherer("standardIndexer"))
+	if err != nil {
+		logger.Fatalln("Failed to create DB for standard indexer")
+	}
+	return &ExternalSubscriberServer{
+		indexer:       indexer.NewDBIndexer(stdDB),
```
aaronbuchwald, line 38: Can we replace this with the splitter design that we discussed? Instead of adding the `indexer.NewDBIndexer` here, the external subscriber server should split the input to a slice of accepted subscribers that it knows nothing about.

extension/grpcindexer/external_subscriber_server.go
```diff
@@ -0,0 +1,85 @@
...
+		return &emptypb.Empty{}, nil
+	}
+
+	// Unmarshal chainID
+	chainID := ids.ID(initRequest.ChainID)
+	// Create parser and store
+	parser, err := e.parserFactory(initRequest.NetworkID, chainID, initRequest.Genesis)
+	if err != nil {
+		e.Logger.Println("Unable to create parser", err)
+		return &emptypb.Empty{}, err
```
aaronbuchwald, line 55: Let's remove all usage of `logger.Println` in place of error handling.

If we hit an error, let's return it (with a formatted error if helpful) and avoid emitting extra logs in these `if err != nil {...` conditions

extension/grpcindexer/external_subscriber_server.go
```diff
@@ -0,0 +1,85 @@
...
+		indexer:       indexer.NewDBIndexer(stdDB),
+		Logger:        logger,
+		parserFactory: parserFactory,
+	}
+}
+
+func (e *ExternalSubscriberServer) Initialize(_ context.Context, initRequest *pb.InitRequest) (*emptypb.Empty, error) {
+	if e.parser != nil {
+		return &emptypb.Empty{}, nil
+	}
```
aaronbuchwald, line 47: We should not receive more than one `Initialize` request. If we do, we should return an error instead of ignoring the request (we only need this handling to accept multiple connections when running a single external server connected to multiple HyperSDK instances, but we shouldn't merge it in that way imo)

extension/grpcindexer/external_subscriber_server.go
```diff
@@ -0,0 +1,85 @@
...
+	// Unmarshal chainID
+	chainID := ids.ID(initRequest.ChainID)
+	// Create parser and store
+	parser, err := e.parserFactory(initRequest.NetworkID, chainID, initRequest.Genesis)
+	if err != nil {
+		e.Logger.Println("Unable to create parser", err)
+		return &emptypb.Empty{}, err
+	}
+	e.parser = parser
+	e.Logger.Println("External Subscriber has initialized the parser associated with MorpheusVM")
```
aaronbuchwald, line 58: We should use the AvalancheGo / zap logging libraries as we do in the rest of the HyperSDK and use `Info` logs instead of `Println`. Let's expect the same log type that's used here https://github.com/ava-labs/hypersdk/blob/87957cdab7a2baa5fcba116a8552c8e30ee56ad4/tests/integration/integration.go#L47

aaronbuchwald (changes requested):
I think there a few changes we should make here. First, the external subscriber should be a splitter from a single HyperVM client that dials the server to an arbitrary number of accepted subscribers. It should not be coupled to a specific indexer or MorpheusVM.

We should follow the style and directory format of AvalancheGo for the proto files and generation script here: https://github.com/ava-labs/avalanchego/blob/master/scripts/protobuf_codegen.sh

Instead of adding a new `main` entrypoint to run the server, we should add this to the HyperSDK CLI, so that it can start an external server for a given VM and the default tx indexing. This should make it easy for people to get started. We should also add an integration test under `tests/integration/integration.go` that spins up an external server and has one of the three VM instances dial it.

RodrigoVillar:
TODO: running `./scripts/tests.integration.sh` in MorpheusVM will fail because the splitter might not be running; we should update the integration tests to either set up a splitter beforehand or to not add an external subscriber

examples/morpheusvm/cmd/morpheus-cli/cmd/root.go
RodrigoVillar, line 1: TODO: spinning up splitter with Morpheus-CLI and then running `scripts/run.sh`
will fail because the first instance of Morpheus-CLI will not give up the lock
on the database, will `scripts/run.sh` needs the DB in order to initialize

RodrigoVillar:
> TODO: running `./scripts/tests.integration.sh` in MorpheusVM will fail because the splitter might not be running; we should update the integration tests to either set up a splitter beforehand or to not add an external subscriber

Update: splitter is now created prior to test suite and therefore, integration tests now pass

examples/morpheusvm/cmd/morpheus-cli/cmd/root.go
RodrigoVillar: > TODO: spinning up splitter with Morpheus-CLI and then running `scripts/run.sh` will fail because the first instance of Morpheus-CLI will not give up the lock on the database, while `scripts/run.sh` needs the DB in order to initialize
> 
> ```go
> rootCmd.PersistentPreRunE = func(*cobra.Command, []string) error {
>     utils.Outf("{{yellow}}database:{{/}} %s\n", dbPath)
>     controller := NewController(dbPath)
>     root, err := cli.New(controller)
>     if err != nil {
> 	    return err
>     }
>     handler = NewHandler(root)
>     return err
> }
> ```

@aaronbuchwald the only change I didn't incorporate was removing `main` due to the CLI issue above

proto/externalsubscriber/externalsubscriber.proto
```diff
@@ -0,0 +1,26 @@
...
+message BlockRequest {
+  bytes block_data = 1;
+  bytes results = 2;
+}
+
+message InitRequest {
+  uint32 network_id = 1;
+  bytes chain_id = 2;
+  bytes genesis = 3;
+}
```
aaronbuchwald, line 26: Let's reorganize these and put init before BlockRequest both here and under the service

scripts/protobuf_codegen.sh
```diff
@@ -0,0 +1,55 @@
...
+if ! buf lint;  then
+    echo "ERROR: protobuf linter failed"
+    exit 1
+fi
+
+echo "Re-generating protobuf..."
+if ! buf generate;  then
+    echo "ERROR: protobuf generation failed"
+    exit 1
+fi
```
aaronbuchwald, line 55: nit: add new line

extension/splitter/splitter.go
```diff
@@ -0,0 +1,55 @@
...
+		tcpPort:    tcpPort,
+		log:        log,
+	}
+}
+
+func (s *Splitter) Start() {
+	go func() {
+		if err := s.gRPCServer.Serve(s.listener); err != nil {
+			s.log.Fatal("failed to serve TCP port", zap.Any("Address", s.tcpPort))
+		}
```
aaronbuchwald, line 46: Let's include the error in the fatal log here

extension/indexer/tx_indexer.go
```diff
@@ -1,116 +0,0 @@
-// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
-// See the file LICENSE for licensing terms.
-
-package indexer
```
aaronbuchwald, line 4: Could we revert this diff? This should make this PR easier to review, and I think the only change needed to this file is the function signature of `Accepted`

extension/splitter/splitter.go
```diff
@@ -0,0 +1,55 @@
...
+	listener   net.Listener
+	tcpPort    string
+	log        logging.Logger
+}
+
+func NewSplitter(e *grpcindexer.ExternalSubscriberServer, log logging.Logger, tcpPort string) *Splitter {
+	lis, err := net.Listen("tcp", tcpPort) // #nosec G102
+	if err != nil {
+		log.Fatal("Could not listen to TCP port", zap.Any("Address", tcpPort))
+	}
```
aaronbuchwald, line 29: Can we propagate the error to the caller instead of using `log.Fatal` here?

extension/grpcindexer/external_subscriber_client.go
```diff
@@ -0,0 +1,80 @@
...
+	log.Debug("Established connection to External Subscriber Server", zap.Any("Server Address", serverAddr))
+	return &ExternalSubscriberClient{
+		client: client,
+		log:    log,
+	}, nil
+}
+
+func (e *ExternalSubscriberClient) Accepted(ctx context.Context, blk *chain.StatefulBlock, results []*chain.Result) error {
+	// Make gRPC call to client
+	e.log.Debug("Sending block over to server", zap.Any("Block", blk))
```
aaronbuchwald, line 63: Is this comment necessary? Not sure it adds any clarification within this function.

What will this log output for a stateful block? I don't think it implements `fmt.Stringer`, so this may result in difficult to read or not very useful output

aaronbuchwald (changes requested):

joshua-kim:
Summarizing our offline conversation - but we can add support for configs in options. We can have the VM be generic on the type of the config specified when hypersdk is initialized (e.g `vm.New[Config]`). During `Initialize`, we can unmarshal the config that avalanchego provides us into the user-defined config + expose it on the VM struct so that when options are initialized, they'll have access to the concrete config type.

PR #1254, opened 2024-08-06
Put the `.so` file in the target directory

richardpringle:
Okay... I don't know how many changes I'm actually going to end up stuffing into this PR, but they will be logically separated by commit


x/programs/simulator/src/lib.rs
```diff
@@ -1,12 +1,35 @@
...
+mod bindings {
+    #![allow(non_upper_case_globals)]
+    #![allow(non_camel_case_types)]
+    #![allow(non_snake_case)]
+    #![allow(dead_code)]
+
+    use wasmlanche_sdk::Address as SdkAddress;
+    // include the generated bindings
+    // reference: https://rust-lang.github.io/rust-bindgen/tutorial-3.html
+    include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
```
richardpringle, line 10: This should be cleaned up to use `PathBuff::new(env!("OUT_DIR"))` and `join` instead of adding `/` by hand.

x/programs/simulator/src/lib.rs
```diff
@@ -1,12 +1,35 @@
...
-include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+mod bindings {
+    #![allow(non_upper_case_globals)]
+    #![allow(non_camel_case_types)]
+    #![allow(non_snake_case)]
+    #![allow(dead_code)]
+
+    use wasmlanche_sdk::Address as SdkAddress;
+    // include the generated bindings
+    // reference: https://rust-lang.github.io/rust-bindgen/tutorial-3.html
```
richardpringle, line 9:
```suggestion
```

x/programs/simulator/src/lib.rs
```diff
@@ -1,12 +1,35 @@
...
+    include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+
+    impl std::ops::Deref for Bytes {
+        type Target = [u8];
+
+        fn deref(&self) -> &Self::Target {
+            unsafe { std::slice::from_raw_parts(self.data, self.length as usize) }
```
richardpringle, line 16: Missing safety comment

x/programs/simulator/src/lib.rs
```diff
@@ -1,12 +1,35 @@
...
+        }
+    }
+
+    impl From<SdkAddress> for Address {
+        fn from(value: SdkAddress) -> Self {
+            Address {
+                // # Safety:
+                // Address is a simple wrapper around an array of bytes
+                // this will fail at compile time if the size is changed
+                address: unsafe { std::mem::transmute::<SdkAddress, [libc::c_uchar; 33]>(value) },
```
samliok, line 26: should use SdkAddress::Len?

x/programs/simulator/src/lib.rs
```diff
@@ -1,12 +1,35 @@
...
-#![allow(non_camel_case_types)]
-#![allow(non_snake_case)]
-// include the generated bindings
-// reference: https://rust-lang.github.io/rust-bindgen/tutorial-3.html
-include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+mod bindings {
+    #![allow(non_upper_case_globals)]
+    #![allow(non_camel_case_types)]
+    #![allow(non_snake_case)]
+    #![allow(dead_code)]
```
samliok, line 5: why allow dead_code?

x/programs/simulator/src/lib.rs
richardpringle: I don't think it matters, but yeah that might be better

x/programs/simulator/src/lib.rs
richardpringle: Soooo much of the generated code isn't actually used. You don't notice when it's in `lib` because it's all marked `pub`, but when you put it in a private mod, it all shows up as dead-code

PR #1259, opened 2024-08-08
Make sure to run rust integration tests

richardpringle:
We need the `--tests` flag to make sure the integration tests are run


richardpringle:
🤔 behaviour seems to be different in CI than locally. This actually runs fewer tests

PR #1260, opened 2024-08-08
DO NOT MERGE

richardpringle:
Just testing some things in CI, don't merge me!


.gitignore
```diff
@@ -72,4 +72,5 @@ Cargo.lock
 x/programs/simulator/libsimulator.h
 x/programs/runtime/debug
 x/programs/runtime/wasm32-unknown-unknown
-x/programs/runtime/.rustc_info.json
```
richardpringle, line 75: Always add newlines at the end of files. I tried adding a new filename here and broke this because there was no new line

res-linux
```diff
@@ -0,0 +1,333 @@
...
+                    180,
+                    207,
+                    37,
+                    41,
+                    25,
+                    168,
+                    156,
+                    135,
+                    220,
+                ],
```
richardpringle, line 44: This changes every time I run the tests. It shouldn't, it should be deterministic

res-linux
```diff
@@ -0,0 +1,333 @@
...
+            0,
+            0,
+            0,
+            0,
+        ],
+        state_cache: RefCell {
+            value: {},
+        },
+    },
+    of: Address(
```
richardpringle, line 281: This value is wrong, it should be all 1s

res-linux
```diff
@@ -0,0 +1,333 @@
...
+            0,
+            0,
+            0,
+            0,
+        ],
+        state_cache: RefCell {
+            value: {},
+        },
+    },
+    to: Address(
```
richardpringle, line 124: This value is wrong, it should be all 1s

res-linux
```diff
@@ -0,0 +1,333 @@
...
+            0,
+            208,
+            8,
+            0,
+            64,
+            255,
+            255,
+            0,
+            0,
+            1,
```
richardpringle, line 142: It looks like the address is actually starting here... That could mean that something is the wrong size?

PR #1263, opened 2024-08-09
Reintroduce the Single-Line Stop Script

containerman17:
**Motivation:**
1. Stopping the chain should offer a user experience (UX) similar to starting it.
2. It's easier to type. With tabs, it's just `./sc[tab]s[tab]`.


examples/morpheusvm/scripts/stop.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+set -e
+
+killall avalanche-network-runner
```
aaronbuchwald, line 7: nit: add new line

aaronbuchwald: We should also add this script back as the way to stop the network mentioned in the README

aaronbuchwald (approved):

PR #1264, opened 2024-08-09
feat: add api for direct state access

### LINKED ISSUE #1226: Direct Current State Access API
Based off of the idea in https://github.com/ava-labs/hypersdk/issues/966

Implement an optional API that provides a direct read of the current HyperSDK state.

najeal:
This Closes #1226 

It adds an optional API to access vm state.
Not sure where this should be tested.

### DIFF
--- a/rpc/jsonrpc_state_server.go
+++ b/rpc/jsonrpc_state_server.go
@@ -33,8 +33,8 @@ type JSONRPCStateServer struct {
 }
 
 func (s *JSONRPCStateServer) ReadState(req *http.Request, args *StateRequest, res *StateResponse) error {
-	_, span := s.stateReader.Tracer().Start(req.Context(), "Server.ReadState")
+	ctx, span := s.stateReader.Tracer().Start(req.Context(), "Server.ReadState")
 	defer span.End()
-	res.Values, res.Errors = s.stateReader.ReadState(req.Context(), args.Keys)
+	res.Values, res.Errors = s.stateReader.ReadState(ctx, args.Keys)
 	return nil
 }
--- a/rpc/jsonrpc_state_server.go
+++ b/rpc/jsonrpc_state_server.go
@@ -1,0 +1,40 @@
+package rpc
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/ava-labs/avalanchego/trace"
+)
+
+type stateReader interface {
+	Tracer() trace.Tracer
+	ReadState(ctx context.Context, keys [][]byte) ([][]byte, []error)
+}
+
+type StateRequest struct {
+	Keys [][]byte
+}
+
+type StateResponse struct {
+	Values [][]byte
+	Errors []error
+}
+
+func NewJSONRPCStateServer(stateReader stateReader) *JSONRPCStateServer {
+	return &JSONRPCStateServer{
+		stateReader: stateReader,
+	}
+}
+
+// JSONRPCStateServer gives direct read access to the vm state
+type JSONRPCStateServer struct {
+	stateReader
+}
+
+func (s *JSONRPCStateServer) ReadState(req *http.Request, args *StateRequest, res *StateResponse) error {
+	_, span := s.stateReader.Tracer().Start(req.Context(), "Server.ReadState")
+	defer span.End()
+	res.Values, res.Errors = s.stateReader.ReadState(req.Context(), args.Keys)
+	return nil
+}
--- a/rpc/jsonrpc_state_server.go
+++ b/rpc/jsonrpc_state_server.go
@@ -10,21 +10,21 @@ import (
 	"github.com/ava-labs/avalanchego/trace"
 )
 
-type stateReader interface {
+type StateReader interface {
 	Tracer() trace.Tracer
 	ReadState(ctx context.Context, keys [][]byte) ([][]byte, []error)
 }
 
-type StateRequest struct {
+type ReadStateRequest struct {
 	Keys [][]byte
 }
 
-type StateResponse struct {
+type ReadStateResponse struct {
 	Values [][]byte
 	Errors []error
 }
 
-func NewJSONRPCStateServer(stateReader stateReader) *JSONRPCStateServer {
+func NewJSONRPCStateServer(stateReader StateReader) *JSONRPCStateServer {
 	return &JSONRPCStateServer{
 		stateReader: stateReader,
 	}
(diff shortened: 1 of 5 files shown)


aaronbuchwald (changes requested):
Thanks for putting this up 🙏 

We're working moving all of the APIs to be left to the VM to opt in/opt out, which may conflict a bit with @joshua-kim 's PR here: https://github.com/ava-labs/hypersdk/pull/1235

I think we can merge this first, but will leave to @joshua-kim

rpc/jsonrpc_state_server.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/ava-labs/avalanchego/trace"
+)
+
+type stateReader interface {
+	Tracer() trace.Tracer
```
joshua-kim, line 14: It seems weird to have the tracer on the interface for this. I think it would make more sense to just make this an implementation detail and pass the tracer through `NewJsonRPCStateServer` when it's constructed.

rpc/jsonrpc_state_server.go
```diff
@@ -0,0 +1,43 @@
...
+	ReadState(ctx context.Context, keys [][]byte) ([][]byte, []error)
+}
+
+type StateRequest struct {
```
joshua-kim, line 18: Make this request/response type consistent with the name of the api (so if we go with `GetStateKey`, this should be a `GetStateKeyRequest`).

rpc/jsonrpc_state_server.go
```diff
@@ -0,0 +1,43 @@
...
+package rpc
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/ava-labs/avalanchego/trace"
+)
+
+type stateReader interface {
```
joshua-kim, line 13: Can we export this interface + maybe we can call this `StateKeyReader`?

vm/dependencies.go
```diff
@@ -51,6 +51,7 @@ type Config struct {
 	ProcessingBuildSkip              int             `json:"processingBuildSkip"`
 	TargetGossipDuration             time.Duration   `json:"targetGossipDuration"`
 	BlockCompactionFrequency         int             `json:"blockCompactionFrequency"`
+	EnableStateReadRPCHandler        bool            `json:"enableStateReadRPCHandler"`
```
joshua-kim, line 54: nit: Make this config key name consistent w/ the API (i.e not `StateRead`)

rpc/jsonrpc_state_server.go
```diff
@@ -0,0 +1,43 @@
...
+
+func NewJSONRPCStateServer(stateReader stateReader) *JSONRPCStateServer {
+	return &JSONRPCStateServer{
+		stateReader: stateReader,
+	}
+}
+
+// JSONRPCStateServer gives direct read access to the vm state
+type JSONRPCStateServer struct {
+	stateReader
```
joshua-kim, line 35: nit: I prefer not embedding this interface here and just making this a normal field

rpc/jsonrpc_state_server.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/ava-labs/avalanchego/trace"
+)
+
+type stateReader interface {
+	Tracer() trace.Tracer
+	ReadState(ctx context.Context, keys [][]byte) ([][]byte, []error)
```
joshua-kim, line 15: nit: Can we call this `GetStateKey`?

rpc/jsonrpc_state_server.go
aaronbuchwald: This is the current style, so I think this is good for this PR and we can separately refactor to avoid this pattern if desired

rpc/jsonrpc_state_server.go
aaronbuchwald: This is the previously used name, so I think it's better to follow the current style for this PR

rpc/jsonrpc_state_server.go
```diff
@@ -0,0 +1,43 @@
...
+
+// JSONRPCStateServer gives direct read access to the vm state
+type JSONRPCStateServer struct {
+	stateReader StateReader
+}
+
+func (s *JSONRPCStateServer) ReadState(req *http.Request, args *ReadStateRequest, res *ReadStateResponse) error {
+	ctx, span := s.stateReader.Tracer().Start(req.Context(), "Server.ReadState")
+	defer span.End()
+	res.Values, res.Errors = s.stateReader.ReadState(ctx, args.Keys)
```
aaronbuchwald, line 41:
```suggestion
	defer span.End()
	
	res.Values, res.Errors = s.stateReader.ReadState(ctx, args.Keys)
```

minor nit

vm/dependencies.go
najeal: I renamed it as `EnableJSONRPCStateHandler`

It follows the mapping I found in the code base
"JSONRPCEndpoint" -> "jsonRPCHandler" -> "NewJSONRPCServer".
So it gives in the PR:
"JSONRPCStateEndpoint" -> "EnableJSONRPCStateHandler" -> "NewJSONRPCStateServer"

vm/vm.go
```diff
@@ -482,6 +482,18 @@ func (vm *VM) Initialize(
 	vm.webSocketServer = webSocketServer
 	vm.handlers[rpc.WebSocketEndpoint] = pubsubServer
 
+	// optional handler
```
aaronbuchwald, line 485:
```suggestion
	// Add optional direct state read handler
```

najeal:
@joshua-kim , I think I applied every suggestions except the ones @aaronbuchwald commented. And I have not renamed `StateReader` as `StateKeyReader`, to follow the current style as suggested. Let me know if you want to rename it anyway 👍

aaronbuchwald (dismissed):

aaronbuchwald:
> @joshua-kim , I think I applied every suggestions except the ones @aaronbuchwald commented. And I have not renamed `StateReader` as `StateKeyReader`, to follow the current style as suggested. Let me know if you want to rename it anyway 👍

Thanks this LGTM, this is the only API optionally added in the VM, so will see what @joshua-kim 's thoughts are there, but I think we can merge this and continue the API refactor we're already doing.

Thanks again!

joshua-kim (approved):

Issue #1217, opened 2024-07-30
Refactor genesis and upgrade bytes out of controller


joshua-kim:
Genesis and upgrade bytes can be removed from the controller code to be injected into the VM to slim down the controller struct.

PR #1271, opened 2024-08-09
Genesis and Rules Refactor Part 1

dboehm-avalabs:
Breaking up the Genesis and Rules, then moving them out of the controller as per #1217


examples/morpheusvm/cmd/morpheus-cli/morpheus-cli
aaronbuchwald, line 1: Should remove this file

examples/morpheusvm/controller/controller.go
```diff
@@ -92,6 +90,8 @@ func (*factory) New(
 			err,
 		)
 	}
+	c.genesis.ChainID = chainID
+	c.genesis.NetworkID = networkID
```
aaronbuchwald, line 94: I think the `chainID` and `networkID` should also be parameters to constructing the genesis instead of setting them here

examples/morpheusvm/controller/controller.go
```diff
@@ -145,7 +142,7 @@ type Controller struct {
 
 func (c *Controller) Rules(t int64) chain.Rules {
 	// TODO: extend with [UpgradeBytes]
-	return c.genesis.Rules(t, c.networkID, c.chainID)
+	return c.genesis.GetRulesFactory().GetRules(t)
```
aaronbuchwald, line 145: Can we remove this from the controller interface so that the controller doesn't need to know about the genesis at all?

Ideally, the VM knows about the genesis/upgrade rules factory etc, and we remove this completely from the controller

vm/rules.go
```diff
@@ -0,0 +1,179 @@
...
+		StorageValueReadUnits:     2,
+		StorageKeyAllocateUnits:   20,
+		StorageValueAllocateUnits: 5,
+		StorageKeyWriteUnits:      10,
+		StorageValueWriteUnits:    3,
+		SponsorStateKeysMaxChunks: []uint16{1},
+	}
+}
+
+func LoadBaseRules(b []byte, _ []byte /* upgradeBytes */) (*BaseRules, error) {
```
aaronbuchwald, line 91: Can we add `networkID` and `chainID` to the params here?

By default, these should be set to those values and only in extreme circumstances when a running network needs to hard fork and change their networkID / chainID would these ever change

examples/morpheusvm/genesis/genesis.go
```diff
@@ -9,90 +9,48 @@ import (
...
+	*vm.BaseRules
 
 	// Allocates
 	CustomAllocation []*CustomAllocation `json:"customAllocation"`
 }
 
+func (g *Genesis) GetRulesFactory() vm.RuleFactory[*vm.BaseRules] {
+	return &RuleFactory{g.BaseRules}
+}
+
```
aaronbuchwald, line 50: Can this be implemented in `vm.BaseRules` so that we don't need to here and it's up to the VM to wrap and override it if desired?

aaronbuchwald (changes requested):

examples/morpheusvm/cmd/morpheus-cli/cmd/key.go
```diff
@@ -186,9 +186,9 @@ var importKeyCmd = &cobra.Command{
 	},
 }
 
-func lookupSetKeyBalance(choice int, address string, uri string, networkID uint32, chainID ids.ID) error {
+func lookupSetKeyBalance(choice int, address string, uri string, _ uint32, _ ids.ID) error {
```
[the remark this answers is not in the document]
dboehm-avalabs: Ran into some interface definitions that couldn't be changed.  Same with the other cases.

examples/morpheusvm/cmd/morpheus-cli/cmd/key.go
```diff
@@ -219,8 +219,8 @@ var setKeyCmd = &cobra.Command{
 	},
 }
 
-func lookupKeyBalance(addr codec.Address, uri string, networkID uint32, chainID ids.ID, _ ids.ID) error {
-	_, err := handler.GetBalance(context.TODO(), brpc.NewJSONRPCClient(uri, networkID, chainID), addr)
+func lookupKeyBalance(addr codec.Address, uri string, _ uint32, _ ids.ID, _ ids.ID) error {
```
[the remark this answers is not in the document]
dboehm-avalabs: as as above, I'd have to make some changes to the base hypersdk which I was trying to avoid.

examples/morpheusvm/controller/controller.go
dboehm-avalabs: I was saving that for part 2 since this was going to be big.

aaronbuchwald (changes requested):

vm/genesis.go
```diff
@@ -0,0 +1,105 @@
...
+		StateBranchFactorGenesis: NewStateBranchFactorGenesis(),
+		CustomAllocation:         []*CustomAllocation{},
+		setBalance:               setBalance,
+	}
+}
+
+func (g *Bech32Genesis) InitializeState(ctx context.Context, tracer trace.Tracer, mu state.Mutable) error {
+	if err := g.StateBranchFactorGenesis.InitializeState(ctx, tracer, mu); err != nil {
+		return err
+	}
```
aaronbuchwald, line 78: Why do we need this?

vm/genesis.go
```diff
@@ -0,0 +1,105 @@
...
+
+	supply := uint64(0)
+	for _, alloc := range g.CustomAllocation {
+		addr, err := codec.ParseAnyHrpAddressBech32(alloc.Address)
+		if err != nil {
+			return fmt.Errorf("%w: %s", err, alloc.Address)
+		}
```
aaronbuchwald, line 85: If we need a function from the VM to set the balance anyways, could we use the same function to parse the address and avoid `ParseAnyHrp...`

vm/genesis.go
dboehm-avalabs: It is an embedded type and we also want to do any initialization that it would do.  In this case, there isn't any, but that might change in future and this is a pretty harmless way to avoid that bug in the future.

aaronbuchwald:
Closing in favor of #1468

PR #1289, opened 2024-08-12
Update Rust and Go Headers

samliok:


license-header
```diff
@@ -0,0 +1,2 @@
+Copyright (C) ${Year}, Ava Labs, Inc. All rights reserved.
+See the file LICENSE for licensing terms.
```
joshua-kim, line 2: nit: add new line at EOF (this can behave in annoying ways with some unix commands like diff for example) because the file spec technically requires a new line to denote EOF.

x/programs/runtime/import_log_debug.go
```diff
@@ -3,6 +3,11 @@
 
 //go:build debug
 
+/*
+ * Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+ * See the file LICENSE for licensing terms.
+ */
+
```
joshua-kim, line 10: Is this right here? Feels like we should be replacing the blob at the top of the file here

x/programs/runtime/import_log_release.go
```diff
@@ -3,6 +3,11 @@
 
 //go:build !debug
 
+/*
+ * Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+ * See the file LICENSE for licensing terms.
+ */
+
```
joshua-kim, line 10: Ref earlier comment

scripts/fix.lint.sh
```diff
@@ -21,8 +21,9 @@ go install -v github.com/palantir/go-license@latest
 
 # alert the user if they do not have $GOPATH properly configured
 check_command go-license
```
joshua-kim, line 23: Wondering if we even need go-license if we're using `hawkeye`

license-header
joshua-kim, line 1: Can we give this file an extension? Looks like a .txt file to me (maybe license.header or license_header.txt... kebab-case isn't normal for file names imo)

license-style
joshua-kim, line 1: I can't tell what type of file this is but similar comment as above.

joshua-kim (dismissed):

scripts/mock.gen.sh
```diff
@@ -43,10 +42,10 @@ do
...
   mockgen -package="${package_name}" -destination="${output_path}" "${src_import_path}" "${interface_name}"
-
-  go-license \
-  --config=./license.yml \
-  "${output_path}"
 done < "$input"
 
+# add license headers to mock files
+# ignore errors since mockgen files give warnings when edited
+hawkeye format || true
```
samliok, line 49: not a fan of this, but otherwise we need go-license and using two license linters seems worse

licenserc.toml
```diff
@@ -0,0 +1,24 @@
...
+useDefaultMapping = true
+
+additionalHeaders = ["license-style.txt"]
+
+[mapping.GO_STYLE]
+extensions = ["go"]
+
+# Properties to fulfill the template.
+[properties]
+Year = 2024
```
joshua-kim, line 24: What's the point of this `Year` variable? It's not like we ever use anything other than the current year so wondering if we can just hard-code it in as part of the license instead.

joshua-kim (approved):

PR #1294, opened 2024-08-14
Add devcontainer.json

containerman17:
A simple `devcontainer.json` to avoid "it works on my machine" issues.


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,7 @@
+ARG INSTALL_AVALANCHEGO_VERSION_SHORT
+FROM avaplatform/avalanchego:${INSTALL_AVALANCHEGO_VERSION_SHORT} AS avalanchego
+
+FROM mcr.microsoft.com/devcontainers/go:1
+
+ARG INSTALL_AVALANCHEGO_VERSION
+COPY --from=avalanchego /avalanchego/build/avalanchego /tmp/hypersdk/avalanchego-${INSTALL_AVALANCHEGO_VERSION}/avalanchego
```
richardpringle, line 7:
```suggestion
COPY --from=avalanchego /avalanchego/build/avalanchego /tmp/hypersdk/avalanchego-${INSTALL_AVALANCHEGO_VERSION}/avalanchego

```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,30 @@
...
+	// "features": {},
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "sudo chown -R $USER /go && sudo chown -R $USER /tmp/hypersdk/",
+	// Configure tool-specific properties.
+	// "customizations": {},
+	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+	// "remoteUser": "root"
+}
```
richardpringle, line 30:
```suggestion
}

```

.devcontainer/devcontainer.json
richardpringle, line 1: Can we remove the comments here? Comments aren't technically valid JSON. I would rather see a README  in the `.devcointainer` directory than a JSON file littered with comments.

richardpringle (changes requested):

.devcontainer/devcontainer.json
containerman17: Agree on the comments. Removed everything except for the commented-out Rust feature.

.devcontainer/devcontainer.json
containerman17: Please check out https://github.com/ava-labs/hypersdk/issues/1312

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,18 @@
+{
+	"name": "Go",
+	"build": {
+		"dockerfile": "Dockerfile",
+		"args": {
+			"INSTALL_AVALANCHEGO_VERSION": "d729e5c7ef9f008c3e89cd7131148ad3acda2e34",
+			"INSTALL_AVALANCHEGO_VERSION_SHORT": "d729e5c7"
```
aaronbuchwald, line 7: Could we switch from using a commit to a specific tag of AvalancheGo, so that it's more readable? We'll then need to keep this up to date when updating AvalancheGo as well

aaronbuchwald:
Who should be using this devcontainer? Supplying a precompiled AvalancheGo binary seems pretty minimal, so it's not clear to me who would need this or want to use it.

containerman17:
Actually yes, only the first build takes a few minutes. Probably not worth the hustle of updating the version. Removed

containerman17:
Even though this PR has been reduced from a full development environment to just a default Golang container, we still need this `devcontainer.json` to address situations like "it doesn't work on my machine."

aaronbuchwald (approved):

PR #1296, opened 2024-08-14
Simulator Cleanup

samliok:
Addressing a couple comments from the Simulator PR
- remove return values from insert + remove callbacks
- gets rid of `SimulatorState` struct in go since that was just a wrapper for `Mutable`
- some other small comments


x/programs/simulator/common/types.h
```diff
@@ -57,9 +57,13 @@ typedef char *(*InsertStateCallback)(void *data, Bytes key, Bytes value);
 typedef char *(*RemoveStateCallback)(void *data, Bytes key);
 
 typedef struct {
-    // this ptr gives context to the callbacks
+    // this is a pointer to the state passed in from rust
+	// it points to the state object on the rust side of the simulator
```
samliok, line 61: this is properly formatted on my IDE and when I download the file as well

richardpringle: It's a hard tab instead of spaces

richardpringle, line 61:
```suggestion
    // this is a pointer to the state passed in from rust
    // it points to the state object on the rust side of the simulator
```

x/programs/simulator/common/types.h
```diff
@@ -57,9 +57,13 @@ typedef char *(*InsertStateCallback)(void *data, Bytes key, Bytes value);
...
     void *stateObj;
+	// this is ptr to the get function
```
richardpringle, line 63:
```suggestion
    // this is ptr to the get function
```

x/programs/simulator/common/types.h
```diff
@@ -57,9 +57,13 @@ typedef char *(*InsertStateCallback)(void *data, Bytes key, Bytes value);
...
     GetStateCallback get_value_callback;
+    // this is ptr to the insert function
     InsertStateCallback insert_callback;
+	// this is a ptr to the delete function
```
richardpringle, line 67:
```suggestion
    // this is a ptr to the delete function
```

x/programs/simulator/src/lib.rs
```diff
@@ -16,6 +16,9 @@ mod bindings {
         type Target = [u8];
 
         fn deref(&self) -> &Self::Target {
+            // # Safety:
+            // These bytes were mallocated by CGo
```
richardpringle, line 20:
```suggestion
            // These bytes were allocated by CGo
```
"allocated" is the more generic term

x/programs/simulator/src/lib.rs
```diff
@@ -16,6 +16,9 @@ mod bindings {
...
+            // They are guaranteed to be valid since for the length of the slice
```
richardpringle, line 21:
```suggestion
            // They are guaranteed to be valid for the length of the slice
```

x/programs/simulator/state/manager.go
```diff
@@ -18,9 +18,14 @@ import (
...
 )
 
-// ensure SimulatorStateManager implements StateManager
 var _ runtime.StateManager = &ProgramStateManager{}
 
+var (
+	ErrUnknownAccount = errors.New("unknown account")
+	balanceKeyBytes   = []byte("balance")
+	programKeyBytes   = []byte("program")
+)
```
richardpringle, line 27: 🥇

richardpringle (changes requested):

richardpringle (approved):
🚀

PR #1297, opened 2024-08-14
Replace hawkeye with golang-based license linter

containerman17:
`google/addlicense` works the same way and doesn't require Cargo to be installed. The only downside is losing licenses on the autogenerated mock files `vm/mock_controller.go` and `chain/mock_rules.go`.


joshua-kim (approved):
I want to get rid of our dependency on mockgen anyways so I'm fine w/ these changes.

joshua-kim (approved):

PR #1304, opened 2024-08-15
Update bytemuck so we can derive NoUninit

richardpringle:
`Pod` and `Zeroable` are more strict than `NoUninit` and we don't actually require that level of strictness but we couldn't specify the crate path before a recent release.


richardpringle:
This passes on Mac but not linux for some reason. I dove into it and the changes in my PR still haven't been published. No idea what's going on but we'll go back to the old way for now, I guess....

richardpringle:
Asked the author to publish my changes and they did... but it appears that I can't reopen this because I force-pushed the branch

PR #1305, opened 2024-08-15
Add Balance Interface to Simulator

samliok:
allows the simulator to set/get the balances of accounts.

decided to expose FFI functions instead of directly setting/getting in rust as to not duplicate the state logic.

### DIFF
--- a/x/programs/simulator/ffi/ffi.go
+++ b/x/programs/simulator/ffi/ffi.go
@@ -165,7 +165,7 @@ func GetBalance(db *C.Mutable, address *C.Address) C.uint64_t {
 
 	state := simState.NewSimulatorState(unsafe.Pointer(db))
 	pState := simState.NewProgramStateManager(state)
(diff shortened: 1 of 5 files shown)


x/programs/simulator/ffi/ffi.go
```diff
@@ -154,4 +154,40 @@ func newCallProgramResponse(result []byte, fuel uint64, err error) C.CallProgram
...
+		panic(ErrInvalidParam)
+	}
+
+	state := simState.NewSimulatorState(unsafe.Pointer(db))
+	pState := simState.NewProgramStateManager(state)
+	account := C.GoBytes(unsafe.Pointer(address), codec.AddressLen)     //nolint:all
+
+	balance, err := pState.GetBalance(SimContext, codec.Address(account))
+	if err != nil {
+		panic(err)
```
samliok, line 172: i think panicing is fine here, saves extra overhead by not passing error messages go <-> rust. plus if it errors, then something strange is wrong anyways

PR #1318, opened 2024-08-16
Fix local usage of shellcheck

RodrigoVillar:
This PR fixes a previous bug in `scripts/tests.shellcheck.sh` where running the script locally would cause it to fail. In particular, this PR removes the expansion brackets which wrapped the `$@` parameter that caused the bug.


aaronbuchwald (approved):
LGTM once this passes CI

PR #1319, opened 2024-08-16
Minor integration test coverage cleanup

aaronbuchwald:
Address @marun 's comments from https://github.com/ava-labs/hypersdk/pull/1303 with the exception of
- https://github.com/ava-labs/hypersdk/pull/1303#discussion_r1720348093
- https://github.com/ava-labs/hypersdk/pull/1303#discussion_r1720225247 (will do a separate PR for this)

### DIFF
--- a/tests/integration/integration.go
+++ b/tests/integration/integration.go
@@ -136,6 +136,9 @@ func setInstances() {
 		require.NoError(err)
 		dname, err := os.MkdirTemp("", nodeID.String()+"-chainData")
 		require.NoError(err)
+		ginkgo.DeferCleanup(func() {
+			os.RemoveAll(dname)
+		})
 		snowCtx := &snow.Context{
 			NetworkID:      networkID,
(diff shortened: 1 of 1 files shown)


tests/integration/integration.go
```diff
@@ -259,12 +264,12 @@ var _ = ginkgo.Describe("[Tx Processing]", ginkgo.Serial, func() {
...
 		})
 
-		ginkgo.By("skip duplicate", func() {
+		ginkgo.By("skip mempool duplicate", func() {
 			_, err := instances[0].cli.SubmitTx(
 				context.Background(),
 				initialTx.Bytes(),
 			)
-			require.Error(err) //nolint:forbidigo
+			require.ErrorContains(err, vm.ErrNotAdded.Error()) //nolint:forbidigo
```
aaronbuchwald, line 272: We're getting the error from the rpc client, so we don't have the exact error type here, but we can still check that it includes the expected error string.

PR #1322, opened 2024-08-18
update txAssertion to use require instead of returning error

aaronbuchwald:
Updates `txAssertion` to use `require.Assertion` instead of returning an error to the caller. This ensures that when an assertion fails, the test output clearly outputs where it came from.


joshua-kim (approved):

aaronbuchwald:
`IgnoreFailureTestingT` is required to support an e2e test that allows tx issuance to error, so adding a GitHub issue to fix the root cause and update the test to remove this pattern.

PR #1330, opened 2024-08-19
MorpheusCLI - Add Deploy Command

RodrigoVillar:
This command extends the MorpheusCLI by adding a `deploy` command. This command, when called, spins up an instance of MorpheusVM using `tmpnet`. This command differs from the existing `run.sh`/`stop.sh` workflow by allowing the user to stop the instance via `Ctrl-C`.

### DIFF
--- a/examples/morpheusvm/cmd/morpheus-cli/cmd/deploy.go
+++ b/examples/morpheusvm/cmd/morpheus-cli/cmd/deploy.go
@@ -1,0 +1,100 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/ava-labs/avalanchego/tests/fixture/tmpnet"
+	"github.com/spf13/cobra"
+
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/consts"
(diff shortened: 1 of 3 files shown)


examples/morpheusvm/cmd/morpheus-cli/cmd/deploy.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/ava-labs/hypersdk/tests/fixture"
+	"github.com/ava-labs/hypersdk/utils"
+)
+
+const (
+	owner      = "morpheus-cli"
+	devNetPort = "9560"
+)
+
+var deployCmd = &cobra.Command{
```
aaronbuchwald, line 30: Can we instead add this via the root handler, so that it can be shared across VMs https://github.com/ava-labs/hypersdk/blob/87957cdab7a2baa5fcba116a8552c8e30ee56ad4/examples/morpheusvm/cmd/morpheus-cli/cmd/root.go#L65 ?

examples/morpheusvm/cmd/morpheus-cli/cmd/deploy.go
```diff
@@ -0,0 +1,100 @@
...
+			ctx,
+			os.Stdout,
+			network,
+			"",
+			avalancheGoPath,
+			avalancheGoPluginDir,
+		); err != nil {
+			utils.Outf(err.Error())
+			return err
+		}
```
aaronbuchwald, line 68: If we need to port this code from the e2e tests, can we wrap the construction into a single function? Similar comment here https://github.com/ava-labs/hypersdk/pull/1320#discussion_r1721813693

PR #1333, opened 2024-08-20
Cleanup find-out branch

samliok:


.devcontainer/devcontainer.json
samliok, line 1: Should we keep this somewhere? Seems good to have if ever spin up another testing enviroment

samliok, line 1: Should we keep this somewhere? Seems good to have if we ever spin up another testing environment

richardpringle:
I think we're going to keep the tutorial around as is. Going to close this for now

Issue #1184, opened 2024-07-23
Pass all gas parameter in `ExternalCallContext`


iFrostizz:
It is currently a bit unpractical to supply a constant `max_units` because some programs called externally may use more than others, and if we offload this to the parameters of the function then it adds more complexity. It would be nice to be able to pass an option to the current `max_units` field of the `ExternalCallContext` and in the case of this value being `None`, any unit that has not been spent would be refunded to the current context.

richardpringle:
We can probably use an `Option<NonZeroU64>` and just treat it as a u64 on the go side. If the value is `0`, that means to take what's needed.

richardpringle:
Instead of passing in an `Option<NonZeroU64>` directly, I think it would be better to use a builder pattern for both `Gas` and `Value`. 
```rust
ctx
    .call_contract_builder()
    .max_fuel(42u64)
    .transfer_value(1_000_000u64)
    .call_function("<function_name>", args)?;
```    
And we could keep 
```rust
ctx.call_function("<function_name>", args)?;
```
as the default usage when you don't need to provide a limited amount of gas or value

iFrostizz:
When using constants, it's clear that passing 0 to forward all gas makes sense because the current behaviour would be not being able to execute anything but there could be footguns if the gas is dynamic and it would be harder to reason about when all the gas is forwarded vs what is passed. Currently done with an enum https://github.com/ava-labs/hypersdk/pull/1417 which avoids this pitfall but the builder pattern looks better.

PR #1417, opened 2024-08-20
`PassAll` gas option for program to program calls

iFrostizz:
Closes https://github.com/ava-labs/hypersdk/issues/1184

Create a new `Gas` type that replaces the current alias by an enum that can either be `PassAll` or `Units(NonZeroU64)`. The new `PassAll` works around the current way of spending gas, namely calling `ConsumeFuel` with the passed `Fuel` parameter and refunding any unused gas by calling `AddFuel` after the execution.


samliok (changes requested):
I think this PR's functionality is important but I don't think a `Gas` enum is the best way to do it. It seems much more intuitive to have gas be a field of `context`. This way we can keep the u64 type(which is nice for dev experience and contract readability) and also have access to the actual amount of gas via the `context`.

iFrostizz:
Good point @samliok ! Thanks

iFrostizz:
Actually I'm not sure if that's totally correct. We already have the `remaining_fuel` method in the implementation of `Program` but it's not gonna work because if the full amount of gas is passed then it will be fetched and the `remainingFuelCost` is consumed before spending the call cost and consuming it in the external program call, so it always end up in an `OutOfFuel` error. That's an interesting point though and it would also be a great solution to just pass `remaining_fuel()`

x/contracts/wasmlanche/src/context.rs
```diff
@@ -334,6 +338,48 @@ impl Context {
...
+    fn with_address(ctx: &'a mut Context, address: Address) -> Self {
+        CallContractBuilder {
+            ctx,
+            address,
+            max_units: Gas::PassAll,
+            value: None,
+        }
+    }
+
+    pub fn with_max_units(mut self, units: u64) -> Self {
```
iFrostizz, line 358: It would be great if this function could only accept units that are > 0, there might be some semantic changes to do

iFrostizz: Otherwise, replicate the `Gas` enum type on the go side too so that when it's the `Unit` variant then it's only interpreted as "pass x units" even when it's 0 to avoid any runtime panic because of the conversion from a `u64` to a `NonZeroU64`.

x/contracts/runtime/contract.go
```diff
@@ -45,6 +45,9 @@ type CallInfo struct {
 	// the serialized parameters that will be passed to the called function
 	Params []byte
 
+	// true if the call passes all gas
+	PassAll bool
+
 	// the maximum amount of fuel allowed to be consumed by wasm for this call
 	Fuel uint64
```
iFrostizz, line 52: Would be interested to know if there could be a cleaner solution. Currently there is a bool and the uint64 that tells the amount of fuel passed, but since having `PassAll = true` and `Fuel > 0` is inconsistent state, this has to be validated through the execution. Is there a better way to simulate this rust enum:
```rust
enum Gas {
    PassAll,
    Units(u64)
}
```

iFrostizz:
Passing the units of gas to program calls to the simulator is still required. We might want to set the initial units of gas in the simulator initialisation but that sounds a little out of scope for this PR.

PR #1331, opened 2024-08-19
Add examples from FAFO

richardpringle:
- **Some docs for programs development (#1302)**
- **Add Balance Interface to Simulator (#1305)**
- **update amm test (#1310)**
- **Dev-container for find-out Friday**
- **Docs as github-pages**
- **Remove out-of-date READMEs**
- **Add minimal wasmlanche-sdk README**
- **Major wasmlanche-docs update**
- **Update `public` macro to not mention implementation details**
- **simulator docs (#1311)**
- **Don't ignore .so files, that's silly**
- **Fix Balance Method Ptrs (#1327)**

PR #1419, opened 2024-08-20
Add tests to AMM Smart Contract

samliok:
Added unit tests using the new simulator.

Found and fixed a bug caught by the unit tests

waiting for #1331


richardpringle (changes requested):
Remove println in go

richardpringle (approved):

PR #1423, opened 2024-08-21
Required Runtime Changes to enable a VM that uses the wasmlanche runtime

dboehm-avalabs:
1. program ID:  the ID needs to become a []byte to support other id's besides just the 32 byte hash.  Even if we want to only support fixed bytes, it would need to be 36 bytes to support the size suffix that hypersdk enforces for state keys.
2. stateaccess: like the program ID's above, the state keys need to have a size associated with them.  Temporarily just add a large size to all state access.


x/programs/rust/wasmlanche-sdk/src/state.rs
```diff
@@ -65,7 +71,7 @@ unsafe impl<K: NoUninit> NoUninit for PrefixedKey<K> {}
 
 const _: fn() = || {
     #[doc(hidden)]
-    struct TypeWithoutPadding([u8; 1 + ::core::mem::size_of::<u32>()]);
+    struct TypeWithoutPadding([u8; 3 + ::core::mem::size_of::<u32>()]);
```
richardpringle, line 74: I should have done this before, but now that we have a random `3` here, I think we should do the following
```rust
type PrefixType = u8;
type MaxChunksType = [u8; size_of::<u16>()];

pub struct PrefixedKey<K: NoUninit> {
    prefix: PrefixType,
    key: K,
    max_chunks: MaxChunksType,
};

const _: fn() = || {
    type TypeForTypeTest = u32;
    const SIZE: usize = ::core::mem::size_of::<PrefixType>() + ::core::mem::size_of::<MaxChunksType>() + ::core::mem::size_of::<TypeForTypeTest>();
    #[doc(hidden]
    struct TypeWithoutPadding([u8; SIZE]);
    let _ = ::core::mem::transmute::<crate::state::PrefixedKey<TypeForTypeTest>, TypeWithoutPadding>;
};

File: x/programs/rust/wasmlanche-sdk/src/state.rs [hunk 2]
```diff
@@ -12,6 +12,9 @@ use bytemuck::NoUninit;
 use sdk_macros::impl_to_pairs;
 use std::{cell::RefCell, collections::HashMap};
 
+// maximum number of chunks that can be stored at the key as big endian u16
+pub const STATE_MAX_CHUNKS: [u8; 2] = [0, 4];
```
richardpringle, line 16:
```suggestion
pub const STATE_MAX_CHUNKS: [u8; 2] = 4u16.to_be_bytes();
```

richardpringle (changes requested):

x/programs/rust/examples/automated-market-maker/src/lib.rs
```diff
@@ -22,8 +22,8 @@ const MAX_GAS: Gas = 10000000;
 
 /// Initializes the pool with the two tokens and the liquidity token
 #[public]
-pub fn init(context: &mut Context, token_x: Program, token_y: Program, liquidity_token: Id) {
-    let lt_program = context.deploy(liquidity_token, &[0, 1]);
+pub fn init(context: &mut Context, token_x: Program, token_y: Program, liquidity_token: Vec<u8>) {
```
samliok, line 25: We should definitely still have a rust type that represents an Id. Could even be a type alias for now

`pub type Id = Vec<u8>`

x/programs/rust/sdk-macros/src/lib.rs
```diff
@@ -357,7 +357,7 @@ pub fn state_schema(input: TokenStream) -> TokenStream {
 
                     const _: fn() = || {
                         #[doc(hidden)]
-                        struct TypeWithoutPadding([u8; 1 + ::core::mem::size_of::<#key_type_name>()]);
+                        struct TypeWithoutPadding([u8; 3 + ::core::mem::size_of::<#key_type_name>()]);
```
samliok, line 360: can we not hardcode this here and potentially use the `STATE_MAX_CHUNKS` const defined below?

samliok (changes requested):

x/programs/rust/sdk-macros/src/lib.rs
richardpringle: I wonder if we can actually re-use some code here. Let me try something...

richardpringle (approved):

x/programs/rust/examples/automated-market-maker/src/lib.rs
richardpringle: done

x/programs/rust/sdk-macros/src/lib.rs
richardpringle: Fixed

PR #1425, opened 2024-08-21
Rebuild go simulator if package content changes

samliok:


meaghanfitzgerald:
is `rerun-if-changed` a `cargo` command or is it some custom script you wrote to check file contents and execute `cargo build` is the contents have changed

iFrostizz:
@meaghanfitzgerald It's an instruction that is understood by cargo to conditionally trigger the build script, here if any file changed under `state/` or `ffi/` which will build the simulator again.

Would it be worth also including the `common` directory so that the bindings.rs file would be regenerated with bindgen if any of the type definition changes ?

richardpringle (changes requested):
Add common and I'll approve

richardpringle (approved):

PR #1430, opened 2024-08-21
Add memory testing

richardpringle:
Improve memory unit tests and add integration tests

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/memory.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/memory.rs
@@ -38,9 +38,7 @@ impl Drop for HostPtr {
             return;
         }
 
-        let len = ALLOCATIONS
-            .with_borrow_mut(|allocations| allocations.remove(&self.0))
-            .expect("attempted to drop invalid host pointer");
+        let len = remove(self.0);
         let layout = Layout::array::<u8>(len).expect("capacity overflow");
 
         unsafe { std::alloc::dealloc(self.0.cast_mut(), layout) };
@@ -52,9 +50,7 @@ impl From<HostPtr> for Vec<u8> {
         // drop will dealloc the bytes
         let host_ptr = ManuallyDrop::new(host_ptr);
 
-        let len = ALLOCATIONS
-            .with_borrow_mut(|allocations| allocations.remove(&host_ptr.0))
-            .expect("attempted to drop invalid host pointer");
+        let len = remove(host_ptr.0);
 
         unsafe { Vec::from_raw_parts(host_ptr.0.cast_mut(), len, len) }
     }
@@ -72,7 +68,7 @@ impl HostPtr {
 /// # Panics
 /// Panics if the pointer exceeds the maximum size of an isize or that the allocated memory is null.
 #[no_mangle]
-pub extern "C" fn alloc(len: usize) -> HostPtr {
+extern "C" fn alloc(len: usize) -> HostPtr {
     assert!(len > 0, "cannot allocate 0 sized data");
     // can only fail if `len > isize::MAX` for u8
     let layout = Layout::array::<u8>(len).expect("capacity overflow");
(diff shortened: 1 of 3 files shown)


x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -52,9 +50,7 @@ impl From<HostPtr> for Vec<u8> {
         // drop will dealloc the bytes
         let host_ptr = ManuallyDrop::new(host_ptr);
 
-        let len = ALLOCATIONS
-            .with_borrow_mut(|allocations| allocations.remove(&host_ptr.0))
-            .expect("attempted to drop invalid host pointer");
+        let len = remove(host_ptr.0);
 
         unsafe { Vec::from_raw_parts(host_ptr.0.cast_mut(), len, len) }
```
samliok, line 55: just checking my understanding -  `Vec::from_raw_parts` will take ownership of the `host_ptr`, meaning it will  automatically drop the Vec once its not longer in use. Can we do this without the first line
`let host_ptr = ManuallyDrop::new(host_ptr);`?

x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -88,22 +84,100 @@ pub extern "C" fn alloc(len: usize) -> HostPtr {
...
     use super::*;
 
     #[test]
-    fn data_allocation() {
-        let len = 1024;
-        let ptr = alloc(len);
-        let vec: Vec<_> = (u8::MIN..=u8::MAX).cycle().take(len).collect();
-        unsafe { std::ptr::copy(vec.as_ptr(), ptr.0.cast_mut(), vec.len()) }
-        assert_eq!(*ptr, vec);
+    #[should_panic(expected = "attempted to deref invalid host pointer")]
```
samliok, line 99: is it standard to hardcode these string error messages?

x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -88,22 +84,100 @@ pub extern "C" fn alloc(len: usize) -> HostPtr {
...
+        drop(HostPtr(ptr));
+
+        ALLOCATIONS.with_borrow(|allocations| {
+            assert!(allocations.get(&ptr).is_none());
+        });
 
-        ALLOCATIONS.with_borrow(|map| assert!(!map.contains_key(&inner)));
+        // overwrites old allocation
+        let data = vec![0x00];
+        assert_eq!(data.as_ptr(), ptr);
```
samliok, line 180: why would this always write into the same location?

samliok (approved):
Couple comments to check my understanding

x/programs/rust/wasmlanche-sdk/src/memory.rs
richardpringle: So, when a `HostPtr` is dropped, it will remove the allocation data from the `ALLOCATIONS` map and attempt to dealloc the pointer by the length from the map. When we convert a `HostPtr` to a `Vec`, we don't want to perform the drop logic, we just want to drop the internal pointer (the stack value) and use the length from the `ALLOCATIONS` map. We use that data to construct the `Vec`. Now it's up to the `Vec` to dealloc on drop or move the data when it needs to grow, etc, etc. It's effectively like saying we're giving control over the bytes from `HostPtr` to `Vec`. 

That make sense?

x/programs/rust/wasmlanche-sdk/src/memory.rs
richardpringle: I tried to put them in a constant and the attribute couldn't read the constant. So I'm not sure if there's a better way to do this. The only reason I checked the message in the first place was because Clippy told me to.

x/programs/rust/wasmlanche-sdk/src/memory.rs
richardpringle: I think there's some level of determinism in the standard allocators. Of course, this definitely isn't guaranteed, but I think it's reasonable to assume that if you're only allocating and deallocating a single byte, an efficient allocator will re-allocate in the same spot.

PR #1435, opened 2024-08-22
Remove RefCell wrapping the state cache

richardpringle:
When writing the state-schema code, I wanted to make the changes as minimal as possible (though the PR was huge). Now that we use `Context` the way that we do (`&mut Context`), we no longer need to wrap the cache in a `RefCell`. Now we can actually construct the `State` struct that wraps the `HashMap` with extra logic and make that a part of the `Context` properties


x/programs/simulator/src/bindings.rs
samliok, line 1: I like this file being separated 🤩

samliok (approved):

PR #1436, opened 2024-08-22
Cleanup MorpheusVM CI workflow dependencies

richardpringle:
E2E tests are heavy and need to be rerun if either the unit tests or linter tests fail. We should wait for these other workflows to pass *before* triggering the e2e tests.


richardpringle:
Checkout the dependency graph here:
https://github.com/ava-labs/hypersdk/actions/runs/10513488864

aaronbuchwald (approved):

PR #1445, opened 2024-08-23
Scripts cleanup fixes

iFrostizz:


scripts/lint.sh
```diff
@@ -25,7 +26,12 @@ else
   TARGET="${1}"
 fi
 
-add_license_headers -check
+ADD_LICENSE=${ADD_LICENSE:-true}
+if [ "$ADD_LICENSE" ]; then
+  add_license_headers -check
+fi
```
iFrostizz, line 32: The `find` is probably wrong when running from anything outside the repo root. We also don't really need to add licenses since it should have already been done by the script at `<root>/scripts/lint.sh`

PR #1447, opened 2024-08-23
state api test

### LINKED ISSUE #1292: Add e2e tests for state api
Add tests for the current state api: https://github.com/ava-labs/hypersdk/pull/1264

najeal:
close #1292 

This PR modifies state api and creates a factory. So it can be added to the  `vm_api_handler_factories`.
The api has been added to the morpheus example and an e2e test has been created.

I had to change the response field `Errors` from []error type to []string type for the json marshaling, then recreate the []errors client side but it may stay a []string.

I see several issues:
1. Currently e2e tests run only the `Ping` test. in `examples/morpheusvm/scripts/run.sh`
```
additional_args+=("--ginkgo.focus=Ping")
```
I tested locally changing the `focus` but we may want to delete this focus to run everything?

2. It should be possible to easily disable or enable the different APIs tests depending on which one you use in your vm.
For example if I choose not to use the state api in morpheus, I want to skip the `ReadState` test


aaronbuchwald:
The run script is used to launch a local network with tmpnet, to run the full e2e tests use `MODE=test ./scripts/run.sh` as in https://github.com/ava-labs/hypersdk/blob/857d2dceef6971721617a0a192d90e942414d15d/.github/workflows/hypersdk-ci.yml#L165.

Added back the ability to easily enable/disable APIs here: https://github.com/ava-labs/hypersdk/pull/1450. Still iterating on this to make it as simple as possible to add services like a current state API, so added a small cleanup PR here: https://github.com/ava-labs/hypersdk/pull/1455 that lists out the further changes I have in mind.

aaronbuchwald:
Changing the errors to a string makes sense to me 👍

tests/workload/apis.go
```diff
@@ -30,3 +31,15 @@ func GetNetwork(ctx context.Context, require *require.Assertions, uris []string,
...
+
+func ReadState(ctx context.Context, require *require.Assertions, uris []string) {
+	for _, uri := range uris {
+		client := state.NewJSONRPCStateClient(uri)
+		values, readerrs, err := client.ReadState(ctx, [][]byte{
+			[]byte(`my-unknown-key`),
+		})
+		require.NoError(err)
+		require.Len(values, 1)
+		require.Len(readerrs, 1)
```
aaronbuchwald, line 43: Hmm it's a bit of a strange problem here that we don't know any other key-value pairs that could be included in the state, so this is all we can test.

Could we move this directly into `e2e.go` ? The functions here are intended for re-use across environments and I don't think that we'll re-use this test in any other environment.

aaronbuchwald (changes requested):
One requested change and then this LGTM - thanks for following up on this change 🙏

aaronbuchwald:
Imo the e2e tests in `e2e.go` should be the APIs shared across all VMs ie the ones that only use a standard set of APIs any VM can use. VMs define the entrypoint to the e2e tests, so if they need to add their own APIs, they can update the config used by `tmpnet` to start up the network to make sure they're enabled.

The annoying thing about ginkgo is that you declare your specs in the format:

```golang
var _ = ginkgo.It(...)
```

so that ginkgo can map out the hierarchy of the spec in advance. If you try to declare an `It` node in any conditional block, it fails. This limits us a bit and forces us to either use ginkgo's labels to filter for the relevant tests to run or break up the imported test coverage into separate components and leave it to the VM to choose which test coverage to import.

Here's an example of what it looks like to import test coverage from AvalancheGo: https://github.com/ava-labs/avalanchego/blob/0117ab96791c9e3b2a35c51b4315e3928dffb983/tests/e2e/e2e_test.go#L14.

If we support a default set of APIs for VMs to choose from, this is pretty easy, we can set them all to be enabled by default or use a default config that enables all of the APIs the e2e tests depend on.

However, if VM developers start to mix and match which of the default APIs they want to enable, then we'll probably need to break this up, so that VM developers can import only the test coverage relevant to them.

api/state/option.go
```diff
@@ -0,0 +1,15 @@
...
+
+package state
+
+import "github.com/ava-labs/hypersdk/vm"
+
+const Namespace = "corestate"
+
+func With() vm.Option {
+	return vm.NewOption(Namespace, func(v *vm.VM, _ []byte) error {
+		vm.WithVMAPIs(JSONRPCStateServerFactory{})(v)
```
aaronbuchwald, line 12: we should add a config that is enabled by default here and json unmarshals the supplied bytes to set the config

najeal: I agree with your comment above https://github.com/ava-labs/hypersdk/pull/1447#issuecomment-2314126322 .

But I am not sure to understand this one. And not sure to understand the purpose of the config you talk about.
Regarding the code, enabling or disabling the apis is done in the controller `New()` factory (A file that will be written by the user). 
If the user does not call `state.With()` the api is disabled.

aaronbuchwald: Ah I see, sorry for the confusion. We're going in the direction of having the `vm/` package provide a default set of APIs. Right now, there's the controller file, which will construct a version with and without options, but ideally there's a simple `vm.DefaultNew` that will construct all of the APIs.

Could you enable the current state API on MorpheusVM and we'll add this to the default set, so that it works with the e2e tests?

To support VMs that do not use the same set of defaults, we'll need to break the test coverage down so that VMs can select the relevant set of tests.

najeal: No problem !
If I am not wrong staye API is already enabled in the PR 👍

aaronbuchwald (dismissed):
LGTM after CI passes

najeal:
Lint was failing, it is now fixed.

aaronbuchwald (approved):

PR #1450, opened 2024-08-23
Add VM Namespaced Options

RodrigoVillar:
This PR extends `Options` by refactoring them as `NamespacedOptions`


api/ws/server.go
```diff
@@ -33,16 +34,25 @@ var (
 	ErrExpired = errors.New("expired")
 )
 
-func WithWebsocketAPI(maxPendingMessages int) vm.Option {
-	return func(v *vm.VM) error {
+type WSConfig struct {
```
joshua-kim, line 37: nit: name this `Config`. In Go the package is the level of abstraction so it seems weird to have a `ws.WsConfig`.

api/ws/server.go
```diff
@@ -33,16 +34,25 @@ var (
...
+	MaxPendingMessages int `json:"maxPendingMessages"`
+}
+
+func WithWebsocketAPI() vm.Option {
+	return func(v *vm.VM, r json.RawMessage) error {
 		actionRegistry, authRegistry := v.Registry()
+		var wsConfig WSConfig
```
joshua-kim, line 44: Should we have defaults set?

examples/morpheusvm/controller/config.go
```diff
@@ -5,10 +5,13 @@ package controller
...
 import (
 	"encoding/json"
+
+	"github.com/ava-labs/hypersdk/api/ws"
 )
 
 type Config struct {
-	StoreTransactions bool `json:"storeTransactions"`
+	StoreTransactions bool        `json:"storeTransactions"`
+	WS                ws.WSConfig `json:"ws"`
```
joshua-kim, line 14: Here it would probably make sense to call this `websocketConfig` or something

vm/vm.go
```diff
@@ -269,8 +269,20 @@ func (vm *VM) Initialize(
...
+		return fmt.Errorf("failed to unmarshal namespaced config: %w", err)
+	}
+
+	allocatedNamespaces := make(map[string]struct{})
+	for _, namespacedOption := range vm.namespacedOptions {
+		_, ok := allocatedNamespaces[namespacedOption.Namespace]
+		if ok {
+			return fmt.Errorf("namespace %s already allocated", namespacedOption.Namespace)
+		}
+		allocatedNamespaces[namespacedOption.Namespace] = struct{}{}
```
aaronbuchwald, line 283: Can we move the check for option namespace collisions to the constructor?

vm/vm.go
```diff
@@ -269,8 +269,20 @@ func (vm *VM) Initialize(
...
+	namespacedConfig := make(map[string]json.RawMessage)
+	if err := json.Unmarshal(controllerConfigBytes, &namespacedConfig); err != nil {
+		return fmt.Errorf("failed to unmarshal namespaced config: %w", err)
+	}
+
+	allocatedNamespaces := make(map[string]struct{})
+	for _, namespacedOption := range vm.namespacedOptions {
+		_, ok := allocatedNamespaces[namespacedOption.Namespace]
+		if ok {
+			return fmt.Errorf("namespace %s already allocated", namespacedOption.Namespace)
```
joshua-kim, line 281: Should we do this check in `New` instead?

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -38,10 +40,21 @@ var _ = ginkgo.BeforeSuite(func() {
...
+	vmConfig := vm.NewConfig()
+	vmConfig.Config = make(map[string]any)
+	vmConfig.Config["storeTransactions"] = true
+	vmConfig.Config["ws"] = ws.WSConfig{
+		MaxPendingMessages: 10_000,
+	}
+
+	configBytes, err := json.Marshal(vmConfig)
+	require.NoError(err)
+
```
aaronbuchwald, line 52: This should be in `integration.go` instead of coming from MorpheusVM

vm/vm.go
```diff
@@ -269,8 +269,20 @@ func (vm *VM) Initialize(
...
 	vm.gossiper = txGossiper
 
-	for _, option := range vm.options {
-		if err := option(vm); err != nil {
+	namespacedConfig := make(map[string]json.RawMessage)
+	if err := json.Unmarshal(controllerConfigBytes, &namespacedConfig); err != nil {
+		return fmt.Errorf("failed to unmarshal namespaced config: %w", err)
+	}
+
+	allocatedNamespaces := make(map[string]struct{})
```
joshua-kim, line 277: nit: this is the idiomatic way to do this in go... but we generally use our implementation of `set.Set` for this in our codebase

api/ws/server.go
```diff
@@ -33,16 +34,25 @@ var (
...
 )
 
-func WithWebsocketAPI(maxPendingMessages int) vm.Option {
-	return func(v *vm.VM) error {
+type WSConfig struct {
+	MaxPendingMessages int `json:"maxPendingMessages"`
+}
+
+func WithWebsocketAPI() vm.Option {
+	return func(v *vm.VM, r json.RawMessage) error {
```
joshua-kim, line 42: Might want to name `r` `configBytes` or something

examples/morpheusvm/controller/controller.go
```diff
@@ -28,28 +28,28 @@ var (
 )
 
 // New returns a VM with the indexer, websocket, and rpc apis enabled.
-func New(options ...vm.Option) *vm.VM {
-	opts := []vm.Option{
-		indexer.WithIndexer(consts.Name, indexer.Endpoint),
-		ws.WithWebsocketAPI(10_000_000),
-		vm.WithVMAPIs(jsonrpc.JSONRPCServerFactory{}),
-		vm.WithControllerAPIs(&jsonRPCServerFactory{}),
+func New(namespacedOptions ...vm.NamespacedOption) *vm.VM {
```
joshua-kim, line 31: Wondering if we should support options that don't require a json namespace

api/indexer/tx_indexer.go
```diff
@@ -28,7 +29,7 @@ var (
 )
 
 func WithIndexer(name string, path string) vm.Option {
-	return func(v *vm.VM) error {
+	return func(v *vm.VM, r json.RawMessage) error {
```
aaronbuchwald, line 32: Let's change the function signature to use `[]byte` - `json.RawMessage` is just an alias for this anyways

examples/morpheusvm/tests/integration/integration_test.go
```diff
@@ -38,10 +40,21 @@ var _ = ginkgo.BeforeSuite(func() {
 
 	randomEd25519AuthFactory := auth.NewED25519Factory(randomEd25519Priv)
 
+	vmConfig := vm.NewConfig()
+	vmConfig.Config = make(map[string]any)
+	vmConfig.Config["storeTransactions"] = true
+	vmConfig.Config["ws"] = ws.WSConfig{
+		MaxPendingMessages: 10_000,
+	}
```
joshua-kim, line 48:
```suggestion
	vmConfig.Config = map[string]any{
	     "storeTransactions": true,
	     "websocketConfig": ws.Config{
	         MaxPendingMessages: 10_000,
	     }
	}
```

examples/morpheusvm/controller/config.go
```diff
@@ -5,10 +5,13 @@ package controller
 
 import (
 	"encoding/json"
+
+	"github.com/ava-labs/hypersdk/api/ws"
 )
 
 type Config struct {
-	StoreTransactions bool `json:"storeTransactions"`
+	StoreTransactions bool        `json:"storeTransactions"`
```
aaronbuchwald, line 13: Let's change this to `txindexer.Config` so that it defines its own config

examples/morpheusvm/controller/controller.go
```diff
@@ -28,28 +28,28 @@ var (
...
+	opts := []vm.NamespacedOption{
+		vm.NewNamespacedOption("indexer", indexer.WithIndexer(consts.Name, indexer.Endpoint)),
+		vm.NewNamespacedOption("ws", ws.WithWebsocketAPI()),
+		vm.NewNamespacedOption("vmAPI", vm.WithVMAPIs(jsonrpc.JSONRPCServerFactory{})),
+		vm.NewNamespacedOption("controllerAPI", vm.WithControllerAPIs(&jsonRPCServerFactory{})),
 	}
```
aaronbuchwald, line 37: Can we change from naming them here to each service providing its own name? Each can export both the option function and the namespace, so that it can be overridden if desired, but this will remove the responsibility from the controller to define the namespace

examples/morpheusvm/controller/controller.go
aaronbuchwald, line 31: Since we are only going to have one option type, we should just leave the name as `vm.Option`

aaronbuchwald (changes requested):

api/ws/server.go
RodrigoVillar: done

api/ws/server.go
RodrigoVillar: done

examples/morpheusvm/tests/integration/integration_test.go
RodrigoVillar: done

examples/morpheusvm/tests/integration/integration_test.go
RodrigoVillar: done

api/indexer/tx_indexer.go
RodrigoVillar: done

examples/morpheusvm/controller/controller.go
RodrigoVillar: done

api/ws/server.go
```diff
@@ -33,41 +37,52 @@ var (
...
-		)
-
-		webSocketFactory := NewWebSocketServerFactory(handler)
-		txRemovedSubscription := subscriptionFuncFactory[vm.TxRemovedEvent]{
-			AcceptF: func(event vm.TxRemovedEvent) error {
-				return server.RemoveTx(event.TxID, event.Err)
-			},
-		}
+type Config struct {
+	MaxPendingMessages int `json:"maxPendingMessages"`
```
aaronbuchwald, line 41: Nice! Separate from this PR, we can add `Enabled` here as well, so that this can be easily disabled via config instead of via code

api/indexer/tx_indexer.go
```diff
@@ -49,11 +55,11 @@ func WithIndexer(name string, path string) vm.Option {
 			indexer: indexer,
 		}
 
-		if err := vm.WithBlockSubscriptions(subscriptionFactory)(v); err != nil {
+		if err := vm.WithBlockSubscriptions(subscriptionFactory)(v, configBytes); err != nil {
```
aaronbuchwald, line 58: `WithBlockSubscriptions` should never actually take in `configBytes` since it should only be called to modify the VM. Same for `WithVMAPIs`.

There should be the namespaced options, which register a service and separately the functions that are not registering a service, but are only modifying the VM. For example, `WithBlockSubscriptions` should not be called directly, but to register a specific service inside of a namespaced option. There's no need for this to take in `configBytes`, so I think we can remove it from the interface.

examples/morpheusvm/controller/controller.go
```diff
@@ -42,14 +42,14 @@ func New(options ...vm.Option) *vm.VM {
 }
 
 // NewWithOptions returns a VM with the specified options
-func NewWithOptions(options ...vm.Option) *vm.VM {
+func NewWithOptions(namespacedOptions ...vm.Option) (*vm.VM, error) {
```
aaronbuchwald, line 45: nit: let's revert this part of the diff

requester/requester.go
```diff
@@ -15,14 +15,14 @@ import (
 	rpc "github.com/gorilla/rpc/v2/json2"
 )
 
-type Option func(*Options)
+type OptionFunc func(*Options)
```
aaronbuchwald, line 18: Let's revert this part of the diff

tests/integration/integration.go
```diff
@@ -155,17 +168,18 @@ func setInstances() {
 		toEngine := make(chan common.Message, 1)
 		db := memdb.New()
 
-		v := createVM(
-			vm.WithManualGossiper(),
-			vm.WithManualBuilder(),
+		v, err := createVM(
+			vm.NewOption("manualGossiper", vm.WithManualGossiper()),
+			vm.NewOption("manualBuilder", vm.WithManualBuilder()),
```
aaronbuchwald, line 173: can we move both of the names into `WithManualGossiper` directly, so that each service is responsible for providing its own namespace.

vm/vm.go
```diff
@@ -269,8 +277,15 @@ func (vm *VM) Initialize(
...
-	for _, option := range vm.options {
-		if err := option(vm); err != nil {
+	namespacedConfig := make(map[string]json.RawMessage)
+	if err := json.Unmarshal(controllerConfigBytes, &namespacedConfig); err != nil {
+		return fmt.Errorf("failed to unmarshal namespaced config: %w", err)
+	}
+
+	for _, Option := range vm.options {
+		config := namespacedConfig[Option.Namespace]
+		snowCtx.Log.Debug("ctx", zap.Any("name", Option.Namespace), zap.Any("Conf", config))
```
aaronbuchwald, line 287: Can we remove this log?

We should typically use `zap.String` or `zap.Stringer` instead of `zap.Any` where possible

vm/option.go
```diff
@@ -11,45 +11,57 @@ import (
...
+}
+
+func NewOption(namespace string, optionFunc OptionFunc) Option {
+	return Option{
+		Namespace:  namespace,
+		OptionFunc: optionFunc,
+	}
+}
+
+func WithManualBuilder() OptionFunc {
```
aaronbuchwald, line 28: There should be two different concepts - an option registers a useful service ie. `indexer.With` registers the indexer service. The functions here are meant to set a field on the VM and can be called from `OptionFunc`s defined by services, but don't need the `configBytes` here since they will never be treated as an `OptionFunc` directly.

In the case of `WithManualBuilder` and `WithManualGossiper` these should really be `WithBuilder` and `WithGossiper` and `WithManual` would call both to register the manual gossiper and builder.

aaronbuchwald (changes requested):

api/ws/server.go
```diff
@@ -33,41 +37,59 @@ var (
...
 
-		if err := vm.WithTxRemovedSubscriptions(txRemovedSubscription)(v); err != nil {
-			return err
-		}
+func OptionFunc(v *vm.VM, configBytes []byte) error {
+	actionRegistry, authRegistry := v.Registry()
+	config := NewDefaultConfig()
+	if err := json.Unmarshal(configBytes, &config); err != nil {
+		return err
+	}
```
aaronbuchwald: Found the issue - need to check for non-zero length `configBytes` here

api/indexer/tx_indexer.go
```diff
@@ -27,8 +29,12 @@ var (
 	_ event.Subscription[*chain.StatelessBlock]        = (*txDBIndexer)(nil)
 )
 
-func WithIndexer(name string, path string) vm.Option {
-	return func(v *vm.VM) error {
+func With(name string, path string) vm.Option {
+	return vm.NewOption(Namespace, OptionFunc(name, path))
+}
+
+func OptionFunc(name string, path string) vm.OptionFunc {
```
aaronbuchwald, line 36: future PR: I think we can remove name and path from the tx indexer and make them constants in this package

examples/morpheusvm/controller/controller.go
```diff
@@ -28,21 +28,28 @@ var (
...
 
-	return NewWithOptions(opts...)
+	v, err := NewWithOptions(opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Register APIs
+	vm.WithVMAPIs(jsonrpc.JSONRPCServerFactory{})(v)
+	vm.WithControllerAPIs(&jsonRPCServerFactory{})(v)
```
aaronbuchwald, line 46: This should stay in `opts`

tests/integration/integration.go
```diff
@@ -120,6 +122,11 @@ func setInstances() {
 	subnetID := ids.GenerateTestID()
 	chainID := ids.GenerateTestID()
 
+	vmConfig := vm.NewConfig()
+	cb, err := json.Marshal(vmConfig)
+	require.NoError(err)
+	configBytes = cb
```
aaronbuchwald, line 128: Can we remove this and revert this part of the diff since this is back to relying on the defaults?

aaronbuchwald (approved):

PR #1457, opened 2024-08-26
External Subscriber Support

RodrigoVillar:
This PR enables HyperVMs (and in particular, MorpheusVM) to have external subscriber support. 

Summary of Changes

- Adds External Subscriber Server and Client implementations
    - Adds `GRPCHandler` which takes in an `ExternalSubscriberServer` and start a gRPC server
    - Adds `pb.go` files for server/client and `protobuf` generation script
- Adds the external subscriber option to default instances of MorpheusVM (i.e. `controller.New()`)
- Extends the existing integration test structure by test for successful interactions between a HyperVM and a given external subscriber server
- Modifies the `VM` struct by adding the `GenesisBytes` field


examples/morpheusvm/controller/controller.go
```diff
@@ -27,13 +28,14 @@ var (
 	_ vm.ControllerFactory = (*factory)(nil)
 )
 
-// New returns a VM with the indexer, websocket, and rpc apis enabled.
+// New returns a VM with the indexer, websocket, rpc, and external subscriber apis enabled;
```
RodrigoVillar, line 31: remove semicolon

RodrigoVillar:
TODO: rebase main; primary focus is to update initialization of external subscriber server since the concept of a MorpheusVM controller/genesis no longer exists

examples/morpheusvm/cmd/morpheus-cli/cmd/start_external_subscriber.go
```diff
@@ -0,0 +1,68 @@
...
+var startExternalSubscriberCommand = &cobra.Command{
+	Use:   "start-external-subscriber",
+	Short: "Start an external subscriber server",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// Initialize global variables
+		logFactory = logging.NewFactory(logging.Config{
+			DisplayLevel: logging.Debug,
+		})
+		l, err := logFactory.Make("startExternalSubscriberCommand")
+		if err != nil {
```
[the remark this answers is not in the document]
aaronbuchwald: Maybe we should use the console logger only from AvalancheGo, will look into this a bit to see if we can more easily export just the functionality we want here

extension/externalsubscriber/grpchandler.go
```diff
@@ -0,0 +1,56 @@
...
+		gRPCServer: s,
+		listener:   lis,
+		tcpPort:    tcpPort,
+		log:        log,
+	}, nil
+}
+
+func (g *GRPCHandler) Start() {
+	go func() {
+		if err := g.gRPCServer.Serve(g.listener); err != nil {
```
aaronbuchwald, line 42: Looks like we ignore this by default in AvalancheGo https://github.com/ava-labs/avalanchego/blob/82bf033c3fb495d9441b752e4f168b3dc58513c8/vms/rpcchainvm/grpcutils/server.go#L101

extension/externalsubscriber/grpchandler.go
```diff
@@ -0,0 +1,56 @@
...
+
+func NewGRPCHandler(e *ExternalSubscriberServer, log logging.Logger, tcpPort string) (*GRPCHandler, error) {
+	lis, err := net.Listen("tcp", tcpPort) // #nosec G102
+	if err != nil {
+		return nil, err
+	}
+	log.Info("listening to tcp port", zap.String("address", tcpPort))
+
+	s := grpc.NewServer()
+	pb.RegisterExternalSubscriberServer(s, e)
```
aaronbuchwald, line 31: I wonder if it would be cleaner to separate the create server / stop server logic from the `grpchandler` and just use `pb.RegisterExternalSubscriberServer`

examples/morpheusvm/cmd/morpheus-cli/cmd/start_external_subscriber.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/ava-labs/hypersdk/event"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/controller"
+	"github.com/ava-labs/hypersdk/extension/externalsubscriber"
+)
+
+var (
+	logFactory          logging.Factory
+	log                 logging.Logger
+	acceptedSubscribers []event.Subscription[externalsubscriber.ExternalSubscriberSubscriptionData]
```
aaronbuchwald, line 22: Looks like this will always be empty, could we add a service here?

extension/externalsubscriber/option.go
```diff
@@ -0,0 +1,62 @@
...
+type Config struct {
+	Enabled       bool   `json:"enabled"`
+	ServerAddress string `json:"serverAddress"`
+}
+
+func NewConfig() Config {
+	return Config{
+		Enabled: false,
+	}
+}
```
aaronbuchwald, line 28: nit: `NewDefaultConfig` and we can just return an empty struct since this defaults to false

extension/externalsubscriber/server.go
```diff
@@ -0,0 +1,105 @@
...
+	if e.parser != nil {
+		return &emptypb.Empty{}, ErrParserAlreadyInitialized
+	}
+	// Create parser and store
+	parser, err := e.createParser(initRequest.Genesis)
+	if err != nil {
+		return &emptypb.Empty{}, err
+	}
+	e.parser = parser
+	e.log.Info("external subscriber has initialized the parser associated with morpheusVM")
```
aaronbuchwald, line 74: Let's fix this log so that it does not mention MorpeusVM

```suggestion
	e.log.Info("initialized external subscriber parser")
```

extension/externalsubscriber/server.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	results, err := chain.UnmarshalResults(b.Results)
+	if err != nil {
+		return &emptypb.Empty{}, err
+	}
+
+	e.log.Info("external subscriber server received a block with",
+		zap.Any("height", blk.Hght),
+	)
```
aaronbuchwald, line 95:
```suggestion
	e.log.Info("external subscriber received accepted block",
		zap.Uint64("height", blk.Hght),
	)
```

aaronbuchwald (changes requested):
Couple final nits and then this LGTM

extension/externalsubscriber/server.go
aaronbuchwald: Let's use `zap.Uint64` as well as changing the log message

aaronbuchwald (approved):

PR #1301, opened 2024-08-15
ABI Generation

containerman17:
Generate a JSON representation of all VM actions and their fields.

This includes all commits from https://github.com/ava-labs/hypersdk/pull/1198, so it should be merged after that.

When reviewing the code, please [compare it with the code in PR 1198](https://github.com/ava-labs/hypersdk/compare/containerman/standardize-marshal-function...containerman17/abi), not with the main branch.

PR #1459, opened 2024-08-27
Abi Generation V2

containerman17:
A clean version of https://github.com/ava-labs/hypersdk/pull/1301

Generate a JSON representation of all VM actions and their fields.

This includes all commits from https://github.com/ava-labs/hypersdk/pull/1198, so it should be merged after that.

When reviewing the code, please [compare it with the code in PR 1198](https://github.com/ava-labs/hypersdk/compare/containerman/standardize-marshal-function...abi), not with the main branch.


vm/vm.go
```diff
@@ -159,6 +162,11 @@ func New(
 		allocatedNamespaces.Add(option.Namespace)
 	}
 
+	abiString, err := codec.GetVMABIString((*actionRegistry).GetRegisteredTypes())
```
containerman17, line 165: I have doubts about placing the `abiString` initialization in `vm.New`, but it throws an error and takes some time (less than 1 ms). I don't have a better alternative, though.

codec/abi.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	result := SingleActionABI{
+		ID:    action.GetTypeID(),
+		Name:  t.Name(),
+		Types: make(map[string][]ABIField),
+	}
+
+	typesleft := []reflect.Type{t}
+	typesAlreadyProcessed := make(map[reflect.Type]bool)
```
aaronbuchwald, line 52: We should use either AvalancheGo's `set.Set` type or `map[type]struct{}` as a set

codec/abi.go
```diff
@@ -0,0 +1,145 @@
...
+				arrayPrefix += "[]"
+				fieldType = fieldType.Elem()
+			}
+
+			typeName := arrayPrefix + fieldType.Name()
+
+			if fieldType.Kind() == reflect.Struct {
+				otherStructsSeen = append(otherStructsSeen, fieldType)
+			} else if fieldType.Kind() == reflect.Ptr {
+				otherStructsSeen = append(otherStructsSeen, fieldType.Elem())
```
aaronbuchwald, line 133: could we document why we're handling some kinds and not others?

For this type of thing, I think we should move towards defining a clear spec (separate PR) since I'm not sure from this code the rationale behind what is vs. is not handled here

aaronbuchwald (changes requested):

api/jsonrpc/server.go
```diff
@@ -139,3 +139,14 @@ func (j *JSONRPCServer) UnitPrices(
 	reply.UnitPrices = unitPrices
 	return nil
 }
+
+type GetABIArgs struct{}
+
+type GetABIReply struct {
+	ABIJSON string `json:"abi"`
```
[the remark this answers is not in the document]
containerman17: It has to be a string because we're comparing hashes. If it were marshaled differently, the hash would change, and the frontend wallet would reject the transaction when we implement ABI hash checking.

tests/workload/apis.go
```diff
@@ -30,3 +32,19 @@ func GetNetwork(ctx context.Context, require *require.Assertions, uris []string,
...
+		require.NotEmpty(abiString)
+
+		var abi []codec.SingleActionABI
+		err = json.Unmarshal([]byte(abiString), &abi)
+		require.NoError(err)
+
+		require.NotEmpty(abi)
+		require.NotEmpty(abi[0].Name)
+	}
+}
```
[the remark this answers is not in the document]
containerman17: Not really, since we're at the HyperSDK level, and the only thing we can be sure of is that the tested VM probably has at least one action defined.

abi/auto_marshal_abi_spec_test.go
```diff
@@ -0,0 +1,619 @@
...
+	To    codec.Address `serialize:"true" json:"to"`
+	Value uint64        `serialize:"true" json:"value"`
+	Memo  []byte        `serialize:"true" json:"memo"`
+}
+
+func (MockActionTransfer) GetTypeID() uint8 {
+	return 2
+}
+
+func TestABISpec(t *testing.T) {
```
aaronbuchwald, line 74: Summary of our offline conversation:

Could we add testing for:

1. `[]abi.TypedField` (doesn't know about the golang type) <-> binary for hash compatibility of the type description
2. Support passing values in via `[]abi.TypedField` so we can add spec tests that the provided values are serialized in the same way as well
3. Split these into shareable test vectors we can re-use across Golang + TypeScript
4. Implement both serialize/deserialize in both Golang + JavaScript so that we can add identity tests as well

aaronbuchwald (changes requested):

codec/address.go
```diff
@@ -105,3 +106,8 @@ func parseAddressBech32(saddr string) (string, Address, error) {
 	}
 	return phrp, Address(p[:AddressLen]), nil
 }
+
+// It marshals the Address as a base64-encoded string.
+func (a Address) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + base64.StdEncoding.EncodeToString(a[:]) + `"`), nil
+}
```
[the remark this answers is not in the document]
containerman17: We'll do this in a separate PR

aaronbuchwald (commented):
Could we switch from including the mock golang types to generating them from the JSON ABI files?

We would then the generated files in their own directory and add a check in CI to make sure that auto-generating them again (allowing them to overwrite the files) does not result in any staged changes.

containerman17:
Should add unpack tests too

abi/testdata/abi.go
```diff
@@ -0,0 +1,96 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package testdata
```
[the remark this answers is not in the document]
containerman17: You are right, moved

cmd/hypersdk-abi-codegen/main.go
```diff
@@ -0,0 +1,63 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
```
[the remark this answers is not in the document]
containerman17: Agree, done

cmd/hypersdk-abi-codegen/main.go
```diff
@@ -0,0 +1,63 @@
...
+	"path/filepath"
+
+	"github.com/ava-labs/hypersdk/abi"
+)
+
+func main() {
+	if len(os.Args) != 3 {
+		fmt.Println("Usage: hypersdk-abi-codegen <input_abi.json> <output_file.go>")
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
containerman17: Agree, using cobra

codec/address.go
```diff
@@ -105,3 +107,31 @@ func parseAddressBech32(saddr string) (string, Address, error) {
...
+
+// It unmarshals the Address from a base64-encoded string.
+func (a *Address) UnmarshalJSON(data []byte) error {
+	// Remove quotes from the string
+	s := string(data)
+	s = strings.Trim(s, "\"")
```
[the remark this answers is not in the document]
containerman17: good one, agree, done

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"github.com/ava-labs/avalanchego/utils/set"
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+
+	reflect "reflect"
+)
+
+type VMABI struct {
```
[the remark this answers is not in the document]
containerman17: No, it has to be a struct with `serialize:"true"`, but good renaming suggestion

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+type SingleActionABI struct {
+	ID    uint8           `serialize:"true" json:"id"`
+	Name  string          `serialize:"true" json:"name"`
+	Types []SingleTypeABI `serialize:"true" json:"types"`
+}
+
+type SingleTypeABI struct {
+	Name   string     `serialize:"true" json:"name"`
+	Fields []ABIField `serialize:"true" json:"fields"`
+}
```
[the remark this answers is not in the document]
containerman17: Renamed that and not only that

abi/auto_marshal_abi_spec_test.go
```diff
@@ -0,0 +1,96 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			// Get object from file
+			err := json.Unmarshal(mustReadFile(t, "testdata/"+tc.name+".json"), tc.object)
+			require.NoError(err)
+
+			// Get spec from file
+			abiJSON := mustReadFile(t, "testdata/abi.json")
+			var abiFromFile VMABI
+			err = json.Unmarshal(abiJSON, &abiFromFile)
+			require.NoError(err)
```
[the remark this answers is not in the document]
containerman17: Done

abi/abi_test.go
```diff
@@ -0,0 +1,46 @@
...
+	})
+	require.NoError(err)
+
+	expectedAbiJSON := mustReadFile(t, "testdata/abi.json")
+	expectedAbi := mustJSONParse[VMABI](t, string(expectedAbiJSON))
+
+	require.Equal(expectedAbi, actualABI)
+}
+
+func TestABIsABI(t *testing.T) {
```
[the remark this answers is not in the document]
containerman17: That's better

abi/abi_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/hypersdk/abi/testdata"
+	"github.com/ava-labs/hypersdk/codec"
+)
+
+func TestABI(t *testing.T) {
```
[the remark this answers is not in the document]
containerman17: TestDescribeVM even better

abi/codegen.go
```diff
@@ -0,0 +1,74 @@
...
+
+			sb.WriteString(fmt.Sprintf("type %s struct {\n", typ.Name))
+			for _, field := range typ.Fields {
+				fieldNameUpperCase := strings.ToUpper(field.Name[0:1]) + field.Name[1:]
+
+				goType := convertToGoType(field.Type)
+				if field.Name[0] >= 'A' && field.Name[0] <= 'Z' {
+					sb.WriteString(fmt.Sprintf("\t%s %s `serialize:\"true\"`\n", fieldNameUpperCase, goType))
+				} else {
+					sb.WriteString(fmt.Sprintf("\t%s %s `serialize:\"true\" json:\"%s\"`\n", fieldNameUpperCase, goType, field.Name))
```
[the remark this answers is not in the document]
containerman17: That's correct!

abi/codegen.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"fmt"
+	"go/format"
+	"strings"
+
+	"github.com/ava-labs/avalanchego/utils/set"
+)
+
+func GenerateGoStructs(abi VMABI, packageName string) (string, error) {
```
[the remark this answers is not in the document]
containerman17: https://gist.github.com/containerman17/d909c7a2b3a33b3de888f94342c681e3

Code is too bloated. Sprintf is better tool in this case

vm/vm.go
```diff
@@ -1250,3 +1251,10 @@ func (vm *VM) Fatal(msg string, fields ...zap.Field) {
 	vm.snowCtx.Log.Fatal(msg, fields...)
 	panic("fatal error")
 }
+
+// GetABI returns the ABI (Application Binary Interface) string for the VM.
+// This ABI string represents the interface of the VM,
+// containing all actions.
+func (vm *VM) GetABI() (abi.VMABI, error) {
+	return abi.GetVMABI((*vm.actionRegistry).GetRegisteredTypes())
```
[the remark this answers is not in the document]
containerman17: I have the same question

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+	"crypto/sha256"
+	"fmt"
+	"strings"
+
+	"github.com/ava-labs/avalanchego/utils/set"
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+
+	reflect "reflect"
```
joshua-kim, line 16: nit: Avoid redundant import aliases

abi/auto_marshal_abi_spec_test.go
```diff
@@ -0,0 +1,102 @@
...
+import (
+	"encoding/hex"
+	"encoding/json"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	_ "embed"
```
joshua-kim, line 15: Why is this imported? Are there side effects of this import in its `init` we depend on?

abi/testdata/abi.abi.json
```diff
@@ -0,0 +1,62 @@
...
+            "types": [
+                {
+                    "name": "VM",
+                    "fields": [
+                        {
+                            "name": "actions",
+                            "type": "[]Action"
+                        }
+                    ]
+                },
```
joshua-kim, line 15: We synced on this offline - but we discussed seeing if we could define types at the ABI level, instead of the action-level so that you don't have to re-define types that are shared between actions.

abi/testdata/strOnly.hex
joshua-kim, line 1: Why have files for these hashes? Can't we just put these in the unit test itself?

vm/vm.go
```diff
@@ -1250,3 +1251,10 @@ func (vm *VM) Fatal(msg string, fields ...zap.Field) {
 	vm.snowCtx.Log.Fatal(msg, fields...)
 	panic("fatal error")
 }
+
+// GetABI returns the ABI (Application Binary Interface) string for the VM.
+// This ABI string represents the interface of the VM,
+// containing all actions.
+func (vm *VM) GetABI() (abi.VM, error) {
```
joshua-kim, line 1258: We don't need to expose this i think, I think `vm.Registry()` lets us do this in the caller

abi/auto_marshal_abi_spec_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+
+	reflect "reflect"
```
joshua-kim, line 20: nit avoid redundant import alias

codec/packer.go
```diff
@@ -155,3 +155,5 @@ func (p *Packer) addErr(err error) {
 func (p *Packer) Offset() int {
 	return p.Packer.Offset
 }
+
+type Bytes []byte
```
joshua-kim, line 159: Why do we need to define this type?

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+	Name  string `serialize:"true" json:"name"`
+	Types []Type `serialize:"true" json:"types"`
+}
+
+type Type struct {
+	Name   string  `serialize:"true" json:"name"`
+	Fields []Field `serialize:"true" json:"fields"`
+}
+
+func DescribeVM(actions []codec.Typed) (VM, error) {
```
joshua-kim, line 66: nit: rename to `New` (since callers will use `abi.New` to construct their abi).

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+		actionABI, err := describeAction(action)
+		if err != nil {
+			return VM{}, err
+		}
+		vmABI = append(vmABI, actionABI)
+	}
+	return VM{Actions: vmABI}, nil
+}
+
+// describeAction generates the VM for a single action.
```
joshua-kim, line 78: nit: This comment looks awkward since it references `generates the VM`... which I think is supposed to be `generates the ABI` or something.

joshua-kim:
Another idea which takes out a lot of complexity of this PR would to be to consider using proto (since we will be able to support gRPC-based apis once https://github.com/ava-labs/avalanchego/pull/3294 is merged).

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+	reflect "reflect"
+)
+
+type VM struct {
+	Actions []Action `serialize:"true" json:"actions"`
+}
+
+var _ codec.Typed = (*VM)(nil)
+
+const ABITypeID = consts.MaxUint8
```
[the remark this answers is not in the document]
containerman17: You’re right, the VM's ABI is separate from the ABI’s ABI, so no collision is possible.

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (VM) GetTypeID() uint8 {
```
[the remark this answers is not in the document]
containerman17: thanks!

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+	return ABITypeID
+}
+
+func (a *VM) Hash() [32]byte {
+	writer := codec.NewWriter(0, consts.NetworkSizeLimit)
+	err := codec.LinearCodec.MarshalInto(a, writer.Packer)
+	if err != nil {
+		// should never happen in prod, safe to panic
+		panic(fmt.Errorf("failed to marshal abi: MarshalInto: %w", err))
```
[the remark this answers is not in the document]
containerman17: This function is defined to follow a similar spec to TS, but you're right, it would be cleaner to have it return an error.

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+
+var _ codec.Typed = (*VM)(nil)
+
+const ABITypeID = consts.MaxUint8
+
+func (VM) GetTypeID() uint8 {
+	return ABITypeID
+}
+
+func (a *VM) Hash() [32]byte {
```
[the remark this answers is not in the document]
containerman17: Just inlined it into the test, no actual need for in in golang implementation

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/ava-labs/avalanchego/utils/set"
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+
+	reflect "reflect"
+)
+
+type VM struct {
+	Actions []Action `serialize:"true" json:"actions"`
```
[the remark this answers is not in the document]
containerman17: No, it's not the default, unfortunately. But I'm happy to hear feedback if anyone thinks JSON tags everywhere are uglier than JSON object keys starting with an uppercase letter.

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+const ABITypeID = consts.MaxUint8
+
+func (VM) GetTypeID() uint8 {
+	return ABITypeID
+}
+
+func (a *VM) Hash() [32]byte {
+	writer := codec.NewWriter(0, consts.NetworkSizeLimit)
+	err := codec.LinearCodec.MarshalInto(a, writer.Packer)
+	if err != nil {
```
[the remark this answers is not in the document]
containerman17: I strongly disagree. I prefer error handling on a separate line - it looks cleaner to me. Do we have a standard for that? It would be nice to agree on one style or another.

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+
+// Field represents a field in the VM (Application Binary Interface).
+type Field struct {
+	// Name of the field, overridden by the json tag if present
+	Name string `serialize:"true" json:"name"`
+	// Type of the field, either a Go type or struct name (excluding package name)
+	Type string `serialize:"true" json:"type"`
+}
+
+// Action represents the VM for an action.
```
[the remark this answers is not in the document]
containerman17: Typo, thanks

abi/abi.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"github.com/ava-labs/avalanchego/utils/set"
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+
+	reflect "reflect"
+)
+
+type VM struct {
```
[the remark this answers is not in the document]
containerman17: Agree here 100%

abi/codegen_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	expected := mustReadFile(t, "mockabi.test.go")
+
+	formatted, err := format.Source(removeCommentLines(expected))
+	require.NoError(err)
+
+	require.Equal(string(formatted), code)
+}
+
+func removeCommentLines(input []byte) []byte {
```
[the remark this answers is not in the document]
containerman17: It's actually only for one particular test, so it doesn't matter much. This specific solution removes an extra \n, so it wouldn't work.

abi/mockabi.test.go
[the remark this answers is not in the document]
containerman17: Test files aren't exported, so it should be fine. I'm also against separating tests or test assets from the code being tested.

abi/auto_marshal_abi_spec_test.go
containerman17: Thanks for catching it! Old code

abi/testdata/abi.abi.json
containerman17: Done!

abi/testdata/strOnly.hex
containerman17: BC we share the testdata folder with TypeScript implementation

vm/vm.go
containerman17: That's a good catch!

abi/auto_marshal_abi_spec_test.go
containerman17: Thanks!

codec/packer.go
containerman17: Yes, we do. Otherwise reflect treats []byte as []uint8

abi/abi.go
containerman17: NewABI even better.

abi/abi.go
containerman17: NewABI even better.

abi/abi.go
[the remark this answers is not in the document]
aaronbuchwald: We use the in-line format when a single error value is returned

abi/abi.go
aaronbuchwald: Let's use `New` so that callers use `abi.New` instead of `abi.NewABI`

abi/abi.go
```diff
@@ -0,0 +1,178 @@
...
+	typesSet := set.Set[string]{}
+
+	for _, action := range actions {
+		actionABI, typeABI, err := describeAction(action)
+		if err != nil {
+			return ABI{}, err
+		}
+		vmActions = append(vmActions, actionABI)
+		for _, t := range typeABI {
+			if !typesSet.Contains(t.Name) {
```
aaronbuchwald, line 58: How do we handle the case that there are two different types with the same name ie. if they're from two different packages?

abi/abi_test.go
```diff
@@ -0,0 +1,58 @@
...
+		MockObjectStringAndBytes{},
+		MockObjectArrays{},
+		MockActionWithTransfer{},
+		MockActionWithTransferArray{},
+		Outer{},
+	})
+	require.NoError(err)
+
+	expectedAbiJSON := mustReadFile(t, "testdata/abi.json")
+	expectedAbi := mustJSONParse[ABI](t, string(expectedAbiJSON))
```
aaronbuchwald, line 31: Could we update this to use `expectedABIJSON` and `expectedABI` ?

abi/abi_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	require.Equal(mustPrintOrderedJSON(t, expectedAbi), mustPrintOrderedJSON(t, actualABI))
```
aaronbuchwald, line 33: Can we use `require.Equal(expectedABI, actualABI)` ?

abi/abi_test.go
```diff
@@ -0,0 +1,58 @@
...
+	bytes, err := json.Marshal(v)
+	require.NoError(t, err)
+
+	var parsed map[string]any
+	require.NoError(t, json.Unmarshal(bytes, &parsed))
+
+	ordered, err := json.Marshal(parsed)
+	require.NoError(t, err)
+	return string(ordered)
+}
```
aaronbuchwald, line 46: Can we remove this in favor of using `require.Equal(...)` ?

abi/codegen_test.go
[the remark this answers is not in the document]
aaronbuchwald: Could we just change this to remove header and strip the first 3 lines of code so that it does exactly what we need and it's clear that it's not a general solution?

abi/auto_marshal_abi_spec_test.go
```diff
@@ -0,0 +1,105 @@
...
+			// Get object from file
+			err := json.Unmarshal(mustReadFile(t, "testdata/"+tc.name+".json"), unmarshaledFromJSON)
+			require.NoError(err)
+
+			// Marshal the object
+			objectPacker := codec.NewWriter(0, consts.NetworkSizeLimit)
+			err = codec.LinearCodec.MarshalInto(unmarshaledFromJSON, objectPacker.Packer)
+			require.NoError(err)
+
+			objectDigest := objectPacker.Bytes()
```
aaronbuchwald, line 82: nit: `objectBytes`

abi/auto_marshal_abi_spec_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+			// Compare with expected hex
+			expectedHex := string(mustReadFile(t, "testdata/"+tc.name+".hex"))
+			expectedHex = strings.TrimSpace(expectedHex)
+			require.Equal(expectedHex, hex.EncodeToString(objectDigest), tc.name)
```
aaronbuchwald, line 87:
```suggestion
			require.Equal(expectedHex, hex.EncodeToString(objectDigest))
```

I think we can drop the name since this is run as a subtest

abi/auto_marshal_abi_spec_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+)
+
+// Combined VM and AutoMarshal spec
+// Used to verify TypeScript implementation
+// Tests added as needed by TypeScript
+// Ensures consistency in marshaling, not testing Go struct marshaling itself
```
aaronbuchwald, line 24: Could we remove this comment block and instead put a comment on each test?

aaronbuchwald:
Could we add a README for the ABI package that describes the ABI format, how it's used, the test vectors in the `testdata/` + how it's intended to be used by any language starting with TypeScript to support frontends ?

abi/abi.go
containerman17: We can't, or we will have to include go package name into the JSON, which will cause troubles implementing it in other languages. I decided to drop support for same-name structs

containerman17: Happy to discuss

abi/auto_marshal_abi_spec_test.go
containerman17: > Could we just change this to remove header and strip the first 3 lines of code so that it does exactly what we need and it's clear that it's not a general solution? 
It's going to be awkward when copyright changes to 3 lines instead of 2

abi/abi_test.go
containerman17: What do you mean?

abi/codegen.go
```diff
@@ -0,0 +1,73 @@
...
+
+	for _, typ := range abi.Types {
+		if processed.Contains(typ.Name) {
+			continue
+		}
+		processed.Add(typ.Name)
+
+		sb.WriteString(fmt.Sprintf("type %s struct {\n", typ.Name))
+		for _, field := range typ.Fields {
+			fieldNameUpperCase := strings.ToUpper(field.Name[0:1]) + field.Name[1:]
```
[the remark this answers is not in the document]
containerman17: Hmm, I don't think so. It's a string, not []byte

tests/workload/apis.go
```diff
@@ -30,3 +31,15 @@ func GetNetwork(ctx context.Context, require *require.Assertions, uris []string,
...
 }
+
+func GetABI(ctx context.Context, require *require.Assertions, uris []string, expectedABI abi.ABI) {
+	for _, uri := range uris {
+		client := jsonrpc.NewJSONRPCClient(uri)
+		actualABI, err := client.GetABI(ctx)
+		require.NoError(err)
+
+		require.GreaterOrEqual(len(actualABI.Actions), 1)
+		require.NotEmpty(actualABI.Actions[0].Action)
```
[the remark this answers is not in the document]
containerman17: I’m strongly against checking against a computed value just to avoid tests that pass by checking `nil == nil`. This rule comes from many hours spent dealing with false-positive tests.

abi/abi.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return actionABI, typesABI, nil
+}
+
+// describeStruct analyzes a struct type and returns its fields and any nested struct types it found
+func describeStruct(t reflect.Type) ([]Field, []reflect.Type, error) {
+	kind := t.Kind()
+
+	if kind != reflect.Struct {
+		return nil, nil, fmt.Errorf("type %s is not a struct", t.String())
```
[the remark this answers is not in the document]
containerman17: I used Name

api/jsonrpc/server.go
```diff
@@ -139,3 +140,20 @@ func (j *JSONRPCServer) UnitPrices(
...
+
+type GetABIArgs struct{}
+
+type GetABIReply struct {
+	ABI abi.ABI `json:"abi"`
+}
+
+func (j *JSONRPCServer) GetABI(_ *http.Request, _ *GetABIArgs, reply *GetABIReply) error {
+	actionRegistry, _ := j.vm.Registry()
+	// Dereference actionRegistry to access the underlying value and call GetRegisteredTypes method
```
aaronbuchwald, line 152:
```suggestion
	// Must dereference aliased type to call GetRegisteredTypes
```

abi/codegen.go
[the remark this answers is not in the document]
aaronbuchwald: Golang will grab the first byte instead of first character if you index into a string this way. Golang be weird, but as discussed offline I think this is ok to leave as is.

abi/README.md
```diff
@@ -0,0 +1,51 @@
...
+Wallets use ABI to display proper action names and field names. To verify ABI implementation in other languages, marshal the ABI into binary, hash it, and compare with the known hash.
+
+## Constraints
+- Actions must have an ID; other structs do not require one.
+- Multiple structs with the same name from different packages are not supported.
+- Maps are not supported; use slices (arrays) instead.
+- Built-in types include Address and Bytes.
+
+## Code Generation
+Use cmd/abigen to automatically generate Go structs from JSON.
```
aaronbuchwald, line 51: Can we add instructions instead of saying that this is handled by `cmd/abigen` ?

abi/README.md
```diff
@@ -0,0 +1,51 @@
...
+                    "type": "uint16"
+                }
+            ]
+        },
+        //...
+}
+```
+
+The ABI consists of two main sections:
+- actions: A list of action definitions, each with an id and action name.
```
aaronbuchwald, line 32: Could we explain here that the action name must also be included as a type ie. they're coupled and we use the dictionary of types to interpret the actions?

abi/README.md
```diff
@@ -0,0 +1,51 @@
...
+- JSON files should align with their corresponding .hex files.
+
+## Hash verification
+Wallets use ABI to display proper action names and field names. To verify ABI implementation in other languages, marshal the ABI into binary, hash it, and compare with the known hash.
+
+## Constraints
+- Actions must have an ID; other structs do not require one.
+- Multiple structs with the same name from different packages are not supported.
+- Maps are not supported; use slices (arrays) instead.
+- Built-in types include Address and Bytes.
```
aaronbuchwald, line 47: What are address and bytes types and why are they special cased?

abi/README.md
```diff
@@ -0,0 +1,51 @@
...
+- types: A list of type definitions, each with a name and a list of fields.
+
+## Implementation
+To create an implementation of this package in any other langauge:
+- Copy the testdata folder.
+- Ensure all marshaling is identical to the Go implementation.
+- JSON files should align with their corresponding .hex files.
+
+## Hash verification
+Wallets use ABI to display proper action names and field names. To verify ABI implementation in other languages, marshal the ABI into binary, hash it, and compare with the known hash.
```
aaronbuchwald, line 41: Could you explain why?

abi/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The ABI consists of two main sections:
+- actions: A list of action definitions, each with an id and action name.
+- types: A list of type definitions, each with a name and a list of fields.
+
+## Implementation
+To create an implementation of this package in any other langauge:
+- Copy the testdata folder.
+- Ensure all marshaling is identical to the Go implementation.
+- JSON files should align with their corresponding .hex files.
```
aaronbuchwald, line 38: Let's add a little more detail here. Could we explain the structure of the testdata folder and how an implementation should use it to ensure compliance?

aaronbuchwald (approved):

PR #1463, opened 2024-08-27
Remove std dependency from wasmlanche-sdk

richardpringle:
- **Consolidate std use**
- **Replace the standard-library hash-map**
- **Remove std dependency from wasmlanche_sdk::memory**
- **Remove std dependency in wasmlanche_sdk::state**
- **Remove std dependency from wasmlanche_sdk::types**
- **Don't use thread_local for allocations**
- **Remove std dependency from wasmlanche-sdk**

### DIFF
--- a/x/programs/rust/wasmlanche-sdk/src/memory.rs
+++ b/x/programs/rust/wasmlanche-sdk/src/memory.rs
@@ -7,7 +7,14 @@
 //! the program. These methods are unsafe as should be used
 //! with caution.
 
-use std::{alloc::Layout, cell::RefCell, collections::HashMap, mem::ManuallyDrop, ops::Deref};
+use std::{
+    alloc::{self, Layout},
+    cell::RefCell,
+    collections::HashMap,
+    mem::ManuallyDrop,
+    ops::Deref,
+    slice,
+};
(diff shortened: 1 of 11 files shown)


x/programs/rust/wasmlanche-sdk/src/memory.rs
```diff
@@ -7,12 +7,28 @@
...
+};
+use hashbrown::HashMap;
+
+/// Map of pointer to the length of its content on the heap
+
+// TODO:
+// `RefCell` shouldn't be necessary here since we need unsafe access anyway.
+// However, it does guarantee no UB on a single-thread.
+
+static mut ALLOCATIONS: LazyCell<RefCell<HashMap<*const u8, usize>>> =
```
samliok, line 30: what does `LazyCell` do here?

samliok (approved):

x/programs/rust/wasmlanche-sdk/src/memory.rs
richardpringle: It instantiates on first access because `HashMap::new` has to allocate (not a const function).

Issue #1159, opened 2024-07-17
Decompose VM Controller


aaronbuchwald:
The VM Controller `Initialize` function currently takes in a large number of parameters and returns a large number as well:

https://github.com/ava-labs/hypersdk/blob/e4dae2c380b685bb509a3d2858720aa5d6a49843/vm/dependencies.go#L101

This couples together a few pieces of functionality that can be more cleanly separated and made easier to re-use across different VM implementations.

I'd suggest that we break this down into distinct components:

## Genesis/Upgrade + Chain Rules

This suggestion is to remove `genesis/upgrade` from the params passed into the VM and remove `Rules(timestamp) chain.Rules` from the controller interface.

This functionality should be moved out to its own component that's responsible for parsing the `genesis/upgrade` bytes and returning the chain rules as needed given a timestamp.

The current VM implementations already pass this off to their `Genesis` structs.

We can then use a factory injection pattern to replace the current VM constructor that just takes in the controller and a version to take in a genesis factory as well.

This would look like:

```golang
func New() *vm.VM {
	return vm.New(
		genesisFactory{},
		&Controller{},
		version.Version,
	)
}
```

The genesis factory should have an interface to construct a genesis:

```golang
type GenesisFactory interface {
	New(genesisBytes []byte, upgradeBytes []byte) (Genesis, error)
}

type Genesis interface {
	Load(context.Context, avatrace.Tracer, state.Mutable) error

	GetStateBranchFactor() merkledb.BranchFactor

	Rules(timestamp int64, networkID uint32, chainID ids.ID) chain.Rules
}
```

## Builder / Gossiper

Similarly, the controller implementations both pass in a default builder/gossiper implementation already provided by the HyperSDK. The only configurability is used to support a test implementation, which can be handled instead by creating a VM with those test implementations passed in rather than passing a config value into `controller.Initialize` in order to determine which builder/gossiper implementation to use.

This would change roughly from the above into:

```golang
func New() *vm.VM {
	return vm.New(
		genesisFactory{},
		builder.Factory{},
		gossiper.Factory{},
		&Controller{},
		version.Version,
	)
}
```

## Action / Auth Registry + Auth Engine

The action/auth registry are the part that most developers will customize with the current layout of the HyperSDK. Currently, these are registered as constants via `init()` functions and returned directly without modification in the `Initialize` function.

We can push this into the VM constructor as well:

```golang
func New() *vm.VM {
	return vm.New(
		genesisFactory{},
[...]

PR #1468, opened 2024-08-28
Refactor genesis no controller

aaronbuchwald:


aaronbuchwald:
This change completes https://github.com/ava-labs/hypersdk/issues/1159

examples/morpheusvm/storage/state_manager.go
RodrigoVillar, line 1: Future PR(?): if we want to take some time towards exploring the concept of a
"middleman" between actions and storage, `StateManager` can perhaps be that middleman

RodrigoVillar (approved):

examples/morpheusvm/storage/state_manager.go
aaronbuchwald: Hmm I think this is two different concepts although it would make a lot of sense to more clearly separate them.

The `StateManager` right now is the state manager for the state that needs to be exposed to the HyperSDK. It doesn't know the internals of the VM's state, but needs to be able to access certain fields, deduct balances to pay for fees, and now add balances as well to handle the genesis allocations.

I think we could re-use the concept of a separating middle layer in actions though.

PR #1469, opened 2024-08-28
Simplify codec.Packer

containerman17:
The goal of this PR is to simplify things for https://github.com/ava-labs/hypersdk/pull/1198 since it already includes many changes. This update removes all the proxy functions. Please merge this before handling https://github.com/ava-labs/hypersdk/pull/1198 and https://github.com/ava-labs/hypersdk/pull/1459.

In the future, I would like to remove `codec.Packer` completely, but that's not part of this update.


codec/packer.go
```diff
@@ -17,166 +17,112 @@ import (
...
-
-func (p *Packer) UnpackByte() byte {
-	return p.p.UnpackByte()
-}
-
-func (p *Packer) PackFixedBytes(b []byte) {
-	p.p.PackFixedBytes(b)
-}
-
+// FIXME: do we need this or UnpackFixedBytes is enough?
```
aaronbuchwald, line 52: Let's keep this, since `UnpackFixedBytes` is not the same as `UnpackAddress`, I think it makes sense to keep both `PackAddress` and `UnpackAddress` with the extra empty address check, so that it's symmetric.

containerman17:
1. Removed all unnecessary changes.
2. Rolled back all performance-related code.
3. Squeezed in a PackInt bug fix in this PR. TLDR: PackInt misbehaves when packing numbers between int32 and int64 on 64-bit systems. [More on that here](https://github.com/ava-labs/hypersdk/pull/1198#issuecomment-2261884573).
4. Removed proxy methods that do nothing.

chain/block.go
```diff
@@ -805,7 +805,7 @@ func (b *StatefulBlock) Marshal() ([]byte, error) {
 	p.PackInt64(b.Tmstmp)
 	p.PackUint64(b.Hght)
 
-	p.PackInt(len(b.Txs))
+	p.PackInt(int32(len(b.Txs))) // FIXME: check if int32 is too large for this use case
```
aaronbuchwald, line 808: It's very unlikely we'll go over max uint16, but it could be close for comfort, so I think we can keep int32. We could certainly switch to using uint32 as this will never be negative (same for the other lengths of things modified in this PR, most of which are also one per tx)

aaronbuchwald, line 808:
```suggestion
	p.PackInt(int32(len(b.Txs)))
```

codec/packer_test.go
```diff
@@ -125,13 +125,13 @@ func TestNewReader(t *testing.T) {
 	vInt := 900
 	wp := NewWriter(5, 5)
 	// Add an int and a bool
-	wp.PackInt(vInt)
+	wp.PackInt(int32(vInt))
```
aaronbuchwald, line 128: nit: declare `vInt` as the desired type above instead of casting twice here and below

codec/packer.go
```diff
@@ -94,65 +70,65 @@ func (p *Packer) UnpackFixedBytes(size int, dest *[]byte) {
...
 
-func (p *Packer) UnpackInt(required bool) int {
-	v := p.p.UnpackInt()
+func (p *Packer) UnpackInt(required bool) int32 {
+	v := p.Packer.UnpackInt()
 	if required && v == 0 {
 		p.addErr(fmt.Errorf("%w: Int field is not populated", ErrFieldNotPopulated))
 	}
-	return int(v)
+	return int32(v)
```
aaronbuchwald, line 115: It's not clear to me why this ever did this type conversion. I think we should just use the `uint32` type that the underlying packer is giving us especially because everywhere we're using it, we could use `uint32`

aaronbuchwald (dismissed):
LGTM - two optional nits

aaronbuchwald (approved):

PR #1472, opened 2024-08-28
Fix memory lock issue for testing

richardpringle:
In this PR, I use conditional compilation to either use `thread_local` or a static `RefCell` with unsafe. The "unsafe" code is safe to use in a single-threaded environment (wasm), but not in a multi-threaded environment. Unit tests are usually run on local architecture in a multi-threaded environment. 

"But @richardpringle, doesn't that mean that you aren't testing the code that will be run in wasm? Isn't that the main code that we care about?" See TODO below.

### TODO
I tested this locally with `cross` to make sure it would run on `wasm32`, but I need to add integration tests so that it runs in CI with every PR. 

@samliok, my suggestion is we merge this code *without* the extra integration tests. I'm going to clean up those tests a little to differenciate between memory tests and testing the macro code.

### DIFF
--- a/x/programs/wasmlanche/src/memory.rs
+++ b/x/programs/wasmlanche/src/memory.rs
@@ -13,22 +13,9 @@ use alloc::{
     alloc::{alloc as allocate, dealloc as deallocate, handle_alloc_error, Layout},
     vec::Vec,
 };
-use core::{
-    cell::{LazyCell, RefCell},
-    mem::ManuallyDrop,
-    ops::Deref,
-    slice,
-};
-use hashbrown::HashMap;
+use core::{mem::ManuallyDrop, ops::Deref, slice};
 
-/// Map of pointer to the length of its content on the heap
-
-// TODO:
-// `RefCell` shouldn't be necessary here since we need unsafe access anyway.
-// However, it does guarantee no UB on a single-thread.
-
-static mut ALLOCATIONS: LazyCell<RefCell<HashMap<*const u8, usize>>> =
-    LazyCell::new(|| RefCell::new(HashMap::new()));
+mod allocations;
 
 #[doc(hidden)]
 /// A pointer where data points to the host.
@@ -40,15 +27,8 @@ impl Deref for HostPtr {
     type Target = [u8];
 
     fn deref(&self) -> &Self::Target {
-        unsafe {
-            let len = ALLOCATIONS
-                .borrow()
-                .get(&self.0)
-                .copied()
-                .expect("attempted to deref invalid host pointer");
-
-            slice::from_raw_parts(self.0, len)
-        }
+        let len = allocations::get(self.0).expect("attempted to deref invalid host pointer");
+        unsafe { slice::from_raw_parts(self.0, len) }
     }
 }
 
@@ -58,7 +38,7 @@ impl Drop for HostPtr {
             return;
         }
 
-        let len = remove(self.0);
+        let len = allocations::remove(self.0).expect("attempted to drop invalid host pointer");
         let layout = Layout::array::<u8>(len).expect("capacity overflow");
 
         unsafe { deallocate(self.0.cast_mut(), layout) };
@@ -70,7 +50,8 @@ impl From<HostPtr> for Vec<u8> {
         // drop will dealloc the bytes
         let host_ptr = ManuallyDrop::new(host_ptr);
 
-        let len = remove(host_ptr.0);
+        let len = allocations::remove(host_ptr.0)
+            .expect("attempted to convert invalid host pointer to a Vec");
 
         unsafe { Vec::from_raw_parts(host_ptr.0.cast_mut(), len, len) }
     }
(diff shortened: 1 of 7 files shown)


x/programs/wasmlanche/src/memory.rs
```diff
@@ -13,22 +13,57 @@ use alloc::{
...
+                static ALLOCATIONS: RefCell<LenMap> = RefCell::new(HashMap::new());
+            }
+
+            fn borrow<F: FnOnce(&LenMap) -> R, R>(f: F) -> R {
+                ALLOCATIONS.with_borrow(f)
+            }
+
+            fn borrow_mut<F: FnOnce(&mut LenMap) -> R, R>(f: F) -> R {
+                ALLOCATIONS.with_borrow_mut(f)
+            }
```
richardpringle, line 51: The secret sauce is in the `borrow` and `borrow_mut` functions. The business logic is the *exact* same in both cases (of the `cfg_if`); it just acts on different types (a `&HashMap<*const u8, size>>>` vs a `Ref<'_, HashMap<*const u8, usize>>` and the mutable equivalents).

x/programs/wasmlanche/src/memory.rs
```diff
@@ -13,22 +13,106 @@ use alloc::{
...
+            unsafe impl Sync for SingletonLenMap {}
+
+
+            impl SingletonLenMap {
+                const fn new() -> Self {
+                    Self(AtomicPtr::new(core::ptr::null_mut()))
+                }
+
+                #[must_use]
+                fn init(&self) -> Guard {
```
samliok, line 62: function name was a bit misleading, seemed like it was initing the `SingletonLenMap`. Maybe call it `acquire_guard` or `init_guard`

samliok: ah actually this is doing two things; both grabbing the guard and initing the map if not created. Still think renaming would make this clearer(`init_gaurd` doesn't make sense anymore but `acquire_guard` could.

x/programs/wasmlanche/src/memory.rs
```diff
@@ -13,22 +13,106 @@ use alloc::{
...
+                    assert!(!LOCK.swap(true, Acquire), "attempted to lock a locked singleton");
```
samliok, line 63: will this wait until the lock can be acquired?

samliok (dismissed):
couple comments, going to approve but maybe some of my comments warrant a change. Also I think we should be wary about relying on compiler directives since that will make code harder to maintain and change in the future. It's fine here, but noting this for me as well

x/programs/wasmlanche/src/memory.rs
richardpringle: No, it panics if you try to lock a locked lock. This code isn't sound in a mutli-threaded environment.

x/programs/wasmlanche/src/memory.rs
```diff
@@ -13,22 +13,106 @@ use alloc::{
...
+use core::{mem::ManuallyDrop, ops::Deref, slice};
+
+/// Safety:
+/// Only safe to use in a single-threaded environment
+mod allocations {
+    use cfg_if::cfg_if;
+    use core::cell::RefCell;
+    use hashbrown::HashMap;
+
+    type LenMap = HashMap<*const u8, usize>;
```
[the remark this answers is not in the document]
richardpringle: That wouldn't work when allocating. You want the HostPtr to be returned, but it would be owned by the map. HostPtr isn't Clone or Copy, but *const u8 is. We don't want to make HostPtr Clone or Copy either as HostPtr's free on Drop.

x/programs/wasmlanche/src/memory.rs
```diff
@@ -13,22 +13,106 @@ use alloc::{
...
+
+    // Map of pointer to the length of its content on the heap
```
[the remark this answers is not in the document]
richardpringle: Whoops. I should probably comment all these functions too

richardpringle:
> couple comments, going to approve but maybe some of my comments warrant a change. Also I think we should be wary about relying on compiler directives since that will make code harder to maintain and change in the future. It's fine here, but noting this for me as well

Yeah, the comments definitely warrant changes, thanks! 

As for compiler directives, I don't think we need to be weary about using them. They are very commonly used (every crate that uses feature flags). I do think we.need to be diligent when they're used though. For instance here, my initial implementation had some code duplication which can become problematic. Now, the types that wrap the map are different, but everything else follows the same code path. 

I definitely need to add more tests though (and comments). 

And now that I'm thinking about it, I don't think the usage of RefCell makes sense as it doesn't protect from data-races. I need to just use the lock.

x/programs/wasmlanche/src/memory.rs
```diff
@@ -13,22 +13,106 @@ use alloc::{
...
+                    if map.is_null() {
+                        let map = Box::into_raw(Box::new(RefCell::new(HashMap::new())));
+                        self.0.store(map, Relaxed);
+                    }
+
+                    Guard
+                }
+
+                fn as_ref(&self) -> &RefCell<LenMap> {
+                    let _guard = self.init();
```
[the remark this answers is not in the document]
richardpringle: Scratch that, not using `RefCell` at all

x/programs/wasmlanche/src/memory.rs
samliok, line 1: should we move this file to memory/memory.rs?

samliok (approved):

x/programs/wasmlanche/src/memory.rs
richardpringle: Nope, then it would be `memory::memory`. It can either be `memory.rs` or `memory/mod.rs`, but I much prefer the former and it's the recommended way of doing things. The reason you see so many people using `mod.rs` is because that was the only way to do things before.

PR #1484, opened 2024-08-30
Add NewOptionWithConfig that auto-unmarshals config bytes into supplied config type

aaronbuchwald:
This PR adds a `NewOptionWithConfig` function and updates all options external to the VM package to use it.

Each option now supplies a namespace string, generic typed default config value, and an option function that will be called with the VM itself and the unmarshalled config value (unmarshals config bytes as JSON into the config provided by the caller).


vm/option.go
```diff
@@ -13,20 +16,38 @@ import (
 
 type RegisterFunc func(*VM)
 
-type OptionFunc func(*VM, []byte) error
+type OptionFunc func(vm *VM, configBytes []byte) error
```
joshua-kim, line 19: Is there ever a reason we want to expose this to the VM? Should we just hide this type within the package and force everyone to define a config per-option? Looks like we're using this for `WithManual` but I think we can just use a an empty struct for that config since it doesn't have one.

vm/option.go
```diff
@@ -13,20 +16,38 @@ import (
...
+
+type OptionConfigFunc[T any] func(vm *VM, config T) error
 
 type Option struct {
 	Namespace  string
 	OptionFunc OptionFunc
 }
 
-func NewOption(namespace string, optionFunc OptionFunc) Option {
+func NewOptionWithBytes(namespace string, optionFunc OptionFunc) Option {
```
joshua-kim, line 28: Looking to see if we can remove this so we can just have a single `NewOption` function signature out of this package since I find this naming confusing - ref earlier comment.

api/indexer/tx_indexer.go
```diff
@@ -32,11 +32,24 @@ var (
 	_ event.Subscription[*chain.StatelessBlock]        = (*txDBIndexer)(nil)
 )
 
+type Config struct {
+	Enabled bool `json:"enabled"`
+}
+
+func NewDefaultConfig() Config {
```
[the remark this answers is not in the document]
aaronbuchwald: I prefer `NewDefaultConfig` here and would use `NewConfig` if the function were supplying arguments to populate each field

vm/option.go
aaronbuchwald: We can un-export this, but the VM does need an option type that is ignorant of the generic type parameter. Then each option can call `NewOption[T]` with its own typed config and it receives a value of that type when it's executed.

aaronbuchwald: We could switch from `[]byte` to `interface{}`, so that it can unmarshal into that.

I'll just un-export this for now, but we do need a layer of indirection to hide the generic type from the VM.

PR #1485, opened 2024-08-30
Swap misnomer stateful and stateless blocks

aaronbuchwald:
As per the title this PR fixes the naming convention of stateful/stateless blocks. Previously, these names were in reverse of what they were actually doing ie. stateful block was stateless and stateless block was stateful.

Additionally, this PR moves what was previously called StatefulBlock functions next to the other StatefulBlock functions within `chain/block.go`.

### DIFF
--- a/extension/externalsubscriber/server.go
+++ b/extension/externalsubscriber/server.go
@@ -27,12 +27,12 @@ var (
 // TODO: switch to eventually using chain.Stateless block
 // Wrapper to pass blocks + results to subscribers
 type ExternalSubscriberSubscriptionData struct {
-	Blk     *chain.StatefulBlock
+	Blk     *chain.StatelessBlock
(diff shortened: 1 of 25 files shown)


extension/externalsubscriber/server.go
```diff
@@ -27,12 +27,12 @@ var (
 // TODO: switch to eventually using chain.Stateless block
```
RodrigoVillar, line 27:
```suggestion
// TODO: switch to eventually using chain.Stateful block
```

aaronbuchwald: This is as it should be. We should switch to using `chain.StatelessBlock` + `chain.Result` ideally in a single type like `ExecutedBlock`, so we don't need to define a type to forward the data here.

joshua-kim (approved):

PR #1486, opened 2024-08-30
Reduce e2e test runtime by reducing tx check interval and reducing StateSyncMinBlocks for e2e setup

aaronbuchwald:
This PR updates the existing functions with the syntax of `Wait[OptionalSuffix](ctx, ...)` to take an additional `time.Duration` argument that specifies the amount of time to wait between running a check condition.

This reduces wasted time during the e2e tests when we're checking for transaction confirmation (previously 500ms) even though once a transaction is produced, it takes ~10ms to build, distribute, and accept a block across the small local network used in tmpnet testing.

Additionally, this test sets `StateSyncMinBlocks` to 512 (avoiding getting too close to the validity window or state history length params) and reduces the number of blocks necessary to force a fall back to state sync from 1024 to 512 to reduce the state sync tests runtime.

### DIFF
--- a/api/indexer/client.go
+++ b/api/indexer/client.go
@@ -6,6 +6,7 @@ package indexer
 import (
 	"context"
 	"strings"
+	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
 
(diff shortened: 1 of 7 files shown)


richardpringle (approved):

richardpringle:
We can admin merge PRs with non-Rust changes like this one for now. 

I'll fix the Rust-CI first thing next week.

PR #1487, opened 2024-09-03
Add simple in-memory block index API

aaronbuchwald:
This PR adds a simple in-memory block index API using a FIFO cache.

This API can be accessed via:

```curl
curl -X POST --data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"blockindexer.getBlock",
    "params" : {
        "blockID": "TE4166k2gREpt4nkXddRpWGEFXTujpK3F9hGx9eTX8A3NNj5t"
    }
}' -H 'content-type:application/json;' 127.0.0.1:9650/ext/bc/morpheusvm/blockindexer

curl -X POST --data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"blockindexer.getBlockByHeight",
    "params" : {
        "height": 355
    }
}' -H 'content-type:application/json;' 127.0.0.1:9650/ext/bc/morpheusvm/blockindexer

curl -X POST --data '{
    "jsonrpc":"2.0",
    "id"     :1,
    "method" :"blockindexer.getLatestBlock",
    "params" : {}
}' -H 'content-type:application/json;' 127.0.0.1:9650/ext/bc/morpheusvm/blockindexer

{
  "jsonrpc": "2.0",
  "result": {
    "block": {
      "parent": "J5THWtbXdUuCFFfxztZWYpPSdsJvEsc28ZJC5n7v7b29uAiqj",
      "timestamp": 1725319871701,
      "height": 355,
      "txs": [],
      "stateRoot": "2J3oY52SgGNdrMMmcQ7TfrzXdVh8BpRdCQBi7CrjDENHDcQD7L"
    }
  },
  "id": 1
}
```

Which should produce output that looks like:

```json
{
  "jsonrpc": "2.0",
  "result": {
    "block": {
      "parent": "27zcMC2eVgFjJToo98aYqb3LQAgSvh9NZUC34GEWvwdpvr4zsw",
      "timestamp": 1725324073240,
      "height": 19,
      "txs": [
        {
          "base": {
            "timestamp": 1725324133000,
            "chainId": "2QSndD3aW5k2iKyuHphy2K2JnVa949qLGUsqHMGSKhikoR4NjW",
            "maxFee": 33400
          },
          "actions": [
            {
              "to": "001b265ade2dc23f7abbd1476f10e20aec8550eebdd2d65e616ceb71404a8b8770",
              "value": 1,
              "memo": ""
            }
          ],
          "auth": {
            "signer": [...], // ignore
            "signature": [...] // ignore
          }
        }
      ],
      "stateRoot": "23W3fHfPYWJ4Hq2pypL6rPwJ7yEPZyhMzWFc3PcH1gZAFrPMBE"
    }
  },
  "id": 1
}
```


README.md
```diff
@@ -51,7 +51,7 @@ Services are created by adding an [Option](./vm/option.go) to the VM. They can b
 ```golang
 // NewWithOptions returns a VM with the specified options
 func New(options ...vm.Option) (*vm.VM, error) {
-	options = append(options, With(), indexer.With()) // Add MorpheusVM API and Indexer
+	options = append(options, With(), txtxindexer.With()) // Add MorpheusVM API and Indexer
```
RodrigoVillar, line 54:
```suggestion
	options = append(options, With(), txindexer.With()) // Add MorpheusVM API and Indexer
```

Possible typo?

api/blockindexer/indexer.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+var (
+	_ event.SubscriptionFactory[*chain.StatefulBlock] = (*subscriptionFactory)(nil)
+	_ event.Subscription[*chain.StatefulBlock]        = (*indexer)(nil)
+)
+
+type Config struct {
+	Enabled             bool `json:"enabled"`
+	AcceptedBlockWindow int  `json:"acceptedBlockWindow"`
```
joshua-kim, line 29: 1. Why is the `blockindexer` separate from the `txindexer`? It seems reasonable to just have a single `indexer` component that can enable either/or indexing for txs/blocks.

2. Why is an in-memory indexer useful? If you have a downstream component reading from the indexer, any downtime in the downstream component can risk reliability loss.
    * If your downstream component is down for some time and > `acceptedBlockWindow` blocks are produced, you have no way of recovering the blocks lost between what downstream last read and the oldest block in the indexer.
    * Would it be more reasonable for the indexer be able to configure a period of time for which it keeps blocks on disk? If we did that we could just say that the downstream component is responsible for not falling behind the server's configured data retention.

api/blockindexer/indexer.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func NewDefaultConfig() Config {
+	return Config{
+		Enabled:             true,
+		AcceptedBlockWindow: 1_000,
+	}
+}
+
+func With() vm.Option {
```
joshua-kim, line 39: I don't know when we adopted this convention but I think `With` reads weirdly - i think `blockindexer.New()` reads a lot better.

api/blockindexer/indexer.go
```diff
@@ -0,0 +1,130 @@
...
+	_ event.SubscriptionFactory[*chain.StatefulBlock] = (*subscriptionFactory)(nil)
+	_ event.Subscription[*chain.StatefulBlock]        = (*indexer)(nil)
+)
+
+type Config struct {
+	Enabled             bool `json:"enabled"`
+	AcceptedBlockWindow int  `json:"acceptedBlockWindow"`
+}
+
+func NewDefaultConfig() Config {
```
joshua-kim, line 32: nit: call this `NewConfig`?

api/blockindexer/server.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/ava-labs/avalanchego/trace"
+
+	"github.com/ava-labs/hypersdk/api"
+	"github.com/ava-labs/hypersdk/chain"
+)
+
+const Endpoint = "/blockindexer"
+
+var (
+	ErrTxNotFound    = errors.New("tx not found")
```
joshua-kim, line 21: nit: unused

api/blockindexer/indexer.go
aaronbuchwald: 1. I thought it was cleaner to separate them, but can re-combine into one if you prefer
2. This API explicitly chooses not to provide reliability for blocks. This was added to provide the minimum API for https://github.com/ava-labs/hypersdk/issues/1513.

We talked about this previously, but the best argument imo for implementing a persistent rolling index is to support running it both in and out of the node. It seemed too strange to me that the VM cleans up blocks as it moves forward and then a separate API keeps its own persistent index.

api/blockindexer/indexer.go
aaronbuchwald: Just following the existing style. I think `With` is just an options convention, but open to switching all of them in a separate PR.

api/blockindexer/indexer.go
aaronbuchwald: Following the existing style here. I do prefer this as is though, `NewDefaultConfig` means no parameters all default values to me and `NewConfig` would take in parameters imo.

We only need `NewDefaultConfig` (at least for now) since we construct the actual config by creating the default and unmarshalling JSON into it.

aaronbuchwald:
> Implementation looks fine. Would it make sense to add tests exercising the new behavior to insure against regression?

The APIs are unfortunately not easily testable atm. Added a GitHub issue to fix this here: https://github.com/ava-labs/hypersdk/issues/1596, but don't think this PR is the place to start.

aaronbuchwald:
> > > Implementation looks fine. Would it make sense to add tests exercising the new behavior to insure against regression?
> > 
> > 
> > The APIs are unfortunately not easily testable atm. Added a GitHub issue to fix this here: #1596, but don't think this PR is the place to start.
> 
> The issue you link to talks about internal implementation details being the limiting factor, but given that we're talking about an API, what's stopping the addition of a golden-path e2e test?

There's a few reasons this is difficult w/ the current code.

I thought you meant writing a unit test, which is difficult because the indexer was coupled to the `chain.StatefulBlock` type, which requires a reference to the VM and a merkledb view. Went ahead and decoupled this by adding https://github.com/ava-labs/hypersdk/pull/1601 to separate out the `ExecutedBlock` type, which includes the results that are actually required from the VM.

For the e2e test, blocks hold a reference to transaction's actions and auth types, which are interfaces, so we can't unmarshal JSON into them. Sending JSON makes for the easiest to consume output for https://github.com/ava-labs/hypersdk/issues/1513, but doesn't work as well for writing an e2e test in go.

Went ahead and moved this to https://github.com/ava-labs/hypersdk/pull/1606 to include the tests based off the `ExecutedBlock` PR. Need to leave this branch as is until the minimal explorer switches over.

PR #1489, opened 2024-09-03
e2e: ignore failure testing only ignore when context has been cancelled

### LINKED ISSUE #1429: Fix State Sync until Cancel Ignores Errors
Update the state sync until cancel e2e test, so that it does not need to ignore errors from tx issuance to pass: https://github.com/ava-labs/hypersdk/blob/2b52fcb48cc1d5399779e2f9855249fe02714846/tests/e2e/e2e.go#L151.

najeal:
Closes #1429 

This modifies the `GenerateUntilCancel` function to use the text context and ignore the errors only when context has been cancelled.


aaronbuchwald:
Thanks for putting this up!

I'm not sure this is better than the original code and it could still be racey when it checks if the context returns an error.

Could we switch to using a channel and defer closing that channel rather than cancelling the context?

najeal:
GenerateUntilCancel renamed as GenerateUntilStop now expects a stop Channel.
The wrapping function creating the goroutine closes the channel to order the goroutine to stop. Then it waits for everything to be stopped properly before exiting the test.

tests/workload/transactions.go
```diff
@@ -107,42 +106,58 @@ func GenerateNBlocks(ctx context.Context, require *require.Assertions, uris []st
...
+	x.req.Errorf(str, args)
 }
 
-func (IgnoreFailureTestingT) FailNow() {}
+func (x IgnoreFailureTestingT) FailNow() {
+	if x.ctx.Err() != nil {
+		return
+	}
+	x.req.FailNow()
+}
```
aaronbuchwald, line 163: Could we remove this if it's no longer used?

tests/workload/transactions.go
```diff
@@ -107,42 +106,58 @@ func GenerateNBlocks(ctx context.Context, require *require.Assertions, uris []st
...
-func GenerateUntilCancel(
-	ctx context.Context,
+func GenerateUntilStop(
+	require *require.Assertions,
 	uris []string,
 	generator TxWorkloadIterator,
+	stopChannel <-chan struct{},
 ) {
 	submitClient := jsonrpc.NewJSONRPCClient(uris[0])
+	ctx := context.Background()
```
aaronbuchwald, line 116: Could we add `ctx` to this function signature and keep passing it in as before? We should just change from cancelling at the end of the function to terminating via the channel, so that we don't need to handle ctx errors

aaronbuchwald (changes requested):
Two small changes requested - after that and CI passes, this LGTM. Thanks for the contribution 🙏

aaronbuchwald (approved):

PR #1491, opened 2024-09-03
Move Call Function into Context

samliok:
Follow up from #1483. 

Changes
Moves `call_function` host call into the context. 
Removes the `Program` struct.


richardpringle (dismissed):

x/programs/wasmlanche/src/context.rs
```diff
@@ -6,17 +6,18 @@ extern crate alloc;
...
+use borsh::{BorshDeserialize, BorshSerialize};
+use displaydoc::Display;
 
 pub type CacheKey = Box<[u8]>;
 pub type CacheValue = Vec<u8>;
 
 /// Representation of the context that is passed to programs at runtime.
 pub struct Context {
-    program: Program,
+    contract_address: Address,
```
samliok, line 20: `contract_address` vs `address`?
```
  address: Address
  actor: Address
```

x/programs/wasmlanche/src/context.rs
```diff
@@ -40,29 +41,37 @@ mod debug {
     impl Debug for Context {
         fn fmt(&self, f: &mut Formatter<'_>) -> Result {
             let Self {
-                program,
+                contract_address,
```
richardpringle, line 44: Fine to leave this for now, but I'm thinking that we should call this `state_address` in the future. The contract code may change and while the code won't actually have an address, I think it would be good to think of contract execution as combining a state-address with a bytecode-id (hash) to perform execution

richardpringle (approved):

PR #1493, opened 2024-09-03
Add a VM with Contract support

dboehm-avalabs:
Create an example VM that has the ability to deploy and run wasm smart contracts.  This VM is based on the morpheus vm, but adds contract support.


examples/vmwithcontracts/actions/call.go
```diff
@@ -0,0 +1,133 @@
...
+	return mconsts.CallProgramID
+}
+
+func (t *Call) StateKeys(_ codec.Address, _ ids.ID) state.Keys {
+	result := state.Keys{}
+	for _, stateKeyPermission := range t.SpecifiedStateKeys {
+		result.Add(stateKeyPermission.Key, stateKeyPermission.Permission)
+	}
+	return result
+}
```
aaronbuchwald, line 57: Note: this only uses the last permissions value specified, which may produce unexpected behavior

examples/vmwithcontracts/actions/call.go
```diff
@@ -0,0 +1,133 @@
...
+	})
+	return [][]byte{result}, err
+}
+
+func (t *Call) ComputeUnits(chain.Rules) uint64 {
+	return t.Fuel / 1000
+}
+
+func (*Call) Size() int {
+	return codec.AddressLen + consts.Uint64Len
```
aaronbuchwald, line 93: We should include the size of the bytes themselves as well

examples/vmwithcontracts/actions/call.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+}
+
+func UnmarshalCallProgram(p *codec.Packer) (chain.Action, error) {
+	var callProgram Call
+	callProgram.Value = p.UnpackUint64(false)
+	callProgram.Fuel = p.UnpackUint64(true)
+	p.UnpackAddress(&callProgram.ContractAddress) // we do not verify the typeID is valid
+	callProgram.Function = p.UnpackString(true)
+	p.UnpackBytes(units.MiB, false, &callProgram.CallData)
```
aaronbuchwald, line 116: Can we move this to a constant above ie. `MaxCallDataSize`

examples/vmwithcontracts/actions/deploy.go
```diff
@@ -0,0 +1,89 @@
...
+		t.address = storage.GetAddressForDeploy(0, t.CreationInfo)
+	}
+	stateKey, _ := keys.Encode(storage.AccountProgramKey(t.address), 36)
+	return state.Keys{
+		string(stateKey): state.All,
+	}
+}
+
+func (*Deploy) StateKeysMaxChunks() []uint16 {
+	return []uint16{storage.BalanceChunks}
```
aaronbuchwald, line 46: Is this correct or leftover from MorpheusVM? If we are placing the contract bytecode at this key, then this should be `len(CreationInfo) / 64` no?

examples/vmwithcontracts/actions/deploy.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (t *Deploy) Execute(
+	ctx context.Context,
+	_ chain.Rules,
+	mu state.Mutable,
+	_ int64,
+	_ codec.Address,
+	_ ids.ID,
+) ([][]byte, error) {
+	result, err := (&storage.ProgramStateManager{Mutable: mu}).NewAccountWithProgram(ctx, t.ProgramID, t.CreationInfo)
```
aaronbuchwald, line 57: Let's separate this to two lines

examples/vmwithcontracts/actions/deploy.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (t *Deploy) Marshal(p *codec.Packer) {
+	p.PackBytes(t.ProgramID)
+	p.PackBytes(t.CreationInfo)
+}
+
+func UnmarshalDeployProgram(p *codec.Packer) (chain.Action, error) {
+	var deployProgram Deploy
+	p.UnpackBytes(36, true, (*[]byte)(&deployProgram.ProgramID))
+	p.UnpackBytes(10*units.MiB, false, &deployProgram.CreationInfo)
```
aaronbuchwald, line 77: The max size for a container sent over the network is 2MB, so we should decrease this to at most 1MB imo and set it as a constant ie. `MaxDeployContractBytes`

examples/vmwithcontracts/actions/deploy.go
```diff
@@ -0,0 +1,89 @@
...
+	deployProgram.address = storage.GetAddressForDeploy(0, deployProgram.CreationInfo)
```
aaronbuchwald, line 78: What's the meaning of the `0` passed in here? It seems we pass in `0` every instance where it's called

examples/vmwithcontracts/actions/publish.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return state.Keys{
+		string(t.id): state.Write | state.Allocate,
+	}
+}
+
+func (t *Publish) StateKeysMaxChunks() []uint16 {
+	return []uint16{uint16(len(t.ContractBytes))}
+}
```
aaronbuchwald, line 48: Need to divide by 64 to put this in terms of chunks

aaronbuchwald: Could we add a comment why this will never overflow as well?

examples/vmwithcontracts/actions/publish.go
```diff
@@ -0,0 +1,87 @@
...
+	return 4 + len(t.ContractBytes)
+}
+
+func (t *Publish) Marshal(p *codec.Packer) {
+	p.PackBytes(t.ContractBytes)
+}
+
+func UnmarshalPublishProgram(p *codec.Packer) (chain.Action, error) {
+	var publishProgram Publish
+	p.UnpackBytes(10*units.MiB, true, &publishProgram.ContractBytes)
```
aaronbuchwald, line 76: Let's use a constant for max size here

examples/vmwithcontracts/actions/publish.go
```diff
@@ -0,0 +1,87 @@
+// Copyright (C) 2023, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package actions
```
aaronbuchwald, line 4: Can we add unit tests for the newly added actions?

examples/vmwithcontracts/storage/recorder.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (r *Recorder) GetStateKeys() state.Keys {
+	result := state.Keys{}
+	for key := range r.ReadState {
+		result.Add(key, state.Read)
+	}
+	for key := range r.WriteState {
+		result.Add(key, state.Write|state.Allocate)
+	}
```
aaronbuchwald, line 59: Can we separate write/allocate by reading in advance of writing the value to check if it's already present?

examples/vmwithcontracts/storage/storage.go
```diff
@@ -0,0 +1,189 @@
...
+// 0x2/ (hypersdk-timestamp)
+// 0x3/ (hypersdk-fee)
+
+const (
+	// Active state
+	balancePrefix   = 0x0
+	heightPrefix    = 0x1
+	timestampPrefix = 0x2
+	feePrefix       = 0x3
+)
```
aaronbuchwald, line 39: Could we put all of the prefixes/keys here and update the state layout comment to include programs storage?

examples/vmwithcontracts/tests/integration/simple.wasm
aaronbuchwald, line 1: Could we switch away from including this binary file in source control?

examples/vmwithcontracts/actions/call.go
dboehm-avalabs: Not sure what you mean by this

examples/vmwithcontracts/actions/call.go
```diff
@@ -0,0 +1,133 @@
...
+		Timestamp:    uint64(timestamp),
+		Fuel:         t.Fuel,
+		Value:        t.Value,
+	})
+	return [][]byte{result}, err
+}
+
+func (t *Call) ComputeUnits(chain.Rules) uint64 {
+	return t.Fuel / 1000
+}
```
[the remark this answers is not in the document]
dboehm-avalabs: arbitrary right now, the fuel just gets very very large, very quickly.  /1000 gave more reasonable compute units.

examples/vmwithcontracts/actions/deploy.go
dboehm-avalabs: that is going to be an issue, program sizes get pretty big pretty quickly

dboehm-avalabs: Not for this particular case, but for publishing

examples/vmwithcontracts/actions/deploy.go
dboehm-avalabs: you need an address type for every address, I was just defaulting to 0

examples/vmwithcontracts/storage/recorder.go
dboehm-avalabs: we could, but it makes it less deterministic if we get some sort of error on the read.

examples/vmwithcontracts/actions/call.go
aaronbuchwald: if `Call` specifies 2 instances of the same state key with a different value for the `Permissions` field then during this iteration the first will be overwritten by the second. If we ever changed this in the future, it would be a breaking change.

dboehm-avalabs: ahh, sure.  This is usually going to be built from a map, but you could have duplicate keys.  It wouldn't overwrite, it would union.  Not sure that is any more expected, but it seems reasonable to me.

vm/vm.go
```diff
@@ -143,6 +143,10 @@ type VM struct {
 	stop  chan struct{}
 }
 
+func (vm *VM) ImmutableState(_ context.Context) (state.Immutable, error) {
+	return vm.State()
+}
```
aaronbuchwald, line 148: This seems to return a reference directly to merkledb, can we specify a single trie view of the last accepted state?

examples/vmwithcontracts/storage/recorder.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+	"errors"
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/utils/set"
+
+	"github.com/ava-labs/hypersdk/state"
+)
+
+type Recorder struct {
```
aaronbuchwald, line 16: (not for this PR) this would be a good utility function for the HyperSDK to provide

aaronbuchwald:
Could we rename from `vmwithcontracts` ? WASMVM (not in love with this) or ContractVM

Any thoughts @richardpringle @samliok ?

richardpringle:
> Could we rename from `vmwithcontracts` ? WASMVM (not in love with this) or ContractVM
> 
> Any thoughts @richardpringle @samliok ?

@aaronbuchwald, we could rename it, but we should be deliberate. One issue with `ProgramsVM` as I heard many times is that it led people to believe that it is a particular VM, instead of just a VM utilizing a set of libraries. Anyone should be able to build a VM with programs, they shouldn't need to "customize the ProgramsVM".

vm/vm.go
```diff
@@ -143,6 +143,10 @@ type VM struct {
 	stop  chan struct{}
 }
 
+func (vm *VM) ImmutableState(_ context.Context) (state.Immutable, error) {
```
aaronbuchwald, line 146:
```suggestion
// TODO: update VM ImmutableState to return an isolated trie view instead of
// reading directly from the on-disk view, which may change mid operation.
// ref. https://github.com/ava-labs/avalanchego/pull/3358/files
func (vm *VM) ImmutableState(_ context.Context) (state.Immutable, error) {
```

Could we also move this to `vm/resolutions.go` to follow the current code organization pattern?

examples/vmwithcontracts/actions/call.go
richardpringle: @aaronbuchwald, I can create an issue to fix this (ie check for uniqueness) after this PR is merged.

examples/vmwithcontracts/actions/deploy.go
richardpringle: @aaronbuchwald, I saw that this comment was resolved without explaining the resolution. Was there an offline resolution?

examples/vmwithcontracts/actions/deploy.go
richardpringle: Will create an issue to pop this out into a constant

examples/vmwithcontracts/actions/publish.go
richardpringle: Will create an issue for this

examples/vmwithcontracts/storage/recorder.go
richardpringle: What do you mean by "utility function"? You mean the whole struct?

vm/vm.go
richardpringle: Not sure what happened here but the suggestion has been applied.

vm/vm.go
richardpringle: I can create an issue for this

api/dependencies.go
```diff
@@ -36,4 +37,5 @@ type VM interface {
 	) (map[ids.NodeID]*validators.GetValidatorOutput, map[string]struct{})
 	GetVerifyAuth() bool
 	ReadState(ctx context.Context, keys [][]byte) ([][]byte, []error)
+	ImmutableState(ctx context.Context) (state.Immutable, error)
```
aaronbuchwald, line 40: Let's fix this: https://github.com/ava-labs/hypersdk/pull/1543

aaronbuchwald (approved):
LGTM to merge in, so we can follow up on the comments in this review.

I do think we should rename the example from `vmwithcontracts` to `contractvm`.

We may want people to use contracts as a part of their VM and not just the only way to interact with their VM, but this example is focused on contracts regardless. Either way - we need a better name than `vmwithcontracts`

vm/vm.go
richardpringle: resolved with your PR https://github.com/ava-labs/hypersdk/pull/1543

examples/vmwithcontracts/storage/recorder.go
richardpringle: https://github.com/ava-labs/hypersdk/issues/1552

richardpringle:
> LGTM to merge in, so we can follow up on the comments in this review.
> 
> I do think we should rename the example from `vmwithcontracts` to `contractvm`.
> 
> We may want people to use contracts as a part of their VM and not just the only way to interact with their VM, but this example is focused on contracts regardless. Either way - we need a better name than `vmwithcontracts`

PR #1497, opened 2024-09-04
Adding Setup Actions to `chaintest`

RodrigoVillar:
This PR extends `ActionTest` by adding setup actions; these actions are executed prior to the action being tested.

For complex VMs, aiming for high coverage via integration tests of actions might require for these actions to be tested in a particular state; while this can be done by modifying the state passed into `ActionTest`, this inherently leads to duplicate code. Adding setup actions allows for developers to easily set up the state required for testing a particular action

### DIFF
--- a/chaintest/action_test_helpers.go
+++ b/chaintest/action_test_helpers.go
@@ -47,13 +47,17 @@ func (i *InMemoryStore) Remove(_ context.Context, key []byte) error {
 	return nil
 }
 
-// ActionTest is a single parameterized test. It calls Execute on the action with the passed parameters
+// ActionTest is a single parameterized test. It calls Execute on the last
+// action in Actions with the passed parameters
 // and checks that all assertions pass.
 type ActionTest struct {
 	Name string
 
-	SetupActions []chain.Action
-	Action       chain.Action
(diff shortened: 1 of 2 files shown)


chaintest/action_test_helpers.go
```diff
@@ -71,6 +72,11 @@ func (test *ActionTest) Run(ctx context.Context, t *testing.T) {
 	t.Run(test.Name, func(t *testing.T) {
 		require := require.New(t)
 
+		for _, act := range test.SetupActions {
+			_, err := act.Execute(ctx, test.Rules, test.State, test.Timestamp, test.Actor, test.ActionID)
+			require.NoError(err)
+		}
+
```
aaronbuchwald, line 79: Is it necessary to separate between `SetupActions` and `Action`? Could we just change `Action` to a slice of actions and execute them sequentially?

RodrigoVillar: I am definitely for that! From DevX pov, I thought it would be better to label these actions as `SetupActions` since their correctness isn't being tested (whereas `Action` is being tested for correctness). However, we can just make `Action` an array and make the developer responsible for its usage (via reading the documentation).

PR #1498, opened 2024-09-04
Consolidate contexts

richardpringle:
- **Use feature flags for bindings**
- **Consolidate contexts**


x/programs/examples/automated-market-maker/src/lib.rs
richardpringle, line 1: This doesn't look amazing at the moment, but we'll make this better with some of the other changes I have planned

x/programs/sdk-macros/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 proc-macro = true
 
 [dependencies]
+cfg-if = "1.0.0"
```
richardpringle, line 10:
```suggestion
```

PR #1501, opened 2024-09-04
Use `cross` to run `wasmlanche` unit tests on a wasm-target

richardpringle:
Add `cross` support here to make sure the tests are run in CI

### DIFF
--- a/x/contracts/wasmlanche/src/memory/allocations.rs
+++ b/x/contracts/wasmlanche/src/memory/allocations.rs
@@ -175,20 +175,22 @@ cfg_if! {
             #[test]
             #[should_panic(expected = "already accessed")]
             fn panic_on_concurrent_access() {
-                static SINGLETON: SingletonLenMap = SingletonLenMap::new();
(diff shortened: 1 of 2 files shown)


x/contracts/wasmlanche/src/memory/allocations.rs
```diff
@@ -175,20 +175,23 @@ cfg_if! {
...
-                    // run enough times to make sure there's a collision
-                    for _ in 0..10000 {
-                        SINGLETON.with_borrow(|map| {
-                            assert!(map.is_empty());
-                        });
+                fn test_with_delay(duration: Duration) -> impl FnOnce() {
+                    static SINGLETON: SingletonLenMap = SingletonLenMap::new();
+
+                    move || {
+                        std::thread::sleep(duration);
```
samliok, line 182: i guess this allows us to test better in multithreaded environment, but wouldn't be relevant when running in our contracts. Why add then?

samliok (dismissed):

PR #1504, opened 2024-09-04
Update package structure

joshua-kim:
Creates an `internal` package to hide code that isn't a part of the public hypersdk api


aaronbuchwald:
Would it make sense to move:

- builder
- gossiper
- network
- pebble
- workers

to internal as well since they are not used in MorpheusVM and not necessary to export?

We should also move tstate out of internal since we do use this in the action transfer tests.

Separate from this PR, I think we could also trim down what's included in utils and storage packages, so that there's nothing exported that is not necessary.

Also, the chain and vm packages are very large now and they are exported because they are required by VMs, but we could consider separating them out and moving the parts that are not required by VM implementers to internal.

Overall, I think it makes sense to move anything that we don't NEED to export to VM implementers to internal to make it more explicit what can be used as a library and try to move from internal to the public API as needed. Good idea 🙏

joshua-kim:
> Also, the chain and vm packages are very large now and they are exported because they are required by VMs, but we could consider separating them out and moving the parts that are not required by VM implementers to internal.

Yeah I want to do this in a follow-up to make review easier. I thought the simple file renames would be easiest to review first.

aaronbuchwald (approved):

PR #1505, opened 2024-09-05
Wrap Host Function Calls

samliok:
wraps host functions so we can mock them during unit tests.


x/programs/wasmlanche/src/context.rs
```diff
@@ -209,23 +245,11 @@ impl Context {
...
-            #[cfg(feature = "bindings")]
-            Context::External(_) => panic!("not supported"),
-        }
-
+    pub fn deploy(&mut self, program_id: ProgramId, account_creation_data: &[u8]) -> Address {
         let ptr =
             borsh::to_vec(&(program_id, account_creation_data)).expect("failed to serialize args");
 
-        let bytes = unsafe { deploy(ptr.as_ptr(), ptr.len()) };
+        let bytes = self.host_accessor_mut().deploy(ptr.as_ptr(), ptr.len());
```
richardpringle, line 252: 🤔, the `host_accessor` should be immutable. We can use interior mutability in `cfg(feature = "test")`

x/programs/wasmlanche/src/memory.rs
```diff
@@ -64,12 +64,25 @@ impl HostPtr {
...
+    }
+}
+
 /// Allocate memory into the instance of Program and return the offset to the
 /// start of the block.
 /// # Panics
 /// Panics if the pointer exceeds the maximum size of an isize or that the allocated memory is null.
 #[no_mangle]
-extern "C" fn alloc(len: usize) -> HostPtr {
+pub extern "C" fn alloc(len: usize) -> HostPtr {
```
richardpringle, line 85:
```suggestion
pub(crate) extern "C" fn alloc(len: usize) -> HostPtr {
```

x/programs/wasmlanche/src/types.rs
```diff
@@ -16,7 +16,7 @@ pub type Id = [u8; ID_LEN];
 pub type Gas = u64;
 
 /// The ID bytes of a program.
-#[derive(BorshSerialize, BorshDeserialize)]
+#[derive(BorshSerialize, BorshDeserialize, Clone)]
```
richardpringle, line 19: Where is the `Clone` needed?

x/programs/wasmlanche/src/types.rs
```diff
@@ -25,6 +25,14 @@ impl From<Box<[u8]>> for ProgramId {
     }
 }
 
+#[cfg(feature = "test")]
+impl ProgramId {
+    #[must_use]
+    pub fn new(bytes: Box<[u8]>) -> Self {
+        Self(bytes)
+    }
+}
```
richardpringle, line 34: This should be a `From<Box<[u8]>> for ProgramId`. Prefer `From` implementation over singe argument constructors.

x/programs/wasmlanche/src/host.rs
```diff
@@ -0,0 +1,258 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+use crate::HostPtr;
+use cfg_if::cfg_if;
+
+#[cfg(feature = "test")]
```
richardpringle, line 7: Can you throw all the `cfg(feature = "test")` stuff into an inline `mod`, then do a:
```rust
#[cfg(feature = "test")]
mod test_stuff {
    // please pick a better name
    // ...
}

#[cfg(feature = "test")]
pub use test_stuff::*;
```

x/programs/wasmlanche/src/host.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        host_ptr
+    }
+
+    #[allow(clippy::unused_self)]
+    pub fn get_remaining_fuel(&self) -> HostPtr {
+        panic!("get_remaining_fuel not implemented in the test context");
+    }
+
+    pub fn send_value(&self, ptr: *const u8, len: usize) -> HostPtr {
```
richardpringle, line 178: All of these methods are unsafe. We should be passing in the proper types and converting them to pointers inside these methods.

x/programs/wasmlanche/src/context.rs
```diff
@@ -350,6 +340,17 @@ impl Context {
             Context::External(ctx) => ctx.call_function(function_name, args),
         }
     }
+
+    #[cfg(feature = "bindings")]
+    #[must_use]
+    pub fn new_external_call_context(&self, address: Address, max_units: Gas, value: u64) -> Self {
+        Self::External(ExternalCallContext::new(
+            self.host_accessor().clone(),
```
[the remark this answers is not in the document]
samliok: cloning is actually fine here since `ExternalCallContext` can only call into `call_function` and won't have access to state in unit tests

x/programs/wasmlanche/src/context.rs
```diff
@@ -468,3 +456,21 @@ struct CallContractArgs<'a> {
...
+        max_value: u64,
+    ) -> Self {
+        Self {
+            target,
+            function: function.as_bytes(),
+            args,
+            max_units,
+            max_value,
+        }
+    }
```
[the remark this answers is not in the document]
samliok: it's used more than once in my follow up pr

x/programs/wasmlanche/src/host.rs
samliok: hmm, we need to keep the function signatures the same for both compiler directives tho?

samliok: nvm, see what ur saying will update

x/programs/wasmlanche/src/context.rs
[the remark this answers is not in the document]
richardpringle: As discussed offline, `ExternalCallContext` should take a mutable borrow to `Context` instead of being part of the enum. Then we can write a wrapper function for `bindings` compilations that takes the `ExternalCallContext` but uses the mutable borrow of `Context` to actually perform the function call. 

I'll start working on that after this lands.

x/programs/wasmlanche/src/context.rs
```diff
@@ -433,29 +408,18 @@ impl ExternalCallContext {
...
-        args,
-        max_units,
-        max_value,
-    };
+    let args = CallContractArgs::new(&address, function_name, args, max_units, max_value);
 
     let args_bytes = borsh::to_vec(&args).expect("failed to serialize args");
 
-    let bytes = unsafe { call_program(args_bytes.as_ptr(), args_bytes.len()) };
+    let bytes = host_accessor.call_program(&args_bytes);
```
richardpringle, line 422: 👍

x/programs/wasmlanche/src/host.rs
```diff
@@ -0,0 +1,253 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+pub struct StateAccessor;
+
+#[cfg(feature = "test")]
```
richardpringle, line 6: Much nicer

x/programs/wasmlanche/src/memory.rs
```diff
@@ -64,12 +64,25 @@ impl HostPtr {
...
+    }
+}
+
 /// Allocate memory into the instance of Program and return the offset to the
 /// start of the block.
 /// # Panics
 /// Panics if the pointer exceeds the maximum size of an isize or that the allocated memory is null.
 #[no_mangle]
-extern "C" fn alloc(len: usize) -> HostPtr {
+pub(crate) extern "C" fn alloc(len: usize) -> HostPtr {
```
richardpringle: 🤔 I thought I just changed this. This diff doesn't look correct

richardpringle (approved):

PR #1528, opened 2024-09-09
Add a set balance mocking function

samliok:


x/programs/test/programs/unit_test/Cargo.toml
samliok, line 1: thinking we could have a unit testing crate that would test our context mocks and other stuff compiled with `#[test]`

x/programs/test/programs/test_directive/Cargo.toml
samliok, line 1: thinking this crate would be for testing `wasmlanche` when compiled with the `test` feature

richardpringle: Not sure I understand? Usually crates in the /x/programs/test/ directory are for testing the runtime.

samliok: hmm yea, what would be a better place to put this? in the `tests` folder of wasmlanche?

richardpringle: Yeah, there's already a test-crate in there

richardpringle (approved):

Issue #1529, opened 2024-09-10
Avoid using internal packages in the examples folder


containerman17:
If you use an internal package in an example VM, you won't be able to copy and run the example independently.

https://github.com/ava-labs/hypersdk/blob/7ad357b38ab5042ac6d066622def39e4944d3b32/examples/morpheusvm/tests/workload/workload.go#L25-L27

containerman17:
@joshua-kim 
This is actually a blocker for https://github.com/ava-labs/hypersdk-starter RN

containerman17:
- [ ] Bring back fees package to root
- [ ] Add a linter checking for internal packages use in examples folder

PR #1530, opened 2024-09-10
Bring package fees back to root

containerman17:
Reference: #1529


fees/manager.go
joshua-kim, line 1: I would prefer if we could keep the struct + functions on `Manager` in this internal package but just move out the definition for `Dimensions` + related static functions.

joshua-kim (approved):

PR #1531, opened 2024-09-10
Compile Contract Bytes Once in Benchmarks

samliok:
Our `testruntime` would compile contracts on every call. While compilation from RUST -> WASM is cached by `cargo`, this flow is not representative of what should happen in a VM when a contract is called.


richardpringle (changes requested):
A bit of a nit, but it'll save us time in the future if we have to make more changes. `SetProgramBytes` is doing too much. It should really be called `CompileProgramAndSetBytes` to accurately describe what it's doing. But instead, I think is would be better to have a `(C|c)ompileProgram` method that returns bytes. Then `SetProgramBytes` can actually take bytes as an argument. 

That way, if in the future we decide to cache the bytes or anything like that, it's clear (at the callsite) what changed.

richardpringle (approved):

richardpringle (approved):

PR #1533, opened 2024-09-10
Expose Program as Child of Runtime

samliok:
programs should be children of runtime as a runtime can hold multiple programs

### DIFF
--- a/x/programs/runtime/util_test.go
+++ b/x/programs/runtime/util_test.go
@@ -189,16 +189,14 @@ func (t *testRuntime) WithValue(value uint64) *testRuntime {
 	return t
 }
 
-// AddProgram compiles [programName] and sets the bytes and account in the state manager
-func (t *testRuntime) AddProgram(programID ProgramID, account codec.Address, programName string) error {
-	err := t.StateManager.(TestStateManager).CompileAndSetProgram(programID, programName)
-	if err != nil {
-		return err
-	}
+// AddProgram compiles [programName] and sets the bytes in the state manager
+func (t *testRuntime) AddProgram(programID ProgramID, programName string) error {
+	return t.StateManager.(TestStateManager).CompileAndSetProgram(programID, programName)
+}
 
+// AddProgramInstance sets creates a new account with the program [programID]
+func (t *testRuntime) AddProgramInstance(programID ProgramID, account codec.Address) {
 	t.StateManager.(TestStateManager).AccountMap[account] = string(programID)
-
-	return nil
 }
 
 func (t *testRuntime) CallProgram(program codec.Address, function string, params ...interface{}) ([]byte, error) {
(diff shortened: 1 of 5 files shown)


x/programs/runtime/util_test.go
```diff
@@ -205,36 +210,42 @@ func (t *testRuntime) CallProgram(program codec.Address, function string, params
...
-				logging.NoLog{}).WithDefaults(CallInfo{Fuel: 10000000}),
-			StateManager: TestStateManager{
-				ProgramsMap: map[string][]byte{},
-				AccountMap:  map[codec.Address]string{account: stringedID},
-				Balances:    map[codec.Address]uint64{},
-				Mu:          test.NewTestDB(),
-			},
-		},
 		Address: account,
+		Runtime: t,
```
samliok, line 234: this hurt my rust brain @richardpringle

x/programs/runtime/util_test.go
```diff
@@ -205,36 +210,42 @@ func (t *testRuntime) CallProgram(program codec.Address, function string, params
 		})
 }
 
-func newTestProgram(ctx context.Context, program string) (*testProgram, error) {
+func newTestRuntime(ctx context.Context) *testRuntime {
+	return &testRuntime{
+		Context: ctx,
+		callContext: NewRuntime(
```
richardpringle, line 216: I know this might not be your code, but why is this called `callContext`?

x/programs/runtime/util_test.go
```diff
@@ -205,36 +210,42 @@ func (t *testRuntime) CallProgram(program codec.Address, function string, params
...
+			NewConfig(),
+			logging.NoLog{}).WithDefaults(CallInfo{Fuel: 10000000}),
+		StateManager: TestStateManager{
+			ProgramsMap: map[string][]byte{},
+			AccountMap:  map[codec.Address]string{},
+			Balances:    map[codec.Address]uint64{},
+			Mu:          test.NewTestDB(),
```
richardpringle, line 223: This is also a terrible name 😞, why not just call it `Db`?

x/programs/runtime/util_test.go
```diff
@@ -194,6 +194,11 @@ func (t *testRuntime) AddProgram(programID ProgramID, programName string) error
 	return t.StateManager.(TestStateManager).CompileAndSetProgram(programID, programName)
 }
 
+// AddProgramInstance creates a new account with the program [programID]
+func (t *testRuntime) AddProgramInstance(programID ProgramID, account codec.Address) {
+	t.StateManager.(TestStateManager).AccountMap[account] = string(programID)
+}
```
richardpringle, line 200: This looks like a footgun for creating inconsistent state. This code is only used in the private context, right? Can we just inline it?

x/programs/runtime/util_test.go
```diff
@@ -205,36 +210,42 @@ func (t *testRuntime) CallProgram(program codec.Address, function string, params
...
 		Address: account,
+		Runtime: t,
 	}
-	err := testProgram.Runtime.AddProgram(ProgramID(stringedID), program)
+
+	err := t.AddProgram(ProgramID(stringedID), program)
 	if err != nil {
 		return nil, err
 	}
+	t.AddProgramInstance(ProgramID(stringedID), account)
```
richardpringle, line 241: I think you can add this logic inside `AddProgram`. I know I said something about functions only doing one thing, but it makes sense to update the programs and accounts maps at the same time.

richardpringle (changes requested):
Haha, don't worry about my comments about field names, just need to inline adding the program-id to the account-map.

richardpringle (approved):

PR #1535, opened 2024-09-11
Benchmark Runtime

samliok:
- Benchmarks a set of example smart contracts.
- Adds `ID` field to `testContract`
- Removes unnecessary `Context.WithCancel` calls

## Basic Contract
This example contract immediately returns 0

```
BenchmarkRuntimeCallProgramBasic-10    	   22760	     65834 ns/op	    2474 B/op	     214 allocs/op
BenchmarkRuntimeCallProgramBasic-10    	   19663	     59600 ns/op	    2480 B/op	     214 allocs/op

```

## Send Native Asset
This example contract sends native balance to another address. This performs a host function call.

```
BenchmarkRuntimeSendValue-10    	   16328	     86647 ns/op	    4415 B/op	     460 allocs/op
BenchmarkRuntimeSendValue-10    	   15646	     73519 ns/op	    4418 B/op	     460 allocs/op
```

## Basic External Call
This example contract calls out to an external contract, thereby spinning up a new `wastime` runtime.

```
BenchmarkRuntimeBasicExternalCalls-10    	    7722	    151352 ns/op	   10485 B/op	    1112 allocs/op
BenchmarkRuntimeBasicExternalCalls-10    	    7597	    146328 ns/op	   10488 B/op	    1112 allocs/op
```

## Swap AMM Token
Performs an AMM swap, calling out to many host function calls as well as external contracts.

```
BenchmarkAmmSwaps-10    	    1456	    943331 ns/op	   87034 B/op	   10448 allocs/op
BenchmarkAmmSwaps-10    	    1621	    835125 ns/op	   87046 B/op	   10450 allocs/op
```

## Short Summary
Spinning up a `wasmtime` runtime is a much larger overhead. Each external call spins up a new runtime so it makes sense the amm is roughly 6 times slower since it has 6 more external calls.


aaronbuchwald (changes requested):
Could we add install instructions to make sure that we can run these locally from a fresh machine?

Also, thoughts on adding benchmarks into CI either with the count set to 1-10 or or as a cronjob, so that we can see if any changes cause them to break?

x/contracts/test/util.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func CompileTest(contractName string) error {
-	cmd := exec.Command("cargo", "build", "-p", contractName, "--release", "--target=wasm32-unknown-unknown", "--target-dir=./target")
+	cmd := exec.Command("cargo", "build", "-p", contractName, "--release", "--target-dir=./target")
```
richardpringle, line 15: Why did you remove --target?

samliok: if i included the --target then our build script wouldn't activate and thus not properly create the `.wasm` file for the examples

samliok:
> Could we add install instructions to make sure that we can run these locally from a fresh machine?

created #1562

richardpringle (changes requested):
Don't worry about the nit if you don't want, but you should get rid of the `build.rs` changes.

richardpringle (approved):

aaronbuchwald (approved):

PR #1539, opened 2024-09-11
Benchmark Morpheus VM Transfer Action

samliok:
Adds a benchmark for the transfer action
```
BenchmarkSimpleTransfer/SimpleTransferBenchmark-10         	  502368	      2367 ns/op	     304 B/op	       8 allocs/op
```


chain/chaintest/action_test_helpers.go
```diff
@@ -81,3 +82,20 @@ func (test *ActionTest) Run(ctx context.Context, t *testing.T) {
 		}
 	})
 }
+
+// RunBenchmark executes the [ActionTest] and make sure all the benchmark assertions pass.
```
aaronbuchwald, line 86: We're using this in a different way here ie. ActionTest does not assume that the state will be re-used `b.N` times.

Could we switch to using a separate type here? We may need to switch to a `CreateState` function so that each benchmark has its own state instead of mutating the same one over and over. If a given benchmark can re-use the same state, it can return the same one from that function.

chain/chaintest/action_test_helpers.go
```diff
@@ -81,3 +82,20 @@ func (test *ActionTest) Run(ctx context.Context, t *testing.T) {
...
+func (test *ActionTest) RunBenchmark(ctx context.Context, b *testing.B) {
+	b.Run(test.Name, func(b *testing.B) {
```
aaronbuchwald, line 88: Can we get rid of `b.Run` here? This is typically used to run subtests / sub-benchmarks, but we're not doing that in this case.

chain/chaintest/action_test_helpers.go
```diff
@@ -81,3 +81,41 @@ func (test *ActionTest) Run(ctx context.Context, t *testing.T) {
...
+}
+
+// Run executes the [ActionBenchmark] and make sure all the benchmark assertions pass.
+func (test *ActionBenchmark) Run(ctx context.Context, b *testing.B) {
+	require := require.New(b)
+
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		state := test.CreateState()
+		b.StartTimer()
```
aaronbuchwald, line 111: To avoid doing this in the loop, could we create a slice of `b.N` states ahead of time?

We're going to be a little bit prone to issues like this: https://github.com/golang/go/issues/20875

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -124,3 +125,41 @@ func TestTransferAction(t *testing.T) {
...
+
+	transferActionTest := &chaintest.ActionBenchmark{
+		Name:  "SimpleTransferBenchmark",
+		Actor: from,
+		Action: &Transfer{
+			To:    to,
+			Value: 1,
+		},
+		CreateState: func() state.Mutable {
+			keys := make(state.Keys)
```
aaronbuchwald, line 144: It looks like this local variable is not used by anything that actually gets returned. This seems to be copy-pasted from the test above. Could you remove it from here and the test above? Running locally confirms it's not needed for the tests to pass.

chain/chaintest/action_test_helpers.go
```diff
@@ -81,3 +81,41 @@ func (test *ActionTest) Run(ctx context.Context, t *testing.T) {
 		}
 	})
 }
+
+// ActionBenchmark is paramatized benchmark. It calls Execute on the action with the passed parameters
+// and checks that all assertions pass. To avoid using shared state between runs, a new
+// state is created for each iteration using the provided `CreateState` function.
```
aaronbuchwald, line 87: nit: `is a parameterized`

aaronbuchwald (approved):

PR #1540, opened 2024-09-11
Ignore `target` dir when checking for license

richardpringle:
This is necessary for running the script locally


joshua-kim (approved):

containerman17:
The PR is breaking the license header check, and now it's giving a file name too long error. This needs to be fixed.

PR #1541, opened 2024-09-12
Optionally test subdirectories

richardpringle:
It should be easy to test a specific package. Not sure if there's a better way to do this, open to suggestions


aaronbuchwald:
You can run:

```sh
go test ./chain/...
```
to test the chain package and all of its subdirectories.

There are specific directories where we may need the constants set by the script to be set correctly (this is for the blst package), but this is the exception not the rule.

PR #1543, opened 2024-09-12
Fix immutable state implementation

aaronbuchwald:
This PR updates the `ImmutableState` function so that:
- it returns a view of the last accepted state that becomes invalid when the state changes instead of silently allowing corruption
- Moves the implementation to `resolutions.go`

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -144,13 +144,6 @@ type VM struct {
 	stop  chan struct{}
 }
 
-// TODO: update VM ImmutableState to return an isolated trie view instead of
-// reading directly from the on-disk view, which may change mid operation.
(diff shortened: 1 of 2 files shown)


vm/vm.go
```diff
@@ -144,13 +144,6 @@ type VM struct {
 	stop  chan struct{}
 }
 
-// TODO: update VM ImmutableState to return an isolated trie view instead of
-// reading directly from the on-disk view, which may change mid operation.
-// ref. https://github.com/ava-labs/avalanchego/pull/3358/files
-func (vm *VM) ImmutableState(_ context.Context) (state.Immutable, error) {
```
joshua-kim, line 150: Is it possible to add a unit test for this bug?

richardpringle (approved):

PR #1558, opened 2024-09-16
Move state/tstate, keys and pubsub packages out of internal

### LINKED ISSUE #1557: Use of internal/pubsub in examples folder
Here is a list of packages currently used in the example VMs. The issue is that internal packages cannot be accessed outside of the repository due to compiler restrictions.

- `github.com/ava-labs/hypersdk/internal/state/tstate`
- `github.com/ava-labs/hypersdk/internal/keys`
- `github.com/ava-labs/hypersdk/internal/pubsub`

To resolve this, we either need to move these packages out of the internal folder or extract their functionality or interfaces elsewhere. Moving the packages seems like the simplest solution.

containerman17:
Closes #1557

Adds a simple linter
```bash
if grep -r "github.com/ava-labs/hypersdk/internal/" ./examples; then
  echo "Error: example VMs should not use internal packages"
  exit 1
fi
```


containerman17:
The manual linter test passed. You can check the result [here](https://github.com/ava-labs/hypersdk/actions/runs/10881269967/job/30189857893?pr=1558).

scripts/lint.sh
```diff
@@ -66,4 +66,10 @@ for test in $TESTS; do
   run "${test}" "${TARGET}"
 done
 
+if grep -r "github.com/ava-labs/hypersdk/internal/" ./examples; then
+  echo "Error: example VMs should not use internal packages"
+  exit 1
+fi
+
```
aaronbuchwald, line 73: Could you add this as a function included in `$TESTS` so that it matches the style of the rest of the linters?

joshua-kim: +1, we should be able to update our `golangci.yml`'s `depguard` config to blacklist this package from the `examples` dir ([ref](https://github.com/OpenPeeDeeP/depguard))

containerman17: I've added a rule to depcheck, but I'm having trouble testing how it works. It's not working for me at all. Will try to come back to it later today

aaronbuchwald:
Gonna go ahead and merge this in to fix the package structure and we can fix the linter separately

aaronbuchwald (approved):

PR #1562, opened 2024-09-16
Add Install Instructions to Example Contract

samliok:


richardpringle (approved):
@samliok, I think it's also worth mentioning that you need `clang` for `cgo`. Not sure if there are any other dependencies

PR #1563, opened 2024-09-16
Run Benchmarks on CI

samliok:
runs all benchmarks with a `benchtime` of 1 to make sure any new pushes didn't break benchmarks


richardpringle (commented):
I think we really only want to run benchmarks on merges to main. There's no pass/fail except for having the benchmarks running longer than the timeout. 

I think it would be better to have a manual trigger for a PR branch if you care about the benchmark results for a particular PR (can use the `workflow_dispatch` event to accomplish this). 

Or we could use Github comments to trigger the workflow on a PR if we want ([example](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#issue_comment))

samliok:
>  There's no pass/fail except for having the benchmarks running longer than the timeout.

Benchmarks pass/fail the same as unit tests? Calling with the `benchtime` arg means the benchmark only runs once essentially treating it like a unit test, so if it runs longer than the timeout something is wrong.


>  if you care about the benchmark results for a particular PR

this action isn't used for getting benchmark results, it's more-so a sanity check making sure your changes didn't unexpectedly break tests.

richardpringle (approved):

Issue #1527, opened 2024-09-09
Switch from Bech32 Address to Hex Address Format


aaronbuchwald:
Bech32 requires passing around an HRP for a small UI "improvement"

We should switch to a simpler hex address format that does not require an HRP

Bech32 address ex: `morpheus1qrzvk4zlwj9zsacqgtufx7zvapd3quufqpxk5rsdd4633m4wz2fdjk97rwu`
Hex address: `0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC`

PR #1567, opened 2024-09-18
Switch from bech32 to hex address format

aaronbuchwald:
This PR switches from bech32 to a hex address format. Closes https://github.com/ava-labs/hypersdk/issues/1527.

Bech32 address format was intended to provide a UI improvement, however, I find it creates a worse UX and DevX.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -15,11 +15,9 @@ type Controller interface {
 	DatabasePath() string
 	Symbol() string
 	Decimals() uint8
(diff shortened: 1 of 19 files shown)


aaronbuchwald:
Note: the hex address used in this PR does not include a checksum. We should add a checksum (ideally similar to EIP55), but need to make sure there are no licensing issues.

joshua-kim (approved):

PR #1574, opened 2024-09-18
Separate proposer monitor from vm into internal package

aaronbuchwald:
This PR separates the proposer monitor from the VM package into a separate internal package.

The proposer monitor provides a layer of abstraction between the HyperSDK's view of its validator set and what's provided by AvalancheGo. This is useful when adding epoch'ing of the validator set and could be used to enable alternative ways to manage the validator set than using AvalancheGo in the future.

Removed all other usage of `snowCtx.ValidatorsState` for now (it was unused, but present in `vm/resolutions.go` and declared as a dependency even though it was not used in `chain/dependencies.go`. Ideally, we'd move away from `*snow.Context` so that we're sure we'll treat the proposer monitor (or an alternative implementation) as the single source of truth.


containerman17:
LGTM as far as I comprehend

containerman17 (approved):

PR #1575, opened 2024-09-19
Move verify context from vm into chain package

aaronbuchwald:
This PR moves `vm/verify_context.go` from the `vm/` package into `chain/`. This avoids a back and forth pattern between the chain and vm packages.

Previously the flow was:

1. Call `vm.GetVerifyContext`
2. vm retrieves a stateful block and decides whether it needs to trace tx conflicts through processing blocks or the accepted chain
3. The verification context was responsible for calling functions on both the stateful block type and on the underlying vm

The new flow moves this directly to `chain.StatefulBlock` which uses the VM only to retrieve details about the last accepted state or retrieve an already verified block.

The responsibilities between chain and VM are still blurred, but this removes one case.


containerman17:
LGTM

PR #1582, opened 2024-09-20
Add support to define VM runtime

joshua-kim:
Adds support to define action runtimes. This can be useful to avoid having to maintain runtime state inside of actions. In the future will expand this interface to support rollbacks/child views.

### DIFF
--- a/examples/vmwithcontracts/actions/call.go
+++ b/examples/vmwithcontracts/actions/call.go
@@ -19,7 +19,7 @@ import (
 	mconsts "github.com/ava-labs/hypersdk/examples/vmwithcontracts/consts"
 )
 
-var _ chain.Action = (*Call)(nil)
+var _ chain.Action[struct{}] = (*Call)(nil)
 
 const MaxCallDataSize = units.MiB
 
@@ -70,8 +70,7 @@ func (t *Call) StateKeysMaxChunks() []uint16 {
 
 func (t *Call) Execute(
 	ctx context.Context,
-	_ chain.Rules,
-	mu state.Mutable,
+	rt chain.Runtime[struct{}],
 	timestamp int64,
 	actor codec.Address,
 	_ ids.ID,
@@ -79,7 +78,7 @@ func (t *Call) Execute(
 	resutBytes, err := t.r.CallContract(ctx, &runtime.CallInfo{
 		Contract:     t.ContractAddress,
 		Actor:        actor,
-		State:        &storage.ContractStateManager{Mutable: mu},
+		State:        &storage.ContractStateManager{Mutable: rt.State},
 		FunctionName: t.Function,
 		Params:       t.CallData,
 		Timestamp:    uint64(timestamp),
@@ -114,8 +113,8 @@ func (t *Call) Marshal(p *codec.Packer) {
 	}
 }
 
-func UnmarshalCallContract(r *runtime.WasmRuntime) func(p *codec.Packer) (chain.Action, error) {
-	return func(p *codec.Packer) (chain.Action, error) {
+func UnmarshalCallContract(r *runtime.WasmRuntime) func(p *codec.Packer) (chain.Action[struct{}], error) {
+	return func(p *codec.Packer) (chain.Action[struct{}], error) {
 		callContract := Call{r: r}
 		callContract.Value = p.UnpackUint64(false)
 		callContract.Fuel = p.UnpackUint64(true)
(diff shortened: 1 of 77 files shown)


examples/vmwithcontracts/actions/call.go
```diff
@@ -19,7 +19,7 @@ import (
 	mconsts "github.com/ava-labs/hypersdk/examples/vmwithcontracts/consts"
 )
 
-var _ chain.Action = (*Call)(nil)
+var _ chain.Action[struct{}] = (*Call)(nil)
```
richardpringle, line 22:
```suggestion
// TODO: inject runtime here instead of attaching to the action
var _ chain.Action[struct{}] = (*Call)(nil)
```

vm/vm.go
```diff
@@ -1243,7 +1246,11 @@ func (vm *VM) restoreAcceptedQueue(ctx context.Context) error {
...
 // issue. It is better to ensure we exit to surface the error.
-func (vm *VM) Fatal(msg string, fields ...zap.Field) {
+func (vm *VM[_]) Fatal(msg string, fields ...zap.Field) {
 	vm.snowCtx.Log.Fatal(msg, fields...)
 	panic("fatal error")
 }
+
+func (vm *VM[T]) GetRuntime() T {
+	return vm.runtime
+}
```
richardpringle, line 1256: 👍

richardpringle (dismissed):
One question, but overall looks good

vm/option.go
```diff
@@ -50,44 +50,44 @@ func NewOption[T any](namespace string, defaultConfig T, optionFunc OptionFunc[T
...
-	return func(vm *VM) {
+func WithVMAPIs[T chain.RuntimeInterface](apiHandlerFactories ...api.HandlerFactory[api.VM[T]]) RegisterFunc[T] {
+	return func(vm *VM[T]) {
 		vm.vmAPIHandlerFactories = append(vm.vmAPIHandlerFactories, apiHandlerFactories...)
 	}
 }
 
-func WithTxRemovedSubscriptions(subscriptions ...event.SubscriptionFactory[TxRemovedEvent]) RegisterFunc {
-	return func(vm *VM) {
+func WithTxRemovedSubscriptions[T chain.RuntimeInterface](subscriptions ...event.SubscriptionFactory[TxRemovedEvent]) RegisterFunc[T] {
```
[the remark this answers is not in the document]
joshua-kim: For this PR my thought was that we would just make the runtime type a required parameter for hypersdk's new ([ref](https://github.com/ava-labs/hypersdk/blob/370360d6c7a2b9a5f3b9ffcdce63356b664e21f1/vm/vm.go#L150)). I don't think we can do an option like I previously thought since the `VM` is parameterized on a generic type so the caller must define it.

aaronbuchwald:
I was thinking of a different direction where we:

- Instead of `T` having `state.Mutable`, `T` would be required to meet the interface that enables creating and committing a tree of revisions (which is currently supported)
- Instead of replacing `state.Mutable` with `T` that includes `state.Mutable`, we'd just give you `T`
- The default would be to use `state.Mutable`, so that if you wanted to you could keep the MorpheusVM implementation the same
- In practice, we'd create a new storage type in MorpheusVM that wraps `state.Mutable` that type would include `GetBalance` , `SubBalance`, and `AddBalance`
- To wrap it, we'd provide a function that wraps `state.Mutable` with the desired type (could be the struct implementing those functions or add a reference to the runtime) 
- Update `transfer.go` so that it calls those functions directly. We could even make the default version include native token balance implementation, so that it can implement the full `StorageManager` interface

richardpringle:
> I was thinking of a different direction where we:
> 
> * Instead of `T` having `state.Mutable`, `T` would be required to meet the interface that enables creating and committing a tree of revisions (which is currently supported)
> * Instead of replacing `state.Mutable` with `T` that includes `state.Mutable`, we'd just give you `T`
> * The default would be to use `state.Mutable`, so that if you wanted to you could keep the MorpheusVM implementation the same
> * In practice, we'd create a new storage type in MorpheusVM that wraps `state.Mutable` that type would include `GetBalance` , `SubBalance`, and `AddBalance`
> * To wrap it, we'd provide a function that wraps `state.Mutable` with the desired type (could be the struct implementing those functions or add a reference to the runtime)
> * Update `transfer.go` so that it calls those functions directly. We could even make the default version include native token balance implementation, so that it can implement the full `StorageManager` interface

I'm not sure this is generic enough. It's not just state that is changing here. I might not fully understand what you're suggesting though

joshua-kim:
Closing for now, unsure how this will be used + wasm contracts has some hacks it can use to still work.

PR #1586, opened 2024-09-24
Refactor runtime suggestion

joshua-kim:


chain/dependencies.go
```diff
@@ -209,7 +211,30 @@ type Marshaler interface {
...
+	Commit()
+}
+
+type View[T PendingView] interface {
+	NewView(scope state.Keys, storage map[string][]byte) T
+	PendingChanges() int
+	OpIndex() int
+	ExportMerkleDBView(ctx context.Context, t trace.Tracer, view state.View) (merkledb.View, error)
+}
+
```
joshua-kim, line 230: These interfaces are based off of the `tstate` functions... I feel like these interfaces are extremely awkward (why do we have two view types? Why do we use `merkledb` types?).

I feel very uncomfortable about exposing these-as-is to a downstream implementation (it would certainly result in cleanup down the road)... but I don't think we can do the refactor/define good interfaces in this PR.

richardpringle: This seems like a premature abstraction, as I'm not quite sure how it would be used. The original feature request was for a way to inject dependencies into the `Execute` call, which I imagine would be useful for specific VMs that have dependencies (the `vmwithcontracts` requires a wasm runtime, but it's the *general* problem that we want to consider). 

Adding a `PendingView` with rollbacks etc without a concrete idea (read example) on the best usage seems premature. That's just my 2-cents

aaronbuchwald: I'd agree w/ this. I was thinking that for now, we would wrap `state.Mutable`.

I think the original feature request is to support adding static dependencies ie. `GetWASMRuntime()`, so that it can pre-allocate WASM runtimes and handle concurrency of those objects.

Overall, it seems like the possible cases we'd support are:
- static dependencies left to the VM to provide
- direct wrapper around `state.Mutable` that supports different accessors, but still operates directly on the same key-value based state
- "extra" state that needs to handle views, commit/abandon revisions as blocks are accepted, and rollbacks if an action/transaction reverts

I think the wrapper around `state.Mutable` could be used to add helper functions like `GetBalance`, `AddBalance,` and `SubBalance` directly onto the state passed into `action.Execute(...)`. We could use the same to provide static dependencies, and then leave handling revisions to a separate piece of work (ex. support an in-memory order book in a more robust way than the original TokenVM action accept handler).

PR #1588, opened 2024-09-24
validate buffer length prior to calling Uint64

tsachiherman:
## Why this should be merged

Code was missing a validation that would ensure value read from database was valid.
Without this code, a corrupted database would likely to result in the process crashing due to out of bound slice lookup.

## How this works

Prior to marshaling the byte slice received from the database layer ( via `binary.BigEndian.Uint64` ), we check to see if the slice length is 8 bytes. If it's any different than 8 bytes, we fail the operation rather then providing the malformed buffer to `Uint64` and letting it panic. 

## How this was tested

Looking for feedback on how to test this further.


state/keys.go
```diff
@@ -14,6 +18,8 @@ const (
 	All              = Read | Allocate | Write
 )
 
+var ErrMalformedEncoding = errors.New("malformed encoding")
```
tsachiherman, line 21: not sure if this is the right place.. but please lmk if you think elsewhere would be a better place.

joshua-kim (approved):

joshua-kim:
I saw there were a few more usages of this, should we also validate here?

```
➜  hypersdk git:(main) grep -r binary.BigEndian.Uint64 . | grep -v test
./fees/dimension.go:		d[i] = binary.BigEndian.Uint64(raw[i*consts.Uint64Len:])
./internal/fees/manager.go:	return binary.BigEndian.Uint64(f.raw[start : start+consts.Uint64Len])
./internal/fees/manager.go:	return binary.BigEndian.Uint64(f.raw[start : start+consts.Uint64Len])
./internal/fees/manager.go:	lastTimeSeconds := int64(binary.BigEndian.Uint64(f.raw[0:consts.Int64Len]))
./internal/window/window.go:		sum, overflow = math.Add(sum, binary.BigEndian.Uint64(w[consts.Uint64Len*i:]))
./internal/window/window.go:	prevUnitsConsumed := binary.BigEndian.Uint64(w[start:])
./internal/window/window.go:	return binary.BigEndian.Uint64(w[WindowSliceSize-consts.Uint64Len:])
./chain/block.go:	parentHeight := binary.BigEndian.Uint64(parentHeightRaw)
./chain/block.go:	parentTimestamp := int64(binary.BigEndian.Uint64(parentTimestampRaw))
./chain/block.go:		acceptedHeight := binary.BigEndian.Uint64(acceptedHeightRaw)
./docs/tutorials/morpheusvm/morpheusvm.md:	return binary.BigEndian.Uint64(v), true, nil
./vm/storage.go:	return binary.BigEndian.Uint64(b), nil
./vm/storage.go:	return binary.BigEndian.Uint64(b), nil
./vm/storage.go:	return binary.BigEndian.Uint64(b), nil
./examples/morpheusvm/storage/storage.go:	return binary.BigEndian.Uint64(v), true, nil
./examples/vmwithcontracts/storage/storage.go:	return binary.BigEndian.Uint64(v), true, nil
./api/indexer/tx_indexer.go:	timestamp := int64(binary.BigEndian.Uint64(v))
./api/indexer/tx_indexer.go:	fee := binary.BigEndian.Uint64(v[consts.Uint64Len+1+fees.DimensionsLen:])
./x/contracts/simulator/state/manager.go:	return binary.BigEndian.Uint64(v), nil
```

tsachiherman:
> I saw there were a few more usages of this, should we also validate here?
> 
> ```
> ➜  hypersdk git:(main) grep -r binary.BigEndian.Uint64 . | grep -v test
> ./fees/dimension.go:		d[i] = binary.BigEndian.Uint64(raw[i*consts.Uint64Len:])
> ./internal/fees/manager.go:	return binary.BigEndian.Uint64(f.raw[start : start+consts.Uint64Len])
> ./internal/fees/manager.go:	return binary.BigEndian.Uint64(f.raw[start : start+consts.Uint64Len])
> ./internal/fees/manager.go:	lastTimeSeconds := int64(binary.BigEndian.Uint64(f.raw[0:consts.Int64Len]))
> ./internal/window/window.go:		sum, overflow = math.Add(sum, binary.BigEndian.Uint64(w[consts.Uint64Len*i:]))
> ./internal/window/window.go:	prevUnitsConsumed := binary.BigEndian.Uint64(w[start:])
> ./internal/window/window.go:	return binary.BigEndian.Uint64(w[WindowSliceSize-consts.Uint64Len:])
> ./chain/block.go:	parentHeight := binary.BigEndian.Uint64(parentHeightRaw)
> ./chain/block.go:	parentTimestamp := int64(binary.BigEndian.Uint64(parentTimestampRaw))
> ./chain/block.go:		acceptedHeight := binary.BigEndian.Uint64(acceptedHeightRaw)
> ./docs/tutorials/morpheusvm/morpheusvm.md:	return binary.BigEndian.Uint64(v), true, nil
> ./vm/storage.go:	return binary.BigEndian.Uint64(b), nil
> ./vm/storage.go:	return binary.BigEndian.Uint64(b), nil
> ./vm/storage.go:	return binary.BigEndian.Uint64(b), nil
> ./examples/morpheusvm/storage/storage.go:	return binary.BigEndian.Uint64(v), true, nil
> ./examples/vmwithcontracts/storage/storage.go:	return binary.BigEndian.Uint64(v), true, nil
> ./api/indexer/tx_indexer.go:	timestamp := int64(binary.BigEndian.Uint64(v))
> ./api/indexer/tx_indexer.go:	fee := binary.BigEndian.Uint64(v[consts.Uint64Len+1+fees.DimensionsLen:])
> ./x/contracts/simulator/state/manager.go:	return binary.BigEndian.Uint64(v), nil
> ```

some of the ones in this list are already handled by this PR. Some others (window) are already have proper prior testing, or doesn't require that testing due to the usage.

aaronbuchwald:
fwiw - we treat database errors or corruption as fatal already, so this should only protect us from bugs that write invalid values into the database and in that case turn a panic into either a fatal error or invalidating the operation trying to use the invalid value. Still a good improvement, just calling out this only protects us in a limited case we should never hit.

Leaving out window because we know the length of the byte array we're using makes sense, what's the reason for leaving it out of `chain/block.go` ?

tsachiherman:
> parentHeight := binary.BigEndian.Uint64(parentHeightRaw)

done.

tsachiherman:
> fwiw - we treat database errors or corruption as fatal already, so this should only protect us from bugs that write invalid values into the database and in that case turn a panic into either a fatal error or invalidating the operation trying to use the invalid value. Still a good improvement, just calling out this only protects us in a limited case we should never hit.

yes, agree. Practically, it would improve the process survivability, giving it a chance to report an error while being alive.
 
> Leaving out window because we know the length of the byte array we're using makes sense, what's the reason for leaving it out of `chain/block.go` ?

done.

aaronbuchwald (approved):

PR #1590, opened 2024-09-26
Fix transfer benchmark

RodrigoVillar:
This PR fixes a current bug within the HyperSDK.

Currently, running `go test -bench=.` within the `actions/` directory of MorpheusVM fails as we are comparing `test.ExpectedOutput` (an array) to `output` (a single instance of `codec.Typed`). 

This PR fixes the bug above by populating the `ExpectedOutputs` field and correcting the comparison mentioned above (we now compare instances of `codec.Typed`)

### DIFF
--- a/examples/morpheusvm/actions/transfer_test.go
+++ b/examples/morpheusvm/actions/transfer_test.go
@@ -142,6 +142,16 @@ func BenchmarkSimpleTransfer(b *testing.B) {
 			To:    to,
 			Value: 1,
 		},
+		ExpectedOutputs: func() []codec.Typed {
+			outputs := make([]codec.Typed, 0)
+			for i := 0; i < b.N; i++ {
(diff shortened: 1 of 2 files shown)


samliok (approved):

examples/morpheusvm/actions/transfer_test.go
```diff
@@ -142,6 +142,16 @@ func BenchmarkSimpleTransfer(b *testing.B) {
 			To:    to,
 			Value: 1,
 		},
+		ExpectedOutputs: func() []codec.Typed {
```
joshua-kim, line 145: Interesting - this test previously panicked. We don't run benchmarks at any cadence... feels like we should have some visibility into this... but it's unrelated to the PR.

joshua-kim (approved):

PR #1592, opened 2024-09-27
Reflect Marshaler

containerman17:
A single CLI binary to manage all the HyperSDK VMs sounds great. However, the current ABI implementation requires regenerating structs with data types for each VM. This PR changes that by generating structs dynamically using reflection.

This package should *NOT* be used within VMs, only in CLI applications. I'm happy to rename it if you have better ideas.  

P.S. I've also added `0x` to addresses to follow the recent changes.
P.S. Please take some extra time during the code review or suggest additional test cases. The code is short but does quite a bit. It is well-tested using the existing `abi` package specs, but I wouldn’t call myself an expert with Go’s `reflect` package.


abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+
+		// Find the type in the ABI
+		abiType := findABIType(inputAbi, abiTypeName)
+		if abiType == nil {
+			// If not found, fallback to interface{}
+			return reflect.TypeOf((*interface{})(nil)).Elem()
+		}
+
+		// Build fields
+		fields := make([]reflect.StructField, len(abiType.Fields))
```
ARR4N, line 135: Is it guaranteed that all types in the ABI are structs? If so, that's the type of information that I think should be included in comments because it can't be inferred from the code.

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+		for i, field := range abiType.Fields {
+			fieldType := getReflectType(field.Type, inputAbi, typeCache)
```
ARR4N, line 137: This variable is unnecessary and the value can just be placed in the struct literal's `Type` field.

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+
+		// Check if type already in cache
+		if cachedType, ok := typeCache[abiTypeName]; ok {
+			return cachedType
+		}
+
+		// Find the type in the ABI
+		abiType := findABIType(inputAbi, abiTypeName)
+		if abiType == nil {
+			// If not found, fallback to interface{}
```
ARR4N, line 130: Why? Please elaborate in the comment.

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+			return reflect.TypeOf((*interface{})(nil)).Elem()
```
ARR4N, line 131: (fyi, no action required) The outcome of this really confused me but there's no other way until Go 1.22, when I recommend changing this to `reflect.TypeFor[interface{}]()`.

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+
+		// Build fields
+		fields := make([]reflect.StructField, len(abiType.Fields))
+		for i, field := range abiType.Fields {
+			fieldType := getReflectType(field.Type, inputAbi, typeCache)
+			fields[i] = reflect.StructField{
+				Name: cases.Title(language.English).String(field.Name),
```
ARR4N, line 139: Oooh, nice! TIL about these packages 🙏

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+		return reflect.TypeOf(int32(0))
+	case "int64":
+		return reflect.TypeOf(int64(0))
+	case "Address":
+		return reflect.TypeOf(codec.Address{})
+	default:
+		if strings.HasPrefix(abiTypeName, "[]") {
+			elemType := getReflectType(strings.TrimPrefix(abiTypeName, "[]"), inputAbi, typeCache)
+			return reflect.SliceOf(elemType)
+		} else if strings.HasPrefix(abiTypeName, "[") {
```
ARR4N, line 107: (nit, subjective style) The `else` is redundant because of the `return` on the line before. Later with the cache check you don't have one.

I think there should at least be consistency. Personally I'd remove it because it adds just that tiny bit of cognitive load for the reader who thinks "else what?".

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+
+	"github.com/ava-labs/hypersdk/abi"
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+)
+
+func DynamicMarshal(inputAbi abi.ABI, typeName string, jsonData string) ([]byte, error) {
```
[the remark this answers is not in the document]
containerman17: My brain keeps switching between TypeScript and Golang (I implement everything in both languages). You are right.

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+	abiType := findABIType(inputAbi, typeName)
+	if abiType == nil {
+		return nil, fmt.Errorf("type %s not found in ABI", typeName)
+	}
+
+	// Create a cache to avoid rebuilding types
+	typeCache := make(map[string]reflect.Type)
+
+	// Create a dynamic struct type
+	dynamicType := getReflectType(typeName, inputAbi, typeCache)
```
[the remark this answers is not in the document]
containerman17: I prefer using full English words for variable names and want to avoid abbreviations like `typ` instead of `type` just for aesthetics. However, since `type` is a reserved keyword, I came up with `dynamicType` as the best alternative.

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+			// Handle fixed-size arrays
+
+			sizeStr := strings.Split(abiTypeName, "]")[0]
+			sizeStr = strings.TrimPrefix(sizeStr, "[")
+
+			size, err := strconv.Atoi(sizeStr)
+			if err != nil {
+				return reflect.TypeOf((*interface{})(nil)).Elem()
+			}
+			elemType := getReflectType(strings.TrimPrefix(abiTypeName, "["+sizeStr+"]"), inputAbi, typeCache)
```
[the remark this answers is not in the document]
containerman17: I was debating that. The string implementation is about 3 times faster and takes less mental effort to read. If you compile the regex separately, you'll regain almost all the speed, but lose readability. I ended up compiling it separately for performance and added the actual regex as a comment where it's used. I do like my string manipulation implementation, but for an open-source product, the regex looks a bit cleaner, yes.

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+			return reflect.ArrayOf(size, elemType)
+		}
+		// For custom types, recursively construct the struct type
+
+		// Check if type already in cache
+		if cachedType, ok := typeCache[abiTypeName]; ok {
```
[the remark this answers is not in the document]
containerman17: Once you start using one-letter variables, it's hard to stop. Here are some real examples from the HyperSDK codebase:

```golang
addr, priv, err := h.h.GetDefaultKey(true)
t.r.CallContract(ctx, &runtime.CallInfo{
w.s.Publish(append([]byte{TxMode}, bytes...), listeners)
c.s.removeConnection(c)
k.D.FillBytes(b)
```

The Go community has pretty much standardized the use of one-letter variables for struct instances. I don’t like it, but I follow it because that’s just how Go works (basically, shut up and use the defaults). Using `i`, `j`, `k` for the first, second, and third loop levels is a classic, but I try to avoid one-letter variables outside of that.

abi/dynamic/reflect_marshal.go
containerman17: Yes, guaranteed. Adding a comment, thanks

abi/dynamic/reflect_marshal.go
```diff
@@ -0,0 +1,162 @@
...
+
+		return structType
+	}
+}
+
+// Helper function to find ABI type
+func findABIType(inputAbi abi.ABI, typeName string) *abi.Type {
+	for i := range inputAbi.Types {
+		if inputAbi.Types[i].Name == typeName {
+			return &inputAbi.Types[i]
```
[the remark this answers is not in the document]
containerman17: That's cleaner, yes, thyanks

abi/dynamic/reflect_marshal.go
containerman17: Ok, now with errors, keeping it.

abi/dynamic/reflect_marshal.go
containerman17: It makes much more sense to return errors here instead of just an empty value, I agree. Yeah, that was a dumb thing to do.

abi/dynamic/reflect_marshal.go
containerman17: Switched to nil, err

abi/dynamic/reflect_marshal.go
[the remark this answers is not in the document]
ARR4N: I don't envy you having to do it in TS too!

abi/dynamic/reflect_marshal.go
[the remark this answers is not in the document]
ARR4N: Is speed necessary in a CLI?

Lose readability because it's not co-located? I happened, by chance, to comment on that in my re-review as there's no need to keep the variable declared up the top.

abi/dynamic/reflect_marshal.go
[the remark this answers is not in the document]
ARR4N: TL;DR I agree with you re their overuse but not that that means they should be avoided. Everything else is just the ramblings of someone who cares way too much about style and cognitive load when reading code.

> because that’s just how Go works

I think you're seeing a misuse of them and agree with you that snippet is _completely_ unreadable. Had I been reviewing it I'd probably have given the inverse of the advice I gave in my comment here.

> Once you start using one-letter variables, it's hard to stop.

Again I think other people are just misusing them. Style guides aren't immutable rules that apply without question, but they seem to be treated as such 🤷  In this particular case my rationale is that even using `typ := typeCache[x]` adds little compared to `t := typeCache[x]` since it's returned on the next line, but I also think `typ` is completely acceptable because it avoids the clutter of `cachedType := typeCache[x]; return cachedType`.

> The Go community has pretty much standardized the use of one-letter variables for struct instances.

This is only in the receivers on methods, in lieu of `this` or `self`, but not for fields (at least not standardised). Personally I don't like the 1/2 character "limit" and occasionally break it when needed.

[Google's style guide on the matter](https://google.github.io/styleguide/go/decisions#variable-names:~:text=Limit%20their%20use%20to%20instances%20where%20the%20full%20word%20is%20obvious%20and%20where%20it%20would%20be%20repetitive%20for%20it%20to%20appear%20in%20place%20of%20the%20single%2Dletter%20variable)

abi/dynamic/reflect_marshal.go
ARR4N: > that was a dumb thing to do

I don't actually think it was dumb as there could be reasons (e.g. you just want the JSON package to decide on the correct native type). I was just wondering what the reason was / prompting discussion about it.

ARR4N (dismissed):

aaronbuchwald (approved):
Couple small nits - LGTM once CI is fixed

abi/dynamic/reflect_marshal.go
```diff
@@ -78,85 +84,90 @@ func DynamicUnmarshal(inputAbi abi.ABI, typeName string, data []byte) (string, e
...
+			if err != nil {
+				return nil, err
+			}
+			return reflect.SliceOf(elemType), nil
+		}
 
+		// golang arrays
+		match := fixedSizeArrayRegex.FindStringSubmatch(abiTypeName) // ^\[(\d+)\](.+)$
+		if match != nil {
+			sizeStr := match[1]
```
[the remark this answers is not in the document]
containerman17: That's exclusively for readability

abi/dynamic/reflect_marshal_test.go
```diff
@@ -67,6 +65,30 @@ func TestDynamicMarshal(t *testing.T) {
...
+	abiJSON := mustReadFile(t, "../testdata/abi.json")
+	var abi abi.ABI
+
+	err := json.Unmarshal(abiJSON, &abi)
+	require.NoError(err)
+
+	t.Run("malformed JSON", func(t *testing.T) {
+		malformedJSON := `{"uint8": 42, "uint16": 1000, "uint32": 100000, "uint64": 10000000000, "int8": -42, "int16": -1000, "int32": -100000, "int64": -10000000000,`
+		_, err := Marshal(abi, "MockObjectAllNumbers", malformedJSON)
+		require.Error(err)
```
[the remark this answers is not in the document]
containerman17: Yes, `require := require.New(t)` is ugly. I'm going to keep it as is though, since it requires jumping through a few hoops to make the proper require available in the subtest.

abi/dynamic/reflect_marshal_test.go
```diff
@@ -74,19 +74,15 @@ func TestDynamicMarshalErrors(t *testing.T) {
...
-	t.Run("wrong struct name", func(t *testing.T) {
-		jsonData := mustReadFile(t, "../testdata/numbers.json")
-		_, err := Marshal(abi, "NonExistentObject", string(jsonData))
-		require.Error(err)
-		require.Contains(err.Error(), "type NonExistentObject not found in ABI")
-	})
+	// Test malformed JSON
+	malformedJSON := `{"uint8": 42, "uint16": 1000, "uint32": 100000, "uint64": 10000000000, "int8": -42, "int16": -1000, "int32": -100000, "int64": -10000000000,`
+	_, err = Marshal(abi, "MockObjectAllNumbers", malformedJSON)
+	require.Contains(err.Error(), "failed to unmarshal JSON data")
```
aaronbuchwald, line 80: Could we use `require.ErrorIs` here as well?

abi/dynamic/reflect_marshal_test.go
```diff
@@ -0,0 +1,94 @@
...
+	abiJSON := mustReadFile(t, "../testdata/abi.json")
+	var abi abi.ABI
+
+	err := json.Unmarshal(abiJSON, &abi)
+	require.NoError(err)
+
+	// Test malformed JSON
+	malformedJSON := `{"uint8": 42, "uint16": 1000, "uint32": 100000, "uint64": 10000000000, "int8": -42, "int16": -1000, "int32": -100000, "int64": -10000000000,`
+	_, err = Marshal(abi, "MockObjectAllNumbers", malformedJSON)
+	require.Contains(err.Error(), "unexpected end of JSON input")
```
aaronbuchwald, line 80: This error comes from `encoding/json` so we cannot use `require.ErrorIs` here

aaronbuchwald (approved):

Issue #1589, opened 2024-09-26
Websocket server hangs on Create token action


containerman17:
When the `createToken` action is submitted, the WebSocket server hangs and stops responding. There are no errors in the logs, and the action itself is executed successfully. However, all WebSocket connections to the node freeze after this action.

### Source Code:
- `createToken` action source: [create_token.go](https://github.com/ava-labs/hypersdk-starter/blob/cfmm-demo/actions/create_token.go)
- Example block listener: [Block Listener Example](https://gist.github.com/containerman17/d3369075e8e89c5ce8b0da0178c54e7b)

### Details:
- Transfer transactions work correctly through WebSocket without any issues.
- The issue happens consistently after every `createToken` action.
- The problem is not related to the client. The same behavior occurs when listening for new blocks in Go, suggesting that the WebSocket server itself hangs, though it's unclear if block production stops or if it's solely a WebSocket issue.
- User get's charged gas fees, so looks like Tx is being processed

### Additional Information:
- **Aaron's suggestion:** This issue could be related to the `pubsub` package used by the WebSocket server. Possible problematic areas include [message_buffer.go#L124](https://github.com/ava-labs/hypersdk/blob/main/pubsub/message_buffer.go#L124) and [message_buffer.go#L116](https://github.com/ava-labs/hypersdk/blob/main/pubsub/message_buffer.go#L116), where a deadlock might occur.

containerman17:
To reproduce, clone https://github.com/ava-labs/hypersdk-starter, check out the `cfmm-demo` branch, and then run `docker compose up -d --build frontend faucet`. Refer to https://github.com/ava-labs/hypersdk-starter/blob/main/README.md for instructions on how to get started. @aaronbuchwald

containerman17:
Replaced by https://github.com/ava-labs/hypersdk/issues/1619

PR #1593, opened 2024-09-27
Minor nits on pubsub implementation

aaronbuchwald:
This PR makes a few minor nits while going through the pubsub implementation to debug https://github.com/ava-labs/hypersdk/issues/1589

### DIFF
--- a/api/ws/server.go
+++ b/api/ws/server.go
@@ -150,10 +150,12 @@ func (w *WebSocketServer) AddTxListener(tx *chain.Transaction, c *pubsub.Connect
 
 	// TODO: limit max number of tx listeners a single connection can create
 	txID := tx.ID()
-	if _, ok := w.txListeners[txID]; !ok {
-		w.txListeners[txID] = pubsub.NewConnections()
+	connections, ok := w.txListeners[txID]
+	if !ok {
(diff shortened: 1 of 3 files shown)


pubsub/server_test.go
```diff
@@ -49,8 +49,6 @@ func (x *counter) dummyProcessTXCallback(b []byte, _ *Connection) {
 // a msg to be sent to all connections. Checks the message was delivered properly
 // and the connection is properly handled when closed.
 func TestServerPublish(t *testing.T) {
-	t.Skip("FLAKY")
```
aaronbuchwald, line 52: Removing `FLAKY` here in favor of seeing and fixing the test failures, since this is causing problems

api/ws/server.go
```diff
@@ -219,7 +221,9 @@ func (w *WebSocketServer) AcceptBlock(b *chain.StatefulBlock) error {
 		if err != nil {
 			return err
 		}
-		w.s.Publish(append([]byte{TxMode}, bytes...), listeners)
```
joshua-kim, line 222: Really not a fan of not using the blank return for errors in the old code... feels like it makes it non-obvious that we're ignoring error values.

joshua-kim (approved):

Issue #1572, opened 2024-09-18
Add Linter to Ensure Examples never use `internal` packages


aaronbuchwald:
After moving packages to internal we accidentally included some in the examples directory, which would be impossible for any external repos to reproduce.

We should add a linter to guarantee that examples never pulls in any packages from internal.

aaronbuchwald:
Separated adding a linter from https://github.com/ava-labs/hypersdk/issues/1557

PR #1594, opened 2024-09-27
Add no internal linter for examples

RodrigoVillar:
This PR revolves #1572 by adding a simple linter at checks for the use of `internal` packages within `examples/`

If the linter passes, it returns the following:
```bash
Ok.
```

If there is an instance of `internal` being used, it returns a similar message as seen below for each violation and exits with code 1:

```bash
Violation: /Users/rodrigo.villar/go/src/github.com/ava-labs/hypersdk/examples/morpheusvm/actions/transfer.go imports from /internal
```

This PR also adds the internal linter as part of the CI tests for HyperSDK

### DIFF
--- a/.github/workflows/hypersdk-ci.yml
+++ b/.github/workflows/hypersdk-ci.yml
@@ -53,6 +53,9 @@ jobs:
       - name: Run actionlint
         shell: bash
         run: scripts/tests.actionlint.sh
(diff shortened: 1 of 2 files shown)


aaronbuchwald:
Would it be possible to look into why depguard is not giving us what we want here?

PR #1597, opened 2024-09-27
Update tx indexer to include tx action outputs

aaronbuchwald:
This PR adds `result.Outputs` to the result of the get transaction API.

This unblocks hypersdk-starter from being blocked on the ws/pubsub package since it's currently the only API that returns transaction results.

This PR also:
- adds back `codec.Bytes` wrapper to support hex encoding `[]byte`
- Switches from manually packing the transaction inside the tx indexer to using the codec reader/writer
- updates the tx assertion in the MorpheusVM workloads to check the transfer result includes the correct receiver balance

### DIFF
--- a/api/indexer/api.go
+++ b/api/indexer/api.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ava-labs/avalanchego/trace"
 
 	"github.com/ava-labs/hypersdk/api"
+	"github.com/ava-labs/hypersdk/codec"
 	"github.com/ava-labs/hypersdk/fees"
 )
 
@@ -52,6 +53,7 @@ type GetTxResponse struct {
 	Success   bool            `json:"success"`
 	Units     fees.Dimensions `json:"units"`
 	Fee       uint64          `json:"fee"`
+	Outputs   []codec.Bytes   `json:"result"`
 }
 
 type Server struct {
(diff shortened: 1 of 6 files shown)


api/indexer/tx_indexer.go
```diff
@@ -122,36 +120,54 @@ func (*txDBIndexer) storeTransaction(
...
 	fee uint64,
+	outputs [][]byte,
 ) error {
-	v := make([]byte, consts.Uint64Len+1+fees.DimensionsLen+consts.Uint64Len)
-	binary.BigEndian.PutUint64(v, uint64(timestamp))
-	if success {
-		v[consts.Uint64Len] = successByte
-	} else {
-		v[consts.Uint64Len] = failureByte
+	outputLength := 1 // Single byte containing number of outputs
```
aaronbuchwald, line 125: Note: `GetMaxActionsPerTx()` returns a `uint8`, which enforces the maximum possible number of actions and corresponding results included in a transaction.

api/indexer/api.go
```diff
@@ -52,6 +53,7 @@ type GetTxResponse struct {
 	Success   bool            `json:"success"`
 	Units     fees.Dimensions `json:"units"`
 	Fee       uint64          `json:"fee"`
+	Outputs   []codec.Bytes   `json:"result"`
```
aaronbuchwald, line 56: Uses wrapped `codec.Bytes` to provide hex format

joshua-kim (approved):

Issue #1595, opened 2024-09-27
Create `BlockWithResults` Type instead of Using `StatefulBlock`


aaronbuchwald:
We currently use `StatefulBlock` when we need both the `StatelessBlock` and the attached results including when sending to the API.

Add a separate type `BlockWithResults` (alternative names could be `ExecutedBlock` or something else if preferred) and use this type instead of `StatefulBlock` whenever we only need the stateless version of the block w/ results.

We should use the new type in APIs that depend on a subscription to accepted blocks and the external subscriber.

This should resolve https://github.com/ava-labs/hypersdk/issues/1481

PR #1601, opened 2024-09-28
Add ExecutedBlock type

aaronbuchwald:
This PR creates an `ExecutedBlock` type that includes the `StatelessBlock` and the execution results including both `[]*chain.Result` and the unit prices ie `fee.Dimensions`. This replaces passing around `chain.StatefulBlock` with the new type.

Addresses https://github.com/ava-labs/hypersdk/issues/1481 and https://github.com/ava-labs/hypersdk/issues/1595


chain/block.go
```diff
@@ -894,6 +896,79 @@ func (b *StatefulBlock) FeeManager() *fees.Manager {
...
+	if err != nil {
+		return nil, err
+	}
+	writer.PackBytes(resultBytes)
+	writer.PackFixedBytes(b.UnitPrices.Bytes())
+
+	return writer.Bytes(), writer.Err()
+}
+
+func UnmarshalExecutedBlock(bytes []byte, parser Parser) (*ExecutedBlock, error) {
```
joshua-kim, line 935: nit: should `Unmarshal` have a receiver on the struct too (or `Marshal` be static)? Seems weird that it's not consistent.

fees/dimension.go
joshua-kim, line 1: What are changes in this file needed for?

chain/block.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/ava-labs/hypersdk/internal/workers"
 	"github.com/ava-labs/hypersdk/state"
 	"github.com/ava-labs/hypersdk/utils"
+
+	publicfees "github.com/ava-labs/hypersdk/fees"
```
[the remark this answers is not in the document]
aaronbuchwald: Yup

chain/block.go
aaronbuchwald: Following current style here, happy to discuss for a separate change

chain/block.go
```diff
@@ -894,6 +896,79 @@ func (b *StatefulBlock) FeeManager() *fees.Manager {
...
+		id:             b.ID(),
+	}
+}
+
+type ExecutedBlock struct {
+	*StatelessBlock `json:"block"`
+	Results         []*Result             `json:"results"`
+	UnitPrices      publicfees.Dimensions `json:"unitPrices"`
+
+	blockBytes []byte
```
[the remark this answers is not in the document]
aaronbuchwald: This just saves marshalling it an extra time. Will switch to doing the extra serialization

chain/block.go
```diff
@@ -894,6 +896,79 @@ func (b *StatefulBlock) FeeManager() *fees.Manager {
...
+	id         ids.ID
```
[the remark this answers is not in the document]
aaronbuchwald: Can un-embed `StatelessBlock` and switch to requiring the caller to go through that type

fees/dimension.go
aaronbuchwald: This was to support JSON encoding/decoding of the `ExecutedBlock` type, which includes this type now.

chain/block.go
```diff
@@ -894,6 +896,79 @@ func (b *StatefulBlock) FeeManager() *fees.Manager {
 	return b.feeManager
 }
 
+func (b *StatefulBlock) ExecutedBlock() *ExecutedBlock {
```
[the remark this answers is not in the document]
aaronbuchwald: I thought about the alternative ordering as well ie. when you execute a stateful block it either produces or populates the fields on an `ExecutedBlock` type. Will take a stab at that and see what it looks like.

As for moving to `vm.Block`, I think we could definitely use a re-org of the chain/vm packages, but for now I think the new block type fits more naturally in the `chain/` package.

re pointers vs. values, this is following the current style, so would prefer not to start making that change in this PR.

chain/block.go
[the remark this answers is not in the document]
aaronbuchwald: This has the annoying side effect of needing to handle the error for both `Marshal()` and `ID()` since it's not stored on the type

fees/dimension.go
```diff
@@ -104,9 +104,11 @@ func (d Dimensions) Greater(o Dimensions) bool {
 	return true
 }
 
+const dimensionsFormatter = "(Bandwidth=%d, Compute=%d, Storage(Read)=%d, Storage(Allocate)=%d, Storage(Write)=%d)"
```
joshua-kim, line 107: nit: move to top of file w/ other consts

chain/block.go
joshua-kim: This is new code introduced as part of the diff... would it be reasonable to do it as part of this PR?

chain/block.go
```diff
@@ -890,10 +892,78 @@ func (b *StatefulBlock) Results() []*Result {
 	return b.results
 }
 
-func (b *StatefulBlock) FeeManager() *fees.Manager {
+func (b *StatefulBlock) FeeManager() *internalfees.Manager {
 	return b.feeManager
 }
 
+type ExecutedBlock struct {
+	Block      *StatelessBlock `json:"block"`
```
joshua-kim, line 900: I would just prefer if we manually copied out the fields in the `StatelessBlock` type (Parent, Timestamp, Height, Txs, StateRoot). `StatelessBlock` has weird signatures (e.g `Size` is a redundant method because you can call `len` on the result of `Marshal`, and the call to `ID()` returns an error but we could handle that more cleanly by just initializing the `ExecutedBlock` with an `ID`) that I want to avoid propagating to upstream code.

chain/block.go
aaronbuchwald: All of the block marshalling / unmarshalling code is consistent with this change, so I'd rather not introduce a new style for one function.

aaronbuchwald: I'm also not sure moving it to a function on the pointer receiver is better. Either way, I don't think it belongs in this PR.

joshua-kim (dismissed):

PR #1615, opened 2024-10-01
Add State Layout

RodrigoVillar:
This PR adds the notion of a `Layout` which helps separate state that should be exclusive to the HyperSDK versus state which any HyperVM is free to modify as it wishes.

When calling `vm.New()`, developers now specify an array of the prefixes which their VM will touch. If any of these prefixes conflicts with either the height, timestamp, or fee prefixes that the HyperSDK specifies, `vm.New()` will return an error. Using this mechanism, the HyperSDK now provides the invariant that as long as `vm.New()` passes, then the prefixes that the developer passed through will not collide with important HyperSDK data.

Summary of Changes:

- Add `layout` package
- Update MorpheusVM + VMWithContracts to provide utilized prefixes 
- Remove `StateManager`
- Update packages throughout HyperSDK to use `BalanceManager` and `layout`


chain/builder.go
```diff
@@ -92,7 +93,7 @@ func BuildBlock(
 	}
 
 	// Compute next unit prices to use
-	feeKey := FeeKey(vm.StateManager().FeeKey())
+	feeKey := FeeKey(layout.FeePrefix())
```
aaronbuchwald, line 96: Rather than using `layout.FeePrefix()` here, which means that this is no longer configurable, could we pass a layout object into the VM, but create a default version of that layout that VMs can use, so they don't need to implement it themselves?

Then I think we can have a `layout.DefaultMetadata` or similar that we pass in to the VM constructor and use `layout.DefaultMinimumPrefix` in the VM to make sure that developers using this no exactly what prefix to start with.

aaronbuchwald (commented):
Can we switch from using the layout package to providing a default implementation of the `MetadataHandler` interface?

I think we should have:
- a default metadata handler that VMs can provide, so they don't need to implement it (passed into the VM)
- a DefaultMinPrefix from the same package, so that any VM using the state manager can use that to set their minimum prefix
- Remove the pattern of passing in `VMPrefixes` to the VM constructor. If a VM wants to sanity check their prefixes, they can call a sanity function directly. Otherwise, there's no need to force this on the VM constructor imo
- Separate the balance handler from the metadata handler, so that we don't need to wrap the default metadata handler into a single `StorageManager` struct

RodrigoVillar:
Closing in favor of #1658

PR #1622, opened 2024-10-02
Add JSON marshalling for fee dimensions

aaronbuchwald:
Add JSON marshaller/unmarshaller for `fees.Dimensions` type.

This improves the JSON output of the block index API: https://github.com/ava-labs/hypersdk/pull/1606#issuecomment-2387672897

### DIFF
--- a/fees/dimension.go
+++ b/fees/dimension.go
@@ -5,6 +5,7 @@ package fees
 
 import (
 	"encoding/binary"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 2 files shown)


tsachiherman (approved):
not sure if `omitempty` would be beneficial here or not.
looks good otherwise.

aaronbuchwald:
> not sure if `omitempty` would be beneficial here or not. looks good otherwise.

No these should always be populated, so don't think we want to use `omitempty` here.

containerman17 (approved):

PR #1624, opened 2024-10-02
add RegistryFactory and deprecate global action, auth and output registries in VMs

tsachiherman:
## What ?

add RegistryFactory and deprecate global action, auth and output registries in VMs

## Why ?

Avoid creation of global registries and pass these along the call stack as part of the server context.


chain/dependencies.go
```diff
@@ -287,3 +287,6 @@ type AuthFactory interface {
 	MaxUnits() (bandwidth uint64, compute uint64)
 	Address() codec.Address
 }
+
+// RegistryFactory is the factory function, provided to the VM initializer that provides the registries for teh actions, auth and output.
+type RegistryFactory func() (actionRegistry ActionRegistry, authRegistry AuthRegistry, outputRegistry OutputRegistry)
```
richardpringle, line 292: Would it be better to have a struct here? 
```go
type Registry struct {
    action ActionRegistry
    auth   AuthRegistry
    output OutputRegistry
}
```
```suggestion
type RegistryFactory func() Registry
```
I'm not sure if we plan on extending this, but I think backwards-compatible changes would be easier with a struct.

richardpringle (commented):
Looks better than before, but I think types where order doesn't matter are typically less error-prone (struct over tuple).

chain/dependencies.go
tsachiherman: sounds good to me.

tsachiherman: done

chain/dependencies.go
```diff
@@ -287,3 +287,12 @@ type AuthFactory interface {
...
 }
+
+type Registry struct {
+	Action ActionRegistry
+	Auth   AuthRegistry
+	Output OutputRegistry
+}
+
+// RegistryFactory is the factory function, provided to the VM initializer that provides the registries for the actions, auth and output.
+type RegistryFactory func() Registry
```
aaronbuchwald, line 298: Can we change from an interface returning a struct of all of these types to an interface that supports `ActionRegistry()`, `AuthRegistry()`, and `OutputRegistry()` individually?

We can have a struct that implements that interface and takes the exact same fields, that way we can pass it around more easily and call a single function on the interface to get the value we care about.

This will also make it more ergonomic when we have a type that needs only the one of those values. The callee can then declare its own dependency and the caller can pass in the existing interface.

tsachiherman: done.

richardpringle (dismissed):

vm/registry.go
```diff
@@ -0,0 +1,32 @@
...
+
+import "github.com/ava-labs/hypersdk/chain"
+
+type Registry struct {
+	actionRegistry chain.ActionRegistry
+	authRegistry   chain.AuthRegistry
+	outputRegistry chain.OutputRegistry
+}
+
+func NewRegistry(action chain.ActionRegistry, auth chain.AuthRegistry, output chain.OutputRegistry) *Registry {
```
aaronbuchwald, line 14: Could we define the struct implementing next to the interface in the chain package?

vm/vm.go
```diff
@@ -165,12 +162,13 @@ func New(
 	}
 
 	return &VM{
+		Registry: *NewRegistry(
+			registry.ActionRegistry(),
+			registry.AuthRegistry(),
+			registry.OutputRegistry()),
```
aaronbuchwald, line 168: Why create a registry from the existing registry? Can we just set the registry to the `chain.Registry` value that was passed in?

aaronbuchwald (changes requested):

vm/registry.go
tsachiherman: done

vm/vm.go
tsachiherman: done

chain/registry.go
```diff
@@ -0,0 +1,38 @@
...
+
+package chain
+
+import "github.com/ava-labs/hypersdk/codec"
+
+type Registry interface {
+	ActionRegistry() *codec.TypeParser[Action]
+	AuthRegistry() *codec.TypeParser[Auth]
+	OutputRegistry() *codec.TypeParser[codec.Typed]
+}
```
joshua-kim, line 12: Why do we need an interface for this? I think we can avoid this and just export `registry` + the fields on it.

joshua-kim, line 12: 1. Why do we need this interface? It feels like this could be a simple struct that just wraps the individual codecs instead of an interface
2. Why even couple the codecs at all w/ a wrapper type instead of passing around each codec individually?

vm/vm.go
```diff
@@ -67,6 +66,8 @@ const (
 )
 
 type VM struct {
+	chain.Registry
```
joshua-kim, line 69: nit: Don't embed this type

vm/resolutions.go
```diff
@@ -75,6 +63,18 @@ func (vm *VM) Logger() logging.Logger {
...
+}
+
+func (vm *VM) AuthCodec() *codec.TypeParser[chain.Auth] {
+	return vm.Registry.AuthRegistry()
+}
+
+func (vm *VM) OutputCodec() *codec.TypeParser[codec.Typed] {
+	return vm.Registry.OutputRegistry()
+}
+
```
joshua-kim, line 77: nit: undo the move

PR #1631, opened 2024-10-03
refactor marshalActions implementation

tsachiherman:
## What's in this PR ?

This PR refactor the transaction's marshaling implementation, and provide separation between the transaction and the inner actions. ( the latter is needed for https://github.com/ava-labs/hypersdk/pull/1610 )

## What are the changes in this PR ?

* a pseudo-type was created `Actions` that wraps `[]Action`.
* `Actions` supports `Size()` and `marshalInto()`.
* The transaction marshaling was unified so that the actual marshaling always takes place in `transaction.marshal()` which is being called by either `transaction.Digest()` or `transaction.Marshal()`. The difference is whether it's being called without/with the signature.
* The `transaction.Sign` method would no longer change the `Auth` member of the receiver object. As before, the returned transaction would contain the `Auth`.

## Additional notes:

Making of `UnmarshalActions` exported was to support upcoming https://github.com/ava-labs/hypersdk/pull/1610.


chain/transaction_test.go
```diff
@@ -117,7 +117,8 @@ func TestMarshalUnmarshal(t *testing.T) {
 
 	signedTx, err := tx.Sign(factory, actionRegistry, authRegistry)
 	require.NoError(err)
-
+	require.Nil(tx.Auth)
+	require.NotNil(signedTx.Auth)
```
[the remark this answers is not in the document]
tsachiherman: done

chain/transaction.go
```diff
@@ -28,71 +28,82 @@ var (
...
-		if err != nil {
-			return nil, err
-		}
-		size += consts.ByteLen + actionSize
+
+	if actionsSize, err := t.Actions.Size(); err != nil {
+		return nil, err
+	} else {
+		size += actionsSize
 	}
```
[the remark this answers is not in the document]
tsachiherman: done

chain/transaction.go
```diff
@@ -28,71 +28,82 @@ var (
...
-	id        ids.ID
-	stateKeys state.Keys
-}
+		digest    []byte
+		bytes     []byte
+		size      int
+		id        ids.ID
+		stateKeys state.Keys
+	}
+)
```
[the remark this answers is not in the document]
tsachiherman: done

chain/transaction.go
```diff
@@ -28,71 +28,82 @@ var (
...
-		}
+	if err := t.marshal(p, false); err != nil {
+		return nil, err
 	}
+
 	return p.Bytes(), p.Err()
 }
 
+// Sign sign the transaction using the provided AuthFactory. The pointer receiver is not being modified by this method,
+// and a new Transaction object would be returned. On success, the returned transaction is the signed transaction.
```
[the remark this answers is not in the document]
tsachiherman: done

chain/transaction.go
```diff
@@ -28,71 +28,82 @@ var (
...
 		return nil, err
 	}
+	// sign the preimage
 	auth, err := factory.Sign(msg)
 	if err != nil {
 		return nil, err
 	}
-	t.Auth = auth
+
+	// fill in the auth on the signed transaction
```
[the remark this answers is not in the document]
tsachiherman: done

chain/transaction.go
```diff
@@ -28,71 +28,82 @@ var (
...
 	msg, err := t.Digest()
 	if err != nil {
 		return nil, err
 	}
+	// sign the preimage
 	auth, err := factory.Sign(msg)
 	if err != nil {
 		return nil, err
 	}
-	t.Auth = auth
```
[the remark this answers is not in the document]
tsachiherman: I figured it would be safer to create a new object for the marshaling rather than (set auth/marshal/clear auth).
i.e. the Sign method parameters suggest it would be a readonly method, as it returns the signed transaction.

chain/transaction.go
```diff
@@ -31,68 +31,71 @@ var (
 type Transaction struct {
 	Base *Base `json:"base"`
 
-	Actions []Action `json:"actions"`
-	Auth    Auth     `json:"auth"`
+	Actions Actions `json:"actions"`
+	Auth    Auth    `json:"auth"`
 
-	digest    []byte
+	preimage  []byte
```
aaronbuchwald, line 37: could we rename this field to `unsignedBytes` and the accessor function to `UnsignedBytes()` ?

chain/transaction.go
```diff
@@ -31,68 +31,71 @@ var (
...
 	return &Transaction{
 		Base:    base,
 		Actions: actions,
 	}
 }
 
-func (t *Transaction) Digest() ([]byte, error) {
-	if len(t.digest) > 0 {
-		return t.digest, nil
+// Digest returns the byte slice representation of the unsigned tx
```
aaronbuchwald, line 51: nit: update the comment to match the function name

chain/transaction.go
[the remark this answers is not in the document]
aaronbuchwald: Yup, much better w/ this change

internal/gossiper/proposer.go
```diff
@@ -193,7 +193,7 @@ func (g *Proposer) HandleAppGossip(ctx context.Context, nodeID ids.NodeID, msg [
 	var seen int
 	for _, tx := range txs {
 		// Verify signature async
-		txDigest, err := tx.Digest()
+		txDigest, err := tx.Preimage()
```
aaronbuchwald, line 196: can we update the local variable names where we're updating the function name so that they match and we get rid of the misnomer? Same comment throughout.

chain/transaction_test.go
```diff
@@ -115,17 +115,20 @@ func TestMarshalUnmarshal(t *testing.T) {
 	err = actionRegistry.Register(&action2{}, unmarshalAction2)
 	require.NoError(err)
 
+	txBeforeSign := tx
+	require.Nil(tx.Auth)
 	signedTx, err := tx.Sign(factory, actionRegistry, authRegistry)
 	require.NoError(err)
-
+	require.Equal(txBeforeSign, tx)
```
aaronbuchwald, line 122: Is it possible for `require.Equal(txBeforeSign, tx)` to fail even if it were still modifying the tx? I think this is checking the same pointer

chain/block.go
```diff
@@ -236,7 +236,7 @@ func (b *StatefulBlock) populateTxs(ctx context.Context) error {
 
 		// Verify signature async
 		if b.vm.GetVerifyAuth() {
-			txDigest, err := tx.Digest()
+			txDigest, err := tx.Preimage()
```
aaronbuchwald, line 239: Could we rename the local variable to align with the updated function name?

aaronbuchwald (changes requested):
Could we update from `Preimage` to `UnsignedBytes()` ? `Preimage()` is a bit unclear because this type includes both the unsigned tx and signed tx, so we can be more explicit with `UnsignedBytes`.

Other than that, I think we should update the names of the local variables from `txDigest` or similar to `unsignedTxBytes` so that it matches the function name.

chain/transaction_test.go
tsachiherman: given that we don't have a deep clone method for the Transaction, I'll just create the "expected" copy of the transaction definition and compare against that.

chain/transaction.go
tsachiherman: done

chain/transaction.go
tsachiherman: done

internal/gossiper/proposer.go
tsachiherman: done

chain/block.go
```diff
@@ -236,7 +236,7 @@ func (b *StatefulBlock) populateTxs(ctx context.Context) error {
 
 		// Verify signature async
 		if b.vm.GetVerifyAuth() {
-			txDigest, err := tx.Digest()
+			txDigest, err := tx.UnsignedBytes()
```
aaronbuchwald, line 239: Can we rename `txDigest` to `unsignedTxBytes` ?

internal/gossiper/proposer.go
```diff
@@ -193,7 +193,7 @@ func (g *Proposer) HandleAppGossip(ctx context.Context, nodeID ids.NodeID, msg [
 	var seen int
 	for _, tx := range txs {
 		// Verify signature async
-		txDigest, err := tx.Digest()
+		txDigest, err := tx.UnsignedBytes()
```
aaronbuchwald, line 196: Can we rename `txDigest` to `unsignedTxBytes` ?

tests/integration/integration.go
```diff
@@ -356,7 +356,7 @@ var _ = ginkgo.Describe("[Tx Processing]", ginkgo.Serial, func() {
 			)
 			// Must do manual construction to avoid `tx.Sign` error (would fail with
 			// 0 timestamp)
-			msg, err := tx.Digest()
+			msg, err := tx.UnsignedBytes()
```
aaronbuchwald, line 359: Can we rename `msg` to `unsignedTxBytes` ?

vm/vm.go
```diff
@@ -924,7 +924,7 @@ func (vm *VM) Submit(
 
 		// Verify auth if not already verified by caller
 		if verifyAuth && vm.config.VerifyAuth {
-			msg, err := tx.Digest()
+			msg, err := tx.UnsignedBytes()
```
aaronbuchwald, line 927: Can we use `tx.Verify(ctx)` here since the next step is `tx.Auth.Verify(ctx, msg)` ?

chain/transaction_test.go
```diff
@@ -115,17 +115,43 @@ func TestMarshalUnmarshal(t *testing.T) {
...
 	require.Equal(len(signedTx.Actions), len(tx.Actions))
 	for i, action := range signedTx.Actions {
 		require.Equal(tx.Actions[i], action)
 	}
 
-	signedDigest, err := signedTx.Digest()
+	signedDigest, err := signedTx.UnsignedBytes()
 	require.NoError(err)
-	txDigest, err := tx.Digest()
+	txDigest, err := tx.UnsignedBytes()
```
aaronbuchwald, line 154: Can we rename these to `unsignedTxBytes` and `originalUnsignedTxBytes` ?

aaronbuchwald (changes requested):
Small rename request for the results of `UnsignedBytes()`, then this LGTM

vm/vm.go
aaronbuchwald: Feel free to ignore this one so it can be cleaned up when we switch to separate types for signed / unsigned txs

chain/block.go
tsachiherman: done

internal/gossiper/proposer.go
tsachiherman: done

tests/integration/integration.go
tsachiherman: done

chain/transaction_test.go
tsachiherman: done

aaronbuchwald (approved):
LGTM

PR #1635, opened 2024-10-04
Simulate a chain of actions

### LINKED ISSUE #1617: Upgrade actions simulation (read-only actions)
1. Simulate an array of actions instead of a single action to closely match the transaction (excluding fees and signing).
2. Include state key permissions in the simulation. For example, throw an error if an action requests read-only permission but attempts to write to a state key (see tState).

containerman17:
Closes #1617

### DIFF
--- a/api/jsonrpc/client.go
+++ b/api/jsonrpc/client.go
@@ -194,21 +194,25 @@ func (cli *JSONRPCClient) GetABI(ctx context.Context) (abi.ABI, error) {
 	return resp.ABI, err
 }
 
-func (cli *JSONRPCClient) Execute(ctx context.Context, actor codec.Address, action chain.Action) ([]byte, error) {
-	actionBytes, err := chain.MarshalTyped(action)
-	if err != nil {
-		return nil, fmt.Errorf("failed to marshal action: %w", err)
+func (cli *JSONRPCClient) Execute(ctx context.Context, actor codec.Address, actions []chain.Action) ([][]byte, error) {
+	actionsMarshaled := make([][]byte, 0)
+	for _, action := range actions {
+		actionBytes, err := chain.MarshalTyped(action)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal action: %w", err)
+		}
+		actionsMarshaled = append(actionsMarshaled, actionBytes)
 	}
 
 	args := &ExecuteActionArgs{
-		Actor:  actor,
-		Action: actionBytes,
+		Actor:   actor,
+		Actions: actionsMarshaled,
 	}
 
 	resp := new(ExecuteActionReply)
-	err = cli.requester.SendRequest(
+	err := cli.requester.SendRequest(
 		ctx,
-		"executeAction",
+		"execute",
 		args,
 		resp,
 	)
@@ -219,7 +223,7 @@ func (cli *JSONRPCClient) Execute(ctx context.Context, actor codec.Address, acti
 		return nil, fmt.Errorf("failed to execute action: %s", resp.Error)
 	}
 
-	return resp.Output, nil
+	return resp.Outputs, nil
 }
 
 func Wait(ctx context.Context, interval time.Duration, check func(ctx context.Context) (bool, error)) error {
--- a/api/jsonrpc/server.go
+++ b/api/jsonrpc/server.go
@@ -192,7 +192,7 @@ func (j *JSONRPCServer) Execute(
 	now := time.Now().UnixMilli()
 
 	storage := make(map[string][]byte)
-	ts := tstate.New(1) //TODO: optimize if needed
+	ts := tstate.New(1)
 
 	for _, action := range actions {
 		// Get expected state keys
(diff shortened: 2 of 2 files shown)


containerman17:
Test is in a gist https://gist.github.com/containerman17/6699d6add767695740f4ed5af11ffb46
Put the test in `examples/morpheusvm/tests/e2e/e2e_test.go` and run `./scripts/run.sh --ginkgo.focus=MorpheusVM`.

api/jsonrpc/server.go
```diff
@@ -175,58 +175,66 @@ func (j *JSONRPCServer) Execute(
...
-	stateKeysWithPermissions := action.StateKeys(args.Actor)
+	storage := make(map[string][]byte)
+	ts := tstate.New(1)
 
-	// flatten the map to a slice of keys
-	storageKeysToRead := make([][]byte, 0)
-	for key := range stateKeysWithPermissions {
-		storageKeysToRead = append(storageKeysToRead, []byte(key))
-	}
+	for _, action := range actions {
```
aaronbuchwald, line 192: Could we update this to exactly match the semantics in `chain/processor.go` to make sure that the behavior aligns with tx execution?

In tx execution, we gather all of the storage keys from all included actions first, create the tstate view instance, and then iterate and execute each action.

This results in a change in behavior between tx execution and simulation where in actual tx execution the state keys provided by action 1 will be made available to action 2.

api/jsonrpc/client.go
```diff
@@ -194,21 +194,25 @@ func (cli *JSONRPCClient) GetABI(ctx context.Context) (abi.ABI, error) {
 	return resp.ABI, err
 }
 
-func (cli *JSONRPCClient) Execute(ctx context.Context, actor codec.Address, action chain.Action) ([]byte, error) {
-	actionBytes, err := chain.MarshalTyped(action)
-	if err != nil {
-		return nil, fmt.Errorf("failed to marshal action: %w", err)
+func (cli *JSONRPCClient) Execute(ctx context.Context, actor codec.Address, actions []chain.Action) ([][]byte, error) {
+	actionsMarshaled := make([][]byte, 0)
```
aaronbuchwald, line 198: nit: can we pre-allocate where we already know the length of the slice? Same comment throughout

api/jsonrpc/client.go
```diff
@@ -194,21 +194,25 @@ func (cli *JSONRPCClient) GetABI(ctx context.Context) (abi.ABI, error) {
...
+		Actor:   actor,
+		Actions: actionsMarshaled,
 	}
 
 	resp := new(ExecuteActionReply)
-	err = cli.requester.SendRequest(
+	err := cli.requester.SendRequest(
 		ctx,
-		"executeAction",
+		"execute",
```
aaronbuchwald, line 215: Could we rename to `executeActions` ?

api/jsonrpc/server.go
```diff
@@ -175,58 +175,66 @@ func (j *JSONRPCServer) Execute(
 	defer span.End()
 
 	actionRegistry := j.vm.ActionRegistry()
-	action, err := (*actionRegistry).Unmarshal(codec.NewReader(args.Action, len(args.Action)))
-	if err != nil {
-		return fmt.Errorf("failed to unmashal action: %w", err)
+	actions := make([]chain.Action, 0)
```
aaronbuchwald, line 178: same comment on pre-allocating where we already know the length of `args.Actions`

api/jsonrpc/server.go
```diff
@@ -175,58 +175,66 @@ func (j *JSONRPCServer) Execute(
...
+		// Get expected state keys
+		stateKeysWithPermissions := action.StateKeys(args.Actor)
 
-	storage := make(map[string][]byte)
-	values, errs := j.vm.ReadState(ctx, storageKeysToRead)
-	for _, err := range errs {
-		if err != nil && !errors.Is(err, database.ErrNotFound) {
-			return fmt.Errorf("failed to read state: %w", err)
+		// flatten the map to a slice of keys
+		storageKeysToRead := make([][]byte, 0)
```
aaronbuchwald, line 197: same comment on pre-allocating where we already know the size of `stateKeysWithPermissions`

api/jsonrpc/server.go
```diff
@@ -175,58 +175,66 @@ func (j *JSONRPCServer) Execute(
...
-	}
+		tsv := ts.NewView(stateKeysWithPermissions, storage)
+
+		output, err := action.Execute(
+			ctx,
+			j.vm.Rules(now),
+			tsv,
+			now,
+			args.Actor,
+			ids.Empty,
```
aaronbuchwald, line 223: This may cause issues for actions that rely on the `actionID` providing a unique ID.

Could we prevent this by using `CreateActionID(ids.Empty, actionIndex)` ? This should work because it will never conflict with a transaction that was actually issued onchain (there should never be a transaction with a zero value hash) and each action in the simulate API will get a unique `actionIndex` value. I think this should give us the best we can do here ie. on-chain execution results may differ with the actual `actionID`, but the `actionID` passed in to the simulation will always be unique.

aaronbuchwald (changes requested):

api/jsonrpc/server.go
aaronbuchwald: Can leave this for now.. if we see issues with the discrepancy in the future we can update it. May also need to add the state keys from state manager in the future.

aaronbuchwald (approved):

PR #1636, opened 2024-10-04
Decouple Spam Functionality from CLI into Throughput Package

samliok:
This PR moves the spam functionality from the CLI into a new `throughput` package. The changes aim to allow us to use spam testing functionality in e2e as well as the cli. 

## Changes:

* Created a new `throughput` package to house spam-related functionality
* Moved spam logic from `cli/spam.go` into the new `throughput` package
  * Kept most of the functionality the same, but split into separate files for clarity
* Updated `examples/morpheusvm` to use the new `throughput` package
* Modified `tests/e2e/e2e.go` to incorporate spam testing using the `throughput` package


throughput/spam.go
```diff
@@ -0,0 +1,454 @@
...
+	}
+
+	webSocketClient, err := ws.NewWebSocketClient(s.uris[0], ws.DefaultHandshakeTimeout, pubsub.MaxPendingMessages, pubsub.MaxReadMessageSize) // we write the max read
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	p := &pacer{ws: webSocketClient}
+	go p.Run(ctx, s.minTxsPerSecond)
+	// This helps from hanging
+	time.Sleep(3 * time.Second)
```
samliok, line 360: need to debug, but something about this logic was finicky with the previous code

throughput/spam.go
```diff
@@ -0,0 +1,428 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+//nolint:gosec
+package throughput
```
samliok, line 5: turned off because gosec calls out math.rand for being weak

```
throughput/spam.go:80:14: G404: Use of weak random number generator (math/rand or math/rand/v2 instead of crypto/rand) (gosec)
```

throughput/spam.go
```diff
@@ -0,0 +1,427 @@
...
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		factories[i] = f
+
+		// Send funds
+		actions := sh.GetTransfer(pk.Address, distAmount, uniqueBytes())
+		// TODO: shouldn't this be using the factory from the account? rather the root key factory?
+		_, tx, err := cli.GenerateTransactionManual(parser, actions, factory, feePerTx)
+		if err != nil {
```
aaronbuchwald, line 351: We're using the root key factory to send funds from the rootKey to the newly created auth factories

throughput/spam.go
aaronbuchwald: Could you add a TODO with enough detail for the auto-generated GitHub issue so that we can follow up?

throughput/spam.go
```diff
@@ -0,0 +1,427 @@
...
+	cctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	for _, issuer := range issuers {
+		issuer.Start(cctx)
+	}
+
+	// set logging
+	t := issuers[0].logStats(cctx)
+	defer t.Stop()
+
```
aaronbuchwald, line 164: Can we move cleanup (`t.Stop()`) into `logStats` by cancelling through `cctx` ?

throughput/spam.go
```diff
@@ -0,0 +1,427 @@
...
+		z = rand.NewZipf(s.zipfSeed, s.sZipf, s.vZipf, uint64(s.numAccounts)-1)
+
+		it                      = time.NewTimer(0)
+		currentTarget           = min(s.txsPerSecond, s.minTxsPerSecond)
+		consecutiveUnderBacklog int
+		consecutiveAboveBacklog int
+
+		stop bool
+	)
+	utils.Outf("{{cyan}}initial target tps:{{/}} %d\n", currentTarget)
```
aaronbuchwald, line 177: (followup) Could we break the section broadcasting txs into its own function as well? Separating the spammer set up into functions much easier to review

throughput/spam.go
```diff
@@ -0,0 +1,427 @@
...
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+	"github.com/ava-labs/hypersdk/fees"
+	"github.com/ava-labs/hypersdk/pubsub"
+	"github.com/ava-labs/hypersdk/utils"
+)
+
+const (
+	amountToTransfer               = 1
```
aaronbuchwald, line 34: Is there a reason for `amountToTransfer` replacing `feePerTx` in a few places?

throughput/spam.go
```diff
@@ -0,0 +1,427 @@
...
+				currentTarget = min(currentTarget+s.txsPerSecondStep, s.txsPerSecond)
+				utils.Outf("{{cyan}}increasing target tps:{{/}} %d\n", currentTarget)
+				consecutiveUnderBacklog = 0
+			} else if currentTarget == s.txsPerSecond && consecutiveUnderBacklog >= successfulRunsToIncreaseTarget {
+				if terminate {
+					utils.Outf("{{green}}reached target tps:{{/}} %d\n", currentTarget)
+					// Cancel the context to stop the issuers
+					cancel()
+				}
+			}
```
aaronbuchwald, line 260: Could we add `terminate` as the first condition in the `if else` branch and add a comment here explaining why this was added (I assume to terminate once hitting desired TPS for the e2e tests) ?

throughput/spam.go
samliok: where do you see it replacing `feePerTx`? I only added it and it shouldn't be replacing it anywhere

aaronbuchwald: Ah I misread, is this just to account for the balance transfer that wasn't accounted for before here: https://github.com/ava-labs/hypersdk/blob/b8104aca01c2025dfb00f847d49e3a4d5d36648c/cli/spam.go#L372 ?

tests/e2e/e2e.go
```diff
@@ -109,6 +115,39 @@ var _ = ginkgo.Describe("[HyperSDK Tx Workloads]", func() {
 	})
 })
 
+var _ = ginkgo.Describe("[HyperSDK Spam Workloads]", func() {
+	ginkgo.It("Spam Workload", func() {
+		if spamKey == nil || spamHelper == nil {
+			return
+		}
```
aaronbuchwald, line 122: (followup) add a TODO to remove this check and make the spamKey/spamHelper required

aaronbuchwald (approved):

throughput/spam.go
samliok: yep!

PR #1640, opened 2024-10-08
Break the unsigned transaction data out of the transaction

tsachiherman:
## What ?

Break the existing `Transaction` struct into `Transaction` and an embedded struct `TransactionData` 

## Why ?

The usage of `Transaction` might not be sufficient to distinguish between the sign and the unsigned transaction.
The unsigned transaction ( i.e. `TransactionData` ) should be used only in the case of a new transaction being created, and before it's being signed.
The signed transaction should be used everywhere else. For that reason, the `Transaction` doesn't come with any constructor of it's own ( only marshaler/unmarshaler ).


chain/transaction.go
```diff
@@ -48,25 +27,24 @@ func NewTx(base *Base, actions Actions) *Transaction {
 	}
 }
 
-// UnsignedBytes returns the byte slice representation of the unsigned tx
-func (t *Transaction) UnsignedBytes() ([]byte, error) {
-	if len(t.unsignedBytes) > 0 {
-		return t.unsignedBytes, nil
+// Bytes returns the byte slice representation of the tx
+func (t *Transaction) Bytes() ([]byte, error) {
```
joshua-kim, line 31: Seems to me like the diff in this function isn't needed

chain/transaction.go
```diff
@@ -275,136 +160,21 @@ func EstimateUnits(r Rules, actions Actions, authFactory AuthFactory) (fees.Dime
...
 	}
-	return t.marshal(p, true)
+	return t.marshal(p)
 }
 
-func (t *Transaction) marshal(p *codec.Packer, marshalSignature bool) error {
+func (t *Transaction) marshal(p *codec.Packer) error {
 	t.Base.Marshal(p)
-	if err := t.Actions.marshalInto(p); err != nil {
+	if err := p.Err(); err != nil {
```
joshua-kim, line 173: I think this diff isn't needed since `Packer` will stop at the first error ([ref](https://github.com/ava-labs/avalanchego/blob/0c4efd743e1d737f4e8970d0e0ebf229ea44406c/utils/wrappers/errors.go#L13)).

chain/signedtxn.go
```diff
@@ -0,0 +1,343 @@
...
+
+	internalfees "github.com/ava-labs/hypersdk/internal/fees"
+)
+
+var (
+	_ emap.Item    = (*SignedTransaction)(nil)
+	_ mempool.Item = (*SignedTransaction)(nil)
+)
+
+type SignedTransaction struct {
```
joshua-kim, line 31: I'm not sure how I feel about this naming. I personally get pretty confused seeing `SignedTransaction` and `Transaction` as I keep assuming `Transaction` is signed. We could consider something else like `SignedTransaction` and `UnsignedTransaction` or `Transaction` and `TransactionData`.

chain/signedtxn.go
```diff
@@ -0,0 +1,343 @@
...
+
+	Auth Auth `json:"auth"`
+
+	bytes     []byte
+	size      int
+	id        ids.ID
+	stateKeys state.Keys
+}
+
+func (t *SignedTransaction) UnsignedTxnBytes() []byte { return t.Transaction.bytes }
```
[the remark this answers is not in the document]
tsachiherman: Regarding the second one, the main advantage in having the `UnsignedTxnBytes` in `SignedTransaction` is that it doesn't need to return a potential `error`. The error case was already handled during the unmarshaler, so no error handling is required by the caller.

chain/signedtxn.go
tsachiherman: It looks like we have only several places the use the unsigned transaction. Therefore, I'm not really opinionated here. I've changed the following:
SignedTransaction -> Transaction
Transaction -> TransactionData

chain/transaction.go
tsachiherman: I see.. so you're saying that adding this would have no practical functional impact since the packet would effectively be "blocked" for any new additions due to the error.

chain/transaction.go
```diff
@@ -105,69 +102,106 @@ func (t *Transaction) Sign(
...
 
-func (t *Transaction) Size() int { return t.size }
+	bytes     []byte
+	size      int
+	id        ids.ID
+	stateKeys state.Keys
+}
 
-func (t *Transaction) ID() ids.ID { return t.id }
+func (s *Transaction) Bytes() []byte { return s.bytes }
```
aaronbuchwald, line 159: Can we change from using `(s *Transaction)` to `(t *Transaction)` to follow the convention of using the lowercase first letter of the type?

aaronbuchwald (changes requested):
Few small nits and then this LGTM - thanks for cleaning this up, this makes it way easier to reason about the fields of a transaction that have been populated

aaronbuchwald (approved):

PR #1647, opened 2024-10-09
Remove marshaling from transfer action

samliok:
This is not needed any more because we can just use the ABI package to provide the correct format

### DIFF
--- a/examples/morpheusvm/vm/vm.go
+++ b/examples/morpheusvm/vm/vm.go
@@ -33,7 +33,7 @@ func init() {
 	errs.Add(
 		// When registering new actions, ALWAYS make sure to append at the end.
 		// Pass nil as second argument if manual marshalling isn't needed (if in doubt, you probably don't)
-		ActionParser.Register(&actions.Transfer{}, actions.UnmarshalTransfer),
+		ActionParser.Register(&actions.Transfer{}, nil),
 
 		// When registering new auth, ALWAYS make sure to append at the end.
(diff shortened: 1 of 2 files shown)


examples/morpheusvm/vm/vm.go
```diff
@@ -33,7 +33,7 @@ func init() {
 	errs.Add(
 		// When registering new actions, ALWAYS make sure to append at the end.
 		// Pass nil as second argument if manual marshalling isn't needed (if in doubt, you probably don't)
-		ActionParser.Register(&actions.Transfer{}, actions.UnmarshalTransfer),
+		ActionParser.Register(&actions.Transfer{}, nil),
```
samliok, line 36: confused by the register function. It uses avalanchego's linearcodec, so where does the new abi package play into here? 

https://github.com/ava-labs/hypersdk/blob/a230d1bbd8c152931c26593eb479c912b75cdc68/codec/type_parser.go#L40

containerman17: It is responsible for defining structs so they can be transferred between languages. If all implementations were in Golang, we wouldn't need an ABI.

containerman17 (changes requested):
I support this, but we need to include the example of marshaling somewhere in the docs.

aaronbuchwald (approved):
LGTM

aaronbuchwald:
> I support this, but we need to include the example of marshaling somewhere in the docs.

I'm alright with dropping this from the docs for now. I don't think anybody should be using this for now and we can either deprecate it completely or add to the docs in the future.

PR #1648, opened 2024-10-10
Add boolean support to the ABI spec

containerman17:
Golang's ABI implementation supports it, but it is not part of the spec, even though it's mentioned in the readme. Adding test cases so [hypersdk-client](https://www.npmjs.com/package/hypersdk-client) can follow up with the implementation.


containerman17:
Implemented in Typescript: https://github.com/ava-labs/hypersdk-client/pull/28

PR #1657, opened 2024-10-10
Include Spam Config Defaults

samliok:
Addressing a follow up from #1636 . This gives the spam script some basic defaults and cleans the code up a little bit.

### DIFF
--- a/throughput/config.go
+++ b/throughput/config.go
@@ -5,7 +5,7 @@ package throughput
 
 import "github.com/ava-labs/hypersdk/auth"
 
-type SpamConfig struct {
+type Config struct {
 	uris             []string
 	key              *auth.PrivateKey
 	sZipf            float64
@@ -17,11 +17,11 @@ type SpamConfig struct {
 	numAccounts      int
 }
 
-func DefaultSpamConfig(
+func NewDefaultConfig(
 	uris []string,
 	key *auth.PrivateKey,
-) *SpamConfig {
-	return &SpamConfig{
+) *Config {
+	return &Config{
 		uris:             uris,
 		key:              key,
 		sZipf:            1.01,
@@ -34,7 +34,7 @@ func DefaultSpamConfig(
 	}
 }
 
-func NewSpamConfig(
+func NewConfig(
 	uris []string,
 	key *auth.PrivateKey,
 	sZipf float64,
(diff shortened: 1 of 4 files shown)


throughput/config.go
```diff
@@ -0,0 +1,59 @@
...
+		key:              key,
+		sZipf:            1.01,
+		vZipf:            2.7,
+		txsPerSecond:     500,
+		minTxsPerSecond:  100,
+		txsPerSecondStep: 200,
+		numClients:       10,
+		numAccounts:      25,
+	}
+}
```
aaronbuchwald, line 35: Could we rename this to `NewDefaultConfig` since it's creating a new instance and call the constructor below?

throughput/config.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package throughput
+
+import "github.com/ava-labs/hypersdk/auth"
+
+type SpamConfig struct {
```
aaronbuchwald, line 8: nit: rename to `Config`

throughput/config.go
```diff
@@ -0,0 +1,59 @@
...
+		vZipf:            2.7,
+		txsPerSecond:     500,
+		minTxsPerSecond:  100,
+		txsPerSecondStep: 200,
+		numClients:       10,
+		numAccounts:      25,
+	}
+}
+
+func NewSpamConfig(
```
aaronbuchwald, line 37: nit: `NewConfig`

throughput/spam.go
```diff
@@ -71,29 +71,30 @@ type Spammer struct {
...
-func NewSpammer(
-	uris []string,
-	key *auth.PrivateKey,
-	balance uint64,
-	sZipf, vZipf float64,
-	txsPerSecond, minTxsPerSecond, txsPerSecondStep, numClients, numAccounts int,
-) *Spammer {
+func NewSpammer(sc *SpamConfig,
+	sh SpamHelper,
+) (*Spammer, error) {
```
aaronbuchwald, line 76: nit: could we make this consistent by either putting both params on a single line (there's not that many anymore) or give each their own line?

aaronbuchwald (changes requested):
Couple small nits, then this LGTM

throughput/spam.go
samliok: oops, yea this should be on its own line

aaronbuchwald (approved):

PR #1658, opened 2024-10-11
Decouple State Manager

RodrigoVillar:
This PR decouples `StateManger` into `BalanceHandler` and `MetadataManager`. In particular, this PR makes explicit the difference between HyperSDK-specific state prefixes and VM-specific state prefixes. 

Summary of Changes:
- Remove `StateManager`
- Add the `metadata` package which gives VM a default `MetadataManager` which they can use
    - Provides a `HasConflictingPrefixes()` sanity function which checks for any prefix conflicts between the HyperSDK-specific state prefixes and the VM-specific prefixes
 - Updates `vm.New()` by having developers pass in both `BalanceHandler` and `MetadataManager` rather than `StateManager` 
 - Updates dependencies which previously used `StateManager` to use either `BalanceHandler` or `MetadataManager` 

This PR was originally #1615, but I decided to close it as the original design of that PR differs from this PR.

### DIFF
--- a/examples/morpheusvm/vm/client.go
+++ b/examples/morpheusvm/vm/client.go
@@ -13,7 +13,6 @@ import (
 	"github.com/ava-labs/hypersdk/chain"
 	"github.com/ava-labs/hypersdk/codec"
 	"github.com/ava-labs/hypersdk/examples/morpheusvm/consts"
-	"github.com/ava-labs/hypersdk/examples/morpheusvm/storage"
 	"github.com/ava-labs/hypersdk/genesis"
 	"github.com/ava-labs/hypersdk/requester"
 	"github.com/ava-labs/hypersdk/utils"
@@ -116,10 +115,6 @@ func (*Parser) OutputCodec() *codec.TypeParser[codec.Typed] {
 
 func (*Parser) AuthCodec() *codec.TypeParser[chain.Auth] {
 	return AuthParser
-}
-
-func (*Parser) StateManager() chain.StateManager {
-	return &storage.StateManager{}
 }
 
 func NewParser(genesis *genesis.DefaultGenesis) chain.Parser {
--- a/internal/gossiper/dependencies.go
+++ b/internal/gossiper/dependencies.go
@@ -33,7 +33,8 @@ type VM interface {
 	Rules(int64) chain.Rules
 	Submit(ctx context.Context, verify bool, txs []*chain.Transaction) []error
 	GetAuthBatchVerifier(authTypeID uint8, cores int, count int) (chain.AuthBatchVerifier, bool)
-	StateManager() chain.StateManager
+	BalanceHandler() chain.BalanceHandler
+	MetadataManager() chain.MetadataManager
 
 	RecordTxsGossiped(int)
 	RecordSeenTxsReceived(int)
--- a/internal/gossiper/dependencies.go
+++ b/internal/gossiper/dependencies.go
@@ -33,8 +33,6 @@ type VM interface {
 	Rules(int64) chain.Rules
 	Submit(ctx context.Context, verify bool, txs []*chain.Transaction) []error
 	GetAuthBatchVerifier(authTypeID uint8, cores int, count int) (chain.AuthBatchVerifier, bool)
-	BalanceHandler() chain.BalanceHandler
-	MetadataManager() chain.MetadataManager
 
 	RecordTxsGossiped(int)
 	RecordSeenTxsReceived(int)
--- a/state/metadata/state_manager.go
+++ b/state/metadata/state_manager.go
@@ -27,6 +27,18 @@ type MetadataManager struct {
 	timestampPrefix []byte
 }
 
+func NewManager(
+	heightPrefix []byte,
+	feePrefix []byte,
+	timestampPrefix []byte,
+) MetadataManager {
+	return MetadataManager{
+		heightPrefix:    heightPrefix,
+		feePrefix:       feePrefix,
+		timestampPrefix: timestampPrefix,
+	}
+}
+
 func NewDefaultManager() MetadataManager {
 	return MetadataManager{
 		heightPrefix:    []byte{defaultHeightPrefix},
--- a/state/metadata/state_manager.go
+++ b/state/metadata/state_manager.go
@@ -19,7 +19,7 @@ const (
 	DefaultMinimumPrefix byte = 0x3
 )
 
-var _ chain.MetadataManager = &MetadataManager{}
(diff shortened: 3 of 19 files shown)


examples/morpheusvm/vm/client.go
```diff
@@ -118,10 +117,6 @@ func (*Parser) AuthCodec() *codec.TypeParser[chain.Auth] {
 	return AuthParser
 }
 
-func (*Parser) StateManager() chain.StateManager {
-	return &storage.StateManager{}
-}
-
```
aaronbuchwald, line 124: Really nice to be able to remove this from the client where it didn't make much sense

internal/gossiper/dependencies.go
```diff
@@ -33,7 +33,8 @@ type VM interface {
 	Rules(int64) chain.Rules
 	Submit(ctx context.Context, verify bool, txs []*chain.Transaction) []error
 	GetAuthBatchVerifier(authTypeID uint8, cores int, count int) (chain.AuthBatchVerifier, bool)
-	StateManager() chain.StateManager
+	BalanceHandler() chain.BalanceHandler
+	MetadataManager() chain.MetadataManager
```
aaronbuchwald, line 37: Seems like this was actually unused in this package, could we remove it instead of renaming it?

state/metadata/state_manager.go
```diff
@@ -0,0 +1,75 @@
...
+
+	for _, p := range prefixes {
+		for vp := range verifiedPrefixes {
+			if bytes.HasPrefix(p, []byte(vp)) || bytes.HasPrefix([]byte(vp), p) {
+				return true
+			}
+		}
+
+		verifiedPrefixes.Add(string(p))
+	}
```
aaronbuchwald, line 72: Is there any reason to use `verifiedPrefixes` as a set here? It seems like we're just using it as a list. The only change this would add is we'd make sure not to check against the same verified prefix more than once, but if we hit any duplicates we should already exit early.

Could we add a short unit test here that covers the cases:
1. no conflicts
2. identical prefixes
3. prefixes contains a prefix of one of the vm prefixes
4. vm prefix contains a prefix of one of the prefixes
5. vm prefixes contains a duplicate

RodrigoVillar: Done

examples/morpheusvm/vm/client.go
RodrigoVillar: Done

internal/gossiper/dependencies.go
RodrigoVillar: Done

state/metadata/state_manager_test.go
```diff
@@ -0,0 +1,72 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package metadata_test
```
aaronbuchwald, line 4: nit: update to `metadata` should only use `_test` suffix if needed to avoid a cyclic dependency

state/metadata/state_manager_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/ava-labs/hypersdk/state/metadata"
+)
+
+func TestConflictingPrefixes(t *testing.T) {
+	require := require.New(t)
+
+	metadataManager := metadata.NewDefaultManager()
+
+	// Test no conflicts
+	require.False(
```
aaronbuchwald, line 20: Can we turn each of these cases into a test struct ie

```golang
type test struct {
	name string
	m MetadataManager
	vmPrefixes [][]byte
	hasConflicts bool
}
```

aaronbuchwald: Here's an example of this approach: https://github.com/ava-labs/hypersdk/blob/bca77398df99cce2920fb35feb5abc61f646c6cd/state/keys_test.go#L35

state/metadata/state_manager.go
aaronbuchwald: Do we need to use a set for `verifiedPrefixes` here rather than a slice?

RodrigoVillar: I used `set.Set` here as it was originally used in the state-factory PR found here (and therefore thought it was the preferred style) https://github.com/ava-labs/hypersdk/blob/a6fdf33ea39140ca0ae3bec5cd63aa51d6fa9dfd/state/layout.go#L100-L112 

I have no preference as to the DS being used and will switch to using a slice instead

state/metadata/state_manager_test.go
RodrigoVillar: Done

state/metadata/state_manager_test.go
RodrigoVillar: Done

state/metadata/state_manager.go
```diff
@@ -0,0 +1,85 @@
...
+
+	for _, p := range prefixes {
+		for _, vp := range verifiedPrefixes {
+			if bytes.HasPrefix(p, vp) || bytes.HasPrefix(vp, p) {
+				return true
+			}
+		}
+
+		verifiedPrefixes = append(verifiedPrefixes, p)
+	}
```
aaronbuchwald, line 82: hmm - not necessary for this PR since performance is not important here. Could we make this check faster?

state/metadata/state_manager_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	require.False(
+		HasConflictingPrefixes(
+			metadataManager,
+			[][]byte{
+				{DefaultMinimumPrefix},
+			},
+		),
+	)
+}
```
aaronbuchwald, line 24: Can we move this to be a test case under the other test by adding a boolean `hasConflict` to the test struct?

aaronbuchwald (changes requested):

state/metadata/state_manager.go
RodrigoVillar: Could you clarify? I'm a bit confused between the statements "performance is not important here" and "can we make this check faster?"

aaronbuchwald: This is called once on startup, so performance is not important here and therefore not needed for this PR. What I meant by "could we make this check faster" is, is there a way to eliminate duplicate work here?

I thought we could do something simpler like include the index in the first loop and check against all later indices. If we are at index i > j, then there's no need to check index i against j again because we already did. This is already what we're doing though because `verifiedPrefixes` starts out as empty, so rather than checking the first prefix against all the others and then decrease the number to check against from there, we check the first prefix against nothing and then the next prefix against one and so on.

aaronbuchwald (approved):

PR #1659, opened 2024-10-11
Remove global variables from spam script

samliok:
per pr title


throughput/spam.go
```diff
@@ -41,14 +42,6 @@ const (
...
 var (
 	maxConcurrency = runtime.NumCPU()
-	issuerWg       sync.WaitGroup
-
-	l            sync.Mutex
-	confirmedTxs uint64
-	totalTxs     uint64
-
-	inflight atomic.Int64
-	sent     atomic.Int64
```
aaronbuchwald, line 51: ❤️

aaronbuchwald: cc @joshua-kim

throughput/spam.go
```diff
@@ -92,6 +87,8 @@ func NewSpammer(sc *Config, sh SpamHelper) (*Spammer, error) {
 		txsPerSecondStep: sc.txsPerSecondStep,
 		numClients:       sc.numClients,
 		numAccounts:      sc.numAccounts,
+
+		txProcessor: &txProcessor,
```
aaronbuchwald, line 91: nit: can we make `txProcessor` a pointer where it's declared?

throughput/processor.go
```diff
@@ -0,0 +1,87 @@
...
+		// We can't error match here because we receive it over the wire.
+		if !strings.Contains(wsErr.Error(), ws.ErrExpired.Error()) {
+			utils.Outf("{{orange}}pre-execute tx failure:{{/}} %v\n", wsErr)
+		}
+	}
+	tp.totalTxs++
+	tp.l.Unlock()
+}
+
+func (tp *txProcessor) logIssuerState(ctx context.Context, issuer *issuer) {
```
aaronbuchwald, line 51: could we switch from passing the whole `issuer` in to including a reference to a client that the tx processor can use to fetch unit prices?

throughput/processor.go
```diff
@@ -0,0 +1,87 @@
...
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/ava-labs/hypersdk/api/ws"
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/utils"
+)
+
+type txProcessor struct {
```
aaronbuchwald, line 19: nit: can we rename to `tracker` ?

aaronbuchwald (changes requested):

aaronbuchwald (approved):

PR #1660, opened 2024-10-11
Indexer on disk

containerman17:


containerman17:
Will get back to it after the DSMR upgrade

PR #1661, opened 2024-10-13
state: Make state.Keys JSON human readable

### LINKED ISSUE #1649: Make `state.Keys` JSON human readable
Update the `state.Keys` JSON marshaling so that the output is human readable.

The `state.Keys` type is a map between requested keys and the corresponding permissions: https://github.com/ava-labs/hypersdk/blob/b3c991f12ec52d308cf18ed426a5390440f75979/state/keys.go#L26

We should update the json marshaling for permissions to use a switch statement between the valid named permissions and the corresponding string ie. `Read` -> `"read"`, `All` -> `"all"`, etc. If an unexpected permission is used, it should return the hex of the single byte permission, so that it's possible to see exactly what valid permissions are included if you interpret it.

This should make `state.Keys` produce a mapping between the string representation of the keys and the human readable permission.

Optional: switch from using the string included in the map to casting to `codec.Bytes` so that we produce a consistent human readable representation of arbitrary byte strings across JSON output.

najeal:
Closes #1649 

This changes the `keysJSON.Perms` field from `[8]permsJSON` to a `map[string]permsJSON`.
Known permissions are now string human readable.
Unknown Permissions are encoded as hexadecimal string.

Few questions coming after this work.
- Why are we going from an internal object having mapping `key -> perm` to a JSON having `perm -> key` (previously perm being an index)
- JSON format has a key `Perms` that seams useless + is title case:
```
{
  "Perms" : {
    "read": "xxxx"
  }
}
```


state/keys.go
```diff
@@ -75,21 +80,48 @@ func (k *Keys) UnmarshalJSON(b []byte) error {
...
+	case "none":
+		perm = None
+	default:
+		res, err := hex.DecodeString(s)
+		if err != nil || len(res) != 1 {
+			return 0, fmt.Errorf("permission %s: %w", s, errInvalidHexadecimalString)
+		}
+		perm = Permissions(res[0])
+	}
+	return perm, nil
```
aaronbuchwald, line 122: Rather than adding this function, could we implement the `TextMarshaler` and `TextUnmarshaler` interfaces on `Permissions` so that we can use `Permissions` in a type we're going to encode to JSON and it will produce the human readable format? See the second example here: https://pkg.go.dev/encoding/json

state/keys.go
```diff
@@ -59,13 +62,15 @@ func (k Keys) ChunkSizes() ([]uint16, bool) {
 type permsJSON []string
 
 type keysJSON struct {
-	Perms [8]permsJSON
+	Perms map[string]permsJSON
```
aaronbuchwald, line 65: Can we change the key type to `codec.Bytes` so that it uses the standard serialization throughout the repo for encoding the binary key values?

Once we do this, we can replace the existing `keysJSON` and `permsJSON` types with a single intermediate type `map[codec.Bytes]Permissions` and call `json.Marshal` / `json.Unmarshal` on the intermediate type

aaronbuchwald (changes requested):

state/keys.go
```diff
@@ -56,40 +61,66 @@ func (k Keys) ChunkSizes() ([]uint16, bool) {
...
 }
 
-type permsJSON []string
-
-type keysJSON struct {
-	Perms [8]permsJSON
-}
+type keysJSON map[string]Permissions
 
+// MarshalJSON marshals Keys has readable JSON.
```
aaronbuchwald, line 66:
```suggestion
// MarshalJSON marshals Keys as readable JSON.
```

state/keys.go
```diff
@@ -56,40 +61,66 @@ func (k Keys) ChunkSizes() ([]uint16, bool) {
...
+	case "allocate":
+		*p = Allocate
+	case "all":
+		*p = All
+	case "none":
+		*p = None
+	default:
+		res, err := hex.DecodeString(str)
+		if err != nil || len(res) != 1 {
+			return fmt.Errorf("permission %s: %w", str, errInvalidHexadecimalString)
```
aaronbuchwald, line 113: Could we break this into two separate error cases so that we don't return an error that includes a nil error in the case where `len(res) != 1` ?

state/keys_test.go
```diff
@@ -181,3 +182,57 @@ func TestKeysMarshalingFuzz(t *testing.T) {
...
+			require.NoError(err)
+			require.Equal(test.perm, perm)
+		})
+	}
+
+	t.Run("InvalidString", func(t *testing.T) {
+		var perm Permissions
+		err := perm.UnmarshalText([]byte{byte(1), byte(10)})
+		require.ErrorIs(err, errInvalidHexadecimalString)
+	})
```
aaronbuchwald, line 227: could we update the struct definition to include `expectedErr` and move this test case alongside the others rather than breaking it out?

aaronbuchwald (changes requested):

state/keys.go
najeal: I never include a `nil` error, we always wrap a `errInvalidHexadecimalString = errors.New("invalid hexadecimal string")` error

state/keys_test.go
najeal: done 👍

aaronbuchwald (approved):

aaronbuchwald (approved):
LGTM - thanks for putting this up 🙏

PR #1662, opened 2024-10-14
Universal CLI

containerman17:
Creates a single binary that can be used with any HyperSDK VM, as long as the VM uses standard functionality.


cmd/hypersdk-cli/README.md
```diff
@@ -0,0 +1,134 @@
...
+## Configuration
+
+The CLI stores configuration in `~/.hypersdk-cli/config.cfg`. This includes:
+- Private key
+- Endpoint URL
+
+## Global Flags
+
+- `--endpoint`: Override the default endpoint for a single command
+- `-o, --output`: Set output format (`human` or `json`)
```
aaronbuchwald, line 22: (optional nit) maybe text instead of human?

cmd/hypersdk-cli/config.go
```diff
@@ -0,0 +1,136 @@
...
+	lines := strings.Split(string(data), "\n")
+	for _, line := range lines {
+		parts := strings.SplitN(line, "=", 2)
+		if len(parts) == 2 {
+			config[strings.TrimSpace(parts[0])] = strings.TrimSpace(parts[1])
+		}
+	}
+
+	return config, nil
+}
```
aaronbuchwald, line 100: Can we switch from reading the config manually to using cobra/viper to read in a config file at the default location?

cmd/hypersdk-cli/config.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	if err := os.WriteFile(configPath, []byte(buf.String()), 0o600); err != nil {
+		return fmt.Errorf("failed to write config file: %w", err)
+	}
+
+	return nil
+}
+
+func decodeFileOrHex(whatever string) ([]byte, error) {
```
aaronbuchwald, line 126: Can we provide a descriptive variable name here? Maybe fileNameOrHex?

cmd/hypersdk-cli/key_set.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		if keyString == "" {
+			return errors.New("--key is required")
+		}
+
+		return checkAndSavePrivateKey(keyString, cmd)
+	},
+}
+
+func checkAndSavePrivateKey(keyString string, cmd *cobra.Command) error {
```
aaronbuchwald, line 47: could we reverse the order of arguments here? We seem to use `cmd` as the first argument everywhere else.

consts/consts.go
```diff
@@ -21,6 +21,7 @@ const (
 	// no more than 50 KiB of overhead but is likely much less)
 	NetworkSizeLimit = 2_044_723 // 1.95 MiB
 
+	// FIXME: should use the standard math.MaxUint8, etc.
```
aaronbuchwald, line 24: Fine for now, but let's avoid adding unrelated changes like this into larger PRs

cmd/hypersdk-cli/endpoint_set.go
```diff
@@ -0,0 +1,69 @@
...
+		if err := updateConfig("endpoint", endpoint); err != nil {
+			return fmt.Errorf("failed to update config: %w", err)
+		}
+
+		client := jsonrpc.NewJSONRPCClient(endpoint)
+		success, err := client.Ping(context.Background())
+		pingErr := ""
+		if err != nil {
+			pingErr = err.Error()
+		}
```
aaronbuchwald, line 39: (unnecessary for this PR) - I think we should support setting the endpoint to a value that is not currently running. If we're going to support that case, should we remove the ping? This currently supports it, but will print out an error result. This is not unexpected if the user knows the endpoint is not currently running, but the UX could be better without the error message.

cmd/hypersdk-cli/read.go
```diff
@@ -0,0 +1,263 @@
...
+
+	return kvPairs, nil
+}
+
+func fillFromInputData(typ abi.Type, kvData map[string]string) (map[string]interface{}, error) {
+	kvPairs := make(map[string]interface{})
+	for _, field := range typ.Fields {
+		value, ok := kvData[field.Name]
+		if !ok {
+			continue
```
aaronbuchwald, line 162: can we switch to returning an error if an unused field is specified?

cmd/hypersdk-cli/transaction.go
```diff
@@ -0,0 +1,224 @@
...
+		_, _, chainID, err := client.Network(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get network info: %w", err)
+		}
+
+		base := &chain.Base{
+			ChainID:   chainID,
+			Timestamp: time.Now().Unix()*1000 + 60*1000,
+			MaxFee:    1_000_000,
+		}
```
aaronbuchwald, line 97: could we add a TODO in the code to handle the timestamp and max fee?

We'll need to use `EstimateUnits` to calculate the max fee and will need to know the maximum expiry time from the chain's rules to make sure the timestamp is valid (will require a new API call).

cmd/hypersdk-cli/transaction.go
```diff
@@ -0,0 +1,224 @@
...
+	p = codec.NewWriter(auth.Size(), consts.NetworkSizeLimit)
+	auth.Marshal(p)
+	authBytes := []byte{auth.GetTypeID()}
+	authBytes = append(authBytes, p.Bytes()...)
+
+	// Combine everything into final signed transaction
+	//nolint:gocritic //append is fine here
+	signedBytes := append(unsignedBytes, authBytes...)
+	return signedBytes, nil
+}
```
aaronbuchwald, line 224: Rather than having tx bytes construction logic here, could we add the required helper into `chain/transaction.go` ?

cmd/hypersdk-cli/transaction.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+
+		wsClient, err := ws.NewWebSocketClient(endpoint, 10*time.Second, 100, 1000000)
+		if err != nil {
+			return fmt.Errorf("failed to create ws client: %w", err)
+		}
+
+		if err := wsClient.RegisterRawTx(signedBytes); err != nil {
+			return fmt.Errorf("failed to register tx: %w", err)
+		}
```
aaronbuchwald, line 111: Can we switch away from using the websocket client here?

aaronbuchwald (changes requested):
🔥  this is awesome 🔥 

just tried it out locally:

```
✅ Read-only execution successful:
sender_balance: 9999999999999
receiver_balance: 1

hypersdk % ./hyper-cli tx Transfer
✔ value: 10█
memo: 12
✅ Transaction successful (txID: 26beD8aVUqmYuAgcDgdruYafHGSgnHGDYVVMrUrLSwbN5ipLj)

receiver_balance: 10
sender_balance: 9999999970990
hypersdk % ./hyper-cli tx Transfer
to: 0x000000000000000000000000000000000000000000000000000000000000000000
value: 1
memo: 1222
✅ Transaction successful (txID: 2hZNKedNDMCTCi6oSVcYNZbpaajSnBJ1ABexYKuBye7sgT3sNz)

sender_balance: 9999999941889
receiver_balance: 11
```

this is a game changer

aaronbuchwald (approved):

PR #1664, opened 2024-10-15
Move Common Test Boilerplate out of Morpheus VM

samliok:
Started writing e2e tests for the updated `vmwithcontracts` and noticed a bunch of common boilerplate code that should be separated out. I think this will also help make things clear for the larger e2e test refactor

PR also removes the tests package from `vmwithcontracts` since its the same as morpheus vm


examples/morpheusvm/tests/workload/workload.go
```diff
@@ -22,78 +21,27 @@ import (
...
-			panic(err)
-		}
-		priv := ed25519.PrivateKey(privBytes)
-		ed25519PrivKeys[i] = priv
-		ed25519AuthFactories[i] = auth.NewED25519Factory(priv)
-		addr := auth.NewED25519Address(priv.PublicKey())
-		ed25519Addrs[i] = addr
-	}
-}
-
```
samliok, line 60: removing this and the genesis configuration is an improvement imo. VM developers are probably going to be fine with these defaults, and if not we can expose a way for them to customize. 

Still some more work to be done cleaning up the `workloadfactory` or removing it all together.

PR #1665, opened 2024-10-16
chain: add Transaction JSON Marshal Unmarshal

### LINKED ISSUE #1633: Add Tx JSON Marshal Support
- include txID in tx JSON
- marshal actions and auth as bytes and encode them as hex via codec.Bytes
- Add TransactionJSON intermediate type that supports MarshalJSON and UnmarshalJSON and stores the action/auth as bytes
- add MarshalJSON implementation to chain.Transaction
- add UnmarshalJSON that takes `chain.Parser` in as a parameter so that it can parse the actions/auth

najeal:
Close #1633 

Few notes:
- After a Marshal -> Unmarshal, the Transaction looses the `Base` data and internals like `bytes`, `size`, `stateKeys`.

- The tests require the file to be part of `chain_test` package instead of `chain`, otherwise it ends up with import cycle (seems to come from `auth` package).
So the created test is outside the `chain` package and `Transaction.id` is unreachable. As a workaround I created a `SetID` function for testing purpose but it is not ideal. One other solution can be to make the `id` a public field.

There are also multiple options to avoid import cycle while testing in `chain` package:
1. Move `Auth` `Rule` `Action` `Object` `Marshaler` `AuthFactory` `AuthbatchVerifier` in a separate dedicated package.
2. Mock the objects coming from `auth` package used in the tests.


chain/transaction.go
```diff
@@ -162,6 +163,8 @@ func (t *Transaction) Size() int { return t.size }
 
 func (t *Transaction) ID() ids.ID { return t.id }
 
+func (t *Transaction) SetID(id ids.ID) { t.id = id }
```
aaronbuchwald, line 166: We should avoid exporting a function that overrides the ID of the transaction since it should always match the contents of the transaction

chain/transaction.go
```diff
@@ -359,6 +362,92 @@ func (t *Transaction) Marshal(p *codec.Packer) error {
 	return t.marshal(p)
 }
 
+func (t *Transaction) MarshalJSON() ([]byte, error) {
+	type txJSON struct {
+		ID      ids.ID `json:"id"`
+		Actions []byte `json:"actions"`
+		Auth    []byte `json:"auth"`
+	}
```
aaronbuchwald, line 370: Can we include `Base` here?

chain/transaction.go
```diff
@@ -359,6 +362,92 @@ func (t *Transaction) Marshal(p *codec.Packer) error {
...
+		return nil, err
+	}
+	authPacker := codec.NewWriter(0, consts.NetworkSizeLimit)
+	authPacker.PackByte(t.Auth.GetTypeID())
+	t.Auth.Marshal(authPacker)
+	if err := authPacker.Err(); err != nil {
+		return nil, err
+	}
+
+	hexActions, err := codec.Bytes(actionsPacker.Bytes()).MarshalText()
```
aaronbuchwald, line 387: We don't need to call `MarshalText()` on this. We can just pass in the `codec.Bytes` type directly and this will be handled in `json.Marshal`

chain/transaction.go
```diff
@@ -359,6 +362,92 @@ func (t *Transaction) Marshal(p *codec.Packer) error {
 	return t.marshal(p)
 }
 
+func (t *Transaction) MarshalJSON() ([]byte, error) {
+	type txJSON struct {
+		ID      ids.ID `json:"id"`
+		Actions []byte `json:"actions"`
+		Auth    []byte `json:"auth"`
```
aaronbuchwald, line 369: These types should be `codec.Bytes` so that marshalling them to hex is handled by `json.Marshal`

chain/transaction.go
```diff
@@ -359,6 +362,92 @@ func (t *Transaction) Marshal(p *codec.Packer) error {
...
+		Auth:    hexAuth,
+	})
+}
+
+func (t *Transaction) UnmarshalJSON(data []byte, parser Parser) error {
+	type txJSON struct {
+		ID      ids.ID `json:"id"`
+		Actions []byte `json:"actions"`
+		Auth    []byte `json:"auth"`
+	}
```
aaronbuchwald, line 408: Can we move the type definition outside of these functions so we don't need to duplicate it?

chain/transaction.go
```diff
@@ -359,6 +362,92 @@ func (t *Transaction) Marshal(p *codec.Packer) error {
...
+	var actionsBytes, authBytes codec.Bytes
+	err = actionsBytes.UnmarshalText(tx.Actions)
+	if err != nil {
+		return fmt.Errorf("%w: cannot unmarshal actions text", err)
+	}
+	err = authBytes.UnmarshalText(tx.Auth)
+	if err != nil {
+		return fmt.Errorf("%w: cannot unmarshal auth text", err)
+	}
+
```
aaronbuchwald, line 425: Same comment for removing `UnmarshalText` in favor of using `codec.Bytes` in the `txJSON` type

chain/transaction.go
```diff
@@ -359,6 +362,92 @@ func (t *Transaction) Marshal(p *codec.Packer) error {
...
+	if err := authReader.Err(); err != nil {
+		return fmt.Errorf("%w: auth packer", err)
+	}
+
+	t.id = tx.ID
+	t.TransactionData = TransactionData{
+		Actions: actions,
+	}
+	t.Auth = auth
+	return nil
```
aaronbuchwald, line 448: Rather than trusting the received txID (which could be a mismatch), we should re-marshal the received tx based on the fields (Base, Actions, and Auth) and unmarshal it to populate the txID.

Ideally we can restrict the number of places where we construct a transaction, so that it's clear all of the required fields are always populated.

Right now the only place we create `Transaction` is in `Sign`. We could create a new function `NewTx(base, actions, auth)` that we call within `Sign` to keep it to a single place where we construct a signed transaction.

chain/transaction_test.go
```diff
@@ -73,6 +75,58 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
...
+				&action2{
+					A: 2,
+					B: 4,
+				},
+			},
+		},
+		Auth: &auth.ED25519{
+			Signer: pk.PublicKey(),
+		},
+	}
```
aaronbuchwald, line 107: Could we use `NewTxData` and `Sign` to produce the signed transaction? This makes sure all fields are populated (currently missing the signature in the ED25519 auth)

aaronbuchwald (changes requested):

chain/transaction.go
```diff
@@ -80,26 +79,7 @@ func (t *TransactionData) Sign(
...
-	p := codec.NewWriter(size, consts.NetworkSizeLimit)
-	if err := signedTransaction.Marshal(p); err != nil {
-		return nil, err
-	}
-	if err := p.Err(); err != nil {
-		return nil, err
-	}
-	p = codec.NewReader(p.Bytes(), consts.MaxInt)
-	return UnmarshalTx(p, actionCodec, authCodec)
+	return newTransaction(t.Base, t.Actions, auth)
```
najeal, line 82: new, `Sign()` method uses `newTransaction()` to initialize a Transaction.
`actionCodec` and `authCodec` are now unnecessary.

chain/transaction.go
```diff
@@ -174,6 +154,28 @@ type Transaction struct {
 	stateKeys state.Keys
 }
 
+// newTransaction creates a Transaction and initializes the private fields.
+func newTransaction(base *Base, actions Actions, auth Auth) (*Transaction, error) {
```
najeal, line 158: `newTransaction` initializes `bytes` `size` and `id` straight after the codec marshal.

chain/transaction_test.go
```diff
@@ -74,6 +77,73 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
...
+	prevUnsignedBytes, err := signedTx.UnsignedBytes()
+	require.NoError(err)
+	unsignedBytes, err := txout.UnsignedBytes()
+	require.NoError(err)
+	require.NoError(txout.Auth.Verify(context.Background(), prevUnsignedBytes))
+	require.NoError(txout.Auth.Verify(context.Background(), unsignedBytes))
+}
+
 // TestMarshalUnmarshal roughly validates that a transaction packs and unpacks correctly
 func TestMarshalUnmarshal(t *testing.T) {
```
najeal, line 148: @aaronbuchwald it looks like this test is incomplete. No unmarshal is done to verify we get back the same transaction data than the original one. I may open a issue to complete it in another PR?

najeal (commented):
few remarks.

chain/transaction_test.go
```diff
@@ -130,6 +200,13 @@ func TestMarshalUnmarshal(t *testing.T) {
 	for i, action := range signedTx.Actions {
 		require.Equal(tx.Actions[i], action)
 	}
+	// validating tx ID, Bytes and Size
+	writerPacker := codec.NewWriter(0, consts.NetworkSizeLimit)
+	err = signedTx.Marshal(writerPacker)
+	require.NoError(err)
+	require.Equal(signedTx.ID(), utils.ToID(writerPacker.Bytes()))
+	require.Equal(signedTx.Bytes(), writerPacker.Bytes())
+	require.Len(writerPacker.Bytes(), signedTx.Size())
```
aaronbuchwald, line 209:
```suggestion
```

Can we remove the explicit length check right after we've already checked the bytes themselves are equivalent?

chain/transaction_test.go
```diff
@@ -130,6 +200,13 @@ func TestMarshalUnmarshal(t *testing.T) {
 	for i, action := range signedTx.Actions {
 		require.Equal(tx.Actions[i], action)
 	}
+	// validating tx ID, Bytes and Size
```
aaronbuchwald, line 203:
```suggestion
```

Could we remove this comment? Should be readable from the code itself

chain/transaction_test.go
aaronbuchwald: re: https://github.com/ava-labs/hypersdk/pull/1665/files#r1819411335 yes, we should add a check for unmarshalling the signed/unsigned tx bytes results in the identical `TxData` / `Transaction`

chain/transaction_test.go
```diff
@@ -74,6 +77,73 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
 	return &action, err
 }
 
+func TestJSONMarshalUnmarshal(t *testing.T) {
+	require := require.New(t)
+
+	txdata := chain.TransactionData{
```
aaronbuchwald, line 83: nit: camel case to `txData`

separate from this PR, but we should be consistent with the variable name throughout this file as well

chain/transaction_test.go
```diff
@@ -74,6 +77,73 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
...
+	b, err := json.Marshal(signedTx)
+	require.NoError(err)
+
+	parser := chaintest.NewParser(nil, actionCodec, authCodec, nil)
+
+	var txout chain.Transaction
+	err = txout.UnmarshalJSON(b, parser)
+	require.NoError(err)
+	// cannot check direct Equal between signedTx and txout
+	// because of the Auth codec.Address that will be different after the unmarshaling.
```
aaronbuchwald, line 132: Why is auth different after unmarshalling?

chain/transaction_test.go
```diff
@@ -74,6 +77,73 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
...
+
+	signedTx, err := txdata.Sign(factory)
+	require.NoError(err)
+
+	b, err := json.Marshal(signedTx)
+	require.NoError(err)
+
+	parser := chaintest.NewParser(nil, actionCodec, authCodec, nil)
+
+	var txout chain.Transaction
```
aaronbuchwald, line 128: nit: rename to `txFromJSON`

chain/transaction_test.go
```diff
@@ -74,6 +77,73 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
...
+	require.Equal(signedTx.Bytes(), txout.Bytes())
+	require.Equal(signedTx.Size(), txout.Size())
+	// verify txout Auth is able to verify unsigned bytes of both original tx and unmarshaled tx.
+	prevUnsignedBytes, err := signedTx.UnsignedBytes()
+	require.NoError(err)
+	unsignedBytes, err := txout.UnsignedBytes()
+	require.NoError(err)
+	require.NoError(txout.Auth.Verify(context.Background(), prevUnsignedBytes))
+	require.NoError(txout.Auth.Verify(context.Background(), unsignedBytes))
+}
```
aaronbuchwald, line 145: Rather than perform every check one by one, could we do a single `require.Equal(signedTx, txFromJSON)` ?

chain/transaction.go
```diff
@@ -174,6 +154,28 @@ type Transaction struct {
...
+	tx := Transaction{
+		TransactionData: TransactionData{
+			Base:    base,
+			Actions: actions,
+		},
+		Auth: auth,
+	}
+	p := codec.NewWriter(0, consts.NetworkSizeLimit)
```
aaronbuchwald, line 166: Could we add a call to `UnsignedBytes` before calling `tx.Marshal(p)` to make sure that the `unsignedBytes` field is initialized as before?

This is correct as is since it will be lazily populated currently, but I'd prefer to keep the old behavior of making sure all private fields are populated when creating a new instance of a transaction.

chain/transaction.go
```diff
@@ -377,6 +379,81 @@ func (t *Transaction) Marshal(p *codec.Packer) error {
 	return t.marshal(p)
 }
 
+type txJSON struct {
+	ID      ids.ID      `json:"id"`
+	Actions codec.Bytes `json:"actions"`
+	Auth    codec.Bytes `json:"auth"`
+	Base    codec.Bytes `json:"base"`
```
aaronbuchwald, line 386: Could we keep `Base` as the original type since we should be able to encode/decode this directly as JSON with no issues?

chain/transaction.go
```diff
@@ -377,6 +379,81 @@ func (t *Transaction) Marshal(p *codec.Packer) error {
...
+	authPacker := codec.NewWriter(0, consts.NetworkSizeLimit)
+	authPacker.PackByte(t.Auth.GetTypeID())
+	t.Auth.Marshal(authPacker)
+	if err := authPacker.Err(); err != nil {
+		return nil, err
+	}
+	basePacker := codec.NewWriter(0, consts.NetworkSizeLimit)
+	if t.Base != nil {
+		t.Base.Marshal(basePacker)
+	}
```
aaronbuchwald, line 407: What's the reason for this check? Base should never be nil.

aaronbuchwald (changes requested):

chain/transaction.go
najeal: it should not, but as it is a pointer I prefered to check it instead of risking a panic 🙂 but anyway we don't need to use the codec as you suggested in another comment 👍

chain/transaction_test.go
najeal: private field `addr` is not Marshalled, looking at the `address()` function is shows:
```
func (d *ED25519) address() codec.Address {
	if d.addr == codec.EmptyAddress {
		d.addr = NewED25519Address(d.Signer)
	}
	return d.addr
}
```
I see the same logic for BLS, ED25519, SECP256R1.

chain/transaction_test.go
najeal: Unfortunatly, I can't because of the `Auth.addr` being different. I thought it was expected behavior.
The way I verified `Auth` correctness, I checked it can verify both the (prev/new)unsignedBytes.

chain/transaction_test.go
aaronbuchwald: Yup, feel free to open a PR directly (no need for an issue)

chain/transaction.go
```diff
@@ -174,6 +154,31 @@ type Transaction struct {
...
+	p := codec.NewWriter(0, consts.NetworkSizeLimit)
+	if _, err := tx.UnsignedBytes(); err != nil {
+		return nil, err
+	}
+	if err := tx.Marshal(p); err != nil {
+		return nil, err
+	}
+	if err := p.Err(); err != nil {
+		return nil, err
+	}
```
aaronbuchwald, line 175: Could we move creating the codec to directly before it's used and use the unsigned bytes to calculate how much memory to pre-allocate for the transaction (same as before this should be `len(unsignedBytes) + consts.ByteLen + auth.Size)` )

aaronbuchwald (approved):

PR #1667, opened 2024-10-16
support procedural test style for VM-defined workloads

tsachiherman:
## What ?

Add the infrastructure that would allow custom-VM to implement their own integration and e2e testing, outside of the workload design pattern.

## Why ?

Custom networks might try to achieve different goals that could not easily be tested using the `workload` patten being used. The workload patten is great when we want to ensure that the custom-VM perform the hyper-SDK functionalities correctly. However, when aiming to test custom network functionalities, this testing model comes short.

## How would we solve that problem ?

Each VM would defined the following:
```golang
// TestsRegistry is global tests registry instance where all the tests are to be registered. Each VM would typically have only
// a single instance of such a registry.
var TestsRegistry = &registry.Registry{}
```

as well as one or more tests:

```golang
var _ = registry.Register(TestsRegistry, "Transfer Transaction", func(t ginkgo.FullGinkgoTInterface, tn tworkload.TestNetwork) error {
 ...
```

Then, when the e2e and integration tests would get executed, a new test node would be added `[Custom VM Tests]` which would invoke these custom test. The network would be the same network created during the e2e/integration tests.
In order to get access to that network, the `workload.TestNetwork` would be used : 
```golang
type TestNetwork interface {
	ConfirmTxs(context.Context, []*chain.Transaction) error
	GenerateTx(context.Context, []chain.Action, chain.AuthFactory) (*chain.Transaction, error)
	URIs() []string
	Configuration() TestNetworkConfiguration
}
```

The implementation of the `TestNetwork` depends on whether we're running e2e or integration testing.
On integration testings, the ConfirmTxs would be executed against the underlying VM directly.
On e2e testings, the the ConfirmTxs would be implemented via the JSON RPC and Indexer.


examples/morpheusvm/tests/e2e/transfer_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/ava-labs/hypersdk/api/jsonrpc"
+	"github.com/ava-labs/hypersdk/auth"
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/vm"
+	"github.com/ava-labs/hypersdk/tests/workload"
+
+	he2e "github.com/ava-labs/hypersdk/tests/e2e"
```
richardpringle, line 20: I don't know how you say this in your head, but I say `he-two-e`... it sounds great haha

tsachiherman: I wasn't that creative here.. I copied that from the e2e.go file..

examples/morpheusvm/tests/e2e/transfer_test.go
```diff
@@ -0,0 +1,58 @@
...
+		[]chain.Action{&actions.Transfer{
+			To:    aother,
+			Value: 1,
+		}},
+		auth.NewED25519Factory(ed25519.PrivateKey(spendingKey.Bytes)),
+	)
+
+	require.NoError(err)
+
+	timeoutCtx, timeoutCtxFnc := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
```
samliok, line 52: we have a const `txCheckInterval`. I think it makes sense to either pass the const in here or remove the const completely and pass in a value from the test.

examples/morpheusvm/tests/e2e/transfer_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/vm"
+	"github.com/ava-labs/hypersdk/tests/workload"
+
+	he2e "github.com/ava-labs/hypersdk/tests/e2e"
+	ginkgo "github.com/onsi/ginkgo/v2"
+)
+
+var _ = he2e.RegisterTest("Transfer Transaction", func(t ginkgo.FullGinkgoTInterface, tn workload.TestNetwork) error {
```
samliok, line 24: i'm concerned this test doesn't add any real value to a hypersdk VM developer. If anything, it adds less value than the current workload tests because the user doesn't have fine grained controls of what a "confirmed" transaction is. In the workload tests, at least devs can define their custom confirmation logic. In these e2e tests all a user gets back is whether or not the transaction was found in the block.

examples/morpheusvm/tests/integration/transfer_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/vm"
+	"github.com/ava-labs/hypersdk/tests/integration"
+	"github.com/ava-labs/hypersdk/tests/workload"
+
+	ginkgo "github.com/onsi/ginkgo/v2"
+)
+
+var _ = integration.RegisterTest("Transfer Transaction", func(t ginkgo.FullGinkgoTInterface, tn workload.TestNetwork) error {
```
samliok, line 24: So the only difference between integration tests and e2e tests are how the transactions are confirmed under the hood. This just adds unnecessary work for the VM developer, and I'd like to remove this code duplication since the testing logic is exactly the same.

tests/e2e/network.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/ava-labs/hypersdk/tests/workload"
+)
+
+var (
+	ErrUnableToConfirmTx = errors.New("unable to confirm transaction")
+	ErrInvalidURI        = errors.New("invalid uri")
+)
+
+const (
+	txCheckInterval = 100 * time.Millisecond
```
samliok, line 23: this is the const i mentioned earlier

examples/morpheusvm/tests/e2e/transfer_test.go
```diff
@@ -0,0 +1,58 @@
...
+	uri := tn.URIs()[0]
+	other, err := ed25519.GeneratePrivateKey()
+	require.NoError(err)
+	aother := auth.NewED25519Address(other.PublicKey())
+
+	lcli := vm.NewJSONRPCClient(uri)
+	parser, err := lcli.Parser(context.Background())
+	require.NoError(err)
+
+	spendingKey, err := tn.WorkloadFactory().GetSpendingKey()
```
samliok, line 36: in these custom e2e tests, we only use the `workloadfactory` for this one method. This means if the user wants to only create custom e2e tests they need to also define the `workloadfactory` interface which is quite confusing 😅

tests/integration/network.go
```diff
@@ -0,0 +1,121 @@
...
+
+	for {
+		allFound := true
+		for _, tx := range txs {
+			err := n.confirmTx(ctx, instance, tx.ID())
+			if err == ErrTxNotFound {
+				allFound = false
+				break
+			}
+			return err
```
samliok, line 47: as mentioned before, simply returning this error is not going to be that helpful for a user

samliok (changes requested):
I'm unconvinced these testing pattern _adds_ value VM developers. Seems like this adds a lot of additional complexity, especially if the developer still needs to implement the `workloadfactory`

examples/morpheusvm/tests/e2e/transfer_test.go
samliok: maybe its worth adding some of the further work you mentioned to this PR. If the `GenesisKeys` are added to the test network then we wouldn't need to couple the `WorkloadFactory` here

examples/morpheusvm/tests/integration/transfer_test.go
aaronbuchwald: Can we create the test registry in one place and support passing the registry into both the integration and e2e environment?

This way VM test writers, can write and register their tests in a single place and then pass the registry in via both the integration and e2e entrypoints to run the VM specific tests in either environment.

tests/registry/registry.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (r *Registry) Add(name string, f TestFunc) {
+	r.tests = append(r.tests, namedTest{Fnc: f, Name: name})
+}
+
+func (r *Registry) List() []namedTest {
+	return r.tests
+}
+
+var testRegistry Registry
```
aaronbuchwald, line 30: Rather than create this global here, could we leave it to each VM to create their own instance and register tests to it?

examples/morpheusvm/tests/transfer.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/tests/workload"
+	"github.com/ava-labs/hypersdk/tests/registry"
+
+	tworkload "github.com/ava-labs/hypersdk/tests/workload"
+	ginkgo "github.com/onsi/ginkgo/v2"
+)
+
+var _ = registry.RegisterTest("Transfer Transaction", func(t ginkgo.FullGinkgoTInterface, tn tworkload.TestNetwork) error {
+	require := require.New(t)
```
aaronbuchwald, line 24: (future PR) - it would be nice to enable tests to declare their funding dependency ie. request that you pass in N keys with balance of at least 1K

examples/morpheusvm/tests/transfer.go
```diff
@@ -0,0 +1,48 @@
...
+		auth.NewED25519Factory(spendingKey),
+	)
+	require.NoError(err)
+
+	timeoutCtx, timeoutCtxFnc := context.WithDeadline(context.Background(), time.Now().Add(2*time.Second))
+	defer timeoutCtxFnc()
+
+	require.NoError(firstNode.SubmitTxs(timeoutCtx, []*chain.Transaction{tx}))
+	require.NoError(firstNode.ConfirmTxs(timeoutCtx, []*chain.Transaction{tx}))
+	return nil
```
aaronbuchwald, line 47: Why split into `Submit` and `Confirm` ? Since we are creating a shared interface between integration and e2e, we won't really be able to reliably check the state of the network between when txs were submitted and confirmed. Ie. we can submit and confirm txs in both environments, but in the e2e environment once the txs are issued we don't have control of when they'll be accepted.

examples/morpheusvm/tests/transfer.go
```diff
@@ -0,0 +1,48 @@
...
+	networkConfig := tn.Configuration().(*workload.NetworkConfiguration)
+	spendingKey := networkConfig.Keys()[0]
+
+	tx, err := firstNode.GenerateTx(context.Background(), []chain.Action{&actions.Transfer{
+		To:    aother,
+		Value: 1,
+	}},
+		auth.NewED25519Factory(spendingKey),
+	)
+	require.NoError(err)
```
aaronbuchwald, line 40: Can we move this functionality from the node to the test network interface?

We should make test writers interact with an interface that works across different environments, which will be easier if we minimize the details left to the test author like selecting which node to issue requests with.

tests/integration/network.go
```diff
@@ -0,0 +1,128 @@
...
+	return out
+}
+
+func (i *instance) SubmitTxs(ctx context.Context, txs []*chain.Transaction) error {
+	errs := i.vm.Submit(ctx, true, txs)
+	if len(errs) == 0 {
+		return nil
+	}
+	return errs[0]
+}
```
aaronbuchwald, line 45: This will only issue the txs to an individual node, whereas in the e2e environment if we issue a tx then it will be immediately gossiped to the rest of the network. We should not expose this detail to test writers.

tests/integration/network.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+}
+
+func (i *instance) URI() string {
+	return i.routerServer.URL
+}
+
+func (i *instance) GenerateTx(ctx context.Context, actions []chain.Action, auth chain.AuthFactory) (*chain.Transaction, error) {
+	// todo :
+	// this implementation need to change so that we can perform it directly without any RPC calls.
```
aaronbuchwald, line 78:
```suggestion
	// TODO: support generating tx without using jsonRPC client
```

tests/workload/network.go
```diff
@@ -0,0 +1,32 @@
...
+
+type TestNetwork interface {
+	Nodes() []TestNode
+	Configuration() TestNetworkConfiguration
+}
+
+type TestNode interface {
+	SubmitTxs(context.Context, []*chain.Transaction) error
+	ConfirmTxs(context.Context, []*chain.Transaction) error
+	GenerateTx(context.Context, []chain.Action, chain.AuthFactory) (*chain.Transaction, error)
```
aaronbuchwald, line 20: Can we remove `TestNode` and support a single `ConfirmTxs` call on `TestNetwork` instead of the node?

We should leave it to each environment (integration/e2e) to handle this since submitting a tx to a single node in integration will result in a different outcome than submitting a tx to a single node in the e2e environment.

Each environment should enforce that marking a tx as confirmed indicates all nodes have marked it as confirmed, so that we can check the state by making API calls directly to the URIs reported by the `TestNetwork`.

tests/workload/network.go
```diff
@@ -0,0 +1,32 @@
...
+
+// TestNetworkConfiguration is an interface, implemented by the custom-vm network test framework
+// that allows the test to store information regarding the test network prior to it's invocation, and
+// retrieve the said information during it's execution. It's vital that all implementations of this
+// interface would keep the data stored immutable as it would be shared across multiple threads.
+type TestNetworkConfiguration interface {
+	GenesisBytes() []byte
+	Name() string
+	Parser() chain.Parser
+}
```
aaronbuchwald, line 32: Can we make this a struct so that VMs can pass in the required fields instead of adding a type that implements this interface?

tests/registry/registry.go
tsachiherman: done. wasn't as intuitive as I was hoping for, but the end results looks nice.

tests/e2e/network.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+type Network struct {
```
samliok, line 29: 👍

examples/morpheusvm/tests/workload/genesis.go
```diff
@@ -60,3 +69,39 @@ func NewDefaultKeys() []ed25519.PrivateKey {
 
 	return testKeys
 }
+
+type NetworkConfiguration struct {
```
samliok, line 73: nit: we should add a type check for NetworkConfiguration

`var _ TestNetworkConfiguration = &NetworkConfiguration{}`

samliok (approved):
two nits and then after aarons comments are addressed LGTM

tests/integration/network.go
tsachiherman: done

examples/morpheusvm/tests/transfer.go
```diff
@@ -0,0 +1,48 @@
...
+	firstNode := tn.Nodes()[0]
+	other, err := ed25519.GeneratePrivateKey()
+	require.NoError(err)
+	aother := auth.NewED25519Address(other.PublicKey())
+
+	networkConfig := tn.Configuration().(*workload.NetworkConfiguration)
+	spendingKey := networkConfig.Keys()[0]
+
+	tx, err := firstNode.GenerateTx(context.Background(), []chain.Action{&actions.Transfer{
+		To:    aother,
```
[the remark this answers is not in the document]
tsachiherman: done

examples/morpheusvm/tests/workload/genesis.go
tsachiherman: done

tests/workload/network.go
tsachiherman: the reason for using an interface here was to allow the casting in the actual test:
```golang
networkConfig := tn.Configuration().(*workload.NetworkConfiguration)
```
just embedding structures would not allow me to do that.
i.e. 
```golang
type TestNetworkConfiguration struct {
	genesisBytes []byte
	name         string
	parser       chain.Parser
}
...
( in the test file )
type NetworkConfiguration struct {
	workload.TestNetworkConfiguration
	keys []ed25519.PrivateKey
}
```
wouldn't let me access the `NetworkConfiguration` by providing a pointer to `workload.TestNetworkConfiguration`. however, if I were to implement an interface, than I would be able to do so.

This would clearly be more important as some VMs would have more setup data that need to be shared with the tests.

examples/morpheusvm/tests/transfer.go
tsachiherman: done

examples/morpheusvm/tests/transfer.go
tsachiherman: done

tests/integration/network.go
tsachiherman: done

tests/workload/network.go
tsachiherman: done

tests/e2e/e2e.go
```diff
@@ -51,39 +54,39 @@ var _ = ginkgo.Describe("[HyperSDK APIs]", func() {
...
 		require.NoError(err)
-		expectedBlockchainID := e2e.GetEnv(tc).GetNetwork().GetSubnet(vmName).Chains[0].ChainID
+		expectedBlockchainID := e2e.GetEnv(tc).GetNetwork().GetSubnet(networkConfig.Name()).Chains[0].ChainID
 		require.Equal(expectedBlockchainID, chainIDFromRPC)
-	})
+	}))
 
-	ginkgo.It("GetNetwork", func() {
-		expectedBlockchainID := e2e.GetEnv(tc).GetNetwork().GetSubnet(vmName).Chains[0].ChainID
+	ginkgo.It("GetNetwork", asyncExecution(func() {
```
aaronbuchwald, line 73: Rather than use `asyncExecution`, could we make sure that the tests do not interfere with each other? I assume this is because we are using the same spending key for all of the tests. The simplest solution to this with this change is to create an isolated spending key for the newly added test registry. The more functionally complete version (good for a separate PR), would be to have each registered test describe the number of keys/amount of funds it requires to execute and then handle distributing funds to it in the background.

tests/integration/integration.go
```diff
@@ -622,7 +618,20 @@ var _ = ginkgo.Describe("[Tx Processing]", ginkgo.Serial, func() {
 	})
 })
 
-func expectBlk(i instance) func(add bool) []*chain.Result {
+var _ = ginkgo.Describe("[Custom VM Tests]", func() {
+	require := require.New(ginkgo.GinkgoT())
+
+	for testRegistry := range registry.GetTestsRegistries() {
```
aaronbuchwald, line 624: Could we pass the test registry into the integration/e2e tests rather than using a global registry this way?

Otherwise, I think this is just adding an extra step to the same pattern as https://github.com/ava-labs/hypersdk/pull/1667#discussion_r1815932572

tests/workload/network.go
```diff
@@ -0,0 +1,56 @@
...
+	ConfirmTxs(context.Context, []*chain.Transaction) error
+	GenerateTx(context.Context, []chain.Action, chain.AuthFactory) (*chain.Transaction, error)
+	URIs() []string
+	Configuration() TestNetworkConfiguration
+}
+
+// TestNetworkConfiguration is an interface, implemented by the custom-vm network test framework
+// that allows the test to store information regarding the test network prior to it's invocation, and
+// retrieve the said information during it's execution. It's vital that all implementations of this
+// interface would keep the data stored immutable as it would be shared across multiple threads.
```
aaronbuchwald, line 22:
```suggestion
// TestNetworkConfiguration is an interface, implemented by VM-specific tests
// to store information regarding the test network prior to it's invocation, and
// retrieve it during execution. All implementations must be thread-safe.
```

tests/integration/network.go
```diff
@@ -0,0 +1,125 @@
...
+		if allFound {
+			return nil
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.NewTimer(txCheckInterval).C:
+			// try again
+		}
+	}
```
aaronbuchwald, line 67: There's no need for retries here. Block building/acceptance will only be triggered by `expectBlk`. If something goes wrong where the tx is not included, it will never be included.

tests/integration/network.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+		}
+		// keep iterating backward.
+		lastAcceptedBlockID = blk.Parent()
+		blk, err = i.vm.GetBlock(ctx, lastAcceptedBlockID)
+		if err != nil {
+			return ErrTxNotFound
+		}
+	}
+}
```
aaronbuchwald, line 117: Can we switch to using the indexer API here instead so we don't need a manual search?

tests/integration/network.go
```diff
@@ -0,0 +1,125 @@
...
+	txCheckInterval = 100 * time.Millisecond
+)
+
+type Network struct {
+	uris []string
+}
+
+func (*Network) ConfirmTxs(ctx context.Context, txs []*chain.Transaction) error {
+	return instances[0].ConfirmTxs(ctx, txs)
+}
```
aaronbuchwald, line 34: This should confirm txs on the entire network. This means that we need to build, verify, and accept the block on the first block, and then issue, verify, and accept it on the rest of the instances in the network ie. `ParseBlock`, `Verify`, `Accept`.

examples/morpheusvm/tests/workload/genesis.go
```diff
@@ -45,10 +53,13 @@ func NewGenesis(keys []ed25519.PrivateKey, minBlockGap time.Duration) *genesis.D
 	genesis.Rules.MaxBlockUnits = fees.Dimensions{1800000, math.MaxUint64, math.MaxUint64, math.MaxUint64, math.MaxUint64}
 	genesis.Rules.MinBlockGap = minBlockGap.Milliseconds()
 
+	genesis.Rules.NetworkID = uint32(1)
+	genesis.Rules.ChainID = ids.GenerateTestID()
```
aaronbuchwald, line 57: What's the reason for these changes?

tests/e2e/network.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return testNetwork
+}
+
+func (n *Network) URIs() (out []string) {
+	for _, node := range n.nodes {
+		out = append(out, node.URI())
+	}
+	return
+}
```
aaronbuchwald, line 56: Can we remove the naked return for consistency with the rest of the codebase?

tests/e2e/network.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (n *Network) ConfirmTxs(ctx context.Context, txs []*chain.Transaction) error {
+	return n.nodes[0].ConfirmTxs(ctx, txs)
+}
```
aaronbuchwald, line 60: This function should make sure that txs are confirmed on all of the nodes in the network. Could we update this to:
1. Issue txs on 1 node
2. Confirm txs have been accepted on that one node
3. Iterate over the remaining nodes and confirm that it has advanced to the same block height/blockID (which should guarantee that it's accepted the same transactions)

Alternatively to 3, we could manually confirm each tx via API call on each node.

aaronbuchwald (changes requested):

tests/integration/integration.go
tsachiherman: Unfortunately, that won't work with ginkgo. we need the tests to be available in the `GetTestRegistries` ( or another global variable ) by the time this method is being invoked.
What I can do is to set this to be a function placeholder that would later on invoke the various tests. However, that would not be a `ginkgo` node and any failure on these tests would not be reported correctly.

tests/integration/network.go
tsachiherman: wasn't the intent of having the `integration` tests was to bypass the indexer and all other rpc endpoints ?
( in contrast to the e2e tests )

examples/morpheusvm/tests/workload/genesis.go
tsachiherman: When we call `GenerateTransaction`, we need to know the `ChainID`, which is included in the `Transaction.Base`.
For `integration` testing, the code in `tests/integration/integration.go` was generating this random ChainID and using it. Instead, I moved it one step up into the testVM.
Doing so allowed me to avoid the unnecessary `parser, err := lcli.Parser(ctx)` call by using the already-known ( or should've-been-known ) values.

tests/e2e/e2e.go
tsachiherman: ok, found that ginkgo has the `Serial` flag that would do the same thing for us.. used that instead.

tests/integration/network.go
tsachiherman: done

tests/integration/network.go
tsachiherman: done

tests/e2e/network.go
tsachiherman: done

tests/e2e/network.go
```diff
@@ -0,0 +1,159 @@
...
+
+func NewNetwork(tc *e2e.GinkgoTestContext) *Network {
+	blockchainID := e2e.GetEnv(tc).GetNetwork().GetSubnet(networkConfig.Name()).Chains[0].ChainID
+	testNetwork := &Network{}
+	for _, uri := range getE2EURIs(tc, blockchainID) {
+		n := &Node{uri: uri, network: testNetwork}
+		testNetwork.nodes = append(testNetwork.nodes, n)
+	}
+	testNetwork.parser = &parser{
+		Parser: networkConfig.Parser(),
```
aaronbuchwald, line 42: Can we add a comment on the `parser` type explaining why we need to wrap the `chain.Parser` type that is already passed in? As I understand it, this is because we cannot currently know the blockchainID in advance since it's the hash of the transaction on the P-Chain that will create the chain.

tests/e2e/network.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (n *Network) GenerateTx(ctx context.Context, actions []chain.Action, auth chain.AuthFactory) (*chain.Transaction, error) {
+	return n.nodes[0].generateTx(ctx, actions, auth)
+}
+
+func (*Network) Configuration() workload.TestNetworkConfiguration {
+	return networkConfig
+}
+
+type Node struct {
```
aaronbuchwald, line 96: Can we remove the `Node` type? This seems to only be used to keep a list of uris and then a pass through back to the network, so I think we can just keep the URIs on the network and move `GenerateTx` to be a function directly on the network.

tests/e2e/network.go
```diff
@@ -0,0 +1,159 @@
...
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(txCheckInterval):
+			}
+			nodeIdx-- // try again the same node.
+			continue
+		}
+	}
+	return nil
+}
```
aaronbuchwald, line 86:
```suggestion
func (n *Network) ConfirmTxs(ctx context.Context, txs []*chain.Transaction) error {
	c := jsonrpc.NewJSONRPCClient(n.uris[0])
	for _, tx := range txs {
		_, err := c.SubmitTx(ctx, tx.Bytes())
		if err != nil {
			return err
		}
	}

	indexerCli := indexer.NewClient(n.uris[0])
	for _, tx := range txs {
		success, _, err := indexerCli.WaitForTransaction(ctx, txCheckInterval, tx.ID())
		if err != nil {
			return err
		}
		if !success {
			return ErrUnableToConfirmTx
		}
	}

	_, targetHeight, _, err := c.Accepted(ctx)
	if err != nil {
		return err
	}

	for _, uri := range n.uris[1:] {
		if err := jsonrpc.Wait(ctx, txCheckInterval, func(ctx context.Context) (bool, error) {
			c := jsonrpc.NewJSONRPCClient(uri)
			_, nodeHeight, _, err := c.Accepted(ctx)
			if err != nil {
				return false, err
			}
			return nodeHeight >= targetHeight, nil
		}); err != nil {
			return err
		}
	}
	return nil
}
```

tests/registry/registry.go
```diff
@@ -0,0 +1,46 @@
...
+func (r *Registry) List() []namedTest {
+	if r == nil {
+		return []namedTest{}
+	}
+	return r.tests
+}
+
+// we need to pre-register all the test registries that are created externally in order to comply with the ginko execution order.
+// i.e. the global `var _ = ginkgo.Describe` used in the integration/e2e tests need to have this field populated before the iteration
+// over the top level nodes.
```
aaronbuchwald, line 35: Can we avoid this by setting this in a global variable / setting via `init` function so that it's identical during test construction and execution?

tests/integration/network.go
```diff
@@ -0,0 +1,200 @@
...
+			err = blk.Accept(ctx)
+			if err != nil {
+				return err
+			}
+			if instances[i].onAccept != nil {
+				instances[i].onAccept(blk)
+			}
+		}
+	}
+	return nil
```
aaronbuchwald, line 101:
```suggestion
	for i := 0; i < len(instances); i++ {
		if i == biggestHeightInstanceIndex {
			continue
		}
		instance := instances[i]
		for {
			height, err := instance.vm.GetLastAcceptedHeight()
			if err != nil {
				return err
			}
			if height == biggestHeight {
				break
			}
			statefulBlock, err := instances[biggestHeightInstanceIndex].vm.GetDiskBlock(ctx, height+1)
			if err != nil {
				return err
			}
			if err := instance.applyBlk(ctx, statefulBlock); err != nil {
				return err
			}
		}
	}
```

aaronbuchwald (changes requested):

examples/morpheusvm/tests/transfer.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/tests/workload"
+	"github.com/ava-labs/hypersdk/tests/registry"
+
+	tworkload "github.com/ava-labs/hypersdk/tests/workload"
+	ginkgo "github.com/onsi/ginkgo/v2"
+)
+
+// TestsRegistry is global tests registry instance where all the tests are to be registered. Each VM would typically have only
+// a single instance of such a registry.
```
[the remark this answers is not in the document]
tsachiherman: done

examples/morpheusvm/tests/transfer.go
```diff
@@ -0,0 +1,49 @@
...
+var TestsRegistry = &registry.Registry{}
+
+var _ = registry.Register(TestsRegistry, "Transfer Transaction", func(t ginkgo.FullGinkgoTInterface, tn tworkload.TestNetwork) error {
```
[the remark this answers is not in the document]
tsachiherman: done

tests/integration/network.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func (*Network) GenerateTx(ctx context.Context, actions []chain.Action, auth chain.AuthFactory) (*chain.Transaction, error) {
+	return instances[0].GenerateTx(ctx, actions, auth)
+}
+
+// SyncronizeNetwork ensures that all the nodes on the network are at the same block height.
```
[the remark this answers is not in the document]
tsachiherman: done

tests/e2e/network.go
tsachiherman: yes. I'll add some comments around that.

tests/e2e/network.go
tsachiherman: done

tests/e2e/network.go
tsachiherman: done

tests/integration/network.go
tsachiherman: done

tests/e2e/network.go
```diff
@@ -0,0 +1,159 @@
...
+	ErrUnableToConfirmTx = errors.New("unable to confirm transaction")
+	ErrInvalidURI        = errors.New("invalid uri")
+)
+
+const (
+	txCheckInterval = 100 * time.Millisecond
+)
+
+type Network struct {
+	nodes  []*Node
```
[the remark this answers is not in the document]
tsachiherman: done

tests/integration/integration.go
```diff
@@ -541,9 +538,12 @@ var _ = ginkgo.Describe("[Tx Processing]", ginkgo.Serial, func() {
...
 
 	ginkgo.It("processes valid index transactions (w/block listening)", func() {
-		// Clear previous txs on instance 0
-		accept := expectBlk(instances[0])
-		accept(false) // don't care about results
+		// syncronize the nodes on the network.
+		// this would clear previous txs on instance 0
+		// and ignore the transaction results while
+		// syncronizing the nodes on the network.
+		testNetwork := &Network{uris: uris}
```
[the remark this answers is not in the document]
tsachiherman: done

tests/registry/registry.go
tsachiherman: I think that it is identical during test construction and execution.

Not sure how to implement that so that it would work `ginkgo` otherwise, as `ginkgo` is managing these in globals as well ( but keep them internal )

aaronbuchwald (approved):

PR #1675, opened 2024-10-23
refactor recorder implementation

tsachiherman:
## What ?

This PR refactor the recorder implementation and move it into the tstate package.

## Why ?

The goal of this PR is to unify the logic responsible for recording ( generating ) and executing ( verifying ) so that there won't be any gaps possible, by design.

## How ?

The `TStateView` would have an additional flag, `simulationMode` determining whether it would act as an execution or recording engine.
The inner `scopeStorage` was also modified to use `state.Immutable` instead of just `map[string][]byte`.  This allow the `TStateView` to treat both uses in a transparent way.

The setting of the `simulationMode` is being done in one of the two places : For execution mode, it would be set in `NewView`. For recording mode, it would be set in `newRecorderView`.

The new struct `TStateRecorder` wraps around `TStateView`, provide the proper initialization and make sure to expose only the required interface.


state/tstate/recorder_test.go
```diff
@@ -0,0 +1,165 @@
...
+			bytes = append(bytes, shaBytes[:]...)
+		}
+		f.Add(bytes)
+	}
+	f.Fuzz(
+		RecorderPermissionValidatorFuzzer,
+	)
+}
+
+func createKeys(prefix byte) (keylist [][]byte) {
```
aaronbuchwald, line 31: nit: could we remove the naked returns and named return variables throughout? We generally avoid this throughout the codebase

state/tstate/recorder_test.go
```diff
@@ -0,0 +1,165 @@
...
+	for i := 0; i < 512; i++ {
+		shaBytes := sha256.Sum256([]byte{prefix, byte(i), byte(i >> 8)})
+		randNewKey := make([]byte, 30, 32)
+		copy(randNewKey, shaBytes[:])
+		keylist = append(keylist, keys.EncodeChunks(randNewKey, 1))
+	}
+	return
+}
+
+func createKeysValues(keys [][]byte) (out map[string][]byte) {
```
aaronbuchwald, line 41: same comment on the return type here

state/tstate/recorder_test.go
```diff
@@ -0,0 +1,165 @@
...
+	for opIdx := byte(0); opIdx < operationCount; opIdx++ {
+		if opType, randBytes, done = nextByte(randBytes); done {
+			return
+		}
+		opType %= 6
+		var keyIdx uint16
+		if keyIdx, randBytes, done = nextUint16(randBytes); done {
+			return
+		}
+		switch opType {
```
aaronbuchwald, line 98: Could we add an enum for the possible operations to make this more readable?

state/tstate/recorder_test.go
```diff
@@ -0,0 +1,165 @@
...
+	return other
+}
+
+func RecorderPermissionValidatorFuzzer(t *testing.T, randBytes []byte) {
+	require := require.New(t)
+	// create a set of keys which would be used for testing.
+	// half of these keys would "exists", where the other won't.
+	existingKeys := createKeys(0)
+	nonExistingKeys := createKeys(1)
+	existingKeyValue := createKeysValues(existingKeys)
```
aaronbuchwald, line 78: Won't this always create the same set of existing / non-existing keys?

Would it make sense to add additional arguments to the fuzzing function to receive random inputs rather than setting them up ourselves?

For example, we can add a single byte argument to get a random input and then use `b` and `b+1` to create existing and non-existing keys.

We should be able to do the same with the operation count as well.

state/tstate/recorder_test.go
```diff
@@ -0,0 +1,165 @@
...
+	var opType byte
+	for opIdx := byte(0); opIdx < operationCount; opIdx++ {
+		if opType, randBytes, done = nextByte(randBytes); done {
+			return
+		}
+		opType %= 6
+		var keyIdx uint16
+		if keyIdx, randBytes, done = nextUint16(randBytes); done {
+			return
+		}
```
aaronbuchwald, line 97: Could we replace these with simply taking a random seed as input to the function and then generating a random number to get the `opType` and index?

state/tstate/recorder_test.go
```diff
@@ -0,0 +1,165 @@
...
+			// validate operation agaist TStateView
+			stateKeys := recorder.GetStateKeys()
+			require.NoError(New(0).NewView(stateKeys, immutableScopeStorage(existingKeyValue).duplicate()).Insert(context.Background(), key, []byte{1, 2, 3}))
+
+			existingKeyValue[string(key)] = []byte{1, 2, 3}
+		case 1: // insert non existing key
+			keyIdx %= uint16(len(nonExistingKeys))
+			key := nonExistingKeys[keyIdx]
+			nonExistingKeys[keyIdx] = []byte{}
+			nonExistingKeys = append(nonExistingKeys[:keyIdx], nonExistingKeys[keyIdx+1:]...)
```
aaronbuchwald, line 113: Could we define a function for removing an element at a specific index? Since we don't need to preserve the ordering of elements, we could do the more efficient operation of swapping the last element in the slice for the deleted `keyIdx` and then re-slicing to make it one element shorter

aaronbuchwald (changes requested):

state/tstate/recorder_test.go
tsachiherman: done

state/tstate/recorder_test.go
tsachiherman: done

state/tstate/recorder_test.go
tsachiherman: done

state/tstate/recorder_test.go
tsachiherman: sure.

state/tstate/tstate_view.go
```diff
@@ -60,10 +76,29 @@ func (ts *TState) NewView(scope state.Keys, storage map[string][]byte) *TStateVi
...
+		pendingChangedKeys: make(map[string]maybe.Maybe[[]byte], 0),
+
+		ops: make([]*op, 0, defaultOps),
+
+		scope:        state.Keys{},
+		scopeStorage: immutableState,
+
+		allocates: make(map[string]uint16, 0),
+		writes:    make(map[string]uint16, 0),
+
```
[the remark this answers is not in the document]
tsachiherman: done

aaronbuchwald (approved):

PR #1679, opened 2024-10-23
Add block assembler + executor

aaronbuchwald:
WIP: add block assembler + executor to DSMR and add implementation of it for the chain package.

This structure should help us refactor the interaction between the `chain/` and `vm/` packages so that:
- `chain` provides a default implementation of transaction/block types that can be used with DSMR
- `vm` uses DSMR package to fulfill `chain.VM` and `snowman.Block` interfaces from AvalancheGo
- `vm` handles writing blocks to disk and notifies subscribers of accepted blocks
- we can separate out state sync from the `vm` package

### DIFF
--- a/dsmr/block.go
+++ b/dsmr/block.go
@@ -111,38 +111,59 @@ func parseChunkProto[T Tx](chunkProto *dsmr.Chunk) (Chunk[T], error) {
 type ChunkFault struct{}
 
 type Block struct {
-	ID        ids.ID `serialize:"true"`
 	ParentID  ids.ID `serialize:"true"`
 	Height    uint64 `serialize:"true"`
 	Timestamp int64  `serialize:"true"`
 
-	// TODO Add TxsRoot?
-	// TODO Add OutgoingWarpRoot?
-	// TODO Do we need separate Available + Executed Chunks?
 	Chunks []*ChunkCertificate `serialize:"true"`
+
+	blkID    ids.ID
+	blkBytes []byte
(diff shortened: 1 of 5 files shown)


dsmr/block.go
```diff
@@ -111,38 +111,60 @@ func parseChunkProto[T Tx](chunkProto *dsmr.Chunk) (Chunk[T], error) {
 type ChunkFault struct{}
 
 type Block struct {
-	ID        ids.ID `serialize:"true"`
```
aaronbuchwald, line 114: We cannot include the block's own ID in the serialization. The blkID is a hash of the block bytes, so including the ID in the block bytes creates a cycle.

dsmr/block.go
```diff
@@ -111,38 +111,60 @@ func parseChunkProto[T Tx](chunkProto *dsmr.Chunk) (Chunk[T], error) {
...
 	Height    uint64 `serialize:"true"`
 	Timestamp int64  `serialize:"true"`
 
-	// TODO Add TxsRoot?
-	// TODO Add OutgoingWarpRoot?
-	// TODO Do we need separate Available + Executed Chunks?
 	Chunks []*ChunkCertificate `serialize:"true"`
+
+	blkID    ids.ID
+	blkBytes []byte
```
aaronbuchwald, line 121: These are left unpopulated for now. It seems this type was not fully created, so leaving a correct block constructor for a follow up

Issue #1645, opened 2024-10-09
Add back `actionID` to `stateKeys()`


RodrigoVillar:
As per the conversation [linked](https://t.me/c/2197747326/394), we most likely want to add back `actionID` argument to the `stateKeys()` interface. 

This change would probably make sense to work on after #1644 is merged

PR #1683, opened 2024-10-23
Add back `actionID` to `StateKeys()`

RodrigoVillar:
As referenced in #1645, this PR adds back the `actionID` parameter back to `StateKeys()`.

Following the convention for when `actionID` previously existed, we call `CreateActionID()` whenever possible. If creating the action ID via `CreateActionID()` isn't possible, we pass in `ids.Empty{}`

### DIFF
--- a/api/jsonrpc/server.go
+++ b/api/jsonrpc/server.go
@@ -204,7 +204,7 @@ func (j *JSONRPCServer) ExecuteActions(
 
 	for actionIndex, action := range actions {
 		// Get expected state keys
-		stateKeysWithPermissions := action.StateKeys(args.Actor)
+		stateKeysWithPermissions := action.StateKeys(args.Actor, ids.Empty)
 
 		// flatten the map to a slice of keys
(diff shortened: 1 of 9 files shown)


api/jsonrpc/server.go
```diff
@@ -204,7 +204,7 @@ func (j *JSONRPCServer) ExecuteActions(
 
 	for actionIndex, action := range actions {
 		// Get expected state keys
-		stateKeysWithPermissions := action.StateKeys(args.Actor)
+		stateKeysWithPermissions := action.StateKeys(args.Actor, ids.Empty)
```
aaronbuchwald, line 207: Can we use `CreateActionID(ids.Empty, actionIndex)` here? This ensures that each simulated action gets a unique `actionID` during the simulation and that it will never overlap with an actual `actionID` that was already used onchain.

aaronbuchwald (approved):

api/jsonrpc/server.go
RodrigoVillar: Done

aaronbuchwald (approved):

PR #1684, opened 2024-10-23
Implement NFT example

richardpringle:
This is an NFT example that implements the majority of the public functions (at least all the mutating ones) from the OpenZeppelin reference implementation. It's meant to be used as a comparison. I intentionally left out the `URI` which could be necessary for practical use.

### DIFF
--- a/x/contracts/examples/nft/Cargo.toml
+++ b/x/contracts/examples/nft/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "nft"
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["lib"]
+
+[dependencies]
+wasmlanche = { workspace = true, features = ["debug"] }
+
+[dev-dependencies]
+wasmlanche = { workspace = true, features = ["debug", "test"] }
+
+[build-dependencies]
+wasmlanche = { workspace = true, features = ["build"] }
+
+[features]
+# bindings = ["wasmlanche/bindings"]
--- a/x/contracts/examples/nft/Cargo.toml
+++ b/x/contracts/examples/nft/Cargo.toml
@@ -16,4 +16,4 @@ wasmlanche = { workspace = true, features = ["debug", "test"] }
 wasmlanche = { workspace = true, features = ["build"] }
 
 [features]
-# bindings = ["wasmlanche/bindings"]
+bindings = ["wasmlanche/bindings"]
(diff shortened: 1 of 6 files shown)


x/contracts/examples/nft/src/lib.rs
```diff
@@ -0,0 +1,688 @@
...
+    let from_balance = balance_of(ctx, from);
+
+    ctx.store((
+        (Owner(token_id), to),
+        (Approval(token_id), Address::ZERO),
+        (Balance(from), from_balance - 1),
+    ))
+    .expect("failed to serialize");
+
+    if to != Address::ZERO {
```
samliok, line 201: the solidity implementation fails if address == zero, maybe move this check up to the top of the function and panic?

x/contracts/examples/nft/src/lib.rs
```diff
@@ -0,0 +1,688 @@
...
+//  * - `tokenId` must not exist.
+//  * - `to` cannot be the zero address.
+//  *
+//  * Emits a {Transfer} event.
+//  */
+// function _mint(address to, uint256 tokenId) internal {
+//     if (to == address(0)) {
+//         revert ERC721InvalidReceiver(address(0));
+//     }
+//     address previousOwner = _update(to, tokenId, address(0));
```
samliok, line 224: the solidity implementation overloads the functionality of the update function for no reason 🤮

x/contracts/examples/nft/src/lib.rs
samliok: ah we call this function in burn nvn

samliok (approved):
Good comparision and the unit tests are very extensive 👍❤️

x/contracts/examples/nft/src/lib.rs
```diff
@@ -0,0 +1,688 @@
...
+
+// /**
+//  * @dev Initializes the contract by setting a `name` and a `symbol` to the token collection.
+//  */
+// constructor(string memory name_, string memory symbol_) {
+//     _name = name_;
+//     _symbol = symbol_;
+// }
+#[public]
+pub fn init(ctx: &mut Context, name: String, symbol: String) {
```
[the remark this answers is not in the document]
richardpringle: Good catch

x/contracts/examples/nft/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+wasmlanche = { workspace = true, features = ["debug"] }
+
+[dev-dependencies]
+wasmlanche = { workspace = true, features = ["debug", "test"] }
+
+[build-dependencies]
+wasmlanche = { workspace = true, features = ["build"] }
+
+[features]
+# bindings = ["wasmlanche/bindings"]
```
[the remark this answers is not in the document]
richardpringle: Maybe I should just turn it on?

samliok (approved):

PR #1685, opened 2024-10-24
Add Getter to `BalanceHandler`

RodrigoVillar:
This PR adds a getter function to `BalanceHandler`


aaronbuchwald (commented):
Could we include the first usage of this (adding balance API call to JSONRPC client/server) so that this is not unused code?

aaronbuchwald (approved):
LGTM after fixing the merge conflict

aaronbuchwald (approved):

PR #1687, opened 2024-10-24
Default Contract Manager Implementation

samliok:
This PR provides a default state layout for VMs that want to use contracts. VMs that use our out of the box implementation don't need to worry about managing and setting up their state keys(although they still can if they want).

### DIFF
--- a/x/contracts/runtime/manager.go
+++ b/x/contracts/runtime/manager.go
@@ -37,9 +37,13 @@ type ContractStateManager struct {
 
 func NewContractStateManager(
 	db state.Mutable,
+	prefix []byte,
 ) *ContractStateManager {
+	// set db to prefixed state space
+	prefixedState := newPrefixStateMutable(prefix, db)
+
 	return &ContractStateManager{
-		db: db,
+		db: prefixedState,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


x/contracts/simulator/state/manager.go
```diff
@@ -5,43 +5,44 @@ package state
...
 type ContractStateManager struct {
-	db state.Mutable
+	db            state.Mutable
+	contractState *runtime.ContractStateManager
 }
 
 func NewContractStateManager(db state.Mutable) *ContractStateManager {
-	return &ContractStateManager{db}
+	contractManager := runtime.NewContractStateManager(
+		runtime.NewPrefixStateMutable(
```
samliok, line 37: just have `NewContractStateManager` take a byte prefix and it can create the prefixed state.

x/contracts/runtime/manager.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/ids"
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/state"
+)
+
+var (
+	_                 ContractManager = &ContractStateManager{}
```
richardpringle, line 19: What's this?

x/contracts/simulator/ffi/ffi.go
```diff
@@ -94,7 +94,7 @@ func CreateContract(db *C.Mutable, path *C.char) C.CreateContractResponse {
 		}
 	}
 
-	err = contractManager.SetContract(context.TODO(), contractID, contractBytes)
+	err = contractManager.SetContractBytes(context.TODO(), runtime.ContractID(contractID[:]), contractBytes)
```
richardpringle, line 97: nit
```suggestion
        contractID := runtime.ContractID(contractID[:])
	err = contractManager.SetContractBytes(context.TODO(), contractID, contractBytes)
```
this is specifically easier to read when separated out on github. If there aren't any real change requests though, we can always change this later.

richardpringle (dismissed):

x/contracts/runtime/manager.go
samliok: type check

richardpringle (approved):

PR #1693, opened 2024-10-25
Add Test Suite for `BalanceHandler`

RodrigoVillar:
This PR adds a test function for `BalanceHandler` that allows VM developers to test their implementation of `BalanceHandler`.

Currently, there's two things that `BalanceHandler` needs to be tested for:

- Proper allocation of state keys via `SponsorStateKeys()`
- Proper modification of account balances

This PR tests the following by providing a `TestBalanceHandler()` function where developers can pass in their implementation of `BalanceHandler`. `TestBalanceHandler()` tests for the cases mentioned above.


chain/chaintest/balance_handler_test_helpers.go
```diff
@@ -0,0 +1,48 @@
...
+	r.NoError(b.AddBalance(context.Background(), addr, store, amount, true))
+
+	stateKeys := b.SponsorStateKeys(addr)
+	ts := tstate.New(1)
+	tsv := ts.NewView(
+		stateKeys,
+		store.Storage,
+	)
+
+	r.NoError(b.AddBalance(context.Background(), addr, tsv, 0, true))
```
aaronbuchwald, line 37: Could we use a non-zero balance?

chain/chaintest/balance_handler_test_helpers.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/codec/codectest"
+	"github.com/ava-labs/hypersdk/state/tstate"
+)
+
+// TestBalanceHandler tests [b] by calling the following methods (in order):
+// - SponsorStateKeys()
+// - AddBalance()
+// - CanDeduct()
```
aaronbuchwald, line 20: Could we complete this comment with all of the steps this test is performing?

aaronbuchwald (changes requested):

chain/chaintest/balance_handler_test_helpers.go
RodrigoVillar: Done

chain/chaintest/balance_handler_test_helpers.go
RodrigoVillar: Done

chain/chaintest/balance_handler_test_helpers.go
joshua-kim, line 1: Can we call this file something else (e.g `balance_handler.go`)? I don't mind us having a file for balance helper but `test` is redundant since this is already known to be a testing file from the package name `chaintest` and `helpers` doesn't give much information to me

chain/chaintest/balance_handler_test_helpers.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/codec/codectest"
+	"github.com/ava-labs/hypersdk/state/tstate"
+)
+
+// TestBalanceHandler tests [b] by doing the following:
```
joshua-kim, line 17: The pattern of `[name]` is something we do in our codebase but I don't think is standard.... in fact it actually breaks indexing in my ide (e.g if I have a struct named `Foo`, calling it `Foo` in a comment indexes it in the IDE, but `[Foo]` will not).

chain/chaintest/balance_handler_test_helpers.go
```diff
@@ -0,0 +1,56 @@
...
+// - Creating a new instance of tstate with permissions from SponsorStateKeys().
+//   - The test account balance is initialized to 0.
+//
+// - Call AddBalance() to increment the test account balance by 1.
+// - Call GetBalance() and require that the test account balance is 1.
+// - Call CanDeduct() to check if the test account balance can be deducted by 1.
+// - Call Deduct() to decrement the test account balance by 1.
+// - Call GetBalance() and require that the test account balance is 0.
+//
+// TestBalanceHandler fails if any of the above method calls returns an error.
```
joshua-kim, line 28: IMO these are implementation details of the function and don't need to be documented

chain/chaintest/balance_handler_test_helpers.go
```diff
@@ -0,0 +1,56 @@
...
+func TestBalanceHandler(t *testing.T, b chain.BalanceHandler) {
```
joshua-kim, line 29: We should pass in the context from the caller. `context.Background` is generally called from `main()`/your test and child contexts are created + passed around the program. As an example if you canceled a context in the calling code any code in here won't be canceled.

chain/chaintest/balance_handler_test_helpers.go
```diff
@@ -0,0 +1,56 @@
...
+	balance, err := b.GetBalance(context.Background(), addr, tsv)
+	r.NoError(err)
+	r.Equal(amount, balance)
+
+	r.NoError(b.CanDeduct(context.Background(), addr, tsv, amount))
+
+	r.NoError(b.Deduct(context.Background(), addr, tsv, amount))
+	balance, err = b.GetBalance(context.Background(), addr, tsv)
+	r.NoError(err)
+	r.Equal(uint64(0), balance)
```
joshua-kim, line 55: I would prefer if we broke this up into subtests of some sort. If this test fails it might be unclear which specific function is broken.

RodrigoVillar: Done

chain/chaintest/balance_handler_test_helpers.go
RodrigoVillar: Done

chain/chaintest/balance_handler_test_helpers.go
RodrigoVillar: Done (as per offline chat, we now refer developers to the BalanceHandler interface to see the invariants `b` must follow)

chain/chaintest/balance_handler_test_helpers.go
RodrigoVillar: Done

chain/chaintest/balance_handler_test_helpers.go
aaronbuchwald: The original implementation was cleaner imho. This is a pretty simple test of getters and setters. If we break it into separate tests, we'd still need to use the setters to test the getters in the same way.

If this test fails at any point, it should be very clear which function had unexpected behavior. Breaking this into subtests with `t.Run` won't provide additional information since we are still operating on the same state with the identical order of operations. @joshua-kim is this what you were suggesting with subtests?

chain/chaintest/balance_handler.go
```diff
@@ -0,0 +1,191 @@
...
+		Actor  codec.Address
+		Amount uint64
+
+		// State keys are not enforced in Setup
+		Setup func(context.Context, *testing.T, state.Mutable)
+		// State keys are enforced in TestFunc
+		TestFunc func(context.Context, codec.Address, state.Mutable, uint64) error
+		// Assertion is run after TestFunc for any additional checks
+		// State keys are not enforced in Assertion
+		Assertion func(context.Context, *testing.T, state.Mutable)
```
joshua-kim, line 39: Table-driven tests are great for when you're trying to just run variations of arguments against a function, e.g:
```
if tt.wantSum != Add(tt.a, tt.b) {
    t.Fatal("wrong result!")
}
```

But in this example where we're writing complex setup code/assertion code in the test vector, there isn't much of a difference between just having separate tests entirely. I think it would actually be much cleaner to just have separate calls to `t.Run` for each test individually instead of trying to force them all to fit the same test-table.

chain/chaintest/balance_handler.go
```diff
@@ -0,0 +1,191 @@
...
+
+		ExpectedBalance uint64
+		ExpectErr       bool
+	}{
+		{
+			Name:   "simpleAddBalance",
```
joshua-kim, line 45: We can avoid camel-casing here - these strings are just a UX improvement when running tests so we should make these human readable. We can also give more descriptive names like "add balance - overflow" for example.

chain/chaintest/balance_handler.go
```diff
@@ -0,0 +1,191 @@
...
+		{
+			Name:   "multipleAccountsCanExist",
+			Actor:  addrTwo,
+			Amount: 2,
+			Setup: func(ctx context.Context, t *testing.T, mu state.Mutable) {
+				r := require.New(t)
+				// Initialize addrOne balance to 1
+				r.NoError(b.AddBalance(ctx, addrOne, mu, 1))
+				// Initialize addrTwo balance to 0
+				r.NoError(b.AddBalance(ctx, addrTwo, mu, 0))
```
joshua-kim, line 145: nit: Don't feel obligated to comment all of this - most of this code is pretty self-describing so the comments aren't really necessary IMO. I think either a high-level comment on the test or a descriptive test name is sufficient

chain/chaintest/balance_handler.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/state/tstate"
+)
+
+// TestBalanceHandler tests b by requiring that it upholds the invariants
+// described in the BalanceHandler interface.
+func TestBalanceHandler(ctx context.Context, t *testing.T, b chain.BalanceHandler) {
+	// Test addresses
+	addrOne := codectest.NewRandomAddress()
+	addrTwo := codectest.NewRandomAddress()
```
joshua-kim, line 25: Another idea is to refactor this so that you can have an array of tests, that take in an array of addresses to run on so you don't have to write separate tests for the multi-address case vs single address case.

chain/chaintest/balance_handler.go
RodrigoVillar: I agree; switching to `t.Run` reduced the code by 25% 😄

chain/chaintest/balance_handler.go
RodrigoVillar: Done

examples/morpheusvm/storage/state_manager_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/ava-labs/hypersdk/chain/chaintest"
+)
+
+func TestBalanceHandler(t *testing.T) {
+	chaintest.TestBalanceHandler(
+		context.Background(),
+		t,
+		&BalanceHandler{},
+	)
```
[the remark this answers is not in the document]
RodrigoVillar: Switched to one line 👍

joshua-kim (dismissed):

chain/chaintest/balance_handler.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/codec/codectest"
+	"github.com/ava-labs/hypersdk/consts"
+	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/state/tstate"
+)
+
+// TestBalanceHandler tests b by requiring that it upholds the invariants
+// described in the BalanceHandler interface.
+func TestBalanceHandler(ctx context.Context, t *testing.T, b chain.BalanceHandler) {
```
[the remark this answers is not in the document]
RodrigoVillar: `b` is stateless (`BalanceHandler` just modifies the `state.Mutable` instance passed in) and so reusing `b` across all tests is equivalent to creating new instances of `b`. From a devX perspective, would argue that passing in `TestBalanceHandler(ctx context.Context, t *testing.T, b chain.BalanceHandler)` is more straightforward than something like `TestAddBalance(ctx context.Context, t *testing.T, AddBalance())`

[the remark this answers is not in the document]
RodrigoVillar: Ah understood; I modified `TestBalanceHandler()` so that it now takes in a factory func

chain/chaintest/balance_handler.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/ava-labs/hypersdk/codec/codectest"
+	"github.com/ava-labs/hypersdk/consts"
+	"github.com/ava-labs/hypersdk/state/tstate"
+)
+
+type BalanceHandlerFactory func() chain.BalanceHandler
+
+// TestBalanceHandler tests b by requiring that it upholds the invariants
+// described in the BalanceHandler interface.
+func TestBalanceHandler(t *testing.T, ctx context.Context, bf BalanceHandlerFactory) {
```
joshua-kim, line 22: nit: i think a distinct type is a little over-kill here since i don't expect we'll be using this anywhere else.
```suggestion
// TestBalanceHandler tests b by requiring that it upholds the invariants
// described in the BalanceHandler interface.
func TestBalanceHandler(t *testing.T, ctx context.Context, bf func() chain.BalanceHandler) {
```

joshua-kim (approved):

PR #1697, opened 2024-10-28
Don't benchmark serializing parameters

richardpringle:
This doesn't have a meaningful effect on the results, but param serialization likely won't be part of the execution of a transaction so we shouldn't benchmark it


samliok (approved):
all good just some nits the linter picked up

samliok (dismissed):

samliok (approved):

PR #1704, opened 2024-10-31
ActionTest assertions in correct order

samliok:
small nit: the expected outcome should be the first arg

### DIFF
--- a/chain/chaintest/action_test_helpers.go
+++ b/chain/chaintest/action_test_helpers.go
@@ -73,8 +73,8 @@ func (test *ActionTest) Run(ctx context.Context, t *testing.T) {
 
 		output, err := test.Action.Execute(ctx, test.Rules, test.State, test.Timestamp, test.Actor, test.ActionID)
 
-		require.ErrorIs(err, test.ExpectedErr)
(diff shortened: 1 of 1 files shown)


aaronbuchwald (dismissed):

chain/chaintest/action_test_helpers.go
```diff
@@ -73,8 +73,8 @@ func (test *ActionTest) Run(ctx context.Context, t *testing.T) {
 
 		output, err := test.Action.Execute(ctx, test.Rules, test.State, test.Timestamp, test.Actor, test.ActionID)
 
-		require.ErrorIs(err, test.ExpectedErr)
-		require.Equal(output, test.ExpectedOutputs)
+		require.ErrorIs(test.ExpectedErr, err)
```
RodrigoVillar, line 76:
```suggestion
		require.ErrorIs(err, test.ExpectedErr)
```

From the spec for `require.ErrorIs()`:

> `func ErrorIs(t TestingT, err error, target error, msgAndArgs ...interface{})`
`ErrorIs` asserts that at least one of the errors in err's chain matches target. This is a wrapper for errors.Is.

aaronbuchwald (approved):

Issue #1278, opened 2024-08-09
Modularize Components


aaronbuchwald:
Support plug and play components, so that VM designers can easily swap out different modules:

- MerkleDB -> state manager that supports maintaining and committing per block views
- Block builder
- Execution Pipeline (https://github.com/ava-labs/hypersdk/issues/736)
- Mempool

aaronbuchwald:
Currently, `StatefulBlock` is combining many responsibilities. It would be nice to break this up into separable components for block execution, expiry replay protection, dynamic state sync, and delayed root calculation / state manager (merkledb -> interface).

### Block Execution

Block execution is currently handled in [innerVerify](https://github.com/ava-labs/hypersdk/blob/7bd1a499af301d94ab33933947b614f64f358d4f/chain/block.go#L407).

This seems like the ideal place to break out a separate executor / processor type and break this function up. This is the place we'd need to pass in alternative tx execution strategies like https://github.com/ava-labs/hypersdk/issues/736.

### Expiry Replay Protection

Transaction expiry is handled via timestamps, so we need to de-duplicate / invalidate any transactions included within a block that are a duplicate of a transaction in the ancestry of the block within the `ValidityWindow` (the maximum lookback the same transaction could have been included in a block. This is broken out in https://github.com/ava-labs/hypersdk/pull/1698.

### Dynamic State Sync

HyperSDK uses dynamic state sync from AvalancheGo, which is well documented here https://github.com/ava-labs/avalanchego/blob/9f0ff018cc1ca2879b59bb3b62c54fdcc1957e9d/x/sync/README.md.

This means that the `StatefulBlock` type needs to handle both normal execution and execution during dynamic state sync. For example, we have a switch statement in `Verify` and `Accept` to handle the different cases.

In Verify, we need to skip block execution because we don't have the full current state required to verify the block: https://github.com/ava-labs/hypersdk/blob/7bd1a499af301d94ab33933947b614f64f358d4f/chain/block.go#L342

In Accept, we need to update the sync target https://github.com/ava-labs/hypersdk/blob/7bd1a499af301d94ab33933947b614f64f358d4f/chain/block.go#L606

[...]

PR #1708, opened 2024-10-31
Refactor chain/vm: move execution off of StatefulBlock

aaronbuchwald:
- move replay protection to its own function within block verification
- move `GetVerifyContext` into `innerVerify` (called in the identical way in three places)
- Move execution from `innerVerify` into `Executor`

This is a start in the direction of https://github.com/ava-labs/hypersdk/issues/1278#issuecomment-2449962374 to separate `StatefulBlock` from the `chain` package, so that it only deals with block parsing, building, and execution.

### DIFF
--- a/chain/processor.go
+++ b/chain/processor.go
@@ -207,7 +207,7 @@ func (e *Executor) Execute(
 // If a block is already accepted, its transactions have already been added
 // to the VM's seen emap and calling [IsRepeat] will return a non-zero value
 // when it should already be considered valid, so we skip this step here.
-func (e *Executor) verifyExpiryReplayProtection(ctx context.Context, b *StatefulBlock, rules Rules, vctx VerifyContext) error {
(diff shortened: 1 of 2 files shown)


chain/processor.go
```diff
@@ -16,15 +22,221 @@ import (
...
+	if b.accepted {
+		return nil
+	}
+
+	oldestAllowed := b.Tmstmp - rules.GetValidityWindow()
+	if oldestAllowed < 0 {
+		// Can occur if verifying genesis
+		oldestAllowed = 0
+	}
+	dup, err := vctx.IsRepeat(ctx, oldestAllowed, b.Txs, set.NewBits(), true)
```
tsachiherman, line 220: This isn't super-efficient, as we're using `vm.seen.Contains` instead of the more appropriate `vm.seen.Any`.
Not an issue for this PR, though.

tsachiherman (approved):
one comment, looks good otherwise.

PR #1709, opened 2024-11-01
add cerification comment

tsachiherman:
Adds a small clarification comment in `JSONRPCServer. LastAccepted` explaining why the return value from `vm.LastAcceptedBlock()` doesn't need to be checked for nil-ness.


RodrigoVillar (approved):
lgtm

api/jsonrpc/server.go
```diff
@@ -121,6 +121,10 @@ type LastAcceptedReply struct {
 
 func (j *JSONRPCServer) LastAccepted(_ *http.Request, _ *struct{}, reply *LastAcceptedReply) error {
 	blk := j.vm.LastAcceptedBlock()
+	// The call above gives a pointer to a chain.StatefulBlock struct.
+	// This pointer is set up during VM.Initialize and is kept updated afterward.
+	// The server handler is created near the end of VM.Initialize, so it's safe to use this
+	// pointer without checking if it's nil.
```
aaronbuchwald, line 127: Can we add this comment to the interface rather than where it's used?

Also, could we change from explaining why it's always non-nil to just state that it is always non-nil?

tsachiherman: I'll move this to the interface. I think that in any case where we're not following the go convention of returning `(value, err)` we should document that explicitly.

aaronbuchwald (approved):

PR #1710, opened 2024-11-02
decouple apis and vm implementation

### LINKED ISSUE #1488: Update VM Option to use decoupled `Opts` struct
Change from the current VM Option as a function taking a pointer directly to the VM:

https://github.com/ava-labs/hypersdk/blob/f8f540c75c900d41b781b7ff0e8965e0a73b0dd7/vm/option.go#L16

to a more common pattern of creating an options struct as in grpc `DialOptions` here: https://github.com/grpc/grpc-go/blob/00d3ec8c71a7ba0f9c755881f0f7147eba5e814c/dialoptions.go#L388.

For the HyperSDK, this has the added advantage that we can ideally re-use a set of options either to run on the VM or to run a set of external services. For example, we could use the external subscriber (https://github.com/ava-labs/hypersdk/pull/1457) to run a set of APIs outside of the node itself and ideally create a different type other than `*vm.VM` that can consume the same options struct to create the same services we'd otherwise be running inside the node.

najeal:
Closes #1488 

This is decoupling VM implementation from the APIs.
it switches from usage of `*vm.VM` to an interface `api.VM`. To implement api.VM a new struct `apiCoreContext` is created (this avoids to add new functions to vm.VM by creating an Adapter).

Instead of directly modifying VM config like done previously, an `Options pattern` is used.

Depending of the completion of the PR, renaming could be considered between those types:
- Opt -> interface for Options Pattern
- Options -> struct for Options Pattern configuration
- optionFunc -> function to configure APIs
- OptionFunc -> function to configure APIs
- Option -> function to configure APIs + namespace


vm/vm.go
```diff
@@ -1208,3 +1224,21 @@ func (vm *VM) Fatal(msg string, fields ...zap.Field) {
...
 	panic("fatal error")
 }
+
+type apiCoreContext struct {
+	*VM
+}
+
+func (c *apiCoreContext) GetDataDir() string {
+	return c.VM.DataDir
+}
```
aaronbuchwald, line 1234: (not necessary for this PR) we should return the data directory that options can use directly ie `filepath.Join(c.VM.DataDir, optNamespace)` rather than leaving this to each option to determine

vm/option.go
```diff
@@ -50,44 +54,72 @@ func NewOption[T any](namespace string, defaultConfig T, optionFunc OptionFunc[T
...
 		struct{}{},
-		func(vm *VM, _ struct{}) error {
-			WithBuilder()(vm)
-			WithGossiper()(vm)
-			return nil
+		func(_ api.VM, _ struct{}) (Opt, error) {
+			return newFuncOption(func(o *Options) {
+				WithBuilder().apply(o)
+				WithGossiper().apply(o)
+			}), nil
```
aaronbuchwald, line 77: (not necessary for this PR) - we should change this to a single exported option `WithManual` there's no need to export the others individually when they can only be used currently to set the gossiper/builder to the manual implementation.

We should also separate the API / subscription related options from this package, so that the API implementations themselves do not depend on the VM package and the VM package can import these options. This way you can use the API / subscription related options without importing the VM package (which should not be necessary to use them)

vm/vm.go
```diff
@@ -1208,3 +1224,21 @@ func (vm *VM) Fatal(msg string, fields ...zap.Field) {
 	vm.snowCtx.Log.Fatal(msg, fields...)
 	panic("fatal error")
 }
+
+type apiCoreContext struct {
+	*VM
+}
```
aaronbuchwald, line 1230: Could we remove `apiCoreContext` and add `GetDataDir` and `GetGenesisBytes` directly to the VM in `resolutions.go` ? Not sure why the intermediate type is needed

aaronbuchwald (approved):

aaronbuchwald (approved):

PR #1719, opened 2024-11-05
Remove Funds map from e2e suite

samliok:
This PR removes the `funds` map and fixes a couple of bugs with the spam script logging. Going to create follow up PRs to further decouple the spam script with the transfer action.


tests/fixture/environment.go
```diff
@@ -28,7 +23,6 @@ func NewTestEnvironment(
 ) *e2e.TestEnvironment {
 	// Run only once in the first ginkgo process
 	nodes := tmpnet.NewNodesOrPanic(flagVars.NodeCount())
-	nodes[0].Flags[config.HTTPPortKey] = config.DefaultHTTPPort
```
aaronbuchwald, line 31: This is intended behavior https://github.com/ava-labs/hypersdk/pull/1320.

We should continue to support a stable network identity at least via the run script

throughput/spam.go
```diff
@@ -154,13 +152,20 @@ func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbo
 	s.tracker.logState(cctx, issuers[0].cli)
 
 	// broadcast transactions
-	s.broadcast(cctx, cancel, sh, accounts, funds, factories, issuers, feePerTx, terminate)
+	err = s.broadcast(cctx, cancel, sh, accounts, factories, issuers, feePerTx, terminate)
+	if err != nil {
+		return err
+	}
```
joshua-kim, line 158: nit: up to you to ignore this comment since this is developer preference - but I prefer using the in-line syntax of `if err := ... ; err != nil {` where it's possible

throughput/spam.go
```diff
@@ -208,6 +213,13 @@ func (s Spammer) broadcast(
 						utils.Outf("{{cyan}}skipping issuance because large backlog detected, decreasing target tps:{{/}} %d\n", currentTarget)
 					} else {
 						utils.Outf("{{cyan}}skipping issuance because large backlog detected, cannot decrease target{{/}}\n")
+						consecutiveFailedDecreaseTarget++
+						if consecutiveFailedDecreaseTarget >= failedRunsToDecreaseTarget {
+							utils.Outf("{{cyan}}skipping issuance because large backlog detected, exiting{{/}}\n")
+							broadcastErr = ErrLargeBacklog
+							stop = true
+							break
+						}
```
joshua-kim, line 222: The changes here seem to be a change unrelated goal of removing the funds map... should these be a separate PR?

throughput/spam.go
samliok: i like the inline syntax, but in this case the line would be too many characters so gonna chose to keep it.
In other cases I agree with this

joshua-kim (approved):

throughput/spam.go
```diff
@@ -154,43 +152,48 @@ func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbo
 	s.tracker.logState(cctx, issuers[0].cli)
 
 	// broadcast transactions
-	s.broadcast(cctx, cancel, sh, accounts, funds, factories, issuers, feePerTx, terminate)
+	err = s.broadcast(cctx, sh, accounts, factories, issuers, feePerTx, terminate)
+	cancel()
```
aaronbuchwald, line 156: Is this intended to cancel the context we just passed into broadcast?

PR #1720, opened 2024-11-05
GetActions in SpamHelper interface

samliok:
To be merged after #1719 

Adds a method `GetActions` to the `SpamHelper` interface. This further decouples the `spammer` from morpheus VM


examples/morpheusvm/throughput/helper.go
```diff
@@ -61,3 +74,13 @@ func (*SpamHelper) GetTransfer(address codec.Address, amount uint64, memo []byte
 		Memo:  memo,
 	}}
 }
```
joshua-kim, line 76: Passing comment... but this is so weird that this is on this type since it's entirely unrelated to the struct

throughput/helper.go
```diff
@@ -37,4 +37,7 @@ type SpamHelper interface {
 	// Memo is used to ensure that each transaction is unique (even if between the same
 	// sender and receiver for the same amount).
 	GetTransfer(address codec.Address, amount uint64, memo []byte) []chain.Action
+
+	// GetActions returns a list of actions the spammer sends to the network.
+	GetActions() []chain.Action
```
joshua-kim, line 42: I feel weird about adding this to the interface... I would feel better if we just returned a list of keys that this created (i.e sh.pks) back to the caller so they could figure out how to create the transactions via `GetTransfer`

examples/morpheusvm/throughput/helper.go
```diff
@@ -61,3 +74,13 @@ func (*SpamHelper) GetTransfer(address codec.Address, amount uint64, memo []byte
...
+
+func (sh *SpamHelper) GetActions() []chain.Action {
+	pkIndex := rand.Int() % len(sh.pks)
```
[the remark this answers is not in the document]
samliok: i think this is safe since we are not writing to this array

throughput/helper.go
samliok: we want to decouple `GetTransfer` from the caller(spam script). 
For a little more context, the VM implements this helper interface. The point of this PR is to allow the spam script to generate txs with arbitrary actions not just `transfer`. It just so happens that  the action we want to test for `morpheus` vm in `GetActions` is the same as `GetTransfer`

joshua-kim (approved):

PR #1743, opened 2024-11-05
Benchmarking

richardpringle:
- **Factor out test crate for use in benchmarks**
- **WIP**
- **Add Rust NFT-mint benchmark**
- **More WIP**


x/contracts/runtime/runtime.go
richardpringle, line 1: Forgot about this change, will put in a separate PR

x/contracts/runtime/runtime_test.go
```diff
@@ -117,17 +118,23 @@ func BenchmarkRuntimeInstance(b *testing.B) {
...
 
-		b.StopTimer()
-		// reset module
-		module, err = rt.callContext.r.getModule(ctx, newInfo, programID)
-		require.NoError(err)
-		b.StartTimer()
+		// // reset module
+		// module, err = rt.callContext.r.getModule(ctx, newInfo, programID)
+		// require.NoError(err)
+		// b.StartTimer()
```
richardpringle, line 137: deleteme

x/contracts/runtime/runtime_test.go
```diff
@@ -117,17 +118,23 @@ func BenchmarkRuntimeInstance(b *testing.B) {
 
 	module, err := rt.callContext.r.getModule(ctx, newInfo, programID)
 	require.NoError(err)
+
+	var inst *ContractInstance
```
richardpringle, line 122: Not needed

x/contracts/runtime/runtime_test.go
```diff
@@ -194,9 +201,14 @@ func BenchmarkRuntimeCallContractBasic(b *testing.B) {
...
 	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		runtime.GC()
+		b.StartTimer()
+
 		result, err := contract.CallWithSerializedParams("get_value", nil)
 		require.NoError(err)
-		require.Equal(uint64(0), into[uint64](result))
+		_ = result
+		// require.Equal(uint64(0), into[uint64](result))
```
richardpringle, line 211: Should keep this, but it should be outside of the timer

x/contracts/runtime/util_test.go
```diff
@@ -198,11 +198,24 @@ func (t *testRuntime) CallContract(contract codec.Address, function string, para
...
+	r := callContext.r
+
+	if r.linkerNeedsInitialization {
+		linker, err := r.hostImports.createLinker(r)
+		if err != nil {
+			panic(err)
+		}
+		r.linker = linker
+		r.linkerNeedsInitialization = false
+	}
```
richardpringle, line 214: Don't need this lazy flag anymore. Not sure why it was ever done this way, the host-imports _should_ be loaded when the runtime is initialized.

x/contracts/wasmlanche/Cargo.toml
```diff
@@ -14,9 +14,11 @@ simulator = { workspace = true, optional = true }
 thiserror = { workspace = true }
 
 [dev-dependencies]
-
-[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
-wasmtime = "14"
+criterion = "0.5.1"
+env_logger = "0.11.5"
+iai = "0.1.1"
```
richardpringle, line 19: Only works on linux, will have to feature flag this out

x/contracts/wasmlanche/tests/bench-crate/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+use wasmlanche::{public, Context};
+
+#[public]
+pub fn always_true(_: &mut Context) -> bool {
+    true
+}
```
richardpringle, line 9: Add fibonacci here

x/contracts/wasmlanche/wasmlanche-test/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+        self.state.get(key)
+    }
+
+    #[inline]
+    pub fn take_result(&mut self) -> Option<Vec<u8>> {
+        self.call_result.take()
+    }
+}
+
+pub struct Builder {
```
samliok, line 58: something feels strange to me regarding the relationship of `Builder` and `TestCrate`. Seems like `TestCrate::new(crate_name)` and `TestCrate::build(&self)` would make more sense. We wouldn't need the use the `builder` struct that could all be done internally.  

Then 
```rust
 b.iter_batched(
            || TestCrate::new("test-crate"),
            |crate| {
                let mut contract = Contract::new(crate); // new would call crate.build
                contract.always_true();
                contract
            },
            BatchSize::PerIteration,
        )
```

x/contracts/wasmlanche/wasmlanche-test/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+    engine: Engine,
+    module: Module,
+    linker: Linker<StoreData>,
+}
+
+impl Builder {
+    pub fn new(crate_name: &'static str) -> Self {
+        thread_local! {
```
samliok, line 66: why is thread_local necessary here?

x/contracts/wasmlanche/wasmlanche-test/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+
+                    let result = memory
+                        .data(&caller)
+                        .get(ptr..ptr + len)
+                        .expect("data should exist")
+                        .to_vec();
+
+                    caller.data_mut().set_result(result);
+                },
+            )
```
samliok, line 159: we could wrap `linker` into our own struct and then separate these `.func_wraps` into a separate file. Defining our imported modules & functions here makes the `Builder::new()` function a bit messy & hard to extend.

x/contracts/wasmlanche/wasmlanche-test/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+        let Self {
+            engine,
+            module,
+            linker,
+        } = self;
+
+        let store_data = StoreData {
+            call_result: None,
+            state: HashMap::default(),
+            limiter: StoreLimitsBuilder::new().memory_size(18 * 0x10000).build(),
```
samliok, line 282: why 18 * 65536 bytes? If no particular reason can we make this a more human readable calculation such as
180 * 1000 = 180 KB

samliok (approved):

x/contracts/wasmlanche/benches/call_contract_mem.rs
```diff
@@ -0,0 +1,21 @@
...
+iai::main!(call_contract, call_nft_mint);
+
+fn call_contract() {
+    let builder = Builder::new("test-crate");
+    Contract::new(builder).always_true();
+}
+
+fn call_nft_mint() {
+    let builder = Builder::new("nft");
+    Nft::new(builder).mint(Address::new(black_box([2; 33])), black_box(0));
```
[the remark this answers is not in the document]
richardpringle: 🤔 I get the sense that their example doesn't make sense. Or maybe it makes sense when there's a finite number of inputs, but iter-batched and friends dynamically generate their "size" (number of times they execute), so it's impossible to optimize that. At best, it could get rid of an indirection, but that would probably be fine.

x/contracts/wasmlanche/wasmlanche-test/src/lib.rs
richardpringle: 1 << 16 is the default page-size. I can leave a comment

x/contracts/wasmlanche/wasmlanche-test/src/lib.rs
richardpringle: `RefCell` isn't `Sync` (required for static). `static mut` is unsafe (and apparently often unsound even in the single-threaded case). That means if I want to memoize the `new` function, I need to either use `std::thread::LocalKey<RefCell>`, a `Mutex`, or a `RwLock`.  The APIs for `std::sync::{Mutex, RwLock}` are a pain because they return results that need to be unwrapped to be used.

I believe that the benchmarks that exist right now are run in a single-thread anyway so there would be no benefit to using a `Sync` type. 

Good question though!

x/contracts/wasmlanche/wasmlanche-test/src/lib.rs
richardpringle: The thing that's strange is that I'm cheating the pattern a little. This should really be two builders and a resultant type. 
```rust
// really a module, engine, and linker, but the engine and linker are singletons for us
let module = ModuleBuilder::new("test-crate").build();
let test_instance = module.instantiate(); 
```
Happy to change it to this if you think that would be clearer.

With the naming from your example above, the `TestCrate` is the type that's doing the work to _construct_ the `Contract`, so it's actually a bit of an anti-pattern to have to pass it into a constructor instead of just having a method called `TestCrate::instantiate_contract` or whatever you want to call it here. 

But in the end, you're right, there was definitely a code-smell here.

x/contracts/wasmlanche/wasmlanche-test/src/lib.rs
richardpringle: I've got a fix for this, I'll add it as a separate commit

richardpringle: 1b4e109ccacfadfbe3250bd3222dd24d8a7087d4

samliok (dismissed):

samliok (approved):

PR #1744, opened 2024-11-06
Stateful block to vm

aaronbuchwald:
This PR moves `StatefulBlock` from the chain package to the VM package.

The `StatefulBlock` type handles the state sync invariants, which is implemented by the syncervm client/server in the VM package, so this should be moved so that the VM package is solely responsible for state sync.

The `StatefulBlock` type also handles determining whether a block has been processed, when a block should be executed in verify vs deferred to handled on accept, etc.

This PR separates all of that logic from the chain package, which is now solely responsible for handling the state transition function of a HyperVM (VM using actions/auth). This requires that it can:
- build a block
- parse a block
- execute a block given a parent state and return a result and newly computed state
- pre-execute a transaction before allowing it in the mempool

The end goal for this refactor is to decouple the VM package as well from knowing the block/result/state types required by the chain package (https://github.com/ava-labs/hypersdk/issues/1278).


chain/chain.go
```diff
@@ -0,0 +1,330 @@
...
+func (c *Chain) ParseBlock(ctx context.Context, b []byte) (*ExecutionBlock, error) {
+	_, span := c.tracer.Start(ctx, "chain.ParseBlock")
+	defer span.End()
+
+	blk, err := UnmarshalBlock(b, c.parser)
+	if err != nil {
+		return nil, err
+	}
+
+	// Setup signature verification job
```
aaronbuchwald, line 291: TODO: only start signature verification for a block we might need to verify

vm/block.go
```diff
@@ -13,313 +13,78 @@ import (
...
 		vm:             vm,
-		executor:       NewExecutor(vm),
-		id:             utils.ToID(source),
-	}
-
-	// If we are parsing an older block, it will not be re-executed and should
-	// not be tracked as a parsed block
-	lastAccepted := b.vm.LastAcceptedBlock()
-	if lastAccepted == nil || b.Hght <= lastAccepted.Hght { // nil when parsing genesis
-		return b, nil
```
aaronbuchwald, line 286: Should add this optimization back into chain. Note: doing this early can cause us to do unnecessary signature verification of any block that we are willing to parse

vm/option.go
```diff
@@ -85,9 +85,3 @@ func WithVMAPIs(apiHandlerFactories ...api.HandlerFactory[api.VM]) RegisterFunc
 		vm.vmAPIHandlerFactories = append(vm.vmAPIHandlerFactories, apiHandlerFactories...)
 	}
 }
-
-func WithTxRemovedSubscriptions(subscriptions ...event.SubscriptionFactory[TxRemovedEvent]) RegisterFunc {
-	return func(vm *VM) {
-		vm.txRemovedSubscriptionFactories = append(vm.txRemovedSubscriptionFactories, subscriptions...)
-	}
-}
```
aaronbuchwald, line 93: Can split this out into a separate PR. Given the limited cases where we can send a notification (only if auth fails, which will only happen for an incorrect/malicious client), clients cannot rely on it for any useful information so we may as well remove it entirely.

vm/resolutions.go
```diff
@@ -79,10 +87,18 @@ func (vm *VM) Rules(t int64) chain.Rules {
...
 	return vm.lastAccepted
 }
 
+func (vm *VM) LastAcceptedBlock() *chain.ExecutedBlock {
+	return vm.lastAccepted.executedBlock
+}
+
+func (vm *VM) LastAcceptedExecutionBlock() *chain.ExecutionBlock {
+	return vm.lastAccepted.ExecutionBlock
+}
```
aaronbuchwald, line 100: Clean up the returned block types

vm/resolutions.go
```diff
@@ -129,16 +138,15 @@ func (vm *VM) Verified(ctx context.Context, b *chain.StatefulBlock) {
...
 			"verified block",
 			zap.Stringer("blkID", b.ID()),
 			zap.Uint64("height", b.Hght),
 			zap.Int("txs", len(b.Txs)),
 			zap.Stringer("parent root", b.StateRoot),
 			zap.Bool("state ready", vm.StateReady()),
-			zap.Any("unit prices", fm.UnitPrices()),
-			zap.Any("units consumed", fm.UnitsConsumed()),
+			zap.Any("unit prices", b.executedBlock.UnitPrices),
+			zap.Any("units consumed", b.executedBlock.UnitsConsumed),
```
aaronbuchwald, line 149: All of these logs and metrics should be moved into the `chain/` package

vm/vm.go
```diff
@@ -324,11 +323,16 @@ func (vm *VM) Initialize(
...
-	vm.acceptedQueue = make(chan *chain.StatefulBlock, vm.config.AcceptorSize)
+	vm.acceptedQueue = make(chan *StatefulBlock, vm.config.AcceptorSize)
 	vm.acceptorDone = make(chan struct{})
 
 	vm.mempool = mempool.New[*chain.Transaction](vm.tracer, vm.config.MempoolSize, vm.config.MempoolSponsorSize)
 
+	vm.chain, err = chain.NewChain(vm, vm.config.ExecutionConfig, vm)
+	if err != nil {
+		return err
+	}
```
aaronbuchwald, line 334: Make this pluggable

vm/vm.go
```diff
@@ -374,17 +378,26 @@ func (vm *VM) Initialize(
...
 		)
 		if err != nil {
 			snowCtx.Log.Error("unable to init genesis block", zap.Error(err))
 			return err
 		}
+		// Set executed block, since we will never execute the genesis block
+		genesisBlk.executedBlock = &chain.ExecutedBlock{
+			BlockID: genesisBlk.ID(),
+			Block:   genesisExecutionBlk.StatelessBlock,
+		}
```
aaronbuchwald, line 400: TODO: clean up init genesis

vm/vm.go
```diff
@@ -846,13 +854,28 @@ func (vm *VM) BuildBlock(ctx context.Context) (snowman.Block, error) {
...
+		return nil, err
+	}
+	executionBlk, executedBlk, view, err := vm.chain.BuildBlock(ctx, preferredView, preferredBlk.ExecutionBlock)
 	if err != nil {
 		// This is a DEBUG log because BuildBlock may fail before
 		// the min build gap (especially when there are no transactions).
 		vm.snowCtx.Log.Debug("BuildBlock failed", zap.Error(err))
 		return nil, err
 	}
+	blk := &StatefulBlock{
```
aaronbuchwald, line 869: TODO: centralize where we create stateful block instances to make it easier to reason about

api/ws/server.go
```diff
@@ -66,11 +66,6 @@ func OptionFunc(v api.VM, config Config) (vm.Opt, error) {
 	)
 
 	webSocketFactory := NewWebSocketServerFactory(handler)
-	txRemovedSubscription := event.SubscriptionFuncFactory[vm.TxRemovedEvent]{
-		AcceptF: func(event vm.TxRemovedEvent) error {
-			return server.RemoveTx(event.TxID, event.Err)
-		},
-	}
```
aaronbuchwald, line 73: This can be factored out to a separate PR if needed. Removed this as part of this change so that `PreExecute` can move into the `chain/` package and would prefer not to force that function to differentiate between an error that can notify the caller the tx will never go through vs. may go through in the future. This is not currently providing any new information to the caller since it only sends this notification for an invalid signature, which the caller could always check itself.

chain/builder.go
```diff
@@ -59,61 +77,53 @@ func HandlePreExecute(log logging.Logger, err error) bool {
...
-	ctx, span := vm.Tracer().Start(ctx, "chain.BuildBlock")
-	defer span.End()
-	log := vm.Logger()
+type buildBlock struct {
+	parentID  ids.ID
+	timestamp int64
+	height    uint64
+	txs       []*Transaction
+	stateRoot ids.ID
+}
```
aaronbuchwald, line 86: add an intermediate type rather than creating a partially initialized instance of a block

chain/processor.go
```diff
@@ -22,97 +21,77 @@ import (
...
 	parentFeeManager := fees.NewManager(feeRaw)
 	feeManager, err := parentFeeManager.ComputeNext(b.Tmstmp, r)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, err
 	}
 
 	// Process transactions
-	results, ts, err := e.ExecuteTxs(ctx, b, e.vm.Tracer(), parentView, feeManager, r)
+	results, ts, err := c.executeTxs(ctx, b, c.tracer, parentView, feeManager, r)
```
aaronbuchwald, line 91: no need to pass this in

vm/config.go
```diff
@@ -39,6 +38,7 @@ type Config struct {
 	ProcessingBuildSkip              int                        `json:"processingBuildSkip"`
 	TargetGossipDuration             time.Duration              `json:"targetGossipDuration"`
 	BlockCompactionFrequency         int                        `json:"blockCompactionFrequency"`
+	ExecutionConfig                  chain.ExecutionConfig      `json:"executionConfig"`
```
aaronbuchwald, line 41: Register chain similar to `ServiceConfig`

aaronbuchwald:
Need to fix state sync signaling when it's ready. Should enable the chain to specify the number of blocks from the state sync point it requires to start.

api/dependencies.go
```diff
@@ -36,9 +36,7 @@ type VM interface {
 		verifySig bool,
 		txs []*chain.Transaction,
 	) (errs []error)
-	// LastAcceptedBlock provides the most recent block that the VM has accepted.
-	// The value returned is guaranteed to be non-nil.
-	LastAcceptedBlock() *chain.StatefulBlock
+	LastAcceptedBlockResult() *chain.ExecutedBlock
```
joshua-kim, line 39: nit: Feel like the rename isn't needed, since the result is technically a field encoded in the block type

chain/assembler.go
```diff
@@ -5,70 +5,46 @@ package chain
...
-type Assembler struct {
-	vm VM
-}
-
-func (a *Assembler) AssembleBlock(
+func (c *Chain) AssembleBlock(
 	ctx context.Context,
-	parent *StatefulBlock,
+	parentView state.View,
+	parent *ExecutionBlock,
```
joshua-kim, line 15: Q: Why we can't just use the `StatelessBlock` type here? Not really critical that we address anything in this file now since this is dead code anyways though

chain/assembler.go
```diff
@@ -5,70 +5,46 @@ package chain
...
-func (a *Assembler) ExecuteBlock(
+func (c *Chain) ExecuteBlock(
 	ctx context.Context,
-	b *StatefulBlock,
-) (*ExecutedBlock, error) {
-	ctx, span := a.vm.Tracer().Start(ctx, "chain.ExecuteBlock")
+	parentView state.View,
+	b *ExecutionBlock,
+) (*ExecutedBlock, state.View, error) {
+	ctx, span := c.tracer.Start(ctx, "chain.ExecuteBlock")
```
joshua-kim, line 46: This is dead code but this function looks odd because the caller of `ExecuteBlock` has everything it already needs to call `Execute` so this function looks like it's not needed in the current implementation if we end up deciding that `Chain` owns `ExecuteBlock`.

chain/builder.go
```diff
@@ -387,62 +395,73 @@ func BuildBlock(
...
-		zap.Int("added", len(b.Txs)),
+		zap.Int("added", len(b.txs)),
 		zap.Int("state changes", ts.PendingChanges()),
 		zap.Int("state operations", ts.OpIndex()),
 		zap.Int64("parent (t)", parent.Tmstmp),
-		zap.Int64("block (t)", b.Tmstmp),
+		zap.Int64("block (t)", b.timestamp),
 	)
-	return b, nil
+	return NewExecutionBlockNoVerify(blk), &ExecutedBlock{
```
joshua-kim, line 461: It feels weird to me that we return both the execution+executed block, can't we just return the executed block result?

chain/chain.go
```diff
@@ -0,0 +1,270 @@
...
+	Mempool() Mempool
+	Logger() logging.Logger
+	Parser
+	RuleFactory() RuleFactory
+	MetadataManager() MetadataManager
+	BalanceHandler() BalanceHandler
+	AuthVerifiers() workers.Workers
+	GetAuthBatchVerifier(authTypeID uint8, cores int, count int) (AuthBatchVerifier, bool)
+	ChainIndex
+}
```
joshua-kim, line 61: It seems like this interface is being used just to hold a bunch of dependencies... if so I would prefer if we just individually passed them as args into `NewChain`.

chain/chain.go
```diff
@@ -0,0 +1,270 @@
...
+	"go.uber.org/zap"
+
+	"github.com/ava-labs/hypersdk/internal/emap"
+	"github.com/ava-labs/hypersdk/internal/workers"
+	"github.com/ava-labs/hypersdk/state"
+
+	internalfees "github.com/ava-labs/hypersdk/internal/fees"
+)
+
+type Chain struct {
```
joshua-kim, line 26: I feel like `Chain` is a weird abstraction to me - it's not clear to me what it's responsibility is and it feels like it's turning into a [god-object](https://en.wikipedia.org/wiki/God_object).

chain/execution_block.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"context"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/set"
+
+	"github.com/ava-labs/hypersdk/internal/workers"
+)
+
+type ExecutionBlock struct {
```
joshua-kim, line 15: I'm wondering if this needs to be exported. It seems to me like an execution block represents a block that's being executed, and I don't know why things would need to depend on it outside of the execution (seems like this is what the `chain` package is currently). I wonder if consumers of `chain` should just depend on the `ExecutedBlock` instead?

chain/assembler.go
aaronbuchwald: We could use `StatelessBlock` in DSMR since execution occurs only accept, so there's no need to kick off async verification separately on parse.

We could kick off async signature verification of transactions received in chunks instead

chain/assembler.go
aaronbuchwald: Will remove and change `Assemble` to `AssembleAndExecute` and return the result and state types as well

chain/builder.go
```diff
@@ -36,6 +37,23 @@ const (
...
+		time.Duration,
+		func(context.Context, *Transaction) (cont bool, rest bool, err error),
+	) error
+
+	StartStreaming(context.Context)
+	PrepareStream(context.Context, int)
+	Stream(context.Context, int) []*Transaction
+	FinishStreaming(context.Context, []*Transaction) int
+}
+
```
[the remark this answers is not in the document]
aaronbuchwald: I can move this back

chain/builder.go
aaronbuchwald: Will change to local variables

chain/chain.go
aaronbuchwald: This is effectively replacing the static functions from the chain package that were used before.

It's not perfect, but much better than the previous code imo. The end goal is to decouple the VM from the block, state, and result types completely.

We could separate the builder/parser/accepter etc, but they should all have the same block/state/result types.

chain/execution_block.go
aaronbuchwald: This is used to handle async signature verification and caching the `txsSet` when checking for repeat transactions.

This PR is only meant to re-structure the code, but keep those functions the same and keep a `StatelessBlock` type that has nothing to do with them.

chain/validity_window.go
```diff
@@ -0,0 +1,142 @@
...
+	if dup.Len() > 0 {
+		return fmt.Errorf("%w: duplicate in ancestry", ErrDuplicateTx)
+	}
+	return nil
+}
+
+func (v *TimeValidityWindow) IsRepeat(
+	ctx context.Context,
+	parentBlk *ExecutionBlock,
+	timestamp int64,
```
aaronbuchwald, line 86: remove unused param

chain/stateless_block.go
```diff
@@ -0,0 +1,150 @@
...
+	// Ensure no leftover bytes
+	if !p.Empty() {
+		return nil, fmt.Errorf("%w: remaining=%d", ErrInvalidObject, len(raw)-p.Offset())
+	}
+	b.bytes = raw
+	b.id = utils.ToID(raw)
+	return &b, p.Err()
+}
+
+func NewGenesisBlock(root ids.ID) (*ExecutionBlock, error) {
```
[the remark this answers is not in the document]
aaronbuchwald: The genesis block is chain specific, so the eventual goal is for the chain type to handle this rather than the VM needing to know about the genesis, but relying on the chain for all other blocks.

chain/builder.go
aaronbuchwald: The block is the actual container being executed and the result is for APIs

vm/block.go
```diff
@@ -13,313 +13,79 @@ import (
...
-
-		// StateRoot should include all allocates made when loading the genesis file
-		StateRoot: root,
-	}
-}
-
 // StatefulBlock is defined separately from "StatelessBlock"
 // in case external packages need to use the stateless block
 // without mocking VM or parent block
 type StatefulBlock struct {
```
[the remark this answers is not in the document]
aaronbuchwald: The VM currently needs to be able to serve the last accepted version of each in `vm/resolutions.go` so we need some different name to differentiate them.

Right now it is:
- LastAcceptedStatefulBlock - currently required by integration tests, syncervm server, and verify context (can remvoe the latter two)
- LastAcceptedBlock - currently required for validity window replay protection in chain package
- LastAcceptedBlockResult - currently required by `Accepted` API and integration tests

Open to any renaming here or changing how to handle the replay protection and populating tx sets / kicking off signature verification.

api/dependencies.go
aaronbuchwald: Open to suggestion to eliminate the need for any of them https://github.com/ava-labs/hypersdk/pull/1744#discussion_r1834517278

chain/chain.go
aaronbuchwald: This is following the existing pattern where the VM is effectively the "god object" that holds all of these dependencies.

I'd prefer to keep it as is in this PR and break it up in a future PR, but can break it up to: `chain.New(vm.Tracer(), vm.Metrics(), vm.Mempool(), ..., vm)` in this PR if you prefer

chain/chain.go
aaronbuchwald: Separated them out into individual components, but kept a single chain type to wrap them together

chain/assembler.go
aaronbuchwald: kept the name as `Assemble`, but included execution in it

vm/resolutions.go
```diff
@@ -129,32 +137,23 @@ func (vm *VM) Verified(ctx context.Context, b *chain.StatefulBlock) {
...
 
 	if b.Processed() {
-		fm := b.FeeManager()
 		vm.snowCtx.Log.Info(
 			"verified block",
-			zap.Stringer("blkID", b.ID()),
-			zap.Uint64("height", b.Hght),
-			zap.Int("txs", len(b.Txs)),
-			zap.Stringer("parent root", b.StateRoot),
+			zap.Stringer("executedBlk", b.executedBlock),
```
joshua-kim, line 142: Can we make the key here "blk" to be consistent w/ the rest of the logs?

vm/verify_context.go
```diff
@@ -24,21 +26,12 @@ func (p *PendingVerifyContext) View(ctx context.Context, verify bool) (state.Vie
...
-
 type AcceptedVerifyContext struct {
-	vm VM
+	vm *VM
 }
 
 // We disregard [verify] because [GetVerifyContext] ensures
 // we will never need to verify a block if [AcceptedVerifyContext] is returned.
-func (a *AcceptedVerifyContext) View(context.Context, bool) (state.View, error) {
+func (a *AcceptedVerifyContext) View(_ context.Context, _ bool) (state.View, error) {
```
joshua-kim, line 35: nit: undo

vm/vm_test.go
```diff
@@ -17,25 +17,30 @@ import (
 	"github.com/ava-labs/hypersdk/chain"
 	"github.com/ava-labs/hypersdk/genesis"
 	"github.com/ava-labs/hypersdk/internal/cache"
-	"github.com/ava-labs/hypersdk/internal/emap"
 	"github.com/ava-labs/hypersdk/internal/mempool"
 	"github.com/ava-labs/hypersdk/internal/trace"
 )
 
 func TestBlockCache(t *testing.T) {
+	t.Skip()
```
joshua-kim, line 25: This is such a bizarre unit test that previously existed... we circumvent vm initialization to test we can fetch blocks from the VM since if it's database doesn't exist, it must be serving requests through cache which feels brittle + too introspect-y. I would be open to just deleting this test instead of skipping it.

aaronbuchwald: Will do

vm/verify_context.go
aaronbuchwald: Will do

vm/resolutions.go
aaronbuchwald: Will do

chain/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+		"time spent waiting for signature verification in verify",
+		r,
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	m := &chainMetrics{
+		buildCapped: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: "chain",
```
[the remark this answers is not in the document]
aaronbuchwald: Yes

chain/validity_window.go
```diff
@@ -0,0 +1,139 @@
...
+	GetExecutionBlock(ctx context.Context, blkID ids.ID) (*ExecutionBlock, error)
+	LastAcceptedBlock() *ExecutionBlock
+}
+
+type TimeValidityWindowBackend interface {
+	Logger() logging.Logger
+	Tracer() trace.Tracer
+	ChainIndex
+}
+
```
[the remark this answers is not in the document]
aaronbuchwald: Will do

chain/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (em *executorMetrics) RecordExecutable() {
+	em.executable.Inc()
+}
+
+func newMetrics() (*prometheus.Registry, *chainMetrics, error) {
+	r := prometheus.NewRegistry()
+
+	rootCalculated, err := metric.NewAverager(
+		"chain_root_calculated",
```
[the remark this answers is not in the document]
aaronbuchwald: Created a separate issue since this PR is moving existing metrics https://github.com/ava-labs/hypersdk/issues/1765

chain/processor.go
```diff
@@ -318,10 +354,50 @@ func (e *Executor) ExecuteTxs(
...
+		return nil
+	}
+	sigJob, err := p.authVerificationWorkers.NewJob(len(block.Txs))
+	if err != nil {
+		return err
+	}
+	block.sigJob = sigJob
+
+	// Setup signature verification job
+	_, sigVerifySpan := p.tracer.Start(ctx, "Chain.AsyncVerify.verifySignatures") //nolint:spancheck
```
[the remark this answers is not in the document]
aaronbuchwald: Ya that's a good idea. Separate PR?

chain/chain.go
```diff
@@ -0,0 +1,121 @@
...
+	builder     *Builder
+	processor   *Processor
+	accepter    *Accepter
+	preExecutor *PreExecutor
+	blockParser *BlockParser
+}
+
+func NewChain(
+	tracer trace.Tracer,
+	gatherer metrics.MultiGatherer,
```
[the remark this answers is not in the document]
aaronbuchwald: Good idea

chain/metrics.go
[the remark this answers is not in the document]
aaronbuchwald: Removed from this PR, but leaving the issue to clean up everywhere else it's used

chain/chain.go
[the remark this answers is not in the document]
aaronbuchwald: Added a note in https://github.com/ava-labs/hypersdk/issues/1765 to remove the special case AvalancheGo metric types throughout the rest of the repo

joshua-kim (approved):

chain/processor.go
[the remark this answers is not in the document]
joshua-kim: Made an issue: https://github.com/ava-labs/hypersdk/issues/1766

PR #1747, opened 2024-11-06
Enforce Contract in `AddBalance`

RodrigoVillar:
This PR updates `AddBalance()` in MorpheusVM so that, if an account doesn't exist and if it wasn't given permission to create an account, that it returns an error. This PR also updates the documentation of `AddBalance()` to make it explicit that if the account doesn't exist and `createAccount = false`, it should return an error.

### DIFF
--- a/chain/dependencies.go
+++ b/chain/dependencies.go
@@ -171,7 +171,7 @@ type BalanceHandler interface {
 	Deduct(ctx context.Context, addr codec.Address, mu state.Mutable, amount uint64) error
 
 	// AddBalance adds [amount] to [addr].
-	// If [addr] doesn not exist and [createAccount] is false, this should return an error.
+	// If [addr] does not exist and [createAccount] is false, this should return an error.
 	AddBalance(ctx context.Context, addr codec.Address, mu state.Mutable, amount uint64, createAccount bool) error
 
 	// GetBalance returns the balance of [addr].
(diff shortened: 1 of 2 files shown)


aaronbuchwald:
What contract in the balance handler are you referring to? Do you just mean based on the name of the function?

RodrigoVillar:
> What contract in the balance handler are you referring to? Do you just mean based on the name of the function?

Yes, the name/interface of the function is part of the contract. By naming the parameter `createAccount`, the contract that's implied is that:

- If the account doesn't exist, but `createAccount = true`, then create the account. Otherwise, I should return an error.

RodrigoVillar:
Also want to add that by currently returning a no-op, this PR blocks #1693 as there's no way to test between setting `createAccount = true` and `createAccount = false`

aaronbuchwald:
> > What contract in the balance handler are you referring to? Do you just mean based on the name of the function?
> 
> Yes, the name/interface of the function is part of the contract. By naming the parameter `createAccount`, the contract that's implied is that:
> 
> * If the account doesn't exist, but `createAccount = true`, then create the account. Otherwise, I should return an error.

I don't think that name explicitly specifies whether it should return an error or treat it as a no-op, so disagree on that point, but do think it makes sense to change the behavior to return an error and add a comment to the interface to specify that behavior.

RodrigoVillar:
Closing in favor of removing the `createAccount` parameter entirely

PR #1758, opened 2024-11-12
Respond to Chunk Signature Requests with Warp Signatures

joshua-kim:
Adds a feature to respond to chunk signature requests w/ warp signatures.


proto/dsmr/dsmr.proto
```diff
@@ -21,13 +21,7 @@ message GetChunkSignatureRequest {
 }
 
 message GetChunkSignatureResponse {
-  // TODO are these fields needed?
-  bytes chunk_id = 1;
-  bytes producer = 2;
-  int64 expiry = 3;
-
-  bytes signer = 4;
```
joshua-kim, line 29: These were in the original DSMR spec, but I don't think they're needed since the chunk metadata can be inferred from the request type (which requester knows), and the signer is known implicitly since the requester knows the node-id of who they're requesting from, so they can lookup the public key in the P-Chain.

x/dsmr/typed_client.go
```diff
@@ -150,10 +173,13 @@ func NewGetChunkClient[T Tx](client *p2p.Client) *TypedClient[*dsmr.GetChunkRequ
 	}
 }
 
-func NewGetChunkSignatureClient(client *p2p.Client) *TypedClient[*dsmr.GetChunkSignatureRequest, *dsmr.GetChunkSignatureResponse, []byte] {
+func NewGetChunkSignatureClient(networKID uint32, chainID ids.ID, client *p2p.Client) *TypedClient[*dsmr.GetChunkSignatureRequest, *dsmr.GetChunkSignatureResponse, []byte] {
```
aaronbuchwald, line 176: nit: networkID

x/dsmr/node_test.go
```diff
@@ -1313,6 +1455,29 @@ func TestAccept_RequestReferencedChunks(t *testing.T) {
 	<-done
 }
 
+func getSignerBitSet(t *testing.T, pChain validators.State, nodeIDs ...ids.NodeID) set.Bits {
```
aaronbuchwald, line 1458: This is super nice, would be great to add this as a convenience function to AvalancheGo

aaronbuchwald (approved):

PR #1762, opened 2024-11-12
fix pubsub flacky tests

tsachiherman:
## What ? 

This PR fix few issues with `TestServerPublish` and `TestServerPublishSpecific` unit tests.

## Details ( main issue )

In `TestServerPublishSpecific`, the test was calling `peekCon, _ := handler.conns.Peek()` in order to get the list of connections. However, that might not have been updated yet ( similar issue in `TestServerPublish` ).

If we'll take a look on `ServeHTTP`, we'll see that:
```golang
func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	// Upgrader.upgrade() is called to upgrade the HTTP connection.
	// No nead to set any headers so we pass nil as the last argument.
	wsConn, err := s.upgrader.Upgrade(w, r, nil)
	if err != nil {
		s.log.Warn("failed to upgrade",
			zap.Error(err),
		)
		return
	}
	s.addConnection(&Connection{
		s:      s,
		conn:   wsConn,
		mb:     NewMessageBuffer(s.log, s.config.MaxPendingMessages, s.config.MaxWriteMessageSize, s.config.MaxMessageWait),
		active: atomic.Bool{},
	})
	s.log.Debug("added pubsub connection", zap.Stringer("addr", wsConn.RemoteAddr()))
}
```

The output to the client ( i.e. `ResponseWriter` ) is being sent before calling `s.addConnection`. That means that a client creating a connection might be able to receive a `webSocket` connection before the connection is being added by the server to the list of available connection.

## Details ( secondary issues )

* The use of `Eventually` was causing a notable test slowdown, since it would *first wait* and then evaluate the condition function. This was addressed by `optimisticEventuallity` which *evaluate first*, than call `Eventually`.
* In `TestServerPublishSpecific`, there was no explicit waiting for the clients to complete before testing the number open connection. This would prevent the observer from figuring out if this is a client side or server side issue.
* In `TestServerPublishSpecific`, the waiting for shutdown had no time limit. This could cause the unit test to "hang up" before aborting with a go testing timeout panic.


pubsub/server_test.go
```diff
@@ -190,12 +196,32 @@ func TestServerRead(t *testing.T) {
...
+	case <-time.After(timeout):
+		return true // timed out
+	}
+}
+
+func optimisticEventuallity(require *require.Assertions, condition func() bool, waitFor time.Duration, tick time.Duration, msgAndArgs ...interface{}) { //nolint:unparam
+	if !condition() {
+		require.Eventually(condition, waitFor, tick, msgAndArgs...)
+	}
+}
```
aaronbuchwald, line 219: is the purpose of this function just to make sure that we run the condition first before waiting?

pubsub/server_test.go
```diff
@@ -254,24 +296,38 @@ func TestServerPublishSpecific(t *testing.T) {
...
-		err := webCon2.SetReadDeadline(time.Now().Add(5 * time.Second))
+		defer clientSideClosedConnection.Done()
+		err := webCon2.SetReadDeadline(time.Now().Add(50 * time.Millisecond))
 		require.NoError(err, "Error setting connection deadline.")
 		// Make sure connection wasn't written too
 		_, _, err = webCon2.ReadMessage()
 		require.Error(err, "Error not thrown.") //nolint:forbidigo
 		netErr, ok := err.(net.Error)
-		require.True(ok, "Error is not a net.Error")
+		require.True(ok, "Error is not a net.Error ( %v )", netErr)
```
aaronbuchwald, line 306: nit: should this be `True` with net error as a separate argument rather than including the formatter in the msg ?

pubsub/server_test.go
```diff
@@ -254,24 +296,38 @@ func TestServerPublishSpecific(t *testing.T) {
...
 		require.True(netErr.Timeout(), "Error is not a timeout error")
+		// close the connection without letter the server know about it.
```
aaronbuchwald, line 308:
```suggestion
		// close the connection without notifying the server
```

pubsub/server_test.go
```diff
@@ -254,24 +296,38 @@ func TestServerPublishSpecific(t *testing.T) {
...
 	err = server.Shutdown(context.TODO())
 	require.NoError(err, "Error shuting down server.")
+
 	// Wait for the server to finish shutting down
-	<-serverDone
+	select {
+	case <-serverDone:
+		// great!
+	case <-time.After(500 * time.Millisecond):
+		require.FailNow("shutting down server takes too lone")
```
aaronbuchwald, line 331:
```suggestion
		require.FailNow("shutting down server takes too long")
```

aaronbuchwald (approved):
few small nits / Qs and then this LGTM

pubsub/server_test.go
tsachiherman: yes, exactly. I'll add a comment above.

PR #1767, opened 2024-11-12
Generalize builder

tsachiherman:
## What ?

Decouple components that are unnecessarily tied to the chain package because they depend on either the block or transaction types. We should remove the pattern of using a resolutions.go and dependencies.go file to pass around mega objects and switch to passing in the required parameters to each type we need to construct.

This PR focuses on decoupling the builder from the chain package.


internal/builder/time.go
```diff
@@ -19,22 +20,32 @@ import (
 // TODO: consider replacing this with AvalancheGo block build metering
 const minBuildGap int64 = 25 // ms
 
-var _ Builder = (*Time)(nil)
+type Mempool interface {
+	Len(context.Context) int // items
+}
+
+type GetPreferedTimestampAndBlockGap func(int64) (int64, int64)
```
aaronbuchwald, line 27: nit: could we add variable names and a comment to make this more easily readable and an error, so that we don't need to use a special case of the latter return value?

tsachiherman: sure!

internal/builder/time.go
```diff
@@ -66,39 +79,39 @@ func (b *Time) nextTime(now int64, preferred int64) int64 {
...
+		b.logger.Debug("unable to acquire waiting lock")
 		return
 	}
-	preferredBlk, err := b.vm.PreferredBlock(context.TODO())
+	now := time.Now().UnixMilli()
+	preferred, gap, err := b.getPreferedTimestampAndBlockGap(now)
 	if err != nil {
-		b.waiting.Store(false)
-		b.vm.Logger().Warn("unable to load preferred block", zap.Error(err))
+		// unable to retrieve block.
```
aaronbuchwald, line 88:
```suggestion
		// unable to retrieve block.
		b.waiting.Store(false)
```

should we keep `b.waiting.Store` ? This error should not happen, but seems we should keep the error handling the same with this change

tsachiherman: yes, good catch. I missed that.

aaronbuchwald (approved):

vm/vm.go
```diff
@@ -324,6 +323,14 @@ func (vm *VM) Initialize(
 
 	vm.mempool = mempool.New[*chain.Transaction](vm.tracer, vm.config.MempoolSize, vm.config.MempoolSponsorSize)
 
+	vm.builder = builder.NewTime(toEngine, snowCtx.Log, vm.mempool, func(t int64) (int64, int64, error) {
```
aaronbuchwald, line 326: Hmm moving this below where we apply options means that we will set the builder to this value even if the option to override the builder is specified. Moving this back above options causes the integration tests (which use the manual builder) to fail.

Seems this change may have introduced a bug in the manual builder that is covered up by this change. We could use `builder.NewTime` rather than manual implementation (clearly we don't need the manual implementation).

PR #1769, opened 2024-11-13
Remove pebble avametrics

aaronbuchwald:
This PR removes the usage of the AvalancheGo multigatherer and `metric.Averager` from pebble

### DIFF
--- a/api/indexer/indexer.go
+++ b/api/indexer/indexer.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/wrappers"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/ava-labs/hypersdk/chain"
 	"github.com/ava-labs/hypersdk/codec"
@@ -45,11 +46,11 @@ func NewIndexer(path string, parser chain.Parser, blockWindow uint64) (*Indexer,
 	if blockWindow > maxBlockWindow {
 		return nil, fmt.Errorf("block window %d exceeds maximum %d", blockWindow, maxBlockWindow)
 	}
-	txDB, _, err := pebble.New(filepath.Join(path, "tx"), pebble.NewDefaultConfig())
+	txDB, err := pebble.New(filepath.Join(path, "tx"), pebble.NewDefaultConfig(), prometheus.NewRegistry())
 	if err != nil {
 		return nil, err
 	}
-	blockDB, _, err := pebble.New(filepath.Join(path, "block"), pebble.NewDefaultConfig())
+	blockDB, err := pebble.New(filepath.Join(path, "block"), pebble.NewDefaultConfig(), prometheus.NewRegistry())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


internal/pebble/metrics.go
```diff
@@ -33,27 +35,28 @@ type metrics struct {
...
-	)
-	if err != nil {
-		return nil, nil, err
-	}
+func newMetrics(r prometheus.Registerer) (*metrics, error) {
 	m := &metrics{
-		writeStall: writeStall,
-		getLatency: getLatency,
+		writeStallCount: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: "pebble",
```
joshua-kim, line 41: Wondering if passing in a namespace from the caller here makes more sense as well.

internal/pebble/metrics.go
```diff
@@ -107,6 +110,10 @@ func newMetrics() (*prometheus.Registry, *metrics, error) {
 	}
 	errs := wrappers.Errs{}
 	errs.Add(
+		r.Register(m.writeStallCount),
+		r.Register(m.writeStallSum),
+		r.Register(m.getCount),
+		r.Register(m.getSum),
```
joshua-kim, line 116: !

vm/vm.go
```diff
@@ -212,13 +212,21 @@ func (vm *VM) Initialize(
...
-	vm.vmDB, err = storage.New(pebbleConfig, vm.snowCtx.ChainDataDir, blockDB, vm.snowCtx.Metrics)
+	vm.vmDB, err = storage.New(pebbleConfig, vm.snowCtx.ChainDataDir, blockDB, blockDBRegistry)
 	if err != nil {
 		return err
 	}
 
-	vm.rawStateDB, err = storage.New(pebbleConfig, vm.snowCtx.ChainDataDir, stateDB, vm.snowCtx.Metrics)
+	rawStateDBRegistry := prometheus.NewRegistry()
+	if err := vm.snowCtx.Metrics.Register("rawstatedb", blockDBRegistry); err != nil {
+		return fmt.Errorf("failed to register blockdb metrics: %w", err)
```
joshua-kim, line 227: 1. Looks like a copy-paste error here and seems like we should be using `rawStateDBRegistry` + in the log.
2. I really don't know much about prometheus - I'm not sure when we're supposed to be sharing a metrics registry vs creating independent ones.

internal/pebble/metrics.go
aaronbuchwald: This change was just intended to remove the multi gatherer usage, other than that it should be a no-op so I didn't add registration of any metrics that was previously ignored or change any namespaces.

api/indexer/indexer.go
```diff
@@ -45,11 +46,11 @@ func NewIndexer(path string, parser chain.Parser, blockWindow uint64) (*Indexer,
 	if blockWindow > maxBlockWindow {
 		return nil, fmt.Errorf("block window %d exceeds maximum %d", blockWindow, maxBlockWindow)
 	}
-	txDB, _, err := pebble.New(filepath.Join(path, "tx"), pebble.NewDefaultConfig())
```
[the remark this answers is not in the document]
aaronbuchwald: This was intended as a no-op since the metrics were ignored before

internal/pebble/metrics.go
aaronbuchwald: ?

vm/vm.go
aaronbuchwald: I think that we should differentiate db metrics for distinct components ie. these are two different db instances with very different workloads, so conflating them will produce much less useful data.

PR #1790, opened 2024-11-15
Ensure Blocks Don't Overpack With Txs

samliok:
The maximum block size that can be sent over the network is `2MiB`.

This PR adds a check to the block builder ensuring the total size of all txs included in a block is less than `1.5MiB`. This extra `.5MiB` of wiggle room allows enough space for the other fields in the blocks to be safely included without exceeding the `2MiB` limit.

### DIFF
--- a/chain/builder.go
+++ b/chain/builder.go
@@ -181,10 +181,11 @@ func (c *Builder) BuildBlock(ctx context.Context, parentView state.View, parent 
 		e := executor.New(streamBatch, c.config.TransactionExecutionCores, MaxKeyDependencies, c.metrics.executorBuildRecorder)
 		pending := make(map[ids.ID]*Transaction, streamBatch)
 		var pendingLock sync.Mutex
-		totalTxSize := 0
+		totalTxsSize := 0
 		for li, ltx := range txs {
 			txsAttempted++
-			if totalTxSize += ltx.Size(); totalTxSize > c.config.TargetTxsSize {
+			totalTxsSize += ltx.Size()
+			if totalTxsSize > c.config.TargetTxsSize {
 				c.log.Debug("Transactions in block exceeded allotted limit ", zap.Int("size", ltx.Size()))
 				restorable = append(restorable, txs[li:]...)
 				break
--- a/chain/builder.go
+++ b/chain/builder.go
@@ -19,6 +19,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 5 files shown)


chain/builder.go
```diff
@@ -181,8 +182,15 @@ func (c *Builder) BuildBlock(ctx context.Context, parentView state.View, parent
 		e := executor.New(streamBatch, c.config.TransactionExecutionCores, MaxKeyDependencies, c.metrics.executorBuildRecorder)
 		pending := make(map[ids.ID]*Transaction, streamBatch)
 		var pendingLock sync.Mutex
+		totalTxSize := 0
 		for li, ltx := range txs {
 			txsAttempted++
+			if totalTxSize += ltx.Size(); totalTxSize > consts.MaxTotalTxSizePerBlock {
+				c.log.Debug("Transactions in block exceeded allotted limit ", zap.Int("size", ltx.Size()))
+				restorable = append(restorable, txs[li:]...)
+				break
```
aaronbuchwald, line 191: We should differentiate between a target during block building and an actual maximum size that is enforced in both block building and block execution/verification.

If this is the maximum size, we should add it into `Execute` as well.

chain/builder.go
```diff
@@ -181,8 +181,15 @@ func (c *Builder) BuildBlock(ctx context.Context, parentView state.View, parent
 		e := executor.New(streamBatch, c.config.TransactionExecutionCores, MaxKeyDependencies, c.metrics.executorBuildRecorder)
 		pending := make(map[ids.ID]*Transaction, streamBatch)
 		var pendingLock sync.Mutex
+		totalTxSize := 0
```
aaronbuchwald, line 184: nit: `totalTxsSize`

chain/builder.go
```diff
@@ -181,8 +181,15 @@ func (c *Builder) BuildBlock(ctx context.Context, parentView state.View, parent
...
 		for li, ltx := range txs {
 			txsAttempted++
+			if totalTxSize += ltx.Size(); totalTxSize > c.config.TargetTxsSize {
+				c.log.Debug("Transactions in block exceeded allotted limit ", zap.Int("size", ltx.Size()))
+				restorable = append(restorable, txs[li:]...)
+				break
+			}
```
aaronbuchwald, line 191: nit: can we move incrementing `totalTxsSize` out of the if condition?

aaronbuchwald (changes requested):

aaronbuchwald (approved):

PR #1791, opened 2024-11-17
Fix two builder regressions

tsachiherman:
## What ?

This PR addresses two regressions introduced in PR 1767.

1. The `Manual` builder specified in the options wouldn't work. This is because the `applyOptions` is implemented as 
```golang
func (vm *VM) applyOptions(o *Options) {
	vm.blockSubscriptionFactories = o.blockSubscriptionFactories
	vm.vmAPIHandlerFactories = o.vmAPIHandlerFactories
	if o.builder {
		vm.builder = builder.NewManual(vm.toEngine, vm.snowCtx.Log)
	}
	if o.gossiper {
		vm.gossiper = gossiper.NewManual(vm)
	}
}
```
which relies on `vm.toEngine` to be initialized. But it wasn't ( by the time `applyOptions` was called ).
Fix : move the initialization of the `vm.toEngine` to the beginning of the `Initialize` function.

2. The builder `builder.NewTime` always override the manual one. This happened due to moving the `builder.NewTime` call from line 264 to line 335 so that the mempool would be initialized.
As a side observation, the creation of a Manual or NewTime is expected to be mutually exclusive. The previous code didn't really taken that into account.
Fix : Let the `applyOptions` create the builder as needed. If no builder was created, a new one would be created.

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -324,7 +324,7 @@ func (vm *VM) Initialize(
 	}
 	err = vm.applyOptions(options)
 	if err != nil {
-		return fmt.Errorf("failed to apply options : %v", err)
+		return fmt.Errorf("failed to apply options : %w", err)
 	}
 
 	vm.chainTimeValidityWindow = chain.NewTimeValidityWindow(vm.snowCtx.Log, vm.tracer, vm)
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -192,6 +192,8 @@ func (vm *VM) Initialize(
 ) error {
 	vm.DataDir = filepath.Join(snowCtx.ChainDataDir, vmDataDir)
 	vm.snowCtx = snowCtx
+	// Init channels before initializing other structs
+	vm.toEngine = toEngine
 	vm.pkBytes = bls.PublicKeyToCompressedBytes(vm.snowCtx.PublicKey)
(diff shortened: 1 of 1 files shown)


vm/vm.go
```diff
@@ -331,13 +330,16 @@ func (vm *VM) Initialize(
...
+	// if we haven't had the builder created by the options, create it now.
+	if vm.builder == nil {
+		vm.builder = builder.NewTime(toEngine, snowCtx.Log, vm.mempool, func(t int64) (int64, int64, error) {
+			blk, err := vm.GetStatefulBlock(context.TODO(), vm.preferred)
+			if err != nil {
+				return 0, 0, err
+			}
+			return blk.Tmstmp, vm.ruleFactory.GetRules(t).GetMinBlockGap(), nil
+		})
+	}
```
aaronbuchwald, line 342: Why this check rather than moving builder initialization before applying options?

Alternatively, could we make this more obviously mutually exclusive by constructing these components inside of `applyOptions` regardless of if it is going to create `NewTime` or `NewManual` ?

tsachiherman: I think both are good ideas. I'll move the `applyOptions` past the initialization of the mempool, have the initialization of the builder done exclusively within the `applyOptions` and replicate the same of the `gossipier`

PR #1796, opened 2024-11-18
Throughput Testing Script

samliok:
### Throughput Testing

Ported over the `vryx-poc` throughput script to use the current sdk. 

Main changes
- Updating the CLI api to accept required parameters for throughput script.
	- `clusterInfo`, `key`, `validityWindow`
	- Parsing helpers for `clusterInfo.yaml` files
- Updated `spam` to properly display `issued/s` metric
- `deploy.devnet.sh` kicks off the throughput testing on AWS


tests/fixture/environment.go
samliok, line 1: removed the hardcoded port because it was giving me issues with e2e tests, will revert back once ready to merge

examples/morpheusvm/scripts/build.sh
samliok, line 1: seperate build script into hypersdk build and cli

examples/morpheusvm/cmd/morpheus-cli/cmd/spam.go
```diff
@@ -29,6 +34,43 @@ var runSpamCmd = &cobra.Command{
...
-		return handler.Root().Spam(ctx, &throughput.SpamHelper{KeyType: args[0]}, spamDefaults)
+
+		if len(clusterInfo) > 0 {
+			_, urisFromFile, err := cli.ReadCLIFile(clusterInfo)
+			if err != nil {
+				utils.Outf("{{red}} failed to read cluster info: %s \n", err)
+				return err
+			}
+			uris := onlyAPIs(urisFromFile)
+			key := "323b1d8f4eed5f0da9da93071b034f2dce9d2d22692c172f3cb252a64ddfafd01b057de320297c29ad0c1f589ea216869cf1938d88c9fbd70d6748323dbf2fa7"
```
samliok, line 45: pass in the key through the cli

examples/morpheusvm/scripts/deploy.devnet.sh
```diff
@@ -0,0 +1,214 @@
...
+if ! aws sts get-caller-identity >/dev/null 2>&1 ; then
+    echo -e "${RED}aws credentials not set. exiting...${NC}"
+    exit 1
+fi
+
+# Set AvalancheGo Build (should have canPop disabled)
+AVALANCHEGO_COMMIT=f03745d187d0c33b927121d4c8da977789b929ac
+
+# Create temporary directory for the deployment
+TMPDIR=/tmp/morpheusvm-deploy
```
samliok, line 43: don't hardcode `morpheus` naming

throughput/spam.go
```diff
@@ -435,3 +436,10 @@ func (s *Spammer) returnFunds(ctx context.Context, cli *jsonrpc.JSONRPCClient, p
 	)
 	return nil
 }
+
+func sleep(it *time.Timer, start time.Time) {
+	// Determine how long to sleep
+	dur := time.Since(start)
+	sleep := max(float64(consts.MillisecondsPerSecond-dur.Milliseconds()), 0)
+	it.Reset(time.Duration(sleep) * time.Millisecond)
+}
```
aaronbuchwald, line 445:
```suggestion
// sleep updates the timer to tick immediately if >= 1s has elapsed
// and otherwise tick after 1s has elapsed since start
func sleep(it *time.Timer, start time.Time) {
	dur := time.Since(start)
	sleep := max(float64(consts.MillisecondsPerSecond-dur.Milliseconds()), 0)
	it.Reset(time.Duration(sleep) * time.Millisecond)
}
```

throughput/tracker.go
```diff
@@ -83,6 +83,10 @@ func (t *tracker) logState(ctx context.Context, cli *jsonrpc.JSONRPCClient) {
 	}()
 }
 
+func (t *tracker) IncrementSent() int64 {
+	return t.sent.Add(1)
+}
+
 func (t *tracker) uniqueBytes() []byte {
-	return binary.BigEndian.AppendUint64(nil, uint64(t.sent.Add(1)))
+	return binary.BigEndian.AppendUint64(nil, uint64(t.IncrementSent()))
```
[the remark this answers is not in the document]
samliok: we don't really need this anymore so just removed this all-together

throughput/issuer.go
```diff
@@ -90,8 +90,8 @@ func (i *issuer) Send(ctx context.Context, actions []chain.Action, factory chain
 
 	// Register transaction and recover upon failure
 	if err := i.ws.RegisterTx(tx); err != nil {
-		i.l.Lock()
 		if i.ws.Closed() {
+			i.l.Lock()
```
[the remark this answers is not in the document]
samliok: no dont think i hit this edge case. But if there was an error and the websocket was open, then we wouldn't unlock which would cause a deadlock

throughput/spam.go
```diff
@@ -213,28 +214,30 @@ func (s Spammer) broadcast(
...
-					if balance < feePerTx {
-						return fmt.Errorf("insufficient funds (have=%d need=%d)", balance, feePerTx)
-					}
+					// balance, err := sh.LookupBalance(sender)
+					// if err != nil {
+					// 	return err
+					// }
+					// if balance < feePerTx {
+					// 	return fmt.Errorf("insufficient funds (have=%d need=%d)", balance, feePerTx)
+					// }
```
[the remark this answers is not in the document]
samliok: i think we just run until out of funds, or until the script terminates. We can't know if the action will use funds so any tracking we do would just be an estimate.

examples/morpheusvm/tests/workload/genesis.go
```diff
@@ -22,7 +22,7 @@ import (
 
 const (
 	// default initial balance for each address
-	InitialBalance uint64 = 10_000_000_000_000
+	InitialBalance uint64 = 3_000_000_000_000_000_000
```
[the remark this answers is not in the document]
samliok: yep it would run out of funds when ran with a large number of accounts and large tps.

examples/morpheusvm/scripts/build.sh
```diff
@@ -19,4 +19,20 @@ MORPHEUSVM_PATH=$(
...
+elif [[ $# -eq 0 ]]; then
+    # Set default binary directory location
+    name="pkEmJQuTUic3dxzg8EYnktwn4W7uCHofNcwiYo458vodAUbY7"
+    BINARY_PATH=$MORPHEUSVM_PATH/build/$name
+else
+    echo "Invalid arguments to build morpheusvm. Requires zero (default location) or one argument to specify binary location."
+    exit 1
+fi
+
+build_project "$MORPHEUSVM_PATH" "morpheusvm" "$BINARY_PATH"
```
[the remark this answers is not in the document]
samliok: build_project uses absolute paths, and the devnet script needs a relative path. Also I feel like build_project adds unnecessary complexity to the build process. I would be in favor of removing it all together.

examples/morpheusvm/scripts/deploy.devnet.sh
samliok: updated to put those parameters at the top of the file

tests/fixture/environment.go
```diff
@@ -45,15 +39,17 @@ func NewTestEnvironment(
 	)
 
 	chainID := testEnv.GetNetwork().GetSubnet(networkConfig.Name()).Chains[0].ChainID
-	setupDefaultChainAlias(testContext, chainID, networkConfig.Name())
+	setupDefaultChainAlias(testContext,
+		testEnv.GetNetwork().Nodes[0].URI,
+		chainID, networkConfig.Name())
```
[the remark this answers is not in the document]
samliok: removed these changes

examples/morpheusvm/scripts/deploy.devnet.sh
```diff
@@ -0,0 +1,218 @@
...
+# Set AvalancheGo Build (should have canPop disabled)
+AVALANCHEGO_COMMIT=f03745d187d0c33b927121d4c8da977789b929ac
+
+# Create temporary directory for the deployment
+TMPDIR=/tmp/$PROJECT_NAME-deploy
+rm -rf $TMPDIR && mkdir -p $TMPDIR
+echo -e "${YELLOW}set working directory:${NC} $TMPDIR"
+
+# Install avalanche-cli
+LOCAL_CLI_COMMIT=18e5259b499bb9b487bb5e1e620d3107e68f2171
```
aaronbuchwald, line 59: same comment here

throughput/config.go
```diff
@@ -36,6 +42,36 @@ func NewDefaultConfig(
 	}
 }
 
+// Default config for using the load test script.
```
aaronbuchwald, line 45: Should this just be moved to the caller since it's just creating an auth factory from the key and adding a check on the URIs? Seems the caller could run these steps and then use the existing `NewDefaultConfig`

throughput/spam.go
```diff
@@ -146,7 +147,7 @@ func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbo
 	s.tracker.logState(cctx, issuers[0].cli)
 
 	// broadcast transactions
-	err = s.broadcast(cctx, sh, accounts, factories, issuers, feePerTx, terminate)
+	err = s.broadcast(cctx, sh, factories, issuers, feePerTx, terminate)
```
aaronbuchwald, line 150: ❤️

throughput/config.go
samliok: yea, we could move `authFactory` to be created by the caller, but the defaults are different than `NewDefaultConfig`. And the default config is for e2e testing so we want it to terminate quickly.

samliok: updated to generate the factory in the caller

aaronbuchwald: Hmm if that's the case can we name them separately rather than calling both default? We could also separate the zipf distribution parameters from the throughput parameters as those are in fact default params.

Perhaps `NewFastConfig` and `NewLongRunningConfig` ?

aaronbuchwald (approved):

Issue #1792, opened 2024-11-17
Workload Test Framework Improvements


aaronbuchwald:
Feedback from @furkan-ux :

- Add examples of how to interact with URIs of the test network
- the test network interface is confusing - it's unclear from the interface how to extract the spending key, generate transactions, and get started
- ConfirmTxs is confusing because it implies that everything about the transaction execution will be checked for you

TODO:
- Add example of interacting with URIs to MorpheusVM transfer workload test
- Add transaction execution result to the return type of `ConfirmTxs` to `ConfirmTxs(context.Context, []*chain.Transaction) ([]*chain.Result, error)` and update example to put it on the caller to confirm the expected result

We should make it easier to grab the private key to use during a workload test rather than casting the network configuration and then grabbing a private key from it. One idea is to add `GetFundedKey() *auth.PrivateKey` to the `workload.TestNetwork` interface  that returns the pre-funded private key to use for the test. Tests can then distribute funds from this initial key as needed and won't need to cast the network configuration to their own type to extract the private key details. Instead, they can use the VM's `AuthProvider` to create an auth factory from the provided private key type.

Further improvements would be to add an argument to `registry.RegisterTest` to specify the minimum balance of the private key, so that the test runner can create isolated keys for each test and distribute the requested funds in the background.

najeal:
This will be done is different PR:

- Enhance PrivateKey access from the workload.TestNetwork: https://github.com/ava-labs/hypersdk/pull/1798
- Add Tx execution result to the return type of `ConfigrmTxs`
- Add example of interacting with URIs to MorpheusVM transfer workload test

PR #1798, opened 2024-11-18
workload test framework improvements: enhance privatekey access

najeal:
Relates to #1792 
This PR Enhance the FundedKey access from the workload tests.


examples/morpheusvm/tests/transfer.go
```diff
@@ -31,14 +30,13 @@ var _ = registry.Register(TestsRegistry, "Transfer Transaction", func(t ginkgo.F
 	require.NoError(err)
 	toAddress := auth.NewED25519Address(other.PublicKey())
 
-	networkConfig := tn.Configuration().(*workload.NetworkConfiguration)
-	spendingKey := networkConfig.Keys()[0]
-
+	authFactory, err := auth.GetFactory(tn.FundedKey())
```
najeal, line 33: The FundedKey can be access using `NetworkConfiguration.PrivateKeys()[0]`.
- On Network interface, I am wondering if it can make sense to create `FundedAuthFactory()` method instead of `FundedKey()`. This looks less "standard" but it prevents the user from having to generate the authFactory.
- Or another way to simplify his life can be to pass an optional (instead of a required) AuthFactory to GenerateTx(). Then the Network.GenerateTx is responsible to set a default AuthFactory using the first PrivateKey it gets from the NetworkConfiguration.

aaronbuchwald: yes we could actually completely replace the `*auth.PrivateKey` type with `AuthFactory` now that it includes an `Address()` function to return the relevant address.

The `PrivateKey` type was introduced to combine both.

examples/morpheusvm/tests/workload/genesis.go
```diff
@@ -68,15 +68,23 @@ func newDefaultKeys() []ed25519.PrivateKey {
...
 }
 
+func (n *NetworkConfiguration) PrivateKeys() []*auth.PrivateKey {
+	keys := make([]*auth.PrivateKey, 0, len(n.keys))
+	for _, key := range n.keys {
+		keys = append(keys, auth.NewPrivateKeyFromED25519(key))
+	}
+	return keys
+}
+
```
aaronbuchwald, line 87: Would it make sense to move the added keys into `workload.DefaultTestNetworkConfiguration` rather than keeping the wrapping `NetworkConfiguration` type?

auth/utils.go
```diff
@@ -22,6 +22,13 @@ type PrivateKey struct {
 	Bytes []byte
 }
 
+func NewPrivateKeyFromED25519(pk ed25519.PrivateKey) *PrivateKey {
```
aaronbuchwald, line 25: nit: should be in ed25519.go since it's specific to ed25519 - although I think we can get rid of this completely (per your comment here https://github.com/ava-labs/hypersdk/pull/1798/files#r1848745797)

examples/morpheusvm/tests/workload/genesis.go
najeal: It depends, the entiere tests of MorpheusVM are manipulating `ed25519` private key right now. Moving it into `workload.DefaultTestNetworkConfiguration` would require to either always store keys as `ed25519` or as generic type `auth.PrivateKey`.
At the first glance, I see no problem in either of the two scenarios.

aaronbuchwald: Yes, I think we can use `auth.PrivateKey` or as mentioned elsewhere `AuthFactory` since it includes `Address()` in the interface now.

examples/morpheusvm/tests/workload/genesis.go
```diff
@@ -68,31 +66,20 @@ func newDefaultKeys() []ed25519.PrivateKey {
...
-			consts.Name,
-			vm.NewParser(genesis)),
-		keys: keys,
-	}, nil
+	return workload.NewDefaultTestNetworkConfiguration(
+		genesisBytes,
+		consts.Name,
+		vm.NewParser(genesis),
+		keys,
+	), nil
```
aaronbuchwald, line 84: Why change from pointer to value here?

tests/integration/network.go
```diff
@@ -86,6 +88,10 @@ func (*Network) SynchronizeNetwork(ctx context.Context) error {
 	return nil
 }
 
+func (n *Network) FundedAuthFactory() chain.AuthFactory {
+	return n.Configuration().AuthFactories()[0]
+}
```
aaronbuchwald, line 93: Seems like it would make more sense to relay the exact return type rather than only support returning the value at index 0? Alternatively, we could just have tests access the auth factory directly through the config rather than adding a helper function to the test network here

tests/workload/network.go
```diff
@@ -31,6 +35,7 @@ type DefaultTestNetworkConfiguration struct {
 	genesisBytes []byte
 	name         string
 	parser       chain.Parser
+	keys         []ed25519.PrivateKey
```
aaronbuchwald, line 38: Can we use `AuthFactory` rather than `ed25519.PrivateKey` here ?

tests/workload/network.go
```diff
@@ -45,11 +50,24 @@ func (d DefaultTestNetworkConfiguration) Parser() chain.Parser {
 	return d.parser
 }
 
+func (d DefaultTestNetworkConfiguration) AuthFactories() []chain.AuthFactory {
+	authFactories := make([]chain.AuthFactory, 0, len(d.keys))
+	for _, key := range d.keys {
+		authFactories = append(authFactories, auth.NewED25519Factory(key))
+	}
+	return authFactories
+}
```
aaronbuchwald, line 59: Can we remove the need for this by just setting the field to be an `AuthFactory` from the start?

tests/workload/network.go
```diff
@@ -45,11 +50,24 @@ func (d DefaultTestNetworkConfiguration) Parser() chain.Parser {
...
+
+func (d DefaultTestNetworkConfiguration) Keys() []ed25519.PrivateKey {
+	return d.keys
+}
+
 // NewDefaultTestNetworkConfiguration creates a new DefaultTestNetworkConfiguration object.
-func NewDefaultTestNetworkConfiguration(genesisBytes []byte, name string, parser chain.Parser) DefaultTestNetworkConfiguration {
+func NewDefaultTestNetworkConfiguration(genesisBytes []byte, name string, parser chain.Parser, keys []ed25519.PrivateKey) DefaultTestNetworkConfiguration {
```
aaronbuchwald, line 66: same comment on using `AuthFactory` from the start

aaronbuchwald (changes requested):

examples/morpheusvm/tests/workload/genesis.go
najeal: It is because NetworkConfiguration embedded workload.DefaultTestNetworkConfiguration value. Do you want to switch to a pointer?

tests/integration/network.go
najeal: I removed the helper function 👍

aaronbuchwald (approved):

PR #1799, opened 2024-11-18
Add tx partitioner

aaronbuchwald:
This PR adds transaction partitioning based on the sponsor address into the DSMR package.

This will be used to partition transactions to be assigned to a specific node for chunk production/verification and targeted transaction gossip.

### DIFF
--- a/x/dsmr/partition.go
+++ b/x/dsmr/partition.go
@@ -125,55 +125,6 @@ func (pp *PrecalculatedPartition[T]) calculateIndexedTxs(txs []T) []*indexedTx[T
 	return indexedTxs
 }
 
-func (pp *PrecalculatedPartition[T]) AssignTxs(txs []T) (map[ids.NodeID][]T, error) {
-	indexedTxs := pp.calculateIndexedTxs(txs)
-
-	assignments := make(map[ids.NodeID][]T)
-	weightIndex := uint64(0)
-	txIndex := 0
-	for _, vdr := range pp.validators {
-		weightIndex += vdr.weight
-		for txIndex < len(indexedTxs) && weightIndex > indexedTxs[txIndex].index {
-			assignments[vdr.nodeID] = append(assignments[vdr.nodeID], indexedTxs[txIndex].tx)
-			txIndex++
-		}
-		if txIndex >= len(indexedTxs) {
-			break
-		}
-	}
-	return assignments, nil
-}
-
-func (pp *PrecalculatedPartition[T]) FilterTxs(nodeID ids.NodeID, txs []T) ([]T, error) {
-	// Return early if there are no transactions to filter
-	if len(txs) == 0 {
-		return nil, nil
-	}
-
-	// Calculate the weight index boundaries for the provided nodeID
-	weightIndex := uint64(0)
-	startWeight := uint64(0)
-	endWeight := uint64(0)
-	for _, vdr := range pp.validators {
-		if vdr.nodeID == nodeID {
-			startWeight = weightIndex
-			endWeight = weightIndex + vdr.weight
-			break
-		}
-		weightIndex += vdr.weight
-	}
-
-	filteredTxs := make([]T, 0, len(txs)/2)
-	for _, tx := range txs {
-		txIndexWeight := calculateSponsorWeightIndex(tx.GetSponsor(), pp.totalWeight)
-		if endWeight > txIndexWeight && startWeight <= txIndexWeight {
-			filteredTxs = append(filteredTxs, tx)
-		}
-	}
-
-	return filteredTxs, nil
-}
-
 type Partition[T Tx] struct {
 	state    validators.State
 	subnetID ids.ID
--- a/x/dsmr/partition.go
+++ b/x/dsmr/partition.go
@@ -1,0 +1,194 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package dsmr
+
+import (
+	"bytes"
+	"context"
+	"encoding/binary"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/snow/validators"
+	"github.com/ava-labs/avalanchego/utils"
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/consts"
+	"github.com/ava-labs/hypersdk/internal/cache"
+)
+
+const partitionCacheSize = 100
+
+type weightedValidator struct {
+	weight uint64
+	nodeID ids.NodeID
+}
+
+type PrecalculatedPartition[T Tx] struct {
+	validators  []*weightedValidator
+	totalWeight uint64
+}
+
+func (w *weightedValidator) Compare(o *weightedValidator) int {
+	return bytes.Compare(w.nodeID[:], o.nodeID[:])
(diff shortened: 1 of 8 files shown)


x/dsmr/partition.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	weightedValidators := make([]*weightedValidator, 0, len(vdrs))
+	totalWeight := uint64(0)
+	for _, vdr := range vdrs {
+		weightedValidators = append(weightedValidators, &weightedValidator{
+			weight: vdr.Weight,
+			nodeID: vdr.NodeID,
+		})
+		totalWeight += vdr.Weight
+	}
```
tsachiherman, line 49: would be marginally better to implement as : 
```golang
	weightedValidators := make([]*weightedValidator, len(vdrs))
	totalWeight := uint64(0)
	for i, vdr := range vdrs {
		weightedValidators[i] = &weightedValidator{
			weight: vdr.Weight,
			nodeID: vdr.NodeID,
		}
		totalWeight += vdr.Weight
	}
        ...
```

x/dsmr/partition.go
```diff
@@ -0,0 +1,195 @@
...
+		totalWeight: totalWeight,
+	}, nil
+}
+
+func calculateSponsorWeightIndex(sponsor codec.Address, totalWeight uint64) uint64 {
+	return binary.BigEndian.Uint64(sponsor[len(sponsor)-consts.Uint64Len:]) % totalWeight
+}
+
+func (pp *PrecalculatedPartition[T]) AssignTx(tx T) (ids.NodeID, bool) {
+	sponsor := tx.GetSponsor()
```
tsachiherman, line 63: add
```golang
if pp.totalWeight ==0 || len(pp.validators) == 0 {
    return ids.NodeID{}, false
}
```

x/dsmr/partition.go
```diff
@@ -0,0 +1,195 @@
...
+		validators:  weightedValidators,
+		totalWeight: totalWeight,
+	}, nil
+}
+
+func calculateSponsorWeightIndex(sponsor codec.Address, totalWeight uint64) uint64 {
+	return binary.BigEndian.Uint64(sponsor[len(sponsor)-consts.Uint64Len:]) % totalWeight
+}
+
+func (pp *PrecalculatedPartition[T]) AssignTx(tx T) (ids.NodeID, bool) {
```
tsachiherman, line 62: the implementation of this function is not very efficient.
instead, in `PrecalculatedPartition` we want to have an array called
``accumulatedTotalWeights` where each elements would contain the total accumulated weight until that validator.
Then, in AssignTx, we want to perform a binary search ( O(logn) ) instead of array scan ( O(n) ).

aaronbuchwald: Great idea

x/dsmr/partition.go
```diff
@@ -0,0 +1,217 @@
...
+	sponsor := tx.GetSponsor()
+	sponsorWeightIndex := calculateSponsorWeightIndex(sponsor, pp.totalWeight)
+	nodeIDIndex := sort.Search(len(pp.validators), func(i int) bool {
+		return pp.validators[i].accumulatedWeight > sponsorWeightIndex
+	})
+
+	// Defensive: this should never happen
+	if nodeIDIndex >= len(pp.validators) {
+		return ids.NodeID{}, false
+	}
```
joshua-kim, line 94: Should we just panic and break loudly as opposed to quietly continuing to operate in a state we shouldn't be in? Wondering if leaving this check will mask a problem in the future if this does end up breaking since this case shouldn't be possible anyways...

Also `sort.Search` only returns `len(slice)` if it's not found, so we should be checking for `==` instead of `>=`.

x/dsmr/partition.go
```diff
@@ -0,0 +1,217 @@
...
+		indexedTxs[i] = &indexedTx[T]{
+			index: sponsorWeightIndex,
+			tx:    tx,
+		}
+	}
+	utils.Sort(indexedTxs)
+	return indexedTxs
+}
+
+func (pp *PrecalculatedPartition[T]) AssignTxs(txs []T) (map[ids.NodeID][]T, error) {
```
joshua-kim, line 128: Why do we need both this and `AssignTx`? Can't we just call `AssignTx` in a loop if it wanted this behavior?

x/dsmr/partition.go
```diff
@@ -0,0 +1,217 @@
...
+			txIndex++
+		}
+		if txIndex >= len(indexedTxs) {
+			break
+		}
+	}
+	return assignments, nil
+}
+
+func (pp *PrecalculatedPartition[T]) FilterTxs(nodeID ids.NodeID, txs []T) ([]T, error) {
```
joshua-kim, line 147: Can't we also get rid of this function as well? Seems like filter could just be implemented by calling `AssignTx` in a loop over `txs` and just filtering out anything not returning `nodeID`.

x/dsmr/partition.go
aaronbuchwald: This is faster, but could be premature optimization

x/dsmr/partition.go
```diff
@@ -0,0 +1,217 @@
...
+
+	return precomputePartition[T](weightedValidators), nil
+}
+
+func calculateSponsorWeightIndex(sponsor codec.Address, totalWeight uint64) uint64 {
+	return binary.BigEndian.Uint64(sponsor[len(sponsor)-consts.Uint64Len:]) % totalWeight
+}
+
+func (pp *PrecalculatedPartition[T]) AssignTx(tx T) (ids.NodeID, bool) {
+	if pp.totalWeight == 0 || len(pp.validators) == 0 {
```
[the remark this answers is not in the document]
aaronbuchwald: In reality, we should never have either case. This is defensive, I don't feel too strongly and am happy to remove either or both.

x/dsmr/partition.go
```diff
@@ -0,0 +1,217 @@
...
+const partitionCacheSize = 100
+
+type weightedValidator struct {
+	weight uint64
+	nodeID ids.NodeID
+	// set accumulatedWeight up to index in precomputePartition
+	accumulatedWeight uint64
+}
+
+type PrecalculatedPartition[T Tx] struct {
```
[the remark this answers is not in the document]
aaronbuchwald: Removing for now to just define the partition function

x/dsmr/partition.go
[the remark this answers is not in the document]
aaronbuchwald: Removed both checks

x/dsmr/partition.go
aaronbuchwald: Removed

x/dsmr/partition.go
aaronbuchwald: Removed

x/dsmr/partition_test.go
```diff
@@ -0,0 +1,223 @@
...
+			tx := createTestPartitionTx(test.txWeight)
+			nodeID, ok := partition.AssignTx(tx)
+			r.True(ok)
+			foundNodeIDIndex := -1
+			for i, vdr := range partition.validators {
+				if vdr.nodeID == nodeID {
+					foundNodeIDIndex = i
+					break
+				}
+			}
```
joshua-kim, line 160: Can't we use `sort.Search` instead of this?

joshua-kim (approved):

PR #1800, opened 2024-11-19
Remove stateful block from dsmr and move execute into node

aaronbuchwald:
This PR removes the `StatefulBlock` type from the DSMR package since DSMR will use `vm/block.go` to implement the `snowman.Block` interface and should implement the same functionality as the chain package.

This moves the `Verify` function from `StatefulBlock` into the node. Alternatively, block building could be defined separately (https://github.com/ava-labs/hypersdk/pull/1744#discussion_r1834630726).

### DIFF
--- a/x/dsmr/node.go
+++ b/x/dsmr/node.go
@@ -207,7 +207,7 @@ func (n *Node[T]) BuildBlock(parent Block, timestamp int64) (Block, error) {
 	return blk, nil
 }
 
-func (*Node[T]) Execute(ctx context.Context, parent Block, block Block) error {
+func (*Node[T]) Execute(ctx context.Context, _ Block, block Block) error {
 	// TODO: Verify header fields
 	// TODO: de-duplicate chunk certificates (internal to block and across history)
 	for _, chunkCert := range block.ChunkCerts {
--- a/x/dsmr/node.go
+++ b/x/dsmr/node.go
@@ -171,7 +171,7 @@ func (n *Node[T]) NewChunk(
 	return chunk, n.storage.AddLocalChunkWithCert(chunk, &chunkCert)
 }
 
-func (n *Node[T]) NewBlock(parent Block, timestamp int64) (Block, error) {
+func (n *Node[T]) BuildBlock(parent Block, timestamp int64) (Block, error) {
 	if timestamp <= parent.Timestamp {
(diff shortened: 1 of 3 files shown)


x/dsmr/node.go
```diff
@@ -207,6 +207,19 @@ func (n *Node[T]) NewBlock(parent Block, timestamp int64) (Block, error) {
 	return blk, nil
 }
 
+func (n *Node[T]) Execute(ctx context.Context, parent Block, block Block) error {
+	// TODO: Verify header fields
+	// TODO: de-duplicate chunk certificates (internal to block and across history)
```
aaronbuchwald, line 212: Use general validity window to de-duplicate chunk certs

x/dsmr/node.go
```diff
@@ -207,6 +207,19 @@ func (n *Node[T]) NewBlock(parent Block, timestamp int64) (Block, error) {
 	return blk, nil
 }
 
+func (*Node[T]) Execute(ctx context.Context, _ Block, block Block) error {
```
joshua-kim, line 210: Isn't this supposed to be verify?

x/dsmr/node.go
```diff
@@ -171,7 +171,7 @@ func (n *Node[T]) NewChunk(
 	return chunk, n.storage.AddLocalChunkWithCert(chunk, &chunkCert)
 }
 
-func (n *Node[T]) NewBlock(parent Block, timestamp int64) (Block, error) {
+func (n *Node[T]) BuildBlock(parent Block, timestamp int64) (Block, error) {
```
joshua-kim, line 174: If we're going to name this `BuildBlock` we should also rename `NewChunk` to `BuildChunk` to be consistent

x/dsmr/node.go
aaronbuchwald: This is meant to be renamed to `Execute` corresponding to https://github.com/ava-labs/hypersdk/pull/1744

`Verify` comes from AvalancheGo's `snowman.Block` interface and includes additional checks. The intent of `Execute` is that we are simply executing the state transition and assume that all extra information is supplied directly to the function ie. parent block, state, etc.

x/dsmr/node.go
aaronbuchwald: Will do

joshua-kim (approved):

Issue #1771, opened 2024-11-13
Generalize expiry based de-duplication


aaronbuchwald:
The validity window is currently used to de-duplicate transactions for expiry based replay protection in the chain package: https://github.com/ava-labs/hypersdk/blob/main/chain/validity_window.go#L34.

This ticket is to create a general purpose version of it that can be re-used for chunk de-duplication and replay protection within the DSMR package.

This should replace the existing validity window: https://github.com/ava-labs/hypersdk/blob/main/chain/validity_window.go#L20 with an implementation that handles arbitrary containers ie. block of txs or chunks.

aaronbuchwald:
We should integrate de-duplication into chunk based block building and block execution. Similarly to tx replay protection, chunks that have already been included should not be included in a newly built block and when executing a chunk based block, any duplicate either within the block or across the ancestry of that block should cause that block to fail verification.

PR #1806, opened 2024-11-21
Rename emap interface method names

tsachiherman:
## What ?

As part of having the shared `validitywindow` implementation being used within the `dsmr`, the entries that would be contained would be better off implementing the same interface.

The dsmr's interface ( i.e. `Tx` ) already using `GetID()` and `GetExpiry()`, which seems to be "better" compared to the `emap.Item`'s  `ID()` and `Expiry()` as there would clash with member names.

This PR is a followup on https://github.com/ava-labs/hypersdk/pull/1782, and part of [Generalize expiry based de-duplication]( https://github.com/ava-labs/hypersdk/issues/1771)

### DIFF
--- a/api/indexer/indexer.go
+++ b/api/indexer/indexer.go
@@ -169,7 +169,7 @@ func (i *Indexer) storeTransactions(blk *chain.ExecutedBlock) error {
 		result := blk.Results[j]
 		if err := i.storeTransaction(
 			batch,
-			tx.ID(),
+			tx.GetID(),
 			blk.Block.Tmstmp,
 			result.Success,
 			result.Units,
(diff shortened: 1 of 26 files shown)


joshua-kim (dismissed):
One pro of this is that the interface function name is unlikely to collide w/ field names (e.g `ID` is a common field name).

aaronbuchwald:
We should be consistent throughout the codebase between using `ID()` and `GetID()` imo. We are bound to use `ID()` to implement `snowman.Block` interface from AvalancheGo, but could push everything else to switch to `GetID()`, which would be fine imo.

aaronbuchwald (approved):

PR #1807, opened 2024-11-21
Add VM accepted/rejected/verified notifications

aaronbuchwald:
This expands on the accepted block notification pattern and adds notifications internal to the VM for Accepted/Rejected/Verified.

This PR:
- updates the gossiper to use the verified block event notifications
- updates the mempool to use the accepted/rejected/verified block notifications
- updates the syncer/chain to use a single accepted block notification and simplifies this code by simply always calling `syncer.Accept` which will invoke the time validity window

Advantages of this pattern are that it is easily extensible and registers exact event dependencies right next to the constructor of each component.

This will also allow us to migrate from creating components within the VM to passing in components following the same option pattern we currently use to assemble your VM.

Downsides of this approach is that by moving event notifications to their receivers it becomes more difficult to see in a single place (block accept for example) every action the VM will take.


vm/vm.go
```diff
@@ -353,6 +379,20 @@ func (vm *VM) Initialize(
...
+			seenValidityWindow, err := vm.syncer.Accept(context.TODO(), b.ExecutionBlock)
+			if err != nil {
+				vm.Fatal("syncer failed to accept block", zap.Error(err))
+			}
+			if seenValidityWindow {
+				vm.seenValidityWindowOnce.Do(func() {
+					close(vm.seenValidityWindow)
+				})
+			}
+			return nil
```
aaronbuchwald, line 393: Rather than check if the `seenValidityWindow` channel is closed, always mark the block as accepted, which in turn pushes it to the time validity window and rely on `seenValidityWindowOnce` to make sure we only close the `seenValidityWindow` channel once.

internal/gossiper/manual.go
```diff
@@ -137,8 +137,6 @@ func (g *Manual[T]) HandleAppGossip(ctx context.Context, nodeID ids.NodeID, msg
 	return nil
 }
 
-func (*Manual[T]) BlockVerified(int64) {}
```
aaronbuchwald, line 140: The no-ops `BlockVerified` and `Queue` should not need to be part of a shared interface between `Manual` and `Proposer` if they register their own notification requirements from the VM rather than rely on the VM to call `BlockVerified` or `Queue` directly

aaronbuchwald:
Follow up from https://github.com/ava-labs/hypersdk/pull/1801#discussion_r1852433958

vm/resolutions.go
```diff
@@ -139,8 +139,12 @@ func (vm *VM) Verified(ctx context.Context, b *StatefulBlock) {
...
 	vm.verifiedL.Unlock()
 	vm.parsedBlocks.Evict(b.ID())
-	vm.mempool.Remove(ctx, b.StatelessBlock.Txs)
-	vm.gossiper.BlockVerified(b.Tmstmp)
+
+	for _, sub := range vm.verifiedSubscriptions {
+		if err := sub.Accept(b.executedBlock); err != nil {
+			vm.Fatal("subscription failed to process verified block", zap.Error(err))
+		}
+	}
```
StephenButtolph, line 147: We do this looping in a bunch of places, should we define a helper that looks like:
```golang
func NotifyAll[T any](e T, subs ...Subscription[T]) error {
	var errs []error
	for _, sub := range subs {
		if err := sub.Accept(e); err != nil {
			errs = append(errs, err)
		}
	}
	return errors.Join(errs)
}
```

vm/resolutions.go
```diff
@@ -255,50 +264,15 @@ func (vm *VM) Accepted(ctx context.Context, b *StatefulBlock) {
...
-			if err != nil {
-				vm.Fatal("syncer failed to accept block", zap.Error(err))
-			}
-			if seenValidityWindow {
-				vm.seenValidityWindowOnce.Do(func() {
-					close(vm.seenValidityWindow)
-				})
-			}
+	for _, sub := range vm.acceptedSubscriptions {
+		if err := sub.Accept(b); err != nil {
```
StephenButtolph, line 268: Should these events be taking in a context? It seems like we are inserting a number of `context.TODO()`s where we could just pass them in as expected.

StephenButtolph (dismissed):
I have a few comments that are probably unrelated to this PR:

---

```golang
// Subscription defines how to consume events
type Subscription[T any] interface {
	// Accept returns fatal errors
	Accept(t T) error
	// Close returns fatal errors
	Close() error
}
```

I feel like `Accept` in the context of a `VM` is pretty confusing (and even outside the context of a VM, having this be called `Accept` seems weird).

Could we consider changing this to `Notify` or `Receive` or `Update` or something? It looks really weird to be calling `Accept` inside of `Reject`.

---

The definition for `SubscriptionFunc` is a bit unusual. I think it would be more idiomatic to define it as:

```golang
type SubscriptionFunc[T any] func(t T) error

func (s SubscriptionFunc[T]) Accept(t T) error {
	return s(t)
}

func (SubscriptionFunc[_]) Close() error {
	return nil
}
```

---

Similarly, I think `SubscriptionFuncFactory` is a bit unusual. Would it make more sense to use:

```golang
type ConstantSubscriptionFactory[T any] struct {
	Subscription Subscription[T]
}

func (s ConstantSubscriptionFactory[T]) New() (Subscription[T], error) {
	return s.Subscription, nil
}
```
So that we don't need to re-define this wrapper for other (non-function) implementations?

---

`vm.Fatal` feels like an anti-pattern. Is there a reason these errors aren't recoverable? And if they aren't recoverable, why do we have to panic rather than notifying avalanchego of this fatal error.

---

All-in-all changes look correct to me. So, going to approve (feel free to take the suggestions or leave them 🤷).

vm/resolutions.go
aaronbuchwald: Updated

vm/resolutions.go
aaronbuchwald: Updated

PR #1810, opened 2024-11-22
Generalize expiry based de-duplication, dsmr

tsachiherman:
## What ?

Integrate the generic de-deduplication logic into the dsmr


x/dsmr/block.go
```diff
@@ -119,3 +127,28 @@ type Block struct {
...
+
+func (b Block) Height() uint64 {
+	return uint64(b.Hght)
+}
+
+func (b Block) Txs() []*ChunkCertificate {
+	return b.ChunkCerts
+}
+
+func (b Block) ContainsTx(id ids.ID) bool {
```
aaronbuchwald, line 147: Can we change this name since chunks are not transactions?

x/dsmr/node.go
```diff
@@ -80,25 +83,36 @@ func New[T Tx](
 			storage: storage,
 		},
 		storage: storage,
-	}, nil
+		log:     logging.NewLogger("dsmr"),
```
aaronbuchwald, line 86: let's pass the log in from the constructor instead of creating one here

x/dsmr/node.go
```diff
@@ -80,25 +83,36 @@ func New[T Tx](
...
+	}
+	node.tracer, err = trace.New(trace.Config{})
```
aaronbuchwald, line 88: same comment with tracer

x/dsmr/node.go
```diff
@@ -172,29 +186,40 @@ func (n *Node[T]) BuildChunk(
...
+		oldestAllowed = 0
+	}
+	dup, err := n.validityWindow.IsRepeat(ctx, parent, chunkCerts, oldestAllowed)
+	if err != nil {
+		return Block{}, err
+	}
+
+	if dup.Len() == len(chunkCerts) {
+		return Block{}, ErrNoAvailableChunkCerts
+	}
```
aaronbuchwald, line 209: I think we can get rid of this extra check for an uncommon case

x/dsmr/node.go
```diff
@@ -263,6 +293,12 @@ func (n *Node[T]) Accept(ctx context.Context, block Block) error {
...
 		}
 	}
+	// update the validity window with the accepted block.
+	n.validityWindow.Accept(block)
+
+	return n.storage.SetMin(block.Tmstmp, chunkIDs)
+}
 
-	return n.storage.SetMin(block.Timestamp, chunkIDs)
+func (n *Node[T]) GetExecutionBlock(ctx context.Context, blkID ids.ID) (validitywindow.ExecutionBlock[*ChunkCertificate], error) {
```
aaronbuchwald, line 302: we can defer this to the VM to maintain a chain index for us that we pass in - similar to how we handle this in the existing `chain/` package

x/dsmr/node.go
tsachiherman: yes, of course.

x/dsmr/node.go
tsachiherman: done.

x/dsmr/block.go
tsachiherman: done

x/dsmr/node.go
tsachiherman: np

x/dsmr/node.go
```diff
@@ -172,19 +193,28 @@ func (n *Node[T]) BuildChunk(
 	return chunk, n.storage.AddLocalChunkWithCert(chunk, &chunkCert)
 }
 
-func (n *Node[T]) BuildBlock(parent Block, timestamp int64) (Block, error) {
-	if timestamp <= parent.Timestamp {
+// BuildBlock(ctx context.Context, parentView state.View, parent *ExecutionBlock) (*ExecutionBlock, *ExecutedBlock, merkledb.View, error)
```
aaronbuchwald, line 196: can we remove this comment?

x/dsmr/node_test.go
```diff
@@ -29,6 +32,26 @@ var (
 	_ Verifier[tx] = (*failVerifier)(nil)
 )
 
+func newTestingLog() logging.Logger {
+	return logging.NewLogger("dsmr")
+}
```
aaronbuchwald, line 37: we can replacing this with `logging.NoLog{}`

x/dsmr/node_test.go
```diff
@@ -29,6 +32,26 @@ var (
...
+
+func newTestingTracer(t *testing.T) trace.Tracer {
+	tracer, err := trace.New(trace.Config{})
+	require.NoError(t, err)
+	return tracer
+}
```
aaronbuchwald, line 43: we can replace this with `trace.Noop`

x/dsmr/node_test.go
```diff
@@ -29,6 +32,26 @@ var (
...
+
+type testingChainIndex struct{}
+
+func (*testingChainIndex) GetExecutionBlock(context.Context, ids.ID) (validitywindow.ExecutionBlock[*ChunkCertificate], error) {
+	return nil, nil
+}
+
+func newChainIndexer() ChainIndex {
+	return &testingChainIndex{}
+}
```
aaronbuchwald: We should add unit tests for the newly added functionality here. What do you think of the following cases?

- skip duplicate chunks in build block
- return an error if all available chunks are duplicates in build block
- duplicate chunk within a block
- duplicate chunk that between the block we are executing and the last accepted block
- duplicate chunk inside of the accepted validity window
- propagate an error if we fail to fetch a block during de-duplication

tsachiherman:
All sounds like great ideas, thanks!

x/dsmr/node_test.go
tsachiherman: done

x/dsmr/node_test.go
tsachiherman: done

x/dsmr/node_test.go
tsachiherman: done : unit test for `skip duplicate chunks in build block`

tsachiherman: done - add cases added.

x/dsmr/node.go
```diff
@@ -208,9 +242,22 @@ func (n *Node[T]) BuildBlock(parent Block, timestamp int64) (Block, error) {
...
 }
 
-func (*Node[T]) Execute(ctx context.Context, _ Block, block Block) error {
+func (n *Node[T]) Execute(ctx context.Context, parentBlock Block, block Block) error {
 	// TODO: Verify header fields
-	// TODO: de-duplicate chunk certificates (internal to block and across history)
+	if block.Tmstmp <= parentBlock.Tmstmp && parentBlock.Hght > 0 {
+		return ErrTimestampNotMonotonicallyIncreasing
+	}
+
```
aaronbuchwald, line 250: Can we remove this check from this PR?

We should enforce timestamp >= parent timestamp (should allow them to be equal in case a block builder pushes the timestamp ahead of wall clock time for some nodes.

We should not allow the case that a malicious node builds a block with a timestamp that we consider valid less than 1s ahead of our wall clock time, but still ahead of our wall clock time, such that when we build a block on top of it, we fail because the current timestamp is ahead of our local timestamp.

We should update the check applied in `BuildBlock` imo.

We should also never execute the genesis block, so the check for `parentBlock.Hght > 0` should be removed.

x/dsmr/node_test.go
```diff
@@ -29,6 +32,29 @@ var (
...
+
+type testingChainIndex struct {
+	blocks map[ids.ID]validitywindow.ExecutionBlock[*ChunkCertificate]
+}
+
+func (ti *testingChainIndex) GetExecutionBlock(_ context.Context, id ids.ID) (validitywindow.ExecutionBlock[*ChunkCertificate], error) {
+	if blk, has := ti.blocks[id]; has {
+		return blk, nil
+	}
+	return nil, nil
```
aaronbuchwald, line 45: Fetching a non-existing block should return an error, can we return `database.ErrNotFound` if the block is not available?

x/dsmr/block.go
```diff
@@ -106,9 +114,9 @@ func ParseChunk[T Tx](chunkBytes []byte) (Chunk[T], error) {
 }
 
 type Block struct {
-	ParentID  ids.ID `serialize:"true"`
-	Height    uint64 `serialize:"true"`
-	Timestamp int64  `serialize:"true"`
+	ParentID ids.ID `serialize:"true"`
+	Hght     uint64 `serialize:"true"`
+	Tmstmp   int64  `serialize:"true"`
```
joshua-kim, line 119: I dislike us naming fields like this because of the interface signature we're trying to implement... is there a way we can get around this?

x/dsmr/block.go
```diff
@@ -119,3 +127,28 @@ type Block struct {
 func (b Block) GetID() ids.ID {
 	return b.blkID
 }
+
+func (b Block) Parent() ids.ID {
```
joshua-kim, line 131: I don't care which we pick, but we should be consistent on the naming of either `Foo()` or `GetFoo()`.

x/dsmr/block.go
```diff
@@ -119,3 +127,28 @@ type Block struct {
...
+	return b.Tmstmp
+}
+
+func (b Block) Height() uint64 {
+	return b.Hght
+}
+
+func (b Block) Txs() []*ChunkCertificate {
+	return b.ChunkCerts
+}
```
joshua-kim, line 145: This seems weird because these are not returning txs

x/dsmr/node.go
```diff
@@ -46,6 +55,10 @@ func New[T Tx](
 	getChunkSignatureClient *p2p.Client,
 	chunkCertificateGossipClient *p2p.Client,
 	validators []Validator,
+	log logging.Logger,
+	tracer trace.Tracer,
```
joshua-kim, line 59: nit: make these (logger + tracer) the first args in this fn

x/dsmr/node.go
```diff
@@ -14,22 +14,31 @@ import (
...
 	ErrEmptyChunk                          = errors.New("empty chunk")
 	ErrNoAvailableChunkCerts               = errors.New("no available chunk certs")
+	ErrAllChunkCertsDuplicate              = errors.New("all chunk certs are duplicated")
 	ErrTimestampNotMonotonicallyIncreasing = errors.New("block timestamp must be greater than parent timestamp")
 )
 
+type (
+	ChainIndex         = validitywindow.ChainIndex[*ChunkCertificate]
+	timeValidityWindow = *validitywindow.TimeValidityWindow[*ChunkCertificate]
+)
```
joshua-kim, line 40: Seems like we're doing this to avoid the caller depending on an internal package. I'm wondering if it even makes sense for `validitywindow` to be in internal at all... should this just be merged into `dsmr`?

x/dsmr/block.go
```diff
@@ -11,14 +11,14 @@ import (
...
+	"github.com/ava-labs/hypersdk/internal/emap"
 	"github.com/ava-labs/hypersdk/utils"
 )
 
 const InitialChunkSize = 250 * 1024
 
 type Tx interface {
-	GetID() ids.ID
-	GetExpiry() int64
+	emap.Item
```
joshua-kim, line 21: This exposes the `internal/emap` package into the caller. I think the previous wrapping pattern where we wrapped this interface w/ a type that implemented the emap interface actually looked cleaner.

internal/validitywindow/validitywindow.go
```diff
@@ -69,6 +70,15 @@ func (v *TimeValidityWindow[Container]) VerifyExpiryReplayProtection(
...
 	}
+	// make sure we have no repeats within the block itself.
+	blkTxsIDs := make(map[ids.ID]bool, len(blk.Txs()))
+	for _, tx := range blk.Txs() {
+		id := tx.GetID()
+		if _, has := blkTxsIDs[id]; has {
+			return fmt.Errorf("%w: duplicate in block", ErrDuplicateContainer)
+		}
+		blkTxsIDs[id] = true
+	}
```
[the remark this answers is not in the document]
tsachiherman: I don't see why testing this twice is an issue on it's own, and I *think* that I have a better solution here;
Testing it in NewExecutionBlock is not ideal from my perspective, as it should be construction step ( i.e. no errors ).
How do you feel about the following:
in ExecutionBlock, we would add the following function:
```golang
func (b *ExecutionBlock) ValidateDuplicateTransactions() error {
    if len(b.Txs) != b.txsSet.Len() {
        ErrDuplicateTx
    }
    return nil
```
than in validitywindow.go `VerifyExpiryReplayProtection`, we can just call this method:
```golang
   ...
   if err := blk.ValidateDuplicateTransactions(); err != nil {
      return err
   }
   ...
```
and remove the test from `NewExecutionBlock`

x/dsmr/block.go
```diff
@@ -119,3 +127,28 @@ type Block struct {
...
+}
+
+func (b Block) Contains(id ids.ID) bool {
+	for _, c := range b.ChunkCerts {
+		if c.ChunkID == id {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
tsachiherman: I'm perfectly fine adding it. will do.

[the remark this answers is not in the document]
tsachiherman: done.

x/dsmr/node.go
```diff
@@ -172,29 +195,40 @@ func (n *Node[T]) BuildChunk(
...
 			continue
 		}
-
 		availableChunkCerts = append(availableChunkCerts, chunkCert)
 	}
 	if len(availableChunkCerts) == 0 {
+		if dup.Len() == len(chunkCerts) && len(chunkCerts) > 0 {
+			return Block{}, ErrAllChunkCertsDuplicate
+		}
 		return Block{}, ErrNoAvailableChunkCerts
```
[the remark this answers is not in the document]
tsachiherman: I thought that you've explicitly asked for a separate error when all the certs are duplicate.. did I get it wrong ?

x/dsmr/block.go
tsachiherman: yes, I agree. let's discuss this as a group ?

x/dsmr/node.go
tsachiherman: @aaronbuchwald asked to remove this section completely.

x/dsmr/block.go
tsachiherman: once merged, we won't need wrapping interface anymore. . unless I'm missing something ?

x/dsmr/block.go
aaronbuchwald: Can we use the term containers?

x/dsmr/node.go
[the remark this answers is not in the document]
aaronbuchwald: Ah I meant a separate test case if we attempt to build an empty block: https://github.com/ava-labs/hypersdk/pull/1810/#discussion_r1857158884

x/dsmr/block.go
aaronbuchwald: I think @joshua-kim 's point is to have the internal type implemented as makes sense in this package and then wrap it with another type that converts between that structure and the required interface when we need to use it in the validity window or expiry map where we need a specific interface. Correct me if I'm wrong @joshua-kim

internal/validitywindow/dependencies.go
```diff
@@ -16,9 +16,9 @@ type ExecutionBlock[Container emap.Item] interface {
 	Timestamp() int64
 	Height() uint64
 	Txs() []Container
-	ContainsTx(ids.ID) bool
+	Contains(ids.ID) bool
 }
 
 type ChainIndex[Container emap.Item] interface {
-	GetExecutionBlock(ctx context.Context, blkID ids.ID) (ExecutionBlock[Container], error)
+	GetExecutionBlock(ctx context.Context, blkID ids.ID) (ExecutionBlock[Container], bool, error)
```
aaronbuchwald, line 23: Why are we adding the boolean here?

internal/validitywindow/validitywindow.go
```diff
@@ -57,8 +58,8 @@ func (v *TimeValidityWindow[Container]) VerifyExpiryReplayProtection(
 	if blk.Height() <= v.lastAcceptedBlockHeight {
 		return nil
 	}
-	parent, err := v.chainIndex.GetExecutionBlock(ctx, blk.Parent())
-	if err != nil {
+	parent, hasBlock, err := v.chainIndex.GetExecutionBlock(ctx, blk.Parent())
+	if err != nil || !hasBlock {
```
aaronbuchwald, line 62: I don't think we should return a boolean for `hasBlock` here, but if we did not have the block, we would not be able correctly verify the block. It's not safe to mark a block as passing verification if we are missing the information required to verify it, we must guarantee we don't reach a point like that in the first place.

x/dsmr/node_test.go
```diff
@@ -994,6 +1022,287 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+	r := require.New(t)
+
+	node := newTestNode(t)
+
+	blk := Block{
+		ParentID: ids.GenerateTestID(),
+		Hght:     1,
+		Tmstmp:   1,
+		blkID:    ids.GenerateTestID(),
+	}
```
joshua-kim, line 1036: We should be building this block through `node.BuildBlock` and calling `Verify` prior to calling `Accept` (see other tests as an example). This is technically not a valid block because it references an unknown parent instead of the last accepted block which would be caught if `Verify` was called on it. Building a block should generally take the form of calling `BuildChunk -> BuildBlock -> Verify -> Accept`

x/dsmr/node_test.go
```diff
@@ -994,6 +1022,287 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+		Hght:     2,
+		Tmstmp:   2,
+		blkID:    ids.GenerateTestID(),
+		ChunkCerts: []*ChunkCertificate{
+			&chunkCert,
+		},
+	}
+	r.NoError(node.Accept(context.Background(), blk))
+
+	r.NoError(node.storage.AddLocalChunkWithCert(chunk, &chunkCert))
```
joshua-kim, line 1063: We shouldn't have to call this, since the chunk cert is already added locally once the chunk is finished being built from `node.BuildChunk`, and remote chunks are added within `Accept` if we don't have them ([ref](https://github.com/ava-labs/hypersdk/blob/8e83b6d7e28b2fd8487e40a4cf0b01b5b2d9de5f/x/dsmr/node_test.go#L1307)). 

I also think this is a weird way of writing the test because we're manually injecting the chunk into the database instead of testing how this would happen through the actual user-flow of dsmr. Maybe a better way would be to test the cases where you are calling `BuildChunk` a second time w/ the same args to make a duplicate chunk, and trying to build + accept the block? We should also test the case where a peer sends us a block referencing a duplicate chunk that we try to verify and fail.

x/dsmr/node_test.go
```diff
@@ -994,6 +1022,287 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+				},
+				expiry,
+				codec.Address{},
+			)
+			r.NoError(err)
+			chunks = append(chunks, chunk)
+			chunksCerts = append(chunksCerts, &cert)
+		}
+		return chunks, chunksCerts
+	}
```
joshua-kim, line 1117: Can we just move this to the test body? It's only used once so I find that this makes the test more confusing to read.

x/dsmr/node_test.go
```diff
@@ -994,6 +1022,287 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+				),
+				validators,
+				genesisBlk,
+				1,
+				1,
+				indexer,
+				valWind,
+			)
+			r.NoError(err)
+
```
joshua-kim, line 1261: Can we use `newTestNode` here? I think it would simplify a lot of the setup

x/dsmr/node_test.go
```diff
@@ -38,6 +40,29 @@ var (
 	chainID = ids.Empty
 )
 
+const testingDefaultValidityWindowDuration = time.Duration(5)
```
joshua-kim, line 43: nit: can we move this to a `const` block at the top of the file?

x/dsmr/node_test.go
```diff
@@ -38,6 +40,29 @@ var (
...
+
+type testingChainIndex struct {
```
joshua-kim, line 45: nit: can we call this something like `testValidityWindow` + an interface check at the top of this file?

internal/validitywindow/dependencies.go
joshua-kim: My guess is that this is to return `true/false`, and use the `error` if something goes wrong instead of just returning the block type + an error, and you have to introspect the error using a check against `database.ErrNotFound`.

internal/validitywindow/syncer.go
```diff
@@ -39,10 +39,11 @@ func (s *Syncer[Container]) start(ctx context.Context, lastAcceptedBlock Executi
 		seenValidityWindow = false
 		validityWindow     = s.getValidityWindow(lastAcceptedBlock.Timestamp())
 		err                error
+		hasBlock           bool
 	)
 	for {
-		parent, err = s.chainIndex.GetExecutionBlock(ctx, parent.Parent())
-		if err != nil {
+		parent, hasBlock, err = s.chainIndex.GetExecutionBlock(ctx, parent.Parent())
```
joshua-kim, line 45: nit: `ok` is usually idiomatic in go for existence checks

x/dsmr/node.go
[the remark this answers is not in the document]
tsachiherman: oh.. yes, I can do that.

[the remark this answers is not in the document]
tsachiherman: done.

x/dsmr/block.go
tsachiherman: I'm going to dodge this one by saying that this method is no longer needed.

x/dsmr/block.go
tsachiherman: done

internal/validitywindow/dependencies.go
tsachiherman: @aaronbuchwald see @joshua-kim request above to change the function signature to `GetExecutionBlock() (ExecutionBlock[T], bool, error) `

internal/validitywindow/validitywindow.go
tsachiherman: you're correct. I'll be adding an error case here which would be returned in case the *hasBlock* is false ( since it shouldn't happen ).

x/dsmr/node_test.go
tsachiherman: sure.

internal/validitywindow/dependencies.go
aaronbuchwald: When is it valid to not have an execution block that the validity window requests?

We should always have a window of blocks that goes past the time validity window. I am not sure of any case where not having a block ie. returning false should not also be treated as an error. Unless there is such a case, I think we're better off just returning an error and not adding to the return type here.

x/dsmr/block.go
```diff
@@ -105,6 +106,41 @@ func ParseChunk[T Tx](chunkBytes []byte) (Chunk[T], error) {
...
+
+func (e ExecutionBlock) Contains(id ids.ID) bool {
+	return e.innerBlock.certSet.Contains(id)
+}
+
+func (e ExecutionBlock) Parent() ids.ID {
+	return e.innerBlock.ParentID
+}
+
+func (e ExecutionBlock) Txs() []*emapChunkCertificate {
```
aaronbuchwald, line 130: Can we rename this to a neutral name instead of txs ?

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,218 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+	for _, testCase := range tests {
+		t.Run(testCase.name, func(t *testing.T) {
+			validationWindow := testCase.validalidityWindowDuration
+			if validationWindow == 0 {
+				validationWindow = testingDefaultValidityWindowDuration
+			}
+
+			var node *Node[tx]
+			nodes, indexer := newNodes(t, 1)
+			node = nodes[0]
```
aaronbuchwald, line 1200: I'm not sure the validity window should be separate from the nodes, since each node maintains its own index

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,218 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+				Timestamp: testCase.timestamp,
+				blkID:     ids.GenerateTestID(),
+			}
+
+			for _, chunkExpiry := range testCase.chunks {
+				_, chunkCert, err := node.BuildChunk(
+					context.Background(),
+					[]tx{
+						{
+							ID:     ids.Empty, // ids.GenerateTestID(),
```
aaronbuchwald, line 1249: Is this meant to be using `ids.GenerateTestID()` ?

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,218 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+				}
+
+				r.NoError(node.Verify(context.Background(), parentBlk, blk))
+
+				r.NoError(node.Accept(context.Background(), blk))
+				indexer.set(blk.GetID(), NewExecutionBlock(blk))
+				parentBlk = blk
+			}
+
+			// create the block so that we can test it against the execute directly.
```
aaronbuchwald, line 1236:
```suggestion
			// create the block so that we can test it against Execute directly.
```

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,218 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+						context.Background(),
+						[]tx{
+							{
+								ID:     ids.Empty,
+								Expiry: int64(chunkExpiry),
+							},
+						},
+						int64(chunkExpiry),
+						codec.Address{},
+					)
```
aaronbuchwald, line 1224: Should we generate actual IDs instead of `ids.Empty` since these will technically be invalid?

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,218 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+					codec.Address{},
+				)
+				r.NoError(err)
+				newBlk.ChunkCerts = append(newBlk.ChunkCerts, &chunkCert)
+			}
+			_, err := node.BuildBlock(context.Background(), parentBlk, testCase.timestamp)
+			r.ErrorIs(err, testCase.buildWantErr)
+
+			r.ErrorIs(node.Verify(context.Background(), parentBlk, newBlk), testCase.verifyWantErr)
+		})
```
aaronbuchwald, line 1263: If `BuildBlock` succeeds, it seems that we are expecting it to build the same block that we have produced locally. Should we be checking that they're the same here?

x/dsmr/node_test.go
tsachiherman: it's not the validity window itself, but rather the validity window duration. The index remains the same.

x/dsmr/node_test.go
tsachiherman: no; but good catch. It needed to be a way to way to reproduce consistent duplicate chunks. I'd updated the code to use `ids.Empty.Prefix(uint64(chunkExpiry))`

x/dsmr/node_test.go
tsachiherman: if the `BuildBlock` won't return any of the expected errors, than it would return a valid block. However, I'm not sure how easy it would be to compare it to the testing block we've created. I'll add ParentID, Height and Timestamp.

x/dsmr/block.go
```diff
@@ -105,6 +106,47 @@ func ParseChunk[T Tx](chunkBytes []byte) (Chunk[T], error) {
 	return c, c.init()
 }
 
+// ExecutionBlock bridge the gap between the dsmr's block implementation and the validity window's execution block interface.
+type ExecutionBlock struct {
+	innerBlock Block
```
joshua-kim, line 111: I think it would be fine to just embed `Block` here since we're just wrapping the data to satisfy the interface impl

x/dsmr/block.go
```diff
@@ -105,6 +106,47 @@ func ParseChunk[T Tx](chunkBytes []byte) (Chunk[T], error) {
...
+	certSet    set.Set[ids.ID]
```
joshua-kim, line 112: nit: rename to `certs`

x/dsmr/block.go
```diff
@@ -105,6 +106,47 @@ func ParseChunk[T Tx](chunkBytes []byte) (Chunk[T], error) {
...
+func (e ExecutionBlock) Contains(id ids.ID) bool {
+	return e.certSet.Contains(id)
+}
+
+func (e ExecutionBlock) Parent() ids.ID {
+	return e.innerBlock.ParentID
+}
+
+func (e ExecutionBlock) Containers() []*emapChunkCertificate {
+	emapChunkCert := make([]*emapChunkCertificate, len(e.innerBlock.ChunkCerts))
```
joshua-kim, line 132: nit: rename to `chunkCerts`. Generally prefer that slices/maps/etc have plural names.

x/dsmr/block.go
```diff
@@ -105,6 +106,47 @@ func ParseChunk[T Tx](chunkBytes []byte) (Chunk[T], error) {
...
+	for i := range emapChunkCert {
+		emapChunkCert[i] = &emapChunkCertificate{*e.innerBlock.ChunkCerts[i]}
```
joshua-kim, line 134: nit: can we use the `append` syntax here?

x/dsmr/block.go
```diff
@@ -105,6 +106,47 @@ func ParseChunk[T Tx](chunkBytes []byte) (Chunk[T], error) {
...
+	}
+	return emapChunkCert
+}
+
+func NewExecutionBlock(innerBlock Block) ExecutionBlock {
```
joshua-kim, line 139: nit: unexport, this should never be used outside of the package because this is a type internal to this package

x/dsmr/block.go
```diff
@@ -116,6 +158,15 @@ type Block struct {
 	blkBytes []byte
 }
 
+func NewBlock(parentID ids.ID, height uint64, timestamp int64, chunkCerts []*ChunkCertificate) Block {
```
joshua-kim, line 161: nit: I don't think we need this diff?

x/dsmr/node.go
```diff
@@ -36,6 +38,7 @@ var (
 
 	ErrEmptyChunk                          = errors.New("empty chunk")
 	ErrNoAvailableChunkCerts               = errors.New("no available chunk certs")
+	ErrAllChunkCertsDuplicate              = errors.New("all chunk certs are duplicated")
```
joshua-kim, line 41: Why is there a specific error type to all chunk certs being duplicated? Shouldn't the verification fail if any chunk cert is a duplicate?

Also I don't think this error type is used

x/dsmr/node.go
```diff
@@ -217,31 +231,41 @@ func (n *Node[T]) BuildChunk(
...
 		return Block{}, ErrTimestampNotMonotonicallyIncreasing
 	}
 
 	chunkCerts := n.storage.GatherChunkCerts()
+	oldestAllowed := max(0, timestamp-int64(n.validityWindowDuration))
+
+	emapChunkCert := make([]*emapChunkCertificate, len(chunkCerts))
+	for i := range emapChunkCert {
+		emapChunkCert[i] = &emapChunkCertificate{*chunkCerts[i]}
+	}
```
joshua-kim, line 245: Similar comment to before of using `append` + renaming

x/dsmr/node.go
```diff
@@ -217,31 +231,41 @@ func (n *Node[T]) BuildChunk(
...
 
-func (n *Node[T]) BuildBlock(parent Block, timestamp int64) (Block, error) {
+func (n *Node[T]) BuildBlock(ctx context.Context, parent Block, timestamp int64) (Block, error) {
 	if timestamp <= parent.Timestamp {
 		return Block{}, ErrTimestampNotMonotonicallyIncreasing
 	}
 
 	chunkCerts := n.storage.GatherChunkCerts()
+	oldestAllowed := max(0, timestamp-int64(n.validityWindowDuration))
+
```
joshua-kim, line 241: nit: spacing is inconsistent, delete line

x/dsmr/node.go
```diff
@@ -217,31 +231,41 @@ func (n *Node[T]) BuildChunk(
...
+	emapChunkCert := make([]*emapChunkCertificate, len(chunkCerts))
+	for i := range emapChunkCert {
+		emapChunkCert[i] = &emapChunkCertificate{*chunkCerts[i]}
+	}
+	dup, err := n.validityWindow.IsRepeat(ctx, NewExecutionBlock(parent), emapChunkCert, oldestAllowed)
```
joshua-kim, line 246: nit: I had to inspect the implementation to realize this is a bitset... can we just call this `bitSet`, or `duplicates`? `dup` I assumed was a `bool` value at first...

x/dsmr/node.go
```diff
@@ -281,6 +305,13 @@ func (n *Node[T]) Verify(ctx context.Context, parent Block, block Block) error {
 		return fmt.Errorf("%w: %s", ErrEmptyBlock, block.GetID())
 	}
 
+	// Find repeats
+	oldestAllowed := max(0, block.Timestamp-int64(n.validityWindowDuration))
+
+	if err := n.validityWindow.VerifyExpiryReplayProtection(ctx, NewExecutionBlock(block), oldestAllowed); err != nil {
+		return err
```
joshua-kim, line 312: I think we should be wrapping this error to check for the duplicate case in tests

x/dsmr/node_test.go
```diff
@@ -1393,17 +1671,20 @@ func newNodes(t *testing.T, n int) []*Node[tx] {
...
+	indexer.set(blk.GetID(), ExecutionBlock{innerBlock: blk})
 
 	for _, n := range result[1:] {
 		require.NoError(t, n.Verify(context.Background(), n.LastAccepted, blk))
 		require.NoError(t, n.Accept(context.Background(), blk))
+		indexer.set(blk.GetID(), ExecutionBlock{innerBlock: blk})
 	}
 
-	return result
+	return result, indexer
```
joshua-kim, line 1689: Do we want this indexer to be shared across all node instances?

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,220 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+		Height:    2,
+		Timestamp: 2,
+		blkID:     ids.GenerateTestID(),
+		ChunkCerts: []*ChunkCertificate{
+			&chunkCert,
+		},
+	}
+	r.NoError(node.Accept(context.Background(), blk))
+
+	_, err = node.BuildBlock(context.Background(), blk, 3)
```
joshua-kim, line 1098: nit: Use `node.LastAccepted` instead of referencing `blk` for the parent to build off of

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,220 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+			{
+				ID:     ids.GenerateTestID(),
+				Expiry: 4,
+			},
+		},
+		4,
+		codec.Address{},
+	)
+	r.NoError(err)
+	r.NoError(node.Accept(context.Background(), blk))
```
joshua-kim, line 1114: I feel like instead of having a single test case that checks:
1. Build a duplicate chunk, have it fail, and build a new chunk and see if it passes

It would just make more sense to have this split into separate test cases:
1. Build a Chunk, it should pass
2. Build a duplicate chunk, it should fail
3. Build a chunk, build a duplicate chunk, should pass with a single chunk in the block

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,220 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+								ID:     ids.Empty.Prefix(uint64(chunkExpiry)),
+								Expiry: int64(chunkExpiry),
+							},
+						},
+						int64(chunkExpiry),
+						codec.Address{},
+					)
+					r.NoError(err)
+					blk.ChunkCerts = append(blk.ChunkCerts, &chunkCert)
+				}
```
joshua-kim, line 1224: Can we use `BuildBlock` after calling `BuildChunk` instead of initializing the block through the struct?

x/dsmr/node_test.go
```diff
@@ -994,11 +1057,220 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
+
+				r.NoError(node.Verify(context.Background(), parentBlk, blk))
+
+				r.NoError(node.Accept(context.Background(), blk))
+				indexer.set(blk.GetID(), NewExecutionBlock(blk))
```
joshua-kim, line 1229: Should dsmr be updating the chain index? Or are we expecting that someone else is responsible for updating it (current behavior)?

x/dsmr/node_test.go
tsachiherman: I *could* create an indexer per node, if you think it would be better.. but it would be a bit of over engineering solution given that it's being used to retrieve previously accepted blocks. Please let me know if you think it's the right move and I'll make it so.

x/dsmr/node_test.go
```diff
@@ -169,6 +197,38 @@ func TestNode_GetChunk_AvailableChunk(t *testing.T) {
 	<-done
 }
 
+func TestIndexerMissingBlock(t *testing.T) {
```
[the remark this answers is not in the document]
tsachiherman: as far as I can tell, this is the only test that we have that cover this case.
( obviously, it shouldn't happen on any real VM that index all it's accepted blocks )

x/dsmr/node_test.go
```diff
@@ -169,6 +197,38 @@ func TestNode_GetChunk_AvailableChunk(t *testing.T) {
...
+		[]tx{{ID: ids.GenerateTestID(), Expiry: 123}},
+		123,
+		codec.Address{123},
+	)
+	r.NoError(err)
+
+	blkNext, err := node.BuildBlock(context.Background(), node.LastAccepted, 4)
+	r.NoError(err)
+
+	r.ErrorIs(node.Verify(context.Background(), node.LastAccepted, blkNext), database.ErrNotFound)
```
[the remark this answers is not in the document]
tsachiherman: it wouldn't be the same transactions since `ids.GenerateTestID()` is random.

internal/validitywindow/dependencies.go
```diff
@@ -15,8 +15,8 @@ type ExecutionBlock[Container emap.Item] interface {
 	Parent() ids.ID
 	Timestamp() int64
 	Height() uint64
-	Txs() []Container
-	ContainsTx(ids.ID) bool
```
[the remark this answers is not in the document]
tsachiherman: requested by @aaronbuchwald

x/dsmr/block.go
tsachiherman: I don't mind to change it to "append" style, just be aware of https://stackoverflow.com/questions/38654729/golang-slice-append-vs-assign-performance

x/dsmr/block.go
tsachiherman: I know that you don't like intuitive constructors... I'll roll back this change. ( we use it in only a single place ).

x/dsmr/node.go
tsachiherman: I'm ok changing this, just please note that the same applied to other placed in the codebase that are already using the `dup` variable name for the same purpose.

aaronbuchwald (changes requested):
As discussed offline, could we update the validity window tests to treat the validity window as a black box rather than the current pattern that requires extra setup?

This way we can test at the following levels:
- add unit tests to time validity window
- blackbox test DSMR assuming the validity window is correctly maintained
- add integration tests when running DSMR with the snow VM

x/dsmr/block.go
```diff
@@ -105,6 +106,47 @@ func ParseChunk[T Tx](chunkBytes []byte) (Chunk[T], error) {
...
+}
+
+func (e validityWindowBlock) Contains(id ids.ID) bool {
+	return e.certs.Contains(id)
+}
+
+func (e validityWindowBlock) Parent() ids.ID {
+	return e.Block.ParentID
+}
+
```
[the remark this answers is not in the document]
tsachiherman: `Timestamp()`, `Height()` and `Parent()` aren't overrides.
The block contains *fields* named `Timestamp`, `Height` and `ParentID`.

The methods exports it so it would implement the interface
```golang
type ExecutionBlock[Container emap.Item] interface {
	Parent() ids.ID
	Timestamp() int64
	Height() uint64
	Containers() []Container
	Contains(ids.ID) bool
}
```

x/dsmr/node_test.go
```diff
@@ -961,11 +1000,57 @@ func TestNode_BuildBlock_IncludesChunks(t *testing.T) {
...
+		[]dsmrtest.Tx{
+			{
+				ID:     ids.GenerateTestID(),
+				Expiry: 4,
+			},
+		},
+		4,
+		codec.Address{},
+	))
+	_, err = node.Accept(context.Background(), blk)
```
aaronbuchwald, line 1042: We should never call `Accept` on the same block twice

x/dsmr/node_test.go
```diff
@@ -961,11 +1000,57 @@ func TestNode_BuildBlock_IncludesChunks(t *testing.T) {
...
+				Expiry: 4,
+			},
+		},
+		4,
+		codec.Address{},
+	))
+
+	blk, err := node.BuildBlock(context.Background(), node.LastAccepted, 2)
+	r.NoError(err)
+	r.NoError(node.Verify(context.Background(), node.LastAccepted, blk))
```
aaronbuchwald, line 1023: I'm not sure this is testing what it's intended to. This should test that the validity window correctly eliminates chunks that are in the chunk pool from block building. However, we build, verify, and accept the block, which will remove the chunk from the chunk pool. Therefore, no chunks are actually "eliminated" by the validity window in this test.

aaronbuchwald (changes requested):
As discussed offline, can we re-write the tests with the following assumptions:

- validity window is a separate component from DSMR in practice and will be responsible for keeping itself up to date
- therefore, it should be tested independently as unit tests for the validity window and within integration tests when used in conjunction with DSMR/chain
- we should add test cases for DSMR package that treat the validity window as a blackbox ie. return a mocked error from VerifyExpiryReplayProtection and IsRepeat and a mocked value of duplicates from IsRepeat that indicates one of the chunks is a duplicate and should be eliminated
- DSMR tests should not be responsible for setting the execution block index (treated as a separate component)

x/dsmr/node_test.go
tsachiherman: good point. This Accept was not needed, so I've just dropped it.

x/dsmr/node_test.go
tsachiherman: I added coverage for that in `TestNode_BuildBlock_IncludesChunks`

tsachiherman:
> * DSMR tests should not be responsible for setting the execution block index (treated as a separate component)

Definitely doable, but would require passing in the underlying `TimeValidityWindow`, so that we could mock the success case. Than, no indexer would be needed.

x/dsmr/node_test.go
```diff
@@ -745,10 +753,11 @@ func TestNode_BuildBlock_IncludesChunks(t *testing.T) {
...
 	tests := []struct {
-		name      string
-		chunks    func(parent Block) []chunk
-		timestamp func(parent Block) int64
-		wantErr   error
+		name               string
+		chunks             func(parent Block) []chunk
+		timestamp          func(parent Block) int64
+		wantErr            error
+		timevaliditywindow TimeValidityWindow[*emapChunkCertificate]
```
aaronbuchwald, line 760: can we use camel case?

internal/validitywindow/validitywindow_test.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidityWindowVerifyExpiryReplayProtection(t *testing.T) {
+	tests := []struct {
+		Name          string
+		Accepted      []executionBlock
+		VerifyBlock   executionBlock
+		OldestAllowed int64
+		ExpectedError error
```
aaronbuchwald, line 25: Can we unexport the field names here?

internal/validitywindow/validitywindow_test.go
```diff
@@ -0,0 +1,292 @@
...
+func TestValidityWindowIsRepeat(t *testing.T) {
+	tests := []struct {
+		Name           string
+		Blocks         []executionBlock
+		Accepted       uint64 // index into Blocks of the last accepted block.
+		ParentBlock    executionBlock
+		Containers     []container
+		OldestAllowewd int64
+		ExpectedError  error
+		ExpectedBits   set.Bits
```
aaronbuchwald, line 83: can we unexport these fields?

internal/validitywindow/validitywindow_test.go
```diff
@@ -0,0 +1,292 @@
...
+	return e.Tmstmp
+}
+
+func (e executionBlock) GetHeight() uint64 {
+	return e.Hght
+}
+
+func (e executionBlock) Containers() []container {
+	return e.Ctrs
+}
```
aaronbuchwald, line 264: If we are going to use `GetXXX`, for the rest of the interface required by validity window, could we do the same with `GetContainers()`?

internal/validitywindow/validitywindow_test.go
aaronbuchwald: Can we add an `Accepted` field to indicate the last block considered accepted as opposed to in processing as we do for `TestValidityWindowIsRepeat` below?

x/dsmr/node_test.go
```diff
@@ -1378,3 +1512,38 @@ func newTestNodes(t *testing.T, n int) []*Node[dsmrtest.Tx] {
...
+	OnIsRepeat func(ctx context.Context,
+		parentBlk validitywindow.ExecutionBlock[*emapChunkCertificate],
+		txs []*emapChunkCertificate,
+		oldestAllowed int64,
+	) (set.Bits, error)
+
+	OnVerifyExpiryReplayProtection func(context.Context, validitywindow.ExecutionBlock[*emapChunkCertificate], int64) error
+
+	defaultIsRepeatErr                     error
+	defaultVerifyExpiryReplayProtectionErr error
```
aaronbuchwald, line 1526: Can we just use the functions and make this a no-op that returns a nil error if they are not populated?

x/dsmr/node_test.go
```diff
@@ -1378,3 +1512,38 @@ func newTestNodes(t *testing.T, n int) []*Node[dsmrtest.Tx] {
...
+	oldestAllowed int64,
+) (set.Bits, error) {
+	if v.OnIsRepeat != nil {
+		return v.OnIsRepeat(ctx, parentBlk, txs, oldestAllowed)
+	}
+	return set.NewBits(), v.defaultIsRepeatErr
+}
+
+func (*testingTimeValidityWindowMock) Accept(validitywindow.ExecutionBlock[*emapChunkCertificate]) {
+}
```
aaronbuchwald, line 1549:
```suggestion
func (*testingTimeValidityWindowMock) Accept(validitywindow.ExecutionBlock[*emapChunkCertificate]) {}
```

x/dsmr/node_test.go
```diff
@@ -1378,3 +1512,38 @@ func newTestNodes(t *testing.T, n int) []*Node[dsmrtest.Tx] {
 
 	return result
 }
+
+type testingTimeValidityWindowMock struct {
```
aaronbuchwald, line 1516: Can we rename this `mockTimeValidityWindow` ?

x/dsmr/node_test.go
aaronbuchwald: The current flow here is build -> verify -> accept and then build another block and confirm that the chunks included in the already accepted block are correctly eliminated so `BuildBlock` returns the expected error.

At that point, the node is in a state of having an empty chunk pool and an updated last accepted block. I guess this is just testing that `BuildBlock` does not build an invalid block with the same chunk that was already included. Isn't that the same as the no available chunk certs in the test above?

aaronbuchwald (changes requested):

internal/validitywindow/validitywindow.go
```diff
@@ -43,32 +44,41 @@ func (v *TimeValidityWindow[Container]) Accept(blk ExecutionBlock[Container]) {
...
 	}
+	// make sure we have no repeats within the block itself.
+	blkContainerIDs := set.NewSet[ids.ID](len(blk.Containers()))
+	for _, container := range blk.Containers() {
+		id := container.GetID()
+		if blkContainerIDs.Contains(id) {
+			return fmt.Errorf("%w: duplicate in block", ErrDuplicateContainer)
+		}
+		blkContainerIDs.Add(id)
+	}
```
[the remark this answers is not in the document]
tsachiherman: You're correct that it's being performed twice : one in `validitywindow.VerifyExpiryReplayProtection` and one in `chain.NewExecutionBlock`. Yet, the functionality is missing in `dsmr.NewValidityWindowBlock`.

This functionality is logically needed for the correctness of the replay protection, although an early testing would work just as well.

my preference would be to keep it in validitywindow so that we can claim that the entire replay-protection is encapsulated there.

tsachiherman:
> The current flow here is build -> verify -> accept and then build another block and confirm that the chunks included in the already accepted block are correctly eliminated so BuildBlock returns the expected error.
> 
> At that point, the node is in a state of having an empty chunk pool and an updated last accepted block. I guess this is just testing that BuildBlock does not build an invalid block with the same chunk that was already included. Isn't that the same as the no available chunk certs in the test above?

No, it's not exactly the same. Please note of two things:
1. in both `node.BuildBlock` we called for a block with a timestamp of 3. The state of the node wasn't altered by the call since `BuildBlock` returned an error.
2. after adding a new chunk and calling `node.BuildBlock` again, we had no error being returned.

That allows us to test that the source for the error is the absence of chunks. ( note that all the chunks in this test have expiry of 4, far beyond the block time ).

internal/validitywindow/validitywindow_test.go
tsachiherman: not sure exactly how to do that, since we really need the included chunks to be set correctly, as it's being tested.

internal/validitywindow/validitywindow.go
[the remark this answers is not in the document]
aaronbuchwald: That makes sense, could we remove the check for this in `ExecutionBlock` in that case, so that we don't duplicate the check?

internal/validitywindow/validitywindow_test.go
```diff
@@ -0,0 +1,292 @@
...
+			},
+			verifyBlock:   newExecutionBlock(2, 3, 3, []int64{2}),
+			oldestAllowed: 2,
+			expectedError: nil,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			r := require.New(t)
+			r.NoError(nil)
```
aaronbuchwald, line 60:
```suggestion
```

internal/validitywindow/validitywindow_test.go
```diff
@@ -0,0 +1,292 @@
...
+				newContainer(3),
+			},
+			oldestAllowewd: 1,
+			expectedError:  database.ErrNotFound,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			r := require.New(t)
+			r.NoError(nil)
```
aaronbuchwald, line 180:
```suggestion
```

internal/validitywindow/validitywindow_test.go
aaronbuchwald: Why can't we do the exact same as we do below and just test `VerifyExpiryReplayProtection` instead of `IsRepeat` ?

aaronbuchwald:
```golang
			r := require.New(t)

			chainIndex := &testChainIndex{}
			tvw := NewTimeValidityWindow(&logging.NoLog{}, trace.Noop, chainIndex)
			r.NotNil(tvw)
			for i, blk := range test.blocks {
				if i <= int(test.accepted) {
					tvw.Accept(blk)
				}
				chainIndex.set(blk.GetID(), blk)
			}
```

tsachiherman: oh - now I understood what you meant. You want that the last entry in the blocks would be the one used for the `VerifyExpiryReplayProtection`

aaronbuchwald (approved):

PR #1811, opened 2024-11-22
Target metrics

aaronbuchwald:
This PR add target specific metrics into the gossiper and renames `g` to `t` as the pointer receiver for `*gossiper.Target`.

### DIFF
--- a/internal/gossiper/metrics.go
+++ b/internal/gossiper/metrics.go
@@ -8,28 +8,54 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+const namespace = "gossiper"
+
 type metrics struct {
-	txsReceived     prometheus.Counter
-	seenTxsReceived prometheus.Counter
-	txsGossiped     prometheus.Counter
+	txsReceived            prometheus.Counter
+	seenTxsReceived        prometheus.Counter
+	txsGossiped            prometheus.Counter
+	selectTxsToGossipCount prometheus.Counter
+	selectTxsToGossipSum   prometheus.Gauge
+	selectedTxsToGossip    prometheus.Counter
+	targetTxsSum           prometheus.Gauge
 }
 
 func newMetrics(r prometheus.Registerer) (*metrics, error) {
 	m := &metrics{
 		txsReceived: prometheus.NewCounter(prometheus.CounterOpts{
-			Namespace: "gossiper",
+			Namespace: namespace,
 			Name:      "txs_received",
 			Help:      "number of txs received over gossip",
 		}),
 		txsGossiped: prometheus.NewCounter(prometheus.CounterOpts{
-			Namespace: "gossiper",
+			Namespace: namespace,
 			Name:      "txs_gossiped",
 			Help:      "number of txs gossiped",
 		}),
 		seenTxsReceived: prometheus.NewCounter(prometheus.CounterOpts{
-			Namespace: "gossiper",
+			Namespace: namespace,
 			Name:      "seen_txs_received",
 			Help:      "number of txs received over gossip that were already seen",
+		}),
+		selectTxsToGossipCount: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "select_txs_to_gossip_count",
+			Help:      "number of times gossiper iterates mempool to select txs for gossip",
+		}),
+		selectTxsToGossipSum: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "select_txs_to_gossip_sum",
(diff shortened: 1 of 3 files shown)


internal/gossiper/metrics.go
```diff
@@ -8,35 +8,65 @@ import (
...
-	txsReceived     prometheus.Counter
-	seenTxsReceived prometheus.Counter
-	txsGossiped     prometheus.Counter
+	txsReceived            prometheus.Counter
+	seenTxsReceived        prometheus.Counter
+	txsGossiped            prometheus.Counter
+	selectTxsToGossipCount prometheus.Counter
+	selectTxsToGossipSum   prometheus.Gauge
+	selectedTxsToGossip    prometheus.Counter
+	targetTxsSum           prometheus.Gauge
```
najeal, line 20: Is there a specific reason to use a Gauge? I feel like only Inc/Add method are used to represent the metric.

internal/gossiper/target.go
```diff
@@ -252,111 +243,115 @@ func (g *Target[T]) HandleAppGossip(ctx context.Context, nodeID ids.NodeID, msg
...
 	defer span.End()
 
-	gossipContainers, err := g.targetStrategy.Target(ctx, txs)
+	start := time.Now()
+	gossipContainers, err := t.targetStrategy.Target(ctx, txs)
 	if err != nil {
 		return err
 	}
 
+	numTxsGossipped := 0
```
joshua-kim, line 341: nit: typo

internal/gossiper/metrics.go
```diff
@@ -8,35 +8,65 @@ import (
...
+			Help:      "number of times gossiper iterates mempool to select txs for gossip",
+		}),
+		selectTxsToGossipSum: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "select_txs_to_gossip_sum",
+			Help:      "sum of time spent iterating mempool to select txs for gossip",
+		}),
+		selectedTxsToGossip: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "selected_txs_to_gossip",
```
joshua-kim, line 52: should we call this `select_` to be consistent w/ the other metrics?

internal/gossiper/metrics.go
```diff
@@ -8,35 +8,65 @@ import (
...
 			Help:      "number of txs received over gossip that were already seen",
 		}),
+		selectTxsToGossipCount: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "select_txs_to_gossip_count",
+			Help:      "number of times gossiper iterates mempool to select txs for gossip",
+		}),
+		selectTxsToGossipSum: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "select_txs_to_gossip_sum",
```
joshua-kim, line 47: it's not obvious this + `target_txs_sum` are latency metrics from the naming - can we put `time` or something in the name to make it more clear?

internal/gossiper/target.go
```diff
@@ -161,68 +151,69 @@ func (g *Target[T]) Force(ctx context.Context) error {
...
 	if mempoolErr != nil {
 		return mempoolErr
 	}
 	if len(txs) == 0 {
-		g.log.Debug("no transactions to gossip")
+		t.log.Debug("no transactions to gossip")
 		return nil
 	}
-	g.log.Debug("gossiping transactions", zap.Int("txs", len(txs)), zap.Duration("t", time.Since(start)))
-	g.metrics.txsGossiped.Add(float64(len(txs)))
```
joshua-kim, line 196: Why did we remove this metric?

internal/gossiper/target.go
```diff
@@ -252,111 +243,115 @@ func (g *Target[T]) HandleAppGossip(ctx context.Context, nodeID ids.NodeID, msg
...
 			return err
 		}
 
 		// Send gossip to specified peers
-		if err := g.client.AppGossip(ctx, common.SendConfig{NodeIDs: gossipContainer.NodeIDs}, b); err != nil {
+		if err := t.client.AppGossip(ctx, common.SendConfig{NodeIDs: gossipContainer.NodeIDs}, b); err != nil {
 			return err
 		}
 	}
+	t.metrics.txsGossiped.Add(float64(numTxsGossipped))
```
joshua-kim, line 354: Is this right? This looks like it will re-accumulate previously counted txs into the metric since we're using `numTxsGossiped`. Shouldn't we be using something like `t.metrics.txsGossiped.Add(float64(len(gossipContainer.Txs))` here?

PR #1814, opened 2024-11-22
tests: init pre-funded local authFactories and URIs usage

najeal:
This PR is proposing a way to init pre-funded isolated authFactories.
It also adds an example of interacting with URIs to MorpheusVM transfer workload test.

### DIFF
--- a/examples/morpheusvm/tests/transfer.go
+++ b/examples/morpheusvm/tests/transfer.go
@@ -26,15 +26,16 @@ var TestsRegistry = &registry.Registry{}
 var _ = registry.Register(TestsRegistry, "Transfer Transaction",
 	func(t ginkgo.FullGinkgoTInterface, tn tworkload.TestNetwork, authFactories ...chain.AuthFactory) {
 		require := require.New(t)
+		ctx := context.Background()
 		targetFactory := authFactories[0]
+		authFactory := tn.Configuration().AuthFactories()[0]
 
 		client := jsonrpc.NewJSONRPCClient(tn.URIs()[0])
-		balance, err := client.GetBalance(context.Background(), targetFactory.Address())
+		balance, err := client.GetBalance(ctx, targetFactory.Address())
 		require.NoError(err)
 		require.Equal(uint64(1000), balance)
 
-		authFactory := tn.Configuration().AuthFactories()[0]
-		tx, err := tn.GenerateTx(context.Background(), []chain.Action{&actions.Transfer{
+		tx, err := tn.GenerateTx(ctx, []chain.Action{&actions.Transfer{
 			To:    targetFactory.Address(),
 			Value: 1,
 		}},
@@ -46,7 +47,7 @@ var _ = registry.Register(TestsRegistry, "Transfer Transaction",
 		defer timeoutCtxFnc()
 		require.NoError(tn.ConfirmTxs(timeoutCtx, []*chain.Transaction{tx}))
 
-		balance, err = client.GetBalance(context.Background(), targetFactory.Address())
+		balance, err = client.GetBalance(ctx, targetFactory.Address())
 		require.NoError(err)
 		require.Equal(uint64(1001), balance)
 	}, 1000)
--- a/examples/morpheusvm/tests/transfer.go
+++ b/examples/morpheusvm/tests/transfer.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	"github.com/ava-labs/hypersdk/api/jsonrpc"
 	"github.com/ava-labs/hypersdk/chain"
 	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
 	"github.com/ava-labs/hypersdk/tests/registry"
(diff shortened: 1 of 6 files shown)


examples/morpheusvm/tests/workload/genesis.go
```diff
@@ -72,7 +74,17 @@ func newDefaultAuthFactories() []chain.AuthFactory {
...
 func NewTestNetworkConfig(minBlockGap time.Duration) (workload.DefaultTestNetworkConfiguration, error) {
 	keys := newDefaultAuthFactories()
-	genesis := newGenesis(keys, minBlockGap)
+	testsLocalAllocations, err := tests.TestsRegistry.GenerateCustomAllocations(func() (chain.AuthFactory, error) {
+		privateKey, err := ed25519.GeneratePrivateKey()
+		if err != nil {
+			return nil, err
+		}
+		return auth.NewED25519Factory(privateKey), nil
+	})
```
najeal, line 83: The lambda function could potentially be moved in `auth` package.

tests/registry/registry.go
```diff
@@ -6,37 +6,63 @@ package registry
...
-func (r *Registry) List() []namedTest {
+func (r *Registry) List() []*namedTest {
 	if r == nil {
-		return []namedTest{}
+		return []*namedTest{}
 	}
 	return r.tests
 }
 
+func (r *Registry) GenerateCustomAllocations(generateAuthFactory func() (chain.AuthFactory, error)) ([]*genesis.CustomAllocation, error) {
```
najeal, line 37: The authFactories creation is delayed because when registering the tests, we don't know which private key type to use.

examples/morpheusvm/tests/workload/genesis.go
aaronbuchwald: Makes sense

examples/morpheusvm/tests/transfer.go
```diff
@@ -24,23 +23,31 @@ import (
...
-	)
-	require.NoError(err)
-
-	timeoutCtx, timeoutCtxFnc := context.WithDeadline(context.Background(), time.Now().Add(30*time.Second))
-	defer timeoutCtxFnc()
-
-	require.NoError(tn.ConfirmTxs(timeoutCtx, []*chain.Transaction{tx}))
-})
+var _ = registry.Register(TestsRegistry, "Transfer Transaction",
+	func(t ginkgo.FullGinkgoTInterface, tn tworkload.TestNetwork, authFactories ...chain.AuthFactory) {
```
aaronbuchwald, line 27: Can we replace `...chain.AuthFactory` with `[]chain.AuthFactory` ? I don't think we need the syntactic sugar here

examples/morpheusvm/tests/transfer.go
```diff
@@ -24,23 +23,31 @@ import (
...
+		require := require.New(t)
+		ctx := context.Background()
+		targetFactory := authFactories[0]
```
aaronbuchwald, line 30: Should we add a require check that the length of `authFactories` is 1 first?

examples/morpheusvm/tests/transfer.go
```diff
@@ -24,23 +23,31 @@ import (
...
+
+		client := jsonrpc.NewJSONRPCClient(tn.URIs()[0])
+		balance, err := client.GetBalance(ctx, targetFactory.Address())
+		require.NoError(err)
+		require.Equal(uint64(1000), balance)
+
+		tx, err := tn.GenerateTx(ctx, []chain.Action{&actions.Transfer{
+			To:    targetFactory.Address(),
+			Value: 1,
+		}},
```
aaronbuchwald, line 41: I think the flow should be `authFactories` are assumed to be pre-funded with a specific amount (1000) in this case, and the test flow should be:
1. confirm the provided balance is as expected
2. generate and confirm transaction _from_ the provided key (rather than the auth factory from the configuration)
3. confirm balance updates after confirming the tx

examples/morpheusvm/tests/workload/genesis.go
aaronbuchwald, line 83: This control flow is a bit confusing. This code is used to generate the genesis, but it's not clear from here where it's going to be used even though it's coupled to the tests registry.

We are using the test registry in the integration and e2e environments where we define two entry points. This is also where we invoke `NewTestNetworkConfig`.

Could we change this function to take in a slice of genesis allocations as an argument (in addition to `minBlockGap`)?

Then we can import the test registry and use the registered custom allocations in the entry point in both `examples/morpheusvm/tests/integration/integration_test.go` and examples/morpheusvm/tests/e2e/e2e_test.go`

aaronbuchwald (changes requested):

examples/morpheusvm/tests/transfer.go
najeal: I have changed the test to have a pre-funded authFactory sending to another pre-funded authFactory 👍 
I thought it was interesting to use _both_ `pre-funded authFactory` and `config authFactory`. So the users see the two possibilities.

PR #1818, opened 2024-11-25
Generate Warp Signatures

joshua-kim:
* Generate warp chunk signatures through acp 118
* Updates avalanchego dependency to use a patch version instead of master


proto/dsmr/dsmr.proto
```diff
@@ -16,14 +16,6 @@ message GetChunkResponse {
...
 }
 
-message GetChunkSignatureRequest {
-  bytes chunk = 1;
-}
-
-message GetChunkSignatureResponse {
-  bytes signature = 1;
-}
-
```
joshua-kim, line 26: I don't know why these were added to begin with since we use acp118 signature requests

x/dsmr/certificate.go
```diff
@@ -66,8 +58,33 @@ func (c *ChunkCertificate) Bytes() []byte {
...
-func (c *ChunkCertificate) Verify(_ context.Context, _ interface{}) error {
-	return c.Signature.Verify()
+func (c *ChunkCertificate) Verify(
+	ctx context.Context,
+	networkID uint32,
+	chainID ids.ID,
+	pChainState validators.State,
+	pChainHeight uint64,
+	quorumNum uint64,
+	quorumDen uint64,
```
joshua-kim, line 69: I think Aaron mentioned we wanted to support arbitrary verification (i.e not just warp). Probably we'll consider supporting this later...

x/dsmr/storage.go
```diff
@@ -40,10 +41,9 @@ func (NoVerifier[T]) Verify(Chunk[T]) error {
 }
 
 type StoredChunkSignature[T Tx] struct {
-	Chunk          Chunk[T]
-	LocalSignature NoVerifyChunkSignatureShare // Decouple signature share / certificate types
```
joshua-kim, line 44: We removed this because if we want no-verify behavior we can achieve this by changing how we handle `Cert` through an interface

x/dsmr/node_test.go
```diff
@@ -1478,6 +1668,120 @@ func getSignerBitSet(t *testing.T, pChain validators.State, nodeIDs ...ids.NodeI
 	return signerBitSet
 }
 
+func Test_Execute(t *testing.T) {
```
joshua-kim, line 1671: Looks like we weren't previously testing `Execute`. In hindsight though - I'm not sure how we want to test failure cases for this because I don't think we should be failing `Execute` if `Accept` doesn't return an error (and we shouldn't be calling `Execute` on a block we haven't accepted).

x/dsmr/storage.go
```diff
@@ -152,33 +152,32 @@ func (s *chunkStorage[T]) SetChunkCert(chunkID ids.ID, cert *ChunkCertificate) e
...
 	if err := s.verifier.Verify(c); err != nil {
-		return NoVerifyChunkSignatureShare{}, err
+		return nil, err
 	}
 	if err := s.putVerifiedChunk(c, nil); err != nil {
-		return NoVerifyChunkSignatureShare{}, err
+		return nil, err
 	}
-	return NoVerifyChunkSignatureShare{}, nil
+	return nil, nil
```
aaronbuchwald, line 169: Should we be producing and returning a local signature here?

x/dsmr/storage.go
```diff
@@ -152,33 +152,32 @@ func (s *chunkStorage[T]) SetChunkCert(chunkID ids.ID, cert *ChunkCertificate) e
...
 		return err
 	}
 	s.chunkEMap.Add([]emapChunk[T]{{chunk: c}})
 
 	chunkCert := &StoredChunkSignature[T]{
-		Chunk:          c,
-		LocalSignature: NoVerifyChunkSignatureShare{}, // TODO: add signer to generate actual signature share
-		Cert:           cert,
+		Chunk: c,
+		Cert:  cert,
```
aaronbuchwald, line 180: We store only the chunk certificate and not a local signature here

x/dsmr/storage.go
```diff
@@ -40,10 +41,9 @@ func (NoVerifier[T]) Verify(Chunk[T]) error {
 }
 
 type StoredChunkSignature[T Tx] struct {
-	Chunk          Chunk[T]
-	LocalSignature NoVerifyChunkSignatureShare // Decouple signature share / certificate types
-	Cert           *ChunkCertificate
-	Available      bool
+	Chunk     Chunk[T]
+	Cert      *ChunkCertificate[T]
+	Available bool
```
aaronbuchwald, line 46: (unrelated to this PR) What's the intent of `Available` ? Why would we need this flag rather than just returning the chunk if it's available?

x/dsmr/storage_test.go
```diff
@@ -121,10 +122,10 @@ func TestStoreAndSaveValidChunk(t *testing.T) {
 	chunkCerts := storage.GatherChunkCerts()
 	require.Empty(chunkCerts)
 
-	chunkCert := &ChunkCertificate{
+	chunkCert := &ChunkCertificate[tx]{
```
aaronbuchwald, line 125: Should we modify these tests to account for when the signature should be present or would you prefer that's tested through the node tests?

x/dsmr/storage.go
```diff
@@ -231,11 +230,11 @@ func (s *chunkStorage[T]) SetMin(updatedMin int64, saveChunks []ids.ID) error {
 // GatherChunkCerts provides a slice of chunk certificates to build
 // a chunk based block.
 // TODO: switch from returning random chunk certs to ordered by expiry
-func (s *chunkStorage[T]) GatherChunkCerts() []*ChunkCertificate {
+func (s *chunkStorage[T]) GatherChunkCerts() []*ChunkCertificate[T] {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
 
-	chunkCerts := make([]*ChunkCertificate, 0, len(s.chunkMap))
+	chunkCerts := make([]*ChunkCertificate[T], 0, len(s.chunkMap))
```
aaronbuchwald, line 237: (unrelated to this PR) Can we remove the below nil check if the cert will always be populated when it's in the map?

x/dsmr/node_test.go
```diff
@@ -1478,6 +1669,120 @@ func getSignerBitSet(t *testing.T, pChain validators.State, nodeIDs ...ids.NodeI
...
+	)
+	r.NoError(err)
+	blk, err := node2.BuildBlock(Block[tx]{
+		ParentID:  ids.GenerateTestID(),
+		Height:    0,
+		Timestamp: 0,
+	}, 1)
+	r.NoError(err)
+	r.NoError(node2.Accept(context.Background(), blk))
+	r.NoError(node2.Execute(context.Background(), blk))
```
aaronbuchwald, line 1783: Should this order be reversed?

x/dsmr/node_test.go
aaronbuchwald: Execute should fail if a an invalid signature, expiry, metadata (parent hash, height, timestamp), or duplicate is included.

Accept should execute the internal block, which should be treated as a fatal error, so we can only really test that it correctly propagates a fatal error.

x/dsmr/node.go
```diff
@@ -133,28 +141,49 @@ func (n *Node[T]) BuildChunk(
...
+		})
+	}
 
-		if err := n.getChunkSignatureClient.AppRequest(ctx, validator.NodeID, request, onResponse); err != nil {
-			return Chunk[T]{}, err
-		}
+	aggregatedMsg, aggregatedStake, totalStake, err := n.chunkSignatureAggregator.AggregateSignatures(
+		ctx,
+		msg,
+		nil,
```
aaronbuchwald, line 170: (optional nit) could we add a comment here to note that justification is left empty? The `justification` field itself is a bit strange imo and think a comment to call out the field can be left nil would reduce cognitive load

x/dsmr/node.go
```diff
@@ -133,28 +141,49 @@ func (n *Node[T]) BuildChunk(
...
+	if warp.VerifyWeight(aggregatedStake, totalStake, n.quorumNum, n.quorumDen) != nil {
+		return Chunk[T]{}, fmt.Errorf("failed to replicate to sufficient stake: %w", err)
 	}
 
-	chunkCert := ChunkCertificate{
+	chunkCert := ChunkCertificate[T]{
 		ChunkID:   chunk.id,
 		Expiry:    chunk.Expiry,
-		Signature: NoVerifyChunkSignature{},
+		Signature: aggregatedMsg.Signature.(*warp.BitSetSignature),
```
aaronbuchwald, line 186: Can we move this typecheck above and return an error if it fails to future proof?

x/dsmr/node.go
```diff
@@ -200,28 +229,37 @@ func (n *Node[T]) BuildBlock(parent Block, timestamp int64) (Block, error) {
...
 }
 
-func (*Node[T]) Execute(ctx context.Context, _ Block, block Block) error {
+func (n *Node[T]) Execute(ctx context.Context, block Block[T]) error {
 	// TODO: Verify header fields
 	// TODO: de-duplicate chunk certificates (internal to block and across history)
 	for _, chunkCert := range block.ChunkCerts {
 		// TODO: verify chunks within a provided context
-		if err := chunkCert.Verify(ctx, struct{}{}); err != nil {
+		if err := chunkCert.Verify(
```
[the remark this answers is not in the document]
joshua-kim: Removed

x/dsmr/node.go
```diff
@@ -200,28 +229,37 @@ func (n *Node[T]) BuildBlock(parent Block, timestamp int64) (Block, error) {
...
+		return Block[T]{}, err
 	}
 
 	blk.blkBytes = packer.Bytes
 	blk.blkID = utils.ToID(blk.blkBytes)
 	return blk, nil
 }
 
-func (*Node[T]) Execute(ctx context.Context, _ Block, block Block) error {
+func (n *Node[T]) Execute(ctx context.Context, block Block[T]) error {
```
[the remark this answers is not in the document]
joshua-kim: I actually didn't understand why we had the parent parameter in the first place... I think if we wanted to keep this fn signature it should be a parent state instead of a block - or we accept the block to execute and assume that `Node` maintains parent state.

x/dsmr/storage.go
joshua-kim: The rationale was that we wouldn't serve `GetChunk` requests unless the chunk was considered "available". That being said... I'm not sure why I added this flag at all because we only store chunks after we've already generated a signature so any chunk written is guaranteed to be available.

x/dsmr/storage.go
joshua-kim: Yeah I'm not sure why this check exists here either. It looks like this should be safe to remove...

x/dsmr/node.go
joshua-kim: Sure

x/dsmr/storage_test.go
joshua-kim: We're now exporting the `ChunkStorage` type outside of this package, but i don't think there's a use-case where people should be manually playing with the internals of the database. I think I would prefer that we test this through the `Node` api

x/dsmr/certificate.go
```diff
@@ -39,35 +39,58 @@ func init() {
...
+	ctx context.Context,
+	storage *ChunkStorage[T],
+	networkID uint32,
+	chainID ids.ID,
+	pChainState validators.State,
+	pChainHeight uint64,
+	quorumNum uint64,
+	quorumDen uint64,
+) error {
+	chunkBytes, _, err := storage.GetChunkBytes(c.Expiry, c.ChunkID)
```
aaronbuchwald, line 71: We should remove this check, we don't need the chunk available in local storage for it to be valid for inclusion within a block

x/dsmr/storage.go
aaronbuchwald: I think we can remove, but don't need to in this PR

x/dsmr/storage.go
aaronbuchwald: We can remove the signature return type completely here.

From the node's perspective, when we receive a signature request we need to verify and store a chunk before returning a response, but the signature handler is now responsible for producing a signature from this node

aaronbuchwald (dismissed):
Left a few comments and I think some of the test cases may be unreachable in production. I could be wrong, otherwise we can either keep them as defensive or remove.

x/dsmr/node.go
```diff
@@ -107,69 +118,103 @@ func (n *Node[T]) BuildChunk(
...
+	}
 
-		<-done
+	if !ok {
+		return Chunk[T]{}, ChunkCertificate{}, errors.New("failed to replicate to sufficient stake")
+	}
+
+	bitSetSignature, ok := aggregatedMsg.Signature.(*warp.BitSetSignature)
+	if !ok {
+		return Chunk[T]{}, ChunkCertificate{}, errors.New("invalid signature type")
```
tsachiherman, line 192: doesn't have to be here, but, maybe move these errors to the top of the file ?

x/dsmr/node_test.go
```diff
@@ -1478,6 +1082,214 @@ func getSignerBitSet(t *testing.T, pChain validators.State, nodeIDs ...ids.NodeI
...
+							},
+							Signature: &warp.BitSetSignature{
+								Signers:   set.NewBits(1, 2, 3).Bytes(),
+								Signature: [96]byte{1, 2, 3},
+							},
+						},
+					},
+				}
+			},
+			wantErr: ErrInvalidWarpSignature,
```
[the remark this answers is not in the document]
joshua-kim: Removed the test case

x/dsmr/node.go
```diff
@@ -264,5 +354,39 @@ func (n *Node[T]) Accept(ctx context.Context, block Block) error {
...
+func (pChain) GetCurrentHeight(context.Context) (uint64, error) {
+	return 0, nil
+}
+
+func (pChain) GetSubnetID(context.Context, ids.ID) (ids.ID, error) {
+	return ids.Empty, nil
+}
+
+func (p pChain) GetValidatorSet(context.Context, uint64, ids.ID) (map[ids.NodeID]*snowValidators.GetValidatorOutput, error) {
+	result := make(map[ids.NodeID]*snowValidators.GetValidatorOutput)
```
tsachiherman, line 377: it would be nice to create the map with the expected size. i.e. `len(p.validators)`

x/dsmr/node_test.go
```diff
@@ -235,45 +173,8 @@ func TestNode_GetChunk_AvailableChunk(t *testing.T) {
...
-			&p2p.NoOpHandler{},
-			ids.EmptyNodeID,
-			ids.EmptyNodeID,
-		),
-		nil,
-	)
-	r.NoError(err)
-
-	chunk, err := node.BuildChunk(
+	node := newNode(t)
```
tsachiherman, line 176: can we call this newTestingNode ? just to distinguish from New()

tsachiherman (approved):

aaronbuchwald (approved):

PR #1825, opened 2024-12-03
Improve NewBlock tests

joshua-kim:


x/dsmr/node_test.go
```diff
@@ -768,196 +768,173 @@ func TestNode_NewBlock_IncludesChunkCerts(t *testing.T) {
...
 		{
 			name: "timestamp equal to parent",
-			parent: Block{
-				ParentID:  ids.GenerateTestID(),
-				Height:    1,
-				Timestamp: 1,
-				blkID:     ids.GenerateTestID(),
+			timestamp: func(parent Block) int64 {
+				return parent.Timestamp
 			},
```
joshua-kim, line 787: It doesn't make sense for us to specify the parent in the test vector because the parent state is now defined by the test node helper.

aaronbuchwald (approved):

PR #1826, opened 2024-12-03
Add Scope Interface

RodrigoVillar:
This PR modifies the `tstate` package by changing `scope` from simply being `state.Keys` to an interface which can be used to pass in `state.Keys` or another type (e.g. `SimulatedKeys`).

This PR also removes the `tstate` recorder.


RodrigoVillar:
Closing in favor of #1834

PR #1831, opened 2024-12-10
Snow VM

aaronbuchwald:
This PR integrates the changes added in https://github.com/ava-labs/hypersdk/pull/1867.

The overall change separates implementing the AvalancheGo `block.VM` and `snowman.Block` interfaces to fulfill DynamicStateSync into a separate, independently tested package.

This splits up the VM into individual pieces that can be overridden and creates a new `Chain` interface that can define three concrete block types
- input - parsed directly from bytes with no further changes or processing
- output - verified against its already verified parent (represents a processing block in consensus)
- accepted - accepted against its verified version and accepted parent (represents latest accepted block)

Additionally, `Chain` implements `Initialize` and `SetConsensusIndex` to take in all of the necessary parameters supplied by AvalancheGo and to set the required subscriptions or hooks needed to operate ie. accepted/verified/rejected block subscriptions, notifications when the VM transitions to `NormalOp`, etc.

After integrating, this PR adds new integration action/auth agnostic integration tests for the updated VM package. This replaces the test coverage previously added through MorpheusVM and cuts down the MorpheusVM integration tests to only the tests written to run against `workload.TestNetwork`.


aaronbuchwald:
Next steps:

1. Clean up dynamic state sync and write snow dynamic state sync tests
2. Move general purpose config out of `snow` package and replace with "context" that can either be a singleton or passed around similar to `require.Equal(t, a, b)` or `r.Equal(a, b)`
3. Update out of date comments
4. Replace "Options" misnomer (Configuration?)
5. Handle cyclic dependency with chain index (chain must currently return last accepted block and potentially re-process blocks)
6. Update VM package + rename to HyperVM + add HyperVM integration tests

PR #1832, opened 2024-12-12
Tiered Storage

RodrigoVillar:
This PR introduces tiered storage into the HyperSDK. By adding tiered storage, users are now refunded if any KV-pairs their TX touches was actually in memory in contrast to the current assumption that data always lives on disk.

TODO:

- [x] Add UT for Scope
- [x] Fix MorpheusVM `e2e` tests


chain/transaction.go
```diff
@@ -361,6 +372,20 @@ func (t *Transaction) Execute(
...
+	if err != nil {
+		return nil, ErrRefundDimensions
+	}
+	refundFee, err := feeManager.Fee(refund)
+	if err != nil {
+		return nil, ErrFailedToComputeRefund
+	}
+	if err := bh.AddBalance(ctx, t.Auth.Sponsor(), ts, refundFee); err != nil {
+		return nil, ErrRefundFailed
+	}
```
aaronbuchwald, line 387: Don't we have all the information needed to handle the refund immediately after fetching, so we can differentiate between tiered storage accessed? Do we need to wait until after execution for some reason?

vm/vm.go
```diff
@@ -1209,6 +1212,30 @@ func (vm *VM) restoreAcceptedQueue(ctx context.Context) error {
...
+			returnValues = append(returnValues, nil)
+			continue
+		}
+		returnValues = append(returnValues, v[:len(v)-consts.Uint64Len])
+		returnErrors = append(returnErrors, nil)
+	}
+
+	return returnValues, returnErrors
+}
+
```
aaronbuchwald, line 1238: Does this need to be coupled to the VM? In effect, we are grabbing the state from the VM's merkledb, could we apply the same transformation we normally do to a merkledb view before giving actions read access to it?

This function gives raw access to the statedb, should we leave to the caller (that knows where the information is used) to strip the suffixes / transform the data as needed?

chain/transaction.go
```diff
@@ -300,7 +302,15 @@ func (t *Transaction) PreExecute(
 	if err != nil {
 		return err
 	}
-	return bh.CanDeduct(ctx, t.Auth.Sponsor(), im, fee)
+
+	var st state.Immutable
```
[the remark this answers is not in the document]
RodrigoVillar: Nope; changed it so now callers are responsible for translating their values

chain/transaction.go
RodrigoVillar: Keeping the refund logic inside of `transaction.Execute()` would be consistent with how we're currently manipulating with the sponsor's balance. It's true that once we've fetched all values from the fetcher (in both the block builder and the validator) that we can immediately compute/apply the refund. However, this would require us to make a call to `BalanceHandler` outside of `transaction.Execute()`, which could be considered an antipattern as the sponsor's balance is manipulated only in `transaction.Execute()`.

As where the refund logic is placed inside `transaction.Execute()`, we could apply the refund after deducting. A property we gain from the current implementation is that refunds are applied only to successful transactions, but I wouldn't mind removing it.

chain/builder.go
```diff
@@ -408,11 +436,16 @@ func (c *Builder) BuildBlock(ctx context.Context, parentView state.View, parent
 	keys.Add(heightKeyStr, state.Write)
 	keys.Add(timestampKeyStr, state.Write)
 	keys.Add(feeKeyStr, state.Write)
-	tsv := ts.NewView(keys, map[string][]byte{
-		heightKeyStr:    binary.BigEndian.AppendUint64(nil, parent.Hght),
-		timestampKeyStr: binary.BigEndian.AppendUint64(nil, uint64(parent.Tmstmp)),
-		feeKeyStr:       parentFeeManager.Bytes(),
-	})
+	tsv = ts.NewView(
```
RodrigoVillar, line 439: After hook

chain/builder.go
```diff
@@ -397,6 +405,26 @@ func (c *Builder) BuildBlock(ctx context.Context, parentView state.View, parent
 		c.metrics.emptyBlockBuilt.Inc()
 	}
 
+	// Before continuing, we append the executing block's suffix to all existing
```
RodrigoVillar, line 408: Before hook

chain/builder.go
```diff
@@ -290,7 +291,12 @@ func (c *Builder) BuildBlock(ctx context.Context, parentView state.View, parent
 				}
 
 				// Execute block
-				tsv := ts.NewView(stateKeys, storage)
+				unsuffixedStorage, hotkeys, err := state.ExtractSuffixes(storage, height, c.config.Epsilon)
```
RodrigoVillar, line 294: Before hook

PR #1834, opened 2024-12-16
Improve `TState`

RodrigoVillar:


internal/fetcher/fetcher_test.go
```diff
@@ -142,12 +150,16 @@ func TestFetchSameKeysSlow(t *testing.T) {
 			stateKeys.Add(keyBase+strconv.Itoa(k), state.Read)
 		}
 		txID := ids.GenerateTestID()
+		keys := make([]string, 0, len(stateKeys))
```
tsachiherman, line 153: converting the state keys map into a keys slice seems to be a common usage pattern; do you think it would make sense to create a function that would encapsulate that ?

RodrigoVillar: For the tests, we can switch to generating database keys instead of generating state keys + immediately stripping their permissions. But agree that we should add a function that converts state keys into a keys slice.

RodrigoVillar: Done

state/dbtest/dbtest.go
```diff
@@ -0,0 +1,38 @@
...
+
+type TestDB struct {
+	storage map[string][]byte
+}
+
+func NewTestDB() *TestDB {
+	return &TestDB{
+		storage: make(map[string][]byte),
+	}
+}
```
aaronbuchwald, line 20: Can we rename this package `statetest` and rename the types here accordingly?

aaronbuchwald (changes requested):
Rather than combine scope, storage, and length into a single interface in this PR, could we decouple it?

ie. scope/immutable:

```golang
type Scope interface {
	Has(key []byte, perm Permissions) bool
}

type Immutable interface {
	GetValue(ctx context.Context, key []byte) (value []byte, err error)
}
```

Then make the constructor of `TStateView`:

```golang
func (ts *TState) NewView(scope state.Scope, storage state.Immutable, preallocateSize int) *TStateView
```

state/tstate/tstate_view.go
```diff
@@ -51,45 +40,29 @@ type TStateView struct {
...
-func (ts *TState) NewView(scope state.Keys, storage map[string][]byte) *TStateView {
-	return newView(ts, scope, immutableScopeStorage(storage), false)
-}
-
-func (*TStateRecorder) newRecorderView(immutableState state.Immutable) *TStateView {
-	return newView(New(0), state.Keys{}, immutableState, true)
 }
 
-func newView(ts *TState, scope state.Keys, storage state.Immutable, simulationMode bool) *TStateView {
+func (ts *TState) NewView(scope state.Scope, storage state.Immutable, preallocateSize int) *TStateView {
```
RodrigoVillar, line 54: Not sure if we should keep `preallocateSize()`. If we stick with removing `Len()` from the `Scope` interface, then we will be always passing in `0` for `preallocateSize` which makes the parameter have no purpose.

state/dbtest/dbtest.go
RodrigoVillar: I removed the package entirely since the only package using it was `TStateRecorder`, which no longer exists as part of this PR.

state/scope.go
```diff
@@ -0,0 +1,51 @@
...
+
+func NewDefaultScope(keys Keys) *DefaultScope {
+	return &DefaultScope{
+		keys: keys,
+	}
+}
+
+func (d *DefaultScope) Has(key []byte, perm Permissions) bool {
+	return d.keys[string(key)].Has(perm)
+}
```
aaronbuchwald, line 27: Could we just implement this interface directly on `Keys` rather than wrapping it?

state/scope.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (d *SimulatedScope) StateKeys() Keys {
+	return d.keys
+}
+
+// Flush clears the keys in the scope
+func (d *SimulatedScope) Flush() {
+	clear(d.keys)
+}
```
aaronbuchwald, line 51: Could this be implemented simply as a type alias that implements `Has` and can typecast back to `state.Keys` to handle `StateKeys()` and use `clear` directly rather than implement `Flush()` ?

state/scope.go
aaronbuchwald: If we do move it back, we should add the typecheck that `Keys` implements `Scope` in `state/keys.go`

state/storage.go
```diff
@@ -0,0 +1,21 @@
...
+package state
+
+import (
+	"context"
+
+	"github.com/ava-labs/avalanchego/database"
+)
+
+// ImmutableStorage implements [state.Immutable], allowing the map to be used
+// as a read-only state source, and making it compatible with error-generating backing storage.
```
aaronbuchwald, line 13: What does `"error-generating backing storage"` mean?

Perhaps:

```suggestion
// ImmutableStorage implements [state.Immutable] by wrapping a key-value map
```

aaronbuchwald (changes requested):

state/scope.go
RodrigoVillar: Done

state/keys.go
```diff
@@ -70,6 +83,17 @@ func (k Keys) WithoutPermissions() []string {
...
+type SimulatedKeys Keys
+
+func (d SimulatedKeys) Has(key []byte, perm Permissions) bool {
+	Keys(d).Add(string(key), perm)
+	return true
+}
+
+func (d SimulatedKeys) StateKeys() Keys {
+	return Keys(d)
+}
```
aaronbuchwald, line 95: Can we move this below all of the functions defined on `Keys` instead of interspersing this between the JSON marshalling functions?

aaronbuchwald (changes requested):

state/keys.go
RodrigoVillar: Done

aaronbuchwald (approved):

PR #1835, opened 2024-12-16
Bonds for Fortified DSMR

joshua-kim:
Implements account bonding for DSMR. HyperSDK will implement the `Bonder` interface which will keep track of account bond balances.

1. Txs are bonded when a node builds a tx into a chunk
2. Txs are un-bonded when the node either accepts a tx that it was tracking, or when the tx expiry passes.

This PR includes a few changes to the `dsmr` package to implement this, namely:

1. Test code common across `dsmr` and `fdsmr` are moved to `dsmrtest`
2. `Accept` now returns the outputted block with chunk details, because these details (i.e txs) are needed to unbond txs in `fdsmr`.
3. Because `Accept` returns the outputted block, we no longer need `BuildChunk` to return chunks/chunk certs to the caller, which was previously exposed for testing (bad practice).

### DIFF
--- a/x/dsmr/node.go
+++ b/x/dsmr/node.go
@@ -18,7 +18,6 @@ import (
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/avalanchego/utils/wrappers"
 	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
-
 	"github.com/ava-labs/hypersdk/codec"
 	"github.com/ava-labs/hypersdk/consts"
 	"github.com/ava-labs/hypersdk/proto/pb/dsmr"
@@ -33,7 +32,6 @@ const (
 
 var (
 	_ snowValidators.State = (*pChain)(nil)
-	_ Interface[Tx]        = (*Node[Tx])(nil)
 
 	ErrEmptyChunk                          = errors.New("empty chunk")
 	ErrNoAvailableChunkCerts               = errors.New("no available chunk certs")
@@ -46,13 +44,6 @@ var (
 	ErrInvalidSignatureType                = errors.New("invalid signature type")
 	ErrFailedToReplicate                   = errors.New("failed to replicate to sufficient stake")
 )
-
-type Interface[T Tx] interface {
-	BuildChunk(ctx context.Context, txs []T, expiry int64, beneficiary codec.Address) error
-	BuildBlock(parent Block, timestamp int64) (Block, error)
-	Verify(ctx context.Context, parent Block, block Block) error
-	Accept(ctx context.Context, block Block) (ExecutedBlock[T], error)
-}
 
 type Validator struct {
 	NodeID    ids.NodeID
--- a/x/dsmr/node.go
+++ b/x/dsmr/node.go
@@ -33,6 +33,7 @@ const (
 
 var (
 	_ snowValidators.State = (*pChain)(nil)
+	_ Interface[Tx]        = (*Node[Tx])(nil)
 
 	ErrEmptyChunk                          = errors.New("empty chunk")
 	ErrNoAvailableChunkCerts               = errors.New("no available chunk certs")
@@ -45,6 +46,13 @@ var (
 	ErrInvalidSignatureType                = errors.New("invalid signature type")
 	ErrFailedToReplicate                   = errors.New("failed to replicate to sufficient stake")
 )
+
+type Interface[T Tx] interface {
+	BuildChunk(ctx context.Context, txs []T, expiry int64, beneficiary codec.Address) error
+	BuildBlock(parent Block, timestamp int64) (Block, error)
+	Verify(ctx context.Context, parent Block, block Block) error
+	Accept(ctx context.Context, block Block) (ExecutedBlock[T], error)
+}
 
 type Validator struct {
 	NodeID    ids.NodeID
@@ -118,9 +126,9 @@ func (n *Node[T]) BuildChunk(
 	txs []T,
 	expiry int64,
 	beneficiary codec.Address,
-) (Chunk[T], ChunkCertificate, error) {
+) error {
 	if len(txs) == 0 {
-		return Chunk[T]{}, ChunkCertificate{}, ErrEmptyChunk
+		return ErrEmptyChunk
 	}
 
 	chunk, err := signChunk[T](
@@ -136,7 +144,7 @@ func (n *Node[T]) BuildChunk(
 		n.Signer,
 	)
 	if err != nil {
-		return Chunk[T]{}, ChunkCertificate{}, fmt.Errorf("failed to sign chunk: %w", err)
+		return fmt.Errorf("failed to sign chunk: %w", err)
 	}
 
 	packer := wrappers.Packer{MaxSize: MaxMessageSize}
@@ -145,12 +153,12 @@ func (n *Node[T]) BuildChunk(
 		Producer: chunk.Producer,
 		Expiry:   chunk.Expiry,
 	}, &packer); err != nil {
-		return Chunk[T]{}, ChunkCertificate{}, fmt.Errorf("failed to marshal chunk reference: %w", err)
+		return fmt.Errorf("failed to marshal chunk reference: %w", err)
 	}
 
 	unsignedMsg, err := warp.NewUnsignedMessage(n.networkID, n.chainID, packer.Bytes)
 	if err != nil {
-		return Chunk[T]{}, ChunkCertificate{}, fmt.Errorf("failed to initialize unsigned warp message: %w", err)
+		return fmt.Errorf("failed to initialize unsigned warp message: %w", err)
 	}
 	msg, err := warp.NewMessage(
 		unsignedMsg,
@@ -159,7 +167,7 @@ func (n *Node[T]) BuildChunk(
 		},
 	)
 	if err != nil {
-		return Chunk[T]{}, ChunkCertificate{}, fmt.Errorf("failed to initialize warp message: %w", err)
+		return fmt.Errorf("failed to initialize warp message: %w", err)
 	}
 
 	canonicalValidators, _, err := warp.GetCanonicalValidatorSet(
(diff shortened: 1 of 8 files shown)


x/dsmr/node.go
```diff
@@ -118,9 +126,9 @@ func (n *Node[T]) BuildChunk(
 	txs []T,
 	expiry int64,
 	beneficiary codec.Address,
-) (Chunk[T], ChunkCertificate, error) {
```
joshua-kim, line 121: This was so gross previously (and I'm to blame...), we previously exposed the chunk/chunk certs that were created into the caller. Realistically this is just leaking implementation details because hypersdk would never care about the chunks that were built - since the mempool/consensus only cares about blocks.

x/dsmr/node_test.go
```diff
@@ -1087,122 +1056,142 @@ func Test_Verify(t *testing.T) {
...
-					ParentID:   ids.GenerateTestID(),
-					Height:     parent.Height + 1,
-					Timestamp:  parent.Timestamp + 1,
-					ChunkCerts: []*ChunkCertificate{&chunkCert},
+					BlockHeader: BlockHeader{
+						ParentID:  ids.GenerateTestID(),
+						Height:    parent.Height + 1,
+						Timestamp: parent.Timestamp + 1,
+					},
+					ChunkCerts: parent.ChunkCerts,
```
joshua-kim, line 1071: Going over this - I think this is either weird/or a bug that we allow duplicate chunk certs to pass verification. For now i'm doing this because it passes tests and it's convenient - but probably something to fix as a follow-up.

x/dsmr/node.go
```diff
@@ -46,6 +47,13 @@ var (
 	ErrFailedToReplicate                   = errors.New("failed to replicate to sufficient stake")
 )
 
+type Interface[T Tx] interface {
+	BuildChunk(ctx context.Context, txs []T, expiry int64, beneficiary codec.Address) error
+	BuildBlock(parent Block, timestamp int64) (Block, error)
+	Verify(ctx context.Context, parent Block, block Block) error
+	Accept(ctx context.Context, block Block) (ExecutedBlock[T], error)
+}
```
aaronbuchwald, line 55: What's the rationale for defining the interface this way? Asking because we've previously discussed removing defining interfaces over a single implementation

x/fdsmr/node.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/x/dsmr"
+)
+
+type Bonder[T dsmr.Tx] interface {
+	// Bond returns if a transaction can be built into a chunk by this node.
+	// If this returns true, Unbond is guaranteed to be called.
+	Bond(tx T) bool
+	// Unbond is called when a tx from an account either expires or is accepted.
```
aaronbuchwald, line 22: Do we guarantee that `Bond` was called on this tx previously? If so, could we add that to the comment

x/fdsmr/node.go
```diff
@@ -0,0 +1,121 @@
...
+type Node[T dsmr.Interface[U], U dsmr.Tx] struct {
+	DSMR   T
+	bonder bonder[U]
+}
+
+func (n *Node[T, U]) BuildChunk(ctx context.Context, txs []U, expiry int64, beneficiary codec.Address) error {
+	if len(txs) == 0 {
+		return n.DSMR.BuildChunk(ctx, txs, expiry, beneficiary)
+	}
+
```
aaronbuchwald, line 46: Should we remove this check since it will be handled the same way (slice of size 0 and skip the loop) regardless of if it's here?

x/fdsmr/node.go
```diff
@@ -0,0 +1,121 @@
...
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	b.pending.Add(tx.GetID())
+
+	duration := time.Until(time.Unix(tx.GetExpiry(), 0))
+	go func() {
+		<-time.After(duration)
+		b.Unbond(tx)
+	}()
```
aaronbuchwald, line 105: Should we move the timestamp based on chain time instead of starting a goroutine for every tx?

x/fdsmr/node.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Unbond unbonds a tx
+func (b *bonder[T]) Unbond(tx T) {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	if !b.pending.Contains(tx.GetID()) {
+		// this tx was either not assigned to us or was already un-bonded
+		return
+	}
```
aaronbuchwald, line 118: When do we expect to hit this case?

x/dsmr/node_test.go
aaronbuchwald: It's a bug fixed in https://github.com/ava-labs/hypersdk/pull/1810. We should not allow duplicates either in a block or across the ancestry.

x/fdsmr/node_test.go
```diff
@@ -0,0 +1,380 @@
...
+		return false
+	}
+
+	pending <- struct{}{}
+	return true
+}
+
+func (b testBonder) Unbond(tx dsmrtest.Tx) {
+	<-b.pending[tx.GetSponsor()]
+}
```
aaronbuchwald, line 380: Can we simplify this if we switch away from using goroutines to mark txs as expired and unbond txs when the chain time advances past their expiry?

aaronbuchwald (changes requested):

x/dsmr/node.go
joshua-kim: I thought about this last night and originally felt this made sense so that we could keep wrapping `dsmr` with more features, but in hindsight maybe it makes more sense for this to live in the `fdsmr` package and say that `dsmr` implements it (i.e `dsmr` is fortifiable).

Another idea is to make `fdsmr` rely on the concrete type of `dsmr` since we only have one implementation but this makes testing look weirder.

x/fdsmr/node.go
joshua-kim: This is referenced the comment on `Bond`, but I can add a similar note to `Unbond`

x/fdsmr/node.go
joshua-kim: This is to handle https://github.com/ava-labs/hypersdk/blob/d1e726c6c4416bcce33f71d1b75d67c029a69cd3/x/fdsmr/node_test.go#L36-L45, where we guarantee that `nil` txs are wired into the underlying dsmr as `nil` as well (instead of a zero-length non-nil slice). This is probably overkill but I thought it made sense to be consistent w/ the caller value.

x/fdsmr/node.go
joshua-kim: Yeah we can use chain timestamps.

x/fdsmr/node.go
joshua-kim: I might not understand something but this is because we call `Unbond` on every tx in the block and only call the underlying bonder if it was a tx that we knew about. I think this is needed because we're not guaranteed that every tx in a block was handled by us.

x/fdsmr/node_test.go
joshua-kim: Yeah we can just use a map.

aaronbuchwald (approved):

x/fdsmr/node.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/ava-labs/avalanchego/ids"
+
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/x/dsmr"
+)
+
+type Interface[T dsmr.Tx] interface {
+	BuildChunk(ctx context.Context, txs []T, expiry int64, beneficiary codec.Address) error
+	Accept(ctx context.Context, block dsmr.Block) (dsmr.ExecutedBlock[T], error)
+}
```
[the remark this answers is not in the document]
joshua-kim: This pattern is done in a few places in the stdlib (i.e `sort.Interface`), but I would be open to calling this something like `fdsmr.Fortifiable`?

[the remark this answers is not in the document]
joshua-kim: Renamed to `DSMR`

x/fdsmr/node.go
```diff
@@ -0,0 +1,100 @@
...
+	for txID, tx := range n.pending {
+		if block.Timestamp <= tx.GetExpiry() {
+			continue
+		}
+
+		if err := n.bonder.Unbond(tx); err != nil {
+			return dsmr.ExecutedBlock[U]{}, err
+		}
+		delete(n.pending, txID)
+	}
```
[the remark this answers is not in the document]
joshua-kim: Added a TODO for this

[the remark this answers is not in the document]
joshua-kim: Added a TODO

x/fdsmr/node.go
```diff
@@ -0,0 +1,99 @@
...
+				continue
+			}
+
+			if err := n.bonder.Unbond(tx); err != nil {
+				return dsmr.ExecutedBlock[U]{}, err
+			}
+		}
+	}
+
+	n.pending.SetMin(block.Timestamp)
```
aaronbuchwald, line 96: `SetMin` returns the elements that were removed, could we use that instead of manually peeking/popping the elements to check if we need to call `Unbond` ?

joshua-kim: DIdn't know this return value existed, pretty nice!

x/fdsmr/node.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		if !ok {
+			continue
+		}
+
+		n.pending.Add(tx)
+		bonded = append(bonded, tx)
+	}
+
+	return n.DSMR.BuildChunk(ctx, bonded, expiry, beneficiary)
```
aaronbuchwald, line 60: (not necessary for this PR) - is `BuildChunk` returning an error considered a fatal error? If not, we need to unbond any transactions that were bonded before building the chunk failed

aaronbuchwald (approved):

PR #1838, opened 2024-12-17
Fix Inconsistent Usage of `consts` in `internal/fees`

RodrigoVillar:
This PR does the following:

- Fixes `manager.lastConsumed()` to use `consts.Int64Len` instead of `consts.IntLen` for the size of timestamp
- Adds a table test for internal fee manager

### DIFF
--- a/internal/fees/manager_test.go
+++ b/internal/fees/manager_test.go
@@ -4,27 +4,52 @@
 package fees
 
 import (
-	"math/rand"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 
-	externalfees "github.com/ava-labs/hypersdk/fees"
+	externalFees "github.com/ava-labs/hypersdk/fees"
 )
 
-func TestUnitsConsumedFuzz(t *testing.T) {
-	r := require.New(t)
-	rand := rand.New(rand.NewSource(0)) //nolint:gosec
+func TestUnitsConsumed(t *testing.T) {
+	t.Run("empty manager consumes", func(t *testing.T) {
+		r := require.New(t)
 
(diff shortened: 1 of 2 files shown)


internal/fees/manager_test.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	externalFees "github.com/ava-labs/hypersdk/fees"
+)
+
+func TestUnitsConsumed(t *testing.T) {
+	t.Run("empty manager consumes", func(t *testing.T) {
```
aaronbuchwald, line 15: Can we change this to a table test instead of using `t.Run` in sequence? This is a bit of a non-standard pattern and I think we could make this a simple table test with an initial input, amount consumed, and expected result

internal/fees/manager_test.go
```diff
@@ -0,0 +1,60 @@
...
+			// Initializing manager
+			manager := NewManager([]byte{})
+			ok, _ := manager.Consume(tt.initialUnits, tt.maxUnits)
+			r.True(ok)
+
+			// Testing manager
+			ok, _ = manager.Consume(tt.unitsToConsume, tt.maxUnits)
+			r.Equal(tt.success, ok)
+			if tt.success {
+				r.Equal(tt.maxUnits, manager.UnitsConsumed())
```
aaronbuchwald, line 56: Could we change from checking that `UnitsConsumed()` matches max units exactly to a separate vector for expected units?

RodrigoVillar: Done

aaronbuchwald (approved):

PR #1839, opened 2024-12-17
Remove Permissions From Fetcher

RodrigoVillar:
This PR removes permissions from `Fetcher` by having it take in a slice of strings (i.e. database keys) instead of `state.Keys`.

This PR also modifies the `Fetcher` tests to natively use database keys instead of state keys.

### DIFF
--- a/state/keys.go
+++ b/state/keys.go
@@ -61,7 +61,7 @@ func (k Keys) ChunkSizes() ([]uint16, bool) {
 	return chunks, true
 }
 
-// WithoutPermissions returns the database keys of k
+// WithoutPermissions returns the keys of k as a slice with permissions removed
 func (k Keys) WithoutPermissions() []string {
 	ks := make([]string, len(k))
 	for key := range k {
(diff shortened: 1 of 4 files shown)


state/keys.go
```diff
@@ -61,6 +61,15 @@ func (k Keys) ChunkSizes() ([]uint16, bool) {
 	return chunks, true
 }
 
+// Strip returns the database keys of k
+func (k Keys) Strip() []string {
```
aaronbuchwald, line 65: Could we rename this to either `Keys()`, `KeysWithoutPermissions()`, or `WithoutPermissions()` ?

RodrigoVillar: Will rename to `WithoutPermissions()`

RodrigoVillar: Done

state/keys.go
```diff
@@ -61,6 +61,15 @@ func (k Keys) ChunkSizes() ([]uint16, bool) {
 	return chunks, true
 }
 
+// WithoutPermissions returns the database keys of k
```
aaronbuchwald, line 64:
```suggestion
// WithoutPermissions returns the keys of k as a slice without the permissions attached
```

aaronbuchwald (approved):

PR #1848, opened 2025-01-02
Update License/Copyright to 2025

RodrigoVillar:
Goodbye 2024, Hello 2025!


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
```
joshua-kim, line 1: Not sure why our linter doesn't catch this (probably because the linter runs against the code on master instead of the commit on the branch) but we'd also have to update the headers in the code themselves.

```
➜  hypersdk git:(goodbye-2024) grep -r 'Copyright (C) 2024' . -l | wc -l
     401
```

LitoMore:
This year means the year when the license starts. It is wrong to modify it.

See https://opensource.stackexchange.com/questions/5778/why-do-licenses-such-as-the-mit-license-specify-a-single-year

joshua-kim:
If we're following that standard, we already already broke it in [this](https://github.com/ava-labs/hypersdk/commit/86d90df4e07287ab2abe8bd4c79b7bc7a9d2a42a) prior commit because hypersdk started in 2023 IIRC. With that being said, we don't use a "standard" license (bsd/mit/et al.) and use a proprietary Avalanche "Ecosystem" license so I'm not sure if this same standard applies, although to your point the referenced law you linked suggests that we should be marking 2023 as the denoted year either way.

LitoMore:
BTW, here is the context of why I noticed this pull request:

- https://github.com/orgs/community/discussions/148296

RodrigoVillar:
For reference, I referred to the following PR for convention: https://github.com/ava-labs/avalanchego/pull/2624

PR #1849, opened 2025-01-03
x/dsmr: Replace the avalanchego codec with canoto

StephenButtolph:


tools.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package hypersdk
+
+// TODO: Once we update to go1.24, we should register canoto as a tool in the
+// go.mod file.
+import (
+	_ "github.com/StephenButtolph/canoto/generate"
+)
```
StephenButtolph, line 10: This allows us to inherit the canoto version from the `go.mod` file - rather than needing to specify the version in every file.

x/dsmr/block.go
```diff
@@ -1,69 +1,65 @@
 // Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
 
+//go:generate go run github.com/StephenButtolph/canoto/canoto --concurrent=false $GOFILE
```
StephenButtolph, line 4: We specify `--concurrent=false` to allow passing structs by value.

x/dsmr/block.go
```diff
@@ -1,69 +1,65 @@
...
 	"github.com/ava-labs/hypersdk/codec"
-	"github.com/ava-labs/hypersdk/consts"
 	"github.com/ava-labs/hypersdk/utils"
 )
 
 const InitialChunkSize = 250 * 1024
 
-type Tx interface {
+type Tx[T any] interface {
+	canoto.FieldMaker[T]
```
StephenButtolph, line 21: This is basically the only real change. Rather than expecting the tx to be serializable with the avalanchego codec (which is an undocumented assumption) the interface now specifies canoto.

I had talked with aaron super briefly, but we may want to replace these with `[]byte` so that the transactions can specify their own encoding.

x/dsmr/block.go
```diff
@@ -1,69 +1,65 @@
...
-		return err
-	}
-
-	c.bytes = packer.Bytes
+func (c *Chunk[T]) init() {
+	c.bytes = c.MarshalCanoto()
 	c.id = utils.ToID(c.bytes)
-
-	return nil
 }
```
StephenButtolph, line 52: This diff warms my heart.

x/dsmr/block.go
```diff
@@ -79,41 +75,46 @@ func signChunk[T Tx](
...
+	return c, nil
 }
 
 type BlockHeader struct {
-	ParentID  ids.ID `serialize:"true"`
-	Height    uint64 `serialize:"true"`
-	Timestamp int64  `serialize:"true"`
+	ParentID  ids.ID `canoto:"fixed bytes,1"`
+	Height    uint64 `canoto:"int,2"`
+	Timestamp int64  `canoto:"int,3"`
```
StephenButtolph, line 108: I decided to use `int` rather than `sint` here - because I don't think we ever actually expect `Timestamp` to be negative.

x/dsmr/block.go
```diff
@@ -79,41 +75,46 @@ func signChunk[T Tx](
...
+
+	canotoData canotoData_BlockHeader
 }
 
 type Block struct {
 	BlockHeader
```
StephenButtolph, line 114: Should this `BlockHeader` be serialized?

x/dsmr/block.go
```diff
@@ -124,8 +125,10 @@ func (b Block) GetID() ids.ID {
 }
 
 // ExecutedBlock contains block data with any referenced chunks reconstructed
-type ExecutedBlock[T Tx] struct {
+type ExecutedBlock[T Tx[T]] struct {
 	BlockHeader
 	ID     ids.ID
-	Chunks []Chunk[T] `serialize:"true"`
+	Chunks []Chunk[T] `canoto:"repeated value,1"`
```
StephenButtolph, line 131: Similarly to `Block` - is it correct for us to only serialize the `Chunks` here?

x/dsmr/certificate.go
```diff
@@ -1,68 +1,48 @@
...
-
-	"github.com/ava-labs/hypersdk/codec"
-
-	acodec "github.com/ava-labs/avalanchego/codec"
-)
-
-const (
-	CodecVersion = 0
-
-	MaxMessageSize = units.KiB
```
StephenButtolph, line 26: Was `MaxMessageSize` used to actually enforce something important? Or was this just used to conform to the avalanchego codec?

x/dsmr/certificate.go
```diff
@@ -1,68 +1,48 @@
...
-	Producer ids.NodeID `serialize:"true"`
-	Expiry   int64      `serialize:"true"`
+type Signature struct {
+	// Signers is a big-endian byte slice encoding which validators signed this
+	// message.
+	Signers   []byte                 `canoto:"bytes,1"`
+	Signature [bls.SignatureLen]byte `canoto:"fixed bytes,2"`
+
+	canotoData canotoData_Signature
 }
```
StephenButtolph, line 33: Because `*warp.BitSetSignature` is implemented in avalanchego - we needed to re-implement it here to be able to serialize it differently.

x/dsmr/p2p.go
```diff
@@ -38,13 +35,9 @@ var (
 		Code:    3,
 		Message: "invalid chunk",
 	}
-
-	_ acp118.Verifier = (*ChunkSignatureRequestVerifier[Tx])(nil)
-	_ p2p.Handler     = (*GetChunkHandler[Tx])(nil)
-	_ p2p.Handler     = (*ChunkCertificateGossipHandler[Tx])(nil)
```
StephenButtolph, line 44: I don't think we have a good type to parameterize `Tx` with here to allow for these assertions... So I removed them. (Similar comment for other interface checks)

x/dsmr/typed_client.go
```diff
@@ -137,23 +134,14 @@ func (getChunkMarshaler[T]) UnmarshalResponse(bytes []byte) (Chunk[T], error) {
...
-	if err := codec.LinearCodec.UnmarshalFrom(
-		&wrappers.Packer{Bytes: response.Chunk},
-		&chunk,
-	); err != nil {
-		return Chunk[T]{}, err
-	}
-
-	return chunk, chunk.init()
+	return ParseChunk[T](response.Chunk)
 }
```
StephenButtolph, line 138: This is almost an unrelated change... This could have previously called to `ParseChunk` as well... But with removal of the `codec.LinearCodec`, I felt like this change made sense here.

x/dsmr/block.go
joshua-kim: I think this will technically pass tests but we should serialize here because the block id/bytes should be derived from the `BlockHeader`

x/dsmr/block.go
joshua-kim: We probably don't have to technically serialize this... we know the `ID` from the `Block` this is derived from

x/dsmr/block.go
```diff
@@ -1,69 +1,63 @@
...
 	"github.com/ava-labs/avalanchego/utils/crypto/bls"
-	"github.com/ava-labs/avalanchego/utils/wrappers"
 	"github.com/ava-labs/avalanchego/vms/platformvm/warp"
 
 	"github.com/ava-labs/hypersdk/codec"
-	"github.com/ava-labs/hypersdk/consts"
 	"github.com/ava-labs/hypersdk/utils"
 )
 
-const InitialChunkSize = 250 * 1024
```
StephenButtolph, line 17: This is why so much memory was being allocated when initializing the chunk.

go.mod
```diff
@@ -1,8 +1,9 @@
 module github.com/ava-labs/hypersdk
 
-go 1.22.8
+go 1.22.9
 
 require (
+	github.com/StephenButtolph/canoto v0.9.0
```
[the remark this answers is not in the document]
StephenButtolph: I had made this as a personal project while OOO. If we wanted to move it under the ava-labs org we could 🤷.

fwiw - I hope we never need to make any changes to it 😅

PR #1852, opened 2025-01-06
Add UT for Transaction `PreExecute()`

RodrigoVillar:
This PR adds various unit tests which test the `PreExecute()` method of `transaction`. This PR also adds a singular mock type for actions and for auth.


chain/transaction_test.go
```diff
@@ -252,3 +334,146 @@ func TestSignRawActionBytesTx(t *testing.T) {
 	require.NoError(err)
 	require.Equal(signedTx.Bytes(), rawSignedTxBytes)
 }
+
+func TestPreExecute(t *testing.T) {
```
aaronbuchwald, line 338: (could be split to separate PR) - we should write unit tests for the pre executor type itself as well, since that is how `PreExecute` will be called from outside this package

chain/transaction_test.go
```diff
@@ -252,3 +334,146 @@ func TestSignRawActionBytesTx(t *testing.T) {
...
+			tx: &chain.Transaction{
+				TransactionData: chain.TransactionData{
+					Base: &chain.Base{
+						Timestamp: 2 * consts.MillisecondsPerSecond,
+					},
+				},
+				Auth: &auth1{},
+			},
+			timestamp: 2 * consts.MillisecondsPerSecond,
+			err:       chain.ErrAuthNotActivated,
```
aaronbuchwald, line 457: Can we add test cases for:

- valid test case
- fee overflow
- insufficient balance

chain/transaction_test.go
```diff
@@ -77,6 +83,82 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
...
+	panic("unimplemented")
+}
+
+type auth1 struct {
+	mockAuth
+}
+
+func (*auth1) ValidRange(_ chain.Rules) (int64, int64) {
+	return 1 * consts.MillisecondsPerSecond, 1 * consts.MillisecondsPerSecond
+}
```
aaronbuchwald, line 160: Can we set the valid range start/end as fields on `auth1` and same for `action3` ? This will make it easier to look at each individual test case and see immediately the start/end values rather than needing to know the start/end on the implementation above

chain/transaction_test.go
```diff
@@ -77,6 +83,82 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
...
+
+func (*mockBalanceHandler) GetBalance(_ context.Context, _ codec.Address, _ state.Immutable) (uint64, error) {
+	panic("unimplemented")
+}
+
+func (*mockBalanceHandler) SponsorStateKeys(_ codec.Address) state.Keys {
+	panic("unimplemented")
+}
+
+type mockAuth struct{}
```
aaronbuchwald, line 120: could we rename this `abstractMockAuth` to match the implementation of the base `mockAction` above?

chain/transaction_test.go
```diff
@@ -252,3 +334,146 @@ func TestSignRawActionBytesTx(t *testing.T) {
...
+			},
+			timestamp: 2 * consts.MillisecondsPerSecond,
+			err:       chain.ErrTimestampTooLate,
+		},
+		{
+			name: "base transaction timestamp too early (62ms > (60 + 1)ms)",
+			tx: &chain.Transaction{
+				TransactionData: chain.TransactionData{
+					Base: &chain.Base{
+						Timestamp: 62 * consts.MillisecondsPerSecond,
```
aaronbuchwald, line 373: Can we set `testRules` above declaring the test cases and use `(testRules.ValidityWindow + 1) * consts.MillisecondsPerSecond` ?

This should make the test case more clear in that it's setting the timestamp to greater than the validity window rather than the reader needing to know the default value of 60

chain/transaction_test.go
```diff
@@ -252,3 +334,146 @@ func TestSignRawActionBytesTx(t *testing.T) {
...
+			},
+			timestamp: consts.MillisecondsPerSecond,
+			err:       chain.ErrTimestampTooEarly,
+		},
+		{
+			name: "base transaction invalid chain id",
+			tx: &chain.Transaction{
+				TransactionData: chain.TransactionData{
+					Base: &chain.Base{
+						ChainID:   ids.ID{1},
```
aaronbuchwald, line 385: Could we set the `chainID` used in the body of the for loop above and also set a variable called `differentChainID` to use here?

chain/transaction_test.go
```diff
@@ -252,3 +334,146 @@ func TestSignRawActionBytesTx(t *testing.T) {
...
+			timestamp: consts.MillisecondsPerSecond,
+			err:       chain.ErrInvalidChainID,
+		},
+		{
+			name: "transaction has too many actions (17 > 16)",
+			tx: &chain.Transaction{
+				TransactionData: chain.TransactionData{
+					Base: &chain.Base{},
+					Actions: func() []chain.Action {
+						actions := make([]chain.Action, 17)
```
aaronbuchwald, line 399: Could we use a variable for the maximum number of actions here rather than requiring the reader to know that 16 is the maximum?

aaronbuchwald (changes requested):

chain/transaction_test.go
RodrigoVillar: Done

chain/transaction_test.go
RodrigoVillar: Done

chain/transaction_test.go
RodrigoVillar: Done

chain/transaction_test.go
RodrigoVillar: Done

chain/transaction_test.go
RodrigoVillar: > Could we set the chainID used in the body of the for loop above

Do you mean the for loop below?

chain/transaction_test.go
RodrigoVillar: Done

chain/transaction_test.go
```diff
@@ -252,3 +404,218 @@ func TestSignRawActionBytesTx(t *testing.T) {
 	require.NoError(err)
 	require.Equal(signedTx.Bytes(), rawSignedTxBytes)
 }
+
+func TestPreExecute(t *testing.T) {
+	// Test values
+	testRules := genesis.NewDefaultRules()
+	maxNumOfActions := 16
```
aaronbuchwald, line 411: This is a rule set by `testRules`, could we use `testRules.MaxActionsPerTx` instead of setting it to 16 here (this works because it's the same as the default value) ?

chain/transaction_test.go
```diff
@@ -77,6 +92,143 @@ func unmarshalAction2(p *codec.Packer) (chain.Action, error) {
...
+
+func (*abstractMockAuth) ValidRange(_ chain.Rules) (int64, int64) {
+	panic("unimplemented")
+}
+
+func (*abstractMockAuth) Verify(_ context.Context, _ []byte) error {
+	panic("unimplemented")
+}
+
+type auth1 struct {
```
aaronbuchwald, line 210: (optional for this PR) - the pattern of extending an abstract type that panics for every function is a bit annoying especially when we implement a new `authX` type for each new test.

Could we switch to use a single mock or test action/auth type that allows tests to set arbitrary values?

aaronbuchwald (changes requested):

chain/transaction_test.go
```diff
@@ -252,3 +404,218 @@ func TestSignRawActionBytesTx(t *testing.T) {
...
+		{
+			name: "valid test case",
+			tx: &chain.Transaction{
+				TransactionData: chain.TransactionData{
+					Base: &chain.Base{},
+				},
+				Auth: &auth1{},
+			},
+			bh: &mockBalanceHandler2{},
+		},
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/transaction_test.go
RodrigoVillar: Done

chain/transaction_test.go
RodrigoVillar: Added to the following issue: https://github.com/ava-labs/hypersdk/issues/1860

RodrigoVillar: Added a single mock/auth type; kept `mockTransferAction` and `action2` instead of solely using `mockAction` since these types have additional field required for codec-related tests

PR #1853, opened 2025-01-07
Implement bonds for fortification

joshua-kim:
Implements bonds by maintaining state of an account's pending + max bond balance

### DIFF
--- a/chain/bond.go
+++ b/chain/bond.go
@@ -1,5 +1,4 @@
 // Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
-
 // See the file LICENSE for licensing terms.
 
 package chain
@@ -7,49 +6,62 @@ package chain
 import (
 	"errors"
 	"fmt"
-	"sync"
 
-	acodec "github.com/ava-labs/avalanchego/codec"
 	"github.com/ava-labs/avalanchego/database"
 	"github.com/ava-labs/avalanchego/utils/wrappers"
 	"github.com/ava-labs/hypersdk/codec"
 	"github.com/ava-labs/hypersdk/x/fdsmr"
 )
 
-var _ fdsmr.Bonder[*Transaction] = (*Bonder)(nil)
+var (
+	_              fdsmr.Bonder[*Transaction] = (*Bonder)(nil)
+	ErrMissingBond                            = errors.New("missing bond")
+)
 
 type BondBalance struct {
 	Pending uint32 `serialize:"true"`
 	Max     uint32 `serialize:"true"`
 }
 
+func NewBonder(db database.Database) *Bonder {
+	return &Bonder{db: db}
+}
+
 // Bonder maintains state of account bond balances to limit the amount of
 // pending transactions per account
 type Bonder struct {
-	codec acodec.Manager
-	db    database.Database
-	lock  sync.Mutex
+	db database.Database
 }
 
 // this needs to be thread-safe if it's called from the api
-func (b *Bonder) SetMaxBalance(address codec.Address, i uint32) error {
-	b.lock.Lock()
-	defer b.lock.Unlock()
-
+func (b *Bonder) GetBalance(address codec.Address) (uint32, uint32, error) {
 	addressBytes := address[:]
 	balance, err := b.getBalance(addressBytes)
 	if err != nil {
-		return err
+		return 0, 0, err
 	}
 
-	balance.Max = i
-	return b.putBalance(addressBytes, balance)
+	return balance.Pending, balance.Max, nil
+}
+
+// this needs to be thread-safe if it's called from the api
+func (b *Bonder) SetMaxBalance(address codec.Address, maxBalance uint32) (uint32, error) {
+	addressBytes := address[:]
+	balance, err := b.getBalance(addressBytes)
+	if err != nil {
+		return 0, err
+	}
+
+	prev := balance.Max
+	balance.Max = maxBalance
+	if err := b.putBalance(addressBytes, balance); err != nil {
+		return 0, err
+	}
+
+	return prev, nil
 }
 
 func (b *Bonder) Bond(tx *Transaction) (bool, error) {
-	b.lock.Lock()
-	defer b.lock.Unlock()
-
 	address := tx.Sponsor()
 	addressBytes := address[:]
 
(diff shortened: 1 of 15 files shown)


chain/transaction.go
```diff
@@ -372,6 +372,10 @@ func (t *Transaction) Execute(
 	}, nil
 }
 
+func (t *Transaction) GetSponsor() codec.Address {
```
tsachiherman, line 375: it's weird that we have both "GetSponsor" and "Sponsor".

joshua-kim: This is because `Sponsor` implements the `mempool.Item` interface while `GetSponsor` implements the `dsmr.Tx` interface. We could either wrap types to avoid this or rename both functions to be the same.

chain/bond.go
```diff
@@ -0,0 +1,134 @@
...
+type BondBalance struct {
+	Pending uint32 `serialize:"true"`
+	Max     uint32 `serialize:"true"`
+}
+
+// Bonder maintains state of account bond balances to limit the amount of
+// pending transactions per account
+type Bonder struct{}
+
+// this needs to be thread-safe if it's called from the api
```
aaronbuchwald, line 38: Should this comment be removed since it's writing to `state.Mutable` which is a merkledb view? We never write concurrently to the merkledb views elsewhere

chain/bond.go
```diff
@@ -0,0 +1,134 @@
...
+	balance, err := getBalance(ctx, mutable, addressBytes)
+	if err != nil {
+		return false, err
+	}
+
+	if balance.Pending >= balance.Max {
+		return false, nil
+	}
+
+	balance.Pending++
```
aaronbuchwald, line 73: Shouldn't this be a fee amount as opposed to incrementing by one? Technically, we only need to make sure that it can pay for bandwidth, not MDF.

aaronbuchwald (changes requested):

chain/bond.go
joshua-kim: Updated to use fee supplied through a parameter

chain/transaction.go
aaronbuchwald: We should rename both functions to be the same imo

x/fdsmr/node.go
```diff
@@ -42,10 +44,10 @@ type Node[T DSMR[U], U dsmr.Tx] struct {
 	pending *eheap.ExpiryHeap[U]
 }
 
-func (n *Node[T, U]) BuildChunk(ctx context.Context, txs []U, expiry int64, beneficiary codec.Address) error {
+func (n *Node[T, U]) BuildChunk(ctx context.Context, mutable state.Mutable, txs []U, expiry int64, beneficiary codec.Address, fee *big.Int) error {
```
aaronbuchwald, line 47: Could we make it clear when `mutable` is a local view of the chunk builder vs. operating on the network wide chain state (merkle trie)?

It's not completely clear to me from this in `BuildChunk` and `Accept` what state we're operating on.

We need to set the max balance in the chain state, so that each chunk builder has an agreed on view of what the max balance is for each potential tx issuer at the start of an epoch.  Each chunk builder must maintain up to date pending balances, so that it can correctly limit transactions if they exceed their cap and needs to maintain a separate state/database for this (cannot write to the network wide state from its own local view).

Should we maintain a database in `fdsmr` so that we can keep the pending balances up to date rather than leaving this to the caller to supply?

joshua-kim: > Should we maintain a database in fdsmr so that we can keep the pending balances up to date rather than leaving this to the caller to supply?

Why would the database live in `fdsmr`? Is this because you want to commit all of the unbond balance updates in a single batch? If so I think this is simpler to implement by just making `Unbond` idempotent

aaronbuchwald: It seems simpler that fdsmr maintaining responsibility for its own state is simpler than pushing it to the caller.

We just need to maintain that fdsmr can restart correctly and know the outstanding balances of any chunks that it has produced in a given epoch and is therefore responsible for.

Another alternative would be that on restart, we iterate over any pending chunks that we've created (which should include all chunks that could still be included in a block both produced by us and otherwise) and refresh our pending balances from this.

internal/chain/bond_test.go
```diff
@@ -0,0 +1,290 @@
...
+	db, err := merkledb.New(
+		context.Background(),
+		memdb.New(),
+		merkledb.Config{BranchFactor: 2},
+	)
+	require.NoError(t, err)
+	return db
+}
+
+type NoAuthFactory struct{}
```
aaronbuchwald, line 241: (optional for this PR / followup) Since we have an auth mock in the chain package, could we move this under `auth/authtest/` so that we can re-use instead of creating it in each place we need?

internal/chain/bond_test.go
```diff
@@ -0,0 +1,290 @@
...
+			tx: func() *chain.Transaction {
+				tx, err := chain.NewTxData(
+					&chain.Base{Timestamp: 123, ChainID: ids.Empty, MaxFee: 456},
+					nil,
+				).Sign(&NoAuthFactory{})
+				require.NoError(t, err)
+
+				return tx
+			}(),
+			feeRate: 0,
```
aaronbuchwald, line 98: (no action necessary) `feeRate = 0` should probably be invalid

aaronbuchwald (approved):
LGTM

Summarizing my understanding of how we'll break down the state here:
- max balance will be included in the chain state (state replicated across the network)
- chunk builders locally maintain a database of pending transactions to ensure they never allow an account to spend more than their max balance

To actually pass `state.Mutable` into fdsmr here, I think we'll need to ensure that the max balance does not change within an epoch or pass in a view that doesn't change from the start of an epoch, so that a chunk builder never ends up in a state where they've included txs that consume bandwidth of 100 and cannot pay for it because they decreased their max balance.

internal/chain/bond.go
```diff
@@ -0,0 +1,159 @@
...
+		return false, nil
+	}
+
+	batch := b.db.NewBatch()
+	if err := putPendingBalance(batch, addressBytes, updatedBalance); err != nil {
+		return false, err
+	}
+
+	txID := tx.GetID()
+	if err := batch.Put(txID[:], binary.BigEndian.AppendUint64(nil, fee)); err != nil {
```
aaronbuchwald, line 86: I'm not sure that writing txs by hash will be performant enough here, but think this is fine for now.

PR #1857, opened 2025-01-08
Remove Error From `window.Roll()`

RodrigoVillar:
This PR removes the error return parameter from `window.Roll()` and modifies the `roll` parameter type from `int64` to `uint64`.

Currently, although `window.Roll()` returns an error, this is always set to `nil` and so for testing/code quality, this PR removes the error return parameter altogether.

Furthermore, the `roll` parameter is currently of type `int64` but after further inspection, we have that `roll = currTime - prevTime`. Assuming that `currTime - prevTime >= 0`, changing `roll` to uint64 is safer as we no longer run the risk of passing in a negative value when [indexing](https://github.com/ava-labs/hypersdk/blob/4c616547551d05aec457fac025f44bfe49d5cce2/internal/window/window.go#L46).

### DIFF
--- a/internal/fees/manager.go
+++ b/internal/fees/manager.go
@@ -81,11 +81,11 @@ func (f *Manager) ComputeNext(currTime int64, r Rules) (*Manager, error) {
 	minUnitPrice := r.GetMinUnitPrice()
 	lastTimeSeconds := int64(binary.BigEndian.Uint64(f.raw[0:consts.Int64Len]))
 	currTimeSeconds := currTime / consts.MillisecondsPerSecond
-	since := currTimeSeconds - lastTimeSeconds
+	since := uint64(currTimeSeconds - lastTimeSeconds)
 	bytes := make([]byte, consts.Int64Len+dimensionStateLen*fees.FeeDimensions)
 	binary.BigEndian.PutUint64(bytes[0:consts.Int64Len], uint64(currTimeSeconds))
 	for i := fees.Dimension(0); i < fees.FeeDimensions; i++ {
-		nextUnitPrice, nextUnitWindow, err := computeNextPriceWindow(
+		nextUnitPrice, nextUnitWindow := computeNextPriceWindow(
 			f.Window(i),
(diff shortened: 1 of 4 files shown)


internal/fees/manager.go
```diff
@@ -81,11 +81,11 @@ func (f *Manager) ComputeNext(currTime int64, r Rules) (*Manager, error) {
 	minUnitPrice := r.GetMinUnitPrice()
 	lastTimeSeconds := int64(binary.BigEndian.Uint64(f.raw[0:consts.Int64Len]))
 	currTimeSeconds := currTime / consts.MillisecondsPerSecond
-	since := currTimeSeconds - lastTimeSeconds
+	since := uint64(currTimeSeconds - lastTimeSeconds)
```
aaronbuchwald, line 84: Can we add an error check here for an underflow?

Seems this would have previously triggered a panic by indexing into a negative index and we should already protect against this elsewhere, but better to either return an error or add a comment that clearly places the responsibility on the caller

aaronbuchwald (approved):
Nice catch! Left one comment to improve edge case handling, but the edge case previously would have triggered a panic, so think this is good to merge as is.

internal/window/window.go
```diff
@@ -34,17 +34,16 @@ type Window [WindowSliceSize]byte
...
-// Assumes that [roll] is greater than or equal to 0
-func Roll(w Window, roll int64) (Window, error) {
+func Roll(w Window, roll uint64) Window {
 	// Note: make allocates a zeroed array, so we are guaranteed
 	// that what we do not copy into, will be set to 0
 	res := [WindowSliceSize]byte{}
 	bound := roll * consts.Uint64Len
 	if bound > WindowSliceSize {
-		return res, nil
+		return res
```
tsachiherman, line 43: this change would miss the overflow protection. to fix it:
```golang
if bound >= WindowSliceSize  || roll >= WindowSliceSize {
```
( note the `>=` on the bound side, as it would prevent a potential oob )

PR #1858, opened 2025-01-09
Add UTs for `PreExecutor.PreExecute()`

RodrigoVillar:
This PR adds UTs for the `PreExecute()` method of `PreExecutor`.


chain/pre_executor_test.go
```diff
@@ -0,0 +1,207 @@
...
+					start: -1,
+					end:   -1,
+				},
+			},
+			validityWindow: &mockValidityWindow{},
+		},
+		{
+			name: "raw fee doesn't exist",
+			err:  database.ErrNotFound,
+		},
```
aaronbuchwald, line 93: Can we include a valid transaction for a test case like this where it will error before dereferencing the transaction?

If the code were to change the order of verification, but the raw fee was still missing, ideally the test case would still pass. By writing the test this way, it makes it brittle and it will fail if we make any ordering changes to the code itself.

aaronbuchwald: Same comment for remaining test cases. To avoid duplicate code, we could construct a single valid tx above the creation of the slice of test cases.

aaronbuchwald: Can we also add a more specific error, so that we won't confuse `database.ErrNotFound` from another part of the code with failing to fetch the raw fee?

chain/pre_executor_test.go
```diff
@@ -0,0 +1,207 @@
...
+	errMockValidityWindow = errors.New("mock validity window error")
+
+	_ chain.Auth           = (*mockAuth)(nil)
+	_ chain.BalanceHandler = (*mockBalanceHandler)(nil)
+	_ chain.ValidityWindow = (*mockValidityWindow)(nil)
+)
+
+type mockValidityWindow struct {
+	isRepeatError error
+	setBits       []int
```
aaronbuchwald, line 40: can we replace `setBits` with a value of type `set.Bits` ? Not sure there's any advantage to handling this within the mocked `IsRepeat` instead of leaving to the caller.

aaronbuchwald (changes requested):

chain/pre_executor_test.go
```diff
@@ -0,0 +1,215 @@
...
+func (*mockValidityWindow) VerifyExpiryReplayProtection(context.Context, validitywindow.ExecutionBlock[*chain.Transaction], int64) error {
+	panic("unimplemented")
+}
+
+func TestPreExecutor(t *testing.T) {
+	testRules := genesis.NewDefaultRules()
+	ruleFactory := genesis.ImmutableRuleFactory{
+		Rules: testRules,
+	}
+	validTX := &chain.Transaction{
```
aaronbuchwald, line 59: Can we camel case this ie. `validTx` ?

aaronbuchwald (changes requested):

chain/pre_executor_test.go
RodrigoVillar: Done

chain/pre_executor_test.go
RodrigoVillar: Done

chain/pre_executor.go
```diff
@@ -40,9 +42,12 @@ func (p *PreExecutor) PreExecute(
 	tx *Transaction,
 ) error {
 	feeRaw, err := view.GetValue(ctx, FeeKey(p.metadataManager.FeePrefix()))
-	if err != nil {
+	if err != nil && err != database.ErrNotFound {
 		return err
 	}
+	if err == database.ErrNotFound {
+		return ErrFeeNotFound
+	}
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/pre_executor_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	_ chain.Auth           = (*mockAuth)(nil)
+	_ chain.BalanceHandler = (*mockBalanceHandler)(nil)
+	_ chain.ValidityWindow = (*mockValidityWindow)(nil)
+)
+
+type mockValidityWindow struct {
+	isRepeatError error
+	setBits       set.Bits
+}
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/pre_executor_test.go
RodrigoVillar: Done

chain/pre_executor_test.go
```diff
@@ -0,0 +1,215 @@
...
+					BranchFactor: merkledb.BranchFactor16,
+					Tracer:       trace.Noop,
+				},
+			)
+			r.NoError(err)
+
+			for k, v := range tt.state {
+				r.NoError(db.Put([]byte(k), v))
+			}
+			r.NoError(db.CommitToDB(ctx))
```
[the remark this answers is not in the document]
RodrigoVillar: Switched the `PreExecutor` from taking in `state.View` to taking in `state.Immutable`

aaronbuchwald (approved):

PR #1859, opened 2025-01-09
pin rust version to 1.83 to avoid breaking changes in 1.84

tsachiherman:
## What ? 

The rust components started generating errors during the CI build due to rust upgrade.

## Workaround

pin rust version to 1.83 to avoid breaking changes in 1.84.

### DIFF
--- a/.github/actions/install-rust/action.yml
+++ b/.github/actions/install-rust/action.yml
@@ -14,8 +14,11 @@ runs:
   using: composite
   steps:
     - name: Install Rust
-      uses: dtolnay/rust-toolchain@stable
+      uses: dtolnay/rust-toolchain@1.83.0
       with:
         targets: ${{ inputs.targets }}
     - if: ${{ inputs.cache == 'true' }}
       uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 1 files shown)


.github/actions/install-rust/action.yml
```diff
@@ -14,8 +14,11 @@ runs:
...
     - name: Install Rust
-      uses: dtolnay/rust-toolchain@stable
+      uses: dtolnay/rust-toolchain@1.83.0
       with:
         targets: ${{ inputs.targets }}
     - if: ${{ inputs.cache == 'true' }}
       uses: Swatinem/rust-cache@v2
+    - name: Install rustfmt
+      shell: bash
+      run: rustup component add rustfmt clippy
```
joshua-kim, line 24: Why is this change needed?

tsachiherman: it wouldn't work without it... From some reason, when specifying an explicit version, it wouldn't install these components ( which we use ). hence, I had to add them explicitly.

joshua-kim (approved):

PR #1862, opened 2025-01-10
Verify chunk signature

tsachiherman:
## What ?

This PR add verification of the received chunk signature.


x/dsmr/block.go
```diff
@@ -82,6 +86,32 @@ func signChunk[T Tx](
...
+	pk, err := bls.PublicKeyFromCompressedBytes(c.Signer[:])
+	if err != nil {
+		return err
+	}
+	packer := wrappers.Packer{Bytes: make([]byte, 0, InitialChunkSize), MaxSize: consts.NetworkSizeLimit}
+	if err := codec.LinearCodec.MarshalInto(c.UnsignedChunk, &packer); err != nil {
+		return err
+	}
+
+	msg, err := warp.NewUnsignedMessage(networkID, chainID, packer.Bytes)
```
aaronbuchwald, line 103: can we use `c.bytes` rather than re-packing the message here?

x/dsmr/block.go
```diff
@@ -16,6 +18,8 @@ import (
 
 const InitialChunkSize = 250 * 1024
 
+var ErrFailedSigVerification = errors.New("failed to verify signature")
```
aaronbuchwald, line 21: Can we make this more detailed ie. at least include the signature type and the type of container the signature was over like `"failed to verify bls chunk signature"`

x/dsmr/p2p.go
```diff
@@ -127,10 +128,18 @@ func (c ChunkSignatureRequestVerifier[T]) Verify(
...
 		return ErrInvalidChunk
 	}
 
+	// TODO:
+	// check if the expiry of this chunk isn't in the past or too far into the future.
+
+	// TODO:
+	// check if the producer was expected to produce this chunk.
+
+	// check to see if this chunk was already accepted.
```
aaronbuchwald, line 142: Can these comments be moved into the verifier?

From p2p perspective, it should only need to know that it must verify the chunk in some context, but doesn't need to know what verification takes place.

x/dsmr/p2p.go
```diff
@@ -127,10 +128,18 @@ func (c ChunkSignatureRequestVerifier[T]) Verify(
 		}
 	}
 
+	// verify that the signature is a valid signature over the unsigned chunk.
```
aaronbuchwald, line 131: Could we remove obvious comments that match the function name ie. parse and verify unless the comment provides additional information?

x/dsmr/block.go
aaronbuchwald: We need to use warp signature verification here to factor in the current validator set as well.

x/dsmr/storage.go
```diff
@@ -32,12 +32,15 @@ type Verifier[T Tx] interface {
...
-type NoVerifier[T Tx] struct{}
+type ChunkVerifier[T Tx] struct {
+	networkID uint32
+	chainID   ids.ID
+}
 
-func (NoVerifier[T]) Verify(Chunk[T]) error {
-	return nil
+func (c ChunkVerifier[T]) Verify(chunk Chunk[T]) error {
+	return chunk.Verify(c.networkID, c.chainID)
```
aaronbuchwald, line 43: Should the verification logic be here so that the functionality is separated from the data?

aaronbuchwald: Disregard - it's currently just a syntactic check, so better to keep it on chunk itself and closer to the sign implementation

x/dsmr/p2p.go
tsachiherman: sure.

x/dsmr/p2p.go
aaronbuchwald: Should we add a TODO for rate limiting chunk production from the given producer?

aaronbuchwald (approved):

PR #1865, opened 2025-01-14
(wip) dsmr: verify chunk producer was supposed to produce the given chunk.

tsachiherman:


x/dsmr/storage.go
```diff
@@ -24,33 +27,91 @@ const (
...
+	c.min = min
 }
 
 func (c ChunkVerifier[T]) Verify(chunk Chunk[T]) error {
-	// TODO:
 	// check if the expiry of this chunk isn't in the past or too far into the future.
+	if chunk.Expiry < c.min {
+		return fmt.Errorf("%w: chunk expiry %d, minimum expiry %d", ErrChunkTooOld, chunk.Expiry, c.min)
+	}
+	if chunk.Expiry >= c.min+maxChunkStorageAheadDuration {
```
aaronbuchwald, line 80: we should use the validity window duration passed into `node` instead of a constant here

x/dsmr/storage.go
```diff
@@ -24,33 +27,91 @@ const (
...
+		return fmt.Errorf("%w: chunk expiry %d, minimum expiry %d", ErrChunkTooFarAhead, chunk.Expiry, c.min)
+	}
 
-	// TODO:
 	// check if the producer was expected to produce this chunk.
+	validatorSet, err := c.chainState.GetValidatorSet(context.Background(), 0, ids.Empty)
```
aaronbuchwald, line 85: (unrelated to this PR) - we may want to define our own interface instead of using one similar to P-Chain state

aaronbuchwald, line 85: We should use a `subnetID` param here rather than fetching with subnetID `ids.Empty`

x/dsmr/storage.go
```diff
@@ -141,14 +202,20 @@ func (s *ChunkStorage[T]) AddLocalChunkWithCert(c Chunk[T], cert *ChunkCertifica
...
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
 	storedChunk, ok := s.chunkMap[chunkID]
 	if !ok {
 		return fmt.Errorf("failed to store cert for non-existent chunk: %s", chunkID)
 	}
+	if cert == nil {
+		return ErrInvalidChunkCertificate
+	}
```
aaronbuchwald, line 215: If we are changing this where the cert cannot be nil, could we remove this check and make it an assumption that this cannot be called with a nil chunk certificate?

x/dsmr/node_test.go
```diff
@@ -435,33 +435,110 @@ func TestNode_AcceptedChunksAvailableOverGetChunk(t *testing.T) {
...
+					validators: []Validator{{
+						NodeID:    nodeID,
+						Weight:    1,
+						PublicKey: pk,
+					}},
+				},
+				1,
+				1,
+			),
+			wantErr:                   ErrInvalidChunk,
```
aaronbuchwald, line 473: Can we use more specific errors in the code and assertions in the test so that we have higher confidence it's failing for the expected reason?

aaronbuchwald (changes requested):

x/dsmr/node_test.go
tsachiherman: yes

x/dsmr/p2p.go
```diff
@@ -124,6 +160,11 @@ func (c ChunkSignatureRequestVerifier[T]) Verify(
 	}
 
 	if err := c.verifier.Verify(chunk); err != nil {
+		for _, e := range verifierAppErrors {
+			if errors.Is(err, e.Error) {
+				return e.AppError
+			}
+		}
```
aaronbuchwald, line 167: Ah I see that this is needed to support testing for an individual error. Apologies for this comment as I didn't realize this was what was required to support it. Could we revert back to remove this special case error handling and simply check for `ErrInvalidChunk` ?

aaronbuchwald (changes requested):

aaronbuchwald (approved):

PR #1868, opened 2025-01-15
Decouple tstate from merkledb

aaronbuchwald:
This PR decouples tstate from merkledb.

TState is responsible for constructing state diffs built on top of `TStateView`s. Each view is constructed on top of its own `state.Immutable` view.

Previously, TState was responsible for creating a merkledb view built on top of a parent. This change makes it so that TState simply calculates the state diff and leaves it to the caller to create a new merkledb view.


state/tstate/tstate_test.go
```diff
@@ -660,14 +641,13 @@ func TestCreateView(t *testing.T) {
 	tsv.Commit()
 
 	// Create merkle view
```
RodrigoVillar, line 643: Nit: remove reference to merkle view

chain/processor.go
```diff
@@ -242,7 +246,8 @@ func (p *Processor) Execute(
 	// Get view from [tstate] after processing all state transitions
```
RodrigoVillar, line 246:
```suggestion
	// Get view from state diff after processing all state transitions
```

RodrigoVillar (changes requested):
One nit; otherwise, LGTM

state/tstate/tstate_test.go
aaronbuchwald: Going to leave as is, since it's still extracting the state diff from tstate

PR #1871, opened 2025-01-16
Improve confirm invalid tx

aaronbuchwald:
This PR adds a step to `ConfirmInvalidTx` which will:
- construct a block on top of the currently preferred block including the invalid tx
- parse and verify the block and assert that it hits the same `targetErr` as for `SubmitTx`

This PR additionally fixes a bug in block verification found in the improved tests where the `parentTimestamp` was used as the `oldestAllowed` value rather than using `b.Tmstmp - r.GetValidityWindow()`.

### DIFF
--- a/chain/processor.go
+++ b/chain/processor.go
@@ -135,11 +135,11 @@ func (p *Processor) Execute(
 	heightKey := HeightKey(p.metadataManager.HeightPrefix())
 	parentHeightRaw, err := parentView.GetValue(ctx, heightKey)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to fetch parent height from state: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


chain/processor.go
```diff
@@ -167,28 +167,28 @@ func (p *Processor) Execute(
 	}
 
 	if isNormalOp {
-		if err := p.validityWindow.VerifyExpiryReplayProtection(ctx, b, parentTimestamp); err != nil {
-			return nil, err
+		oldestAllowed := max(0, b.Tmstmp-r.GetValidityWindow())
```
tsachiherman, line 170: should the validity window object itself be responsible for calculating the "oldest allowed" ? It seems that we had a bug due to improper code duplication.

tsachiherman (commented):
looks good. one minor comment.

PR #1872, opened 2025-01-16
improve snow fuzz test and remove use of fzgen fuzzing library

aaronbuchwald:
Removes `fzgen` library in favor of native go fuzzing library, which generates more useful inputs.

The previous code using `fzgen` for the initial seed was repeatedly calling `BuildBlock` because it was generating the same step on each invocation. Changing to the native fuzzing library immediately caught testing bugs cause by:

- JSON marshal/unmarshal invalid data returned a different error than our target parse invalid block error
- needed to update preference after accepting a non-preferred chain

This PR removes `ParseInvalidBlockBytes` as it did not provide very useful coverage and fixes the handling of `acceptNonPreferredChain`.

Note: calling `SetPreference` after accepting a non-preferred chain matches the behavior of the consensus engine, which  calls `SetPreference` to the current preference of consensus after successfully recording a poll: https://github.com/ava-labs/avalanchego/blob/5cb8b4363dfda46a3490e31c16a9f4c454f6acfe/snow/engine/snowman/voter.go#L67

### DIFF
--- a/snow/vm_test.go
+++ b/snow/vm_test.go
@@ -23,7 +23,6 @@ import (
 	"github.com/ava-labs/avalanchego/utils/set"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
-	"github.com/thepudds/fzgen/fuzzer"
 	"golang.org/x/exp/slices"
 
 	"github.com/ava-labs/hypersdk/chainindex"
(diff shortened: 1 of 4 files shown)


snow/vm_test.go
```diff
@@ -914,23 +947,23 @@ func TestDynamicStateSync_FinishOnAcceptedAncestor(t *testing.T) {
...
-		fz.Fill(&randSource)
+	maxFuzzSteps := 50
+	// Cap the number of steps to take by using byte as the type
+	f.Fuzz(func(t *testing.T, randSource int64, byteSteps []byte) {
 		rand := rand.New(rand.NewSource(randSource)) //nolint:gosec
 
 		ctx := context.Background()
 		ce := NewTestConsensusEngineWithRand(t, rand, &TestBlock{outputPopulated: true, acceptedPopulated: true})
 
-		for i := byte(0); i < numSteps; i++ {
```
aaronbuchwald, line 930: Previous behavior for seed 1 was to repeatedly call `BuildBlock` ie. step = 0.

PR #1873, opened 2025-01-17
Improve dynamic state sync transition comments

aaronbuchwald:
Update the comments around `chainLock` used to synchronize between finishing dynamic state sync and entering normal operation.

### DIFF
--- a/snow/vm.go
+++ b/snow/vm.go
@@ -104,7 +104,7 @@ type VM[I Block, O Block, A Block] struct {
 	version              string
 
 	// chainLock provides a synchronization point between state sync and normal operation.
-	// To complete dynamic state sync, we may:
+	// To complete dynamic state sync, we must:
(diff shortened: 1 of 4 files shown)


snow/vm.go
```diff
@@ -104,9 +103,17 @@ type VM[I Block, O Block, A Block] struct {
 	preReadyAcceptedSubs []event.Subscription[I]
 	version              string
 
-	// chainLock must be held to process a block with chain (Build/Verify/Accept).
-	chainLock       sync.Mutex
-	chain           Chain[I, O, A]
+	// chainLock provides a synchronization point between state sync and normal operation.
+	// To complete dynamic state sync, we may:
```
aaronbuchwald, line 107:
```suggestion
	// To complete dynamic state sync, we must:
```

PR #1882, opened 2025-01-21
Improve Fee Manager / Window

RodrigoVillar:
This PR modifies the following:

- Currently, the `ComputeNext` method returns an error, but this error is always `nil`. Therefore, this PR removes the error return parameter from `ComputeNext`. This PR also modifies any code which calls `ComputeNext` to no longer account for an error.
- As mentioned [here](https://github.com/ava-labs/hypersdk/pull/1857#discussion_r1907882049), the current overflow check for the window `Roll()` function is insufficient. This PR modifies the overflow check to account for all possible cases of overflow.
    - This PR also adds a regression test to check for window overflow.

### DIFF
--- a/internal/window/window_test.go
+++ b/internal/window/window_test.go
@@ -107,7 +107,7 @@ func TestWindowRollOverflow(t *testing.T) {
 			r := require.New(t)
 
 			window := Window{}
-			for i := 0; i < WindowSize; i++ {
+			for i := 0; i < WindowSliceSize; i++ {
 				window[i] = consts.MaxUint8
 			}
 


internal/window/window.go
```diff
@@ -39,7 +39,7 @@ func Roll(w Window, roll uint64) Window {
 	// that what we do not copy into, will be set to 0
 	res := [WindowSliceSize]byte{}
 	bound := roll * consts.Uint64Len
-	if bound > WindowSliceSize {
+	if bound >= WindowSliceSize || roll >= WindowSliceSize {
```
aaronbuchwald, line 42: if we are trying to defend against an overflow here, could we do this more simply?

For example:

```
if roll > WindowSize { ... }
```

then we can just return the empty byte slice

aaronbuchwald: Should we add a regression test for this as well?

RodrigoVillar: Done

aaronbuchwald (approved):

PR #1885, opened 2025-01-22
Break up processor nits

aaronbuchwald:
This PR applies a few nits on top of https://github.com/ava-labs/hypersdk/pull/1883

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -89,13 +89,6 @@ func (c *Chain) Execute(
 	return c.processor.Execute(ctx, parentView, b, isNormalOp)
 }
 
-func (c *Chain) AsyncVerify(
-	ctx context.Context,
-	b *ExecutionBlock,
-) error {
-	return c.processor.asyncVerify(ctx, b)
-}
-
 func (c *Chain) PreExecute(
 	ctx context.Context,
 	parentBlk *ExecutionBlock,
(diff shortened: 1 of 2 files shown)


chain/chain.go
```diff
@@ -89,13 +89,6 @@ func (c *Chain) Execute(
 	return c.processor.Execute(ctx, parentView, b, isNormalOp)
 }
 
-func (c *Chain) AsyncVerify(
```
aaronbuchwald, line 92: Remove since this is no longer used

chain/processor.go
```diff
@@ -36,7 +36,6 @@ type ExecutionBlock struct {
 	// to preallocate memory
 	authCounts map[uint8]int
 	txsSet     set.Set[ids.ID]
-	sigJob     workers.Job
```
aaronbuchwald, line 39: Switch from setting on the block type and waiting later to returning and using later. This should be less error prone since if we don't use it again, we'll get a warning.

chain/processor.go
```diff
@@ -46,9 +45,10 @@ type OutputBlock struct {
 	ExecutionResults ExecutionResults
 }
 
-type BlockContext struct {
-	height    uint64
-	timestamp int64
+type blockContext struct {
```
aaronbuchwald, line 48: un-exporting this for now since it's only used in this package

RodrigoVillar (approved):
LGTM

PR #1890, opened 2025-01-23
Add UTs to Processor `Execute()`

RodrigoVillar:
This PR adds the following:

- UTs to test the `Execute()` method of the processor
- Adds additional errors for UTs and expands functionality of existing mocks


chain/processor_test.go
```diff
@@ -0,0 +1,318 @@
...
+				timestampKey: binary.BigEndian.AppendUint64(nil, 0),
+				feeKey:       {},
+			},
+			block: &chain.ExecutionBlock{
+				StatelessBlock: &chain.StatelessBlock{
+					Hght:      1,
+					Tmstmp:    testRules.GetMinEmptyBlockGap(),
+					StateRoot: invalidStateRoot,
+				},
+			},
```
aaronbuchwald, line 237: Can we switch all instances to using an existing constructor where possible?

This ensures that we're using the actual code to produce a fully populated block where possible and reduces the number of paths where we may create a block instance with some fields left unpopulated (this will leave the bytes, id, and auth counts fields unpopulated).

chain/processor_test.go
```diff
@@ -0,0 +1,318 @@
...
+			}
+			r.NoError(db.CommitToDB(ctx))
+
+			if tt.includeRoot {
+				root, err := db.GetMerkleRoot(ctx)
+				r.NoError(err)
+				tt.block.StateRoot = root
+			}
+
+			_, err = testChain.Execute(ctx, db, tt.block, tt.isNormalOp)
```
aaronbuchwald, line 314: Can we change to use a function to return the block based off of a `merkledb.View` constructed from the state?

This way each test case clearly specifies whether it is using the parent's merkle root or a random value as opposed to relying on overriding it via `includeRoot` here.

aaronbuchwald (changes requested):

internal/logging/noop.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package logging
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,318 @@
...
+				feeKey:       {},
+			},
+			block: validBlock,
+			workers: &workers.MockWorkers{
+				OnNewJob: func(_ int) (workers.Job, error) {
+					return &workers.MockJob{
+						OnWaitError: errMockWorkerJob,
+					}, nil
+				},
+			},
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,318 @@
...
+						return &workers.MockJob{}, nil
+					},
+				}
+			}
+
+			if tt.validityWindow == nil {
+				tt.validityWindow = &validitywindowtest.MockTimeValidityWindow[*chain.Transaction]{}
+			}
+
+			testChain, err := chain.NewChain(
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,383 @@
...
+	testRuleFactory := genesis.ImmutableRuleFactory{Rules: testRules}
+	createValidBlock := func(root ids.ID) (*chain.StatelessBlock, error) {
+		return chain.NewStatelessBlock(
+			ids.Empty,
+			time.Now().UnixMilli(),
+			1,
+			nil,
+			root,
+		)
+	}
```
joshua-kim, line 66: I would just eat the cost of duplicate code here and declare these per test - even if these are types that are immutable sharing state across tests is harder to maintain a mental model of + can lead to issues if it becomes mutable in the future.

chain/processor_test.go
```diff
@@ -0,0 +1,383 @@
...
+			name: "failed to get fee",
+			state: map[string][]byte{
+				heightKey:    binary.BigEndian.AppendUint64(nil, 0),
+				timestampKey: binary.BigEndian.AppendUint64(nil, 0),
+			},
+			createBlock: createValidBlock,
+			expectedErr: chain.ErrFailedToFetchParentFee,
+		},
+		{
+			name: "failed VerifyExpiryReplayProtection",
```
joshua-kim, line 212: nit: name this lowercase to be consistent, e.g `fails replay protection`

chain/processor_test.go
```diff
@@ -0,0 +1,383 @@
...
+			r := require.New(t)
+			ctx := context.Background()
+
+			if tt.validityWindow == nil {
+				tt.validityWindow = &validitywindowtest.MockTimeValidityWindow[*chain.Transaction]{}
+			}
+
+			if tt.workers == nil {
+				tt.workers = workers.NewSerial()
+			}
```
joshua-kim, line 335: Why do we override this from the test? IMO it's better to have a single source of truth and define this in the test table instead of overriding it w/ a default in the test body

chain/processor_test.go
```diff
@@ -0,0 +1,383 @@
...
+			state: map[string][]byte{
+				heightKey:    binary.BigEndian.AppendUint64(nil, 0),
+				timestampKey: binary.BigEndian.AppendUint64(nil, 0),
+				feeKey:       {},
+			},
+			createBlock: createValidBlock,
+		},
+		{
+			name: "block timestamp too late",
+			createBlock: func(root ids.ID) (*chain.StatelessBlock, error) {
```
joshua-kim, line 88: IMO returning this error is awkward, can we use `require.NoError(t, <type>)` to guarantee no error to the caller?

chain/processor_test.go
```diff
@@ -0,0 +1,383 @@
...
+			root,
+		)
+	}
+
+	tests := []struct {
+		name           string
+		validityWindow chain.ValidityWindow
+		isNormalOp     bool
+		createBlock    createBlock
+		state          map[string][]byte
```
joshua-kim, line 73: Since the function ` Execute` that we're testing operates on a `merkledb.View`, should we use that instead of a `map`?

chain/processor_test.go
```diff
@@ -0,0 +1,383 @@
...
+		workers        workers.Workers
```
joshua-kim, line 74: nit: move this to be adjacent to `validityWindow` so that dependencies and parameters are grouped to be adjacent with each other

chain/processor_test.go
```diff
@@ -0,0 +1,383 @@
...
+		expectedErr    error
+	}{
+		{
+			name: "valid test case",
+			state: map[string][]byte{
+				heightKey:    binary.BigEndian.AppendUint64(nil, 0),
+				timestampKey: binary.BigEndian.AppendUint64(nil, 0),
+				feeKey:       {},
```
joshua-kim, line 82: q: Why is this value the empty value? Is this for convenience because the zero value `{}` is equivalent to a 0 varint?

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Updated tests to declare the workers/validity window that they use

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

chain/processor.go
```diff
@@ -383,7 +383,7 @@ func (p *Processor) createBlockContext(
 	}
 	parentHeight, err := database.ParseUInt64(parentHeightRaw)
 	if err != nil {
-		return blockContext{}, fmt.Errorf("failed to parse parent height from state: %w", err)
+		return blockContext{}, fmt.Errorf("%w %w", ErrFailedToParseParentHeight, err)
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: By setting the raw fee value to empty, this is equivalent to a fresh fee manager being loaded in (and which should be fine since we're not testing the FM in this PR)

https://github.com/ava-labs/hypersdk/blob/a9801b84ce8c7572a5484042a215ef8374a231ba/internal/fees/manager.go#L33-L35

chain/processor_test.go
```diff
@@ -0,0 +1,383 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package chain_test
```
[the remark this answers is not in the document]
RodrigoVillar: There's a cyclic dependency between the `chain` package and the `genesis` package

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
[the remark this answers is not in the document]
joshua-kim: The cyclic dependency is more of a symptom of the underlying problem which is that the tests depend on `auth` since we use their signing implementations. This dependency issue is easily fixable if we just create a test auth implementation - there really isn't a reason why this package needs to exist. I'm fine with us doing this in a different PR though, since this package already exists in the current state of the repo

chain/metrics.go
```diff
@@ -53,7 +53,7 @@ func (em *executorMetrics) RecordExecutable() {
 	em.executable.Inc()
 }
 
-func newMetrics(reg *prometheus.Registry) (*chainMetrics, error) {
+func NewMetrics(reg *prometheus.Registry) (*chainMetrics, error) {
```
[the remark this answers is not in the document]
RodrigoVillar: As per offline convo, going to export `NewMetrics()` and `ChainMetrics` for now. A future PR which resolves the `chain` test cyclic dependencies should un-export these.

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+
+type mockAuthVM struct{}
+
+func (*mockAuthVM) GetAuthBatchVerifier(_ uint8, _ int, _ int) (chain.AuthBatchVerifier, bool) {
+	return nil, false
+}
+
+func (*mockAuthVM) Logger() logging.Logger {
+	panic("unimplemented")
+}
```
joshua-kim, line 53: nit: move test implementations to the bottom of this file

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+
+func TestProcessorExecute(t *testing.T) {
+	createValidBlock := func(root ids.ID) *chain.StatelessBlock {
+		block, err := chain.NewStatelessBlock(
+			ids.Empty,
+			time.Now().UnixMilli(),
```
joshua-kim, line 59: nit: avoid non-deterministic behavior in unit tests where possible - this test should pass if we just use `time.Time{}`

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+			db: func() merkledb.MerkleDB {
+				db := createDB()
+				require.NoError(t, db.Put([]byte(heightKey), []byte{}))
+				return db
+			}(),
+			createBlock: createValidBlock,
+			expectedErr: chain.ErrFailedToParseParentHeight,
+		},
+		{
+			name:           "block height is not one more than parent height (2 != 0 + 1)",
```
joshua-kim, line 153: nit: test name is introspecting too much, the assertion message will tell us what is expected/returned, we don't care about the values and only care that the height is the previous height + 1.
```suggestion
			name:           "block height is not one more than parent height",
```

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+				require.NoError(t, db.Put([]byte(timestampKey), binary.BigEndian.AppendUint64(nil, 0)))
+				return db
+			}(),
+			createBlock: createValidBlock,
+			expectedErr: chain.ErrFailedToFetchParentFee,
+		},
+		{
+			name: "fails replay protection",
+			validityWindow: &validitywindowtest.MockTimeValidityWindow[*chain.Transaction]{
+				OnVerifyExpiryReplayProtection: func(_ context.Context, _ validitywindow.ExecutionBlock[*chain.Transaction]) error {
```
joshua-kim, line 273: nit: delete unnamed param names, not sure why our linter doesn't catch this
```suggestion
				OnVerifyExpiryReplayProtection: func(context.Context, validitywindow.ExecutionBlock[*chain.Transaction]) error {
```

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+			validityWindow: &validitywindowtest.MockTimeValidityWindow[*chain.Transaction]{},
+			workers:        workers.NewSerial(),
+			db: func() merkledb.MerkleDB {
+				db := createDB()
+				require.NoError(t, db.Put([]byte(heightKey), binary.BigEndian.AppendUint64(nil, 0)))
+				require.NoError(t, db.Put([]byte(timestampKey), binary.BigEndian.AppendUint64(nil, 0)))
+				require.NoError(t, db.Put([]byte(feeKey), []byte{}))
+				return db
+			}(),
+			createBlock: func(_ ids.ID) *chain.StatelessBlock {
```
joshua-kim, line 342:
```suggestion
			createBlock: func(ids.ID) *chain.StatelessBlock {
```

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+	timestampKey = string(chain.TimestampKey([]byte{1}))
+
+	errMockVerifyExpiryReplayProtection = errors.New("mock validity window error")
+)
+
+type createBlock func(parentRoot ids.ID) *chain.StatelessBlock
+
+type mockAuthVM struct{}
+
+func (*mockAuthVM) GetAuthBatchVerifier(_ uint8, _ int, _ int) (chain.AuthBatchVerifier, bool) {
```
joshua-kim, line 47:
```suggestion
func (*mockAuthVM) GetAuthBatchVerifier(uint8, int, int) (chain.AuthBatchVerifier, bool) {
```

RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+		return db
+	}
+
+	tests := []struct {
+		name           string
+		validityWindow chain.ValidityWindow
+		workers        workers.Workers
+		isNormalOp     bool
+		db             merkledb.MerkleDB
+		createBlock    createBlock
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+var (
+	_ chain.AuthVM = (*mockAuthVM)(nil)
+
+	heightKey    = string(chain.HeightKey([]byte{0}))
+	timestampKey = string(chain.TimestampKey([]byte{1}))
+
+	errMockVerifyExpiryReplayProtection = errors.New("mock validity window error")
+)
+
+type createBlock func(parentRoot ids.ID) *chain.StatelessBlock
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+			validityWindow: &validitywindowtest.MockTimeValidityWindow[*chain.Transaction]{},
+			workers:        workers.NewSerial(),
+			db: func() merkledb.MerkleDB {
+				db := createDB()
+				require.NoError(t, db.Put([]byte(heightKey), binary.BigEndian.AppendUint64(nil, 0)))
+				require.NoError(t, db.Put([]byte(timestampKey), binary.BigEndian.AppendUint64(nil, 0)))
+				require.NoError(t, db.Put([]byte(feeKey), []byte{}))
+				return db
+			}(),
+			createBlock: createValidBlock,
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,435 @@
...
+			context.Background(),
+			memdb.New(),
+			merkledb.Config{
+				BranchFactor: merkledb.BranchFactor16,
+				Tracer:       trace.Noop,
+			},
+		)
+		require.NoError(t, err)
+		return db
+	}
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
joshua-kim: This test is still inherently flaky, it could be refactored to accept caller-provided timestamps but it's outside of the scope of the PR

joshua-kim (approved):

chain/processor_test.go
```diff
@@ -0,0 +1,618 @@
...
+			root, err := tt.db.GetMerkleRoot(ctx)
+			r.NoError(err)
+
+			_, err = processor.Execute(
+				ctx,
+				tt.db,
+				chain.NewExecutionBlock(tt.newBlockF(root)),
+				tt.isNormalOp,
+			)
+			r.ErrorIs(err, tt.expectedErr)
```
aaronbuchwald, line 605: (separate PR) - could we confirm the post state is what we expect as well ie. we should make sure that the metadata is updated correctly and that tx/action execution (include successful and failed w/ rollbacks) has the expected effect

aaronbuchwald (changes requested):

chain/processor_test.go
```diff
@@ -0,0 +1,618 @@
...
+						Tracer:       trace.Noop,
+					},
+				)
+				require.NoError(t, err)
+				return db
+			}(),
+			newBlockF: func(root ids.ID) *chain.StatelessBlock {
+				block, err := chain.NewStatelessBlock(
+					ids.Empty,
+					genesis.NewDefaultRules().GetMinEmptyBlockGap(),
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,618 @@
...
+					context.Background(),
+					memdb.New(),
+					merkledb.Config{
+						BranchFactor: merkledb.BranchFactor16,
+						Tracer:       trace.Noop,
+					},
+				)
+				require.NoError(t, err)
+				return db
+			}(),
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,618 @@
...
+			expectedErr: chain.ErrTimestampTooLate,
+		},
+		{
+			name:           "verify signatures fails",
+			validityWindow: &validitywindowtest.MockTimeValidityWindow[*chain.Transaction]{},
+			workers: func() workers.Workers {
+				w := workers.NewParallel(0, 0)
+				w.Stop()
+				return w
+			}(),
```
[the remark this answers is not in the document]
RodrigoVillar: As per offline convo, I believe we can get rid of this test. `Stop()` is called when the VM is shutting down and since that case is outside the scope of `Execute()`, we shouldn't need to test for it.

chain/processor_test.go
```diff
@@ -0,0 +1,618 @@
...
+	timestampKey = string(chain.TimestampKey([]byte{1}))
+
+	errMockVerifyExpiryReplayProtection = errors.New("mock validity window error")
+)
+
+func TestProcessorExecute(t *testing.T) {
+	tests := []struct {
+		name           string
+		validityWindow chain.ValidityWindow
+		workers        workers.Workers
```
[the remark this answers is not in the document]
RodrigoVillar: Removed `workers` from tests

aaronbuchwald (commented):
Could we update the tests to make sure that a block that fails for reason X will only fail verification for reason X where possible?

We should try to make it so that if we changed the order of verification of let's say the timestamp and block height (which shouldn't change block validity at all and would be a perfectly fine although unnecessary change), the tests would ideally still pass. As the tests are currently structured, some of them would start to fail with a different reason because the test blocks are invalid for multiple reasons, when they are just trying to test for one condition.

chain/processor_test.go
```diff
@@ -0,0 +1,507 @@
...
+				return v
+			}(),
+			newBlockF: func(parentRoot ids.ID) *chain.StatelessBlock {
+				block, err := chain.NewStatelessBlock(
+					ids.Empty,
+					testRules.GetMinEmptyBlockGap(),
+					1,
+					[]*chain.Transaction{
+						func() *chain.Transaction {
+							p, err := ed25519.GeneratePrivateKey()
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,507 @@
...
+				return v
+			}(),
+			newBlockF: func(parentRoot ids.ID) *chain.StatelessBlock {
+				block, err := chain.NewStatelessBlock(
+					ids.Empty,
+					testRules.GetMinEmptyBlockGap(),
+					1,
+					[]*chain.Transaction{
+						func() *chain.Transaction {
+							tx, err := chain.NewTransaction(
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,507 @@
...
+								&chain.Base{},
+								[]chain.Action{},
+								&mockAuth{
+									typeID: 1,
+								},
+							)
+							require.NoError(t, err)
+							return tx
+						}(),
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/processor_test.go
```diff
@@ -0,0 +1,507 @@
...
+			}(),
+			newBlockF: func(root ids.ID) *chain.StatelessBlock {
+				block, err := chain.NewStatelessBlock(
+					ids.Empty,
+					testRules.GetMinEmptyBlockGap(),
+					1,
+					nil,
+					root,
+				)
+				require.NoError(t, err)
```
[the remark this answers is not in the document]
RodrigoVillar: Yes as per https://github.com/ava-labs/hypersdk/pull/1890#discussion_r1930962971; I've modified the table tests to take in the assertion from the calling subtest.

chain/processor_test.go
```diff
@@ -0,0 +1,536 @@
...
+					root,
+					&block.Context{},
+				)
+				r.NoError(err)
+				return block
+			},
+			expectedErr: errMockVerifyExpiryReplayProtection,
+		},
+		{
+			name:           "failed to execute txs",
```
aaronbuchwald, line 354: This is a good case to include. There are a few places where transaction failure should invalidate the block and propagate the error and a few where it should be treated as a failed/reverted, but valid transaction. Separate from this PR, we should cover cases like a transaction reverting is valid, whereas `StateKeys` returning invalid values, which must be caught in pre-execute invalidates the block.

chain/processor_test.go
```diff
@@ -0,0 +1,536 @@
...
+					root,
+					&block.Context{},
+				)
+				r.NoError(err)
+				return block
+			},
+			expectedErr: chain.ErrFailedToParseParentTimestamp,
+		},
+		{
+			name:           "non-empty block timestamp less than parent timestamp with gap",
```
aaronbuchwald, line 234:
```suggestion
			name:           "non-empty block - timestamp less than parent timestamp with gap",
```

chain/processor_test.go
```diff
@@ -0,0 +1,536 @@
...
+					parentRoot,
+					&block.Context{},
+				)
+				r.NoError(err)
+				return block
+			},
+			expectedErr: chain.ErrTimestampTooEarly,
+		},
+		{
+			name:           "empty block timestamp less than parent timestamp with gap",
```
aaronbuchwald, line 272:
```suggestion
			name:           "empty block - timestamp less than parent timestamp with gap",
```

chain/processor_test.go
```diff
@@ -0,0 +1,536 @@
...
+						),
+					},
+					[]chain.Action{},
+					&mockAuth{typeID: 1},
+				)
+				r.NoError(err)
+
+				block, err := chain.NewStatelessBlock(
+					ids.Empty,
+					0,
```
aaronbuchwald, line 260: Could we use `MinBlockGap() - 1` rather than 0 to make this a more explicit boundary test case?

chain/processor_test.go
```diff
@@ -0,0 +1,536 @@
...
+					timestampKey: binary.BigEndian.AppendUint64(nil, 0),
+					feeKey:       {},
+				})
+				r.NoError(err)
+				return v
+			},
+			newBlockF: func(r *require.Assertions, parentRoot ids.ID) *chain.StatelessBlock {
+				block, err := chain.NewStatelessBlock(
+					ids.Empty,
+					0,
```
aaronbuchwald, line 286: Could we make this `MinEmptyBlockGap() - 1` to make this a more explicit boundary test case?

aaronbuchwald (changes requested):

chain/processor_test.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Ack

aaronbuchwald (approved):

PR #1894, opened 2025-01-27
dsmr: replace hard coded validators with ChainState

tsachiherman:
## What ?

This PR replaces the hard coded `validators` list and few more parameters that were previously hard-coded into the node with a dynamic `ChainState` interface.

The node test is currently have an implementation for that interface, which is used during testing.

## Node creation

The node constructor which used to look like this:
```golang
func New[T Tx](
	log logging.Logger,
	nodeID ids.NodeID,
	networkID uint32,
	chainID ids.ID,
	pk *bls.PublicKey,
	signer warp.Signer,
	chunkStorage *ChunkStorage[T],
	getChunkClient *p2p.Client,
	getChunkSignatureClient *p2p.Client,
	chunkCertificateGossipClient *p2p.Client,
	validators []Validator, // TODO remove hard-coded validator set
	lastAccepted Block,
	quorumNum uint64,
	quorumDen uint64,
	timeValidityWindow TimeValidityWindow[*emapChunkCertificate],
	validityWindowDuration time.Duration,
)
```

looks now like that:
```golang
func New[T Tx](
	log logging.Logger,
	nodeID ids.NodeID,
	chainState ChainState,
	pk *bls.PublicKey,
	signer warp.Signer,
	chunkStorage *ChunkStorage[T],
	getChunkClient *p2p.Client,
	getChunkSignatureClient *p2p.Client,
	chunkCertificateGossipClient *p2p.Client,
	lastAccepted Block,
	timeValidityWindow TimeValidityWindow[*emapChunkCertificate],
	validityWindowDuration time.Duration,
)
```

and the `ChainState` is defined as : 
```golang
type ChainState interface {
	GetNetworkParams() (networdID uint32, subnetID, chainID ids.ID)
	GetSignatureParams(ctx context.Context) (validators warp.CanonicalValidatorSet, quorumNum uint64, quorumDen uint64, err error)
	GetValidatorSet(ctx context.Context) (map[ids.NodeID]*snowValidators.GetValidatorOutput, error)
}
```


x/dsmr/node.go
```diff
@@ -50,6 +49,12 @@ var (
 	ErrFailedToReplicate                   = errors.New("failed to replicate to sufficient stake")
 )
 
+type ChainState interface {
+	GetNetworkParams() (networdID uint32, subnetID, chainID ids.ID)
+	GetSignatureParams(ctx context.Context) (validators warp.CanonicalValidatorSet, quorumNum uint64, quorumDen uint64, err error)
+	GetLatestValidatorSet(ctx context.Context) (map[ids.NodeID]*snowValidators.GetValidatorOutput, error)
+}
```
aaronbuchwald, line 56: Can we break this into individual accessors rather than grouping them based on what we will need at different points? This way when we need `networkID` and `chainID`, we can access those two rather than grouping all the network params defined by this function together.

In the future, we may want to make the rules configurable based on the block timestamp and fetch the validator set based off of the P-Chain height (both fine to leave as is for now). WIP for adding ProposerVM block context into the block is here: https://github.com/ava-labs/hypersdk/pull/1897.

For now, can we break this up to fetch the immutable params individually ie. `networkID`, `subnetID`, `chainID`, `quorumNum`, and `quorumDenom`. Can we also switch to two functions to fetch the canonical validator set and check whether a specific nodeID is a validator (can add P-Chain height as a parameter in the future)?

aaronbuchwald (changes requested):

tests/e2e/e2e.go
```diff
@@ -182,7 +182,8 @@ var _ = ginkgo.Describe("[HyperSDK Syncing]", ginkgo.Serial, func() {
 		})
 		ginkgo.By("Pause the node", func() {
 			// TODO: remove the need to call SaveAPIPort from the test
-			require.NoError(syncNode.SaveAPIPort())
+			// Tsachi : The following line was commented out, as it's no longer supported by avalanchego.
+			// require.NoError(syncNode.SaveAPIPort())
```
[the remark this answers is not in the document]
tsachiherman: I was trying to figure out if this is no longer needed. Can you take a look https://github.com/ava-labs/avalanchego/pull/3650/files and let me know what you think ?

x/dsmr/node.go
```diff
@@ -372,7 +356,11 @@ func (n *Node[T]) Accept(ctx context.Context, block Block) (ExecutedBlock[T], er
 				}
 
 				// TODO better request strategy
-				nodeID := n.validators[rand.Intn(len(n.validators))].NodeID //nolint:gosec
+				validators, _, _, err := n.chainState.GetSignatureParams(ctx)
+				if err != nil {
+					return ExecutedBlock[T]{}, fmt.Errorf("failed to retrieve validators list: %w", err)
+				}
+				nodeID := validators.Validators[rand.Intn(len(validators.Validators))].NodeIDs[0] //nolint:gosec
```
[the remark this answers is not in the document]
tsachiherman: no ;-(
I think that this entire code potion need to be refactored so that it would attempt to fetch, in parallel, all the chunks from multiple validators. I'm not sure if using the stake is wise, as it would not scale well.

Given that more work in needed here, I had opted into avoiding making any further interface decisions here.

examples/morpheusvm/go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/ava-labs/hypersdk/examples/morpheusvm
 go 1.22.8
 
 require (
-	github.com/ava-labs/avalanchego v1.11.13-0.20241230212828-6dea1b366756
-	github.com/ava-labs/hypersdk v0.0.1
+	github.com/ava-labs/avalanchego v1.12.3-warp-verify5.0.20250129181352-a1aff06d0fef
```
joshua-kim, line 6: Why do we have to update avalanchego to use the new `localsigner`? I think we can replace the usage of the hard-coded validator set w/out a version upgrade

x/dsmr/node.go
```diff
@@ -50,6 +49,12 @@ var (
 	ErrFailedToReplicate                   = errors.New("failed to replicate to sufficient stake")
 )
 
+type ChainState interface {
+	GetNetworkParams() (networdID uint32, subnetID, chainID ids.ID)
+	GetSignatureParams(ctx context.Context) (validators warp.CanonicalValidatorSet, quorumNum uint64, quorumDen uint64, err error)
```
joshua-kim, line 54: Why are we making this a part of this interface? If it's to support forking live networks I think this is a premature abstraction and as a tradeoff it bloats this interface. IMO this interface should just return validator sets

x/dsmr/node_test.go
```diff
@@ -1520,3 +1510,55 @@ func newTestNodes(t *testing.T, n int) []*Node[dsmrtest.Tx] {
...
 }
+
+type testChainState struct {
+	validators []Validator
+	quorumNum  uint64
+	quorumDen  uint64
+}
+
+func (*testChainState) GetNetworkParams() (uint32, ids.ID, ids.ID) {
+	return networkID /*subnetID*/, ids.Empty, chainID
```
joshua-kim, line 1521: nit: don't leave these comments

x/dsmr/node_test.go
```diff
@@ -1520,3 +1510,55 @@ func newTestNodes(t *testing.T, n int) []*Node[dsmrtest.Tx] {
...
+		) (map[ids.NodeID]*validators.GetValidatorOutput, error) {
+			return t.getValidatorSet(), nil
+		},
+	}
+
+	canonicalValidators, err := warp.GetCanonicalValidatorSetFromSubnetID(
+		ctx,
+		pChain,
+		0,
+		ids.Empty, /*subnetID*/
```
joshua-kim, line 1554: same nit here

x/dsmr/node_test.go
```diff
@@ -1520,3 +1510,55 @@ func newTestNodes(t *testing.T, n int) []*Node[dsmrtest.Tx] {
...
+	validators []Validator
+	quorumNum  uint64
+	quorumDen  uint64
+}
+
+func (*testChainState) GetNetworkParams() (uint32, ids.ID, ids.ID) {
+	return networkID /*subnetID*/, ids.Empty, chainID
+}
+
+func (t *testChainState) getValidatorSet() map[ids.NodeID]*validators.GetValidatorOutput {
```
joshua-kim, line 1524: I feel like this introduces complexity, can we just cache this in the `testChainState` struct instead of having to re-call this function every time you want to get the validator set

joshua-kim (changes requested):

x/dsmr/node.go
tsachiherman: done

examples/morpheusvm/go.mod
tsachiherman: It's not about the upgrade to `localsigner`. The goal here was to be able to verify the signature `c.Signature.Verify` without providing it with pseudo interface. ( i.e. provide it with the data it truly needs to consume ). see https://github.com/ava-labs/avalanchego/pull/3679 for more details.

x/dsmr/node.go
joshua-kim: If we want to support live networks won't we need some timestamp/block height parameter in this interface then?

x/dsmr/node_test.go
tsachiherman: The method is being used only in a single place ( within `GetCanonicalValidatorSet` ), and exclusively used for testing purposes. It doesn't seems that we would gain much by caching the map here. We could use a map as the backing storage ( instead of a slice ).

tests/e2e/e2e.go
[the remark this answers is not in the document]
aaronbuchwald: Hmm it looks like with the new behavior the URI to access the node may change, so this does not look correct to me.

In the test above, we have saved the `syncNodeURI` and added it to `uris`, which we use throughout the tests below, so if that URI is not guaranteed to be reachable after the restart we may fail.

If AvalancheGo deprecated support for this feature, then we'll need to make sure to either switch away from saving them to `uris` and `syncNodeURI` and creating them on demand from the updated node instance or update those variables in place after the restart.

x/dsmr/node_test.go
```diff
@@ -1526,6 +1516,79 @@ func newTestNodes(t *testing.T, n int) []*Node[dsmrtest.Tx] {
...
+		GetValidatorSetF: func(
+			context.Context,
+			uint64,
+			ids.ID,
+		) (map[ids.NodeID]*validators.GetValidatorOutput, error) {
+			return t.getValidatorSet(), nil
+		},
+	}
+
+	canonicalValidators, err := warp.GetCanonicalValidatorSetFromSubnetID(
```
aaronbuchwald, line 1571: if we need this, can we include it in the `testChainState` type rather than constructing it on demand in this function?

x/dsmr/node.go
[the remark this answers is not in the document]
aaronbuchwald: Makes sense

tests/e2e/e2e.go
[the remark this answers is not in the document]
aaronbuchwald: This seems to explain the failing e2e sync test: https://github.com/ava-labs/hypersdk/actions/runs/13078267879/job/36495962625?pr=1894#step:4:312

x/dsmr/node_test.go
tsachiherman: sure. it would require a constructor, but no prob.

x/dsmr/node_test.go
```diff
@@ -1532,6 +1504,86 @@ func newTestNodes(t *testing.T, n int) []*Node[dsmrtest.Tx] {
...
 }
 
+type testChainState struct {
+	validatorstest.State
+	validators []Validator
+	quorumNum  uint64
+	quorumDen  uint64
+}
+
+func newTestChainState(validatorsSlice []Validator, quorumNum, quorumDen uint64) *testChainState { //nolint:unparam
```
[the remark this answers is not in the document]
tsachiherman: in the test, we only use (1,1) for the quorum params. Its ok, since we don't really want to test that, but the linter alert anyway.

aaronbuchwald (approved):

PR #1895, opened 2025-01-27
Move timestamp verification into validity window package

aaronbuchwald:
This PR moves timestamp verification checks into the time validity window package. Same code will be shared across tx and chunks.

### DIFF
--- a/internal/validitywindow/validitywindow.go
+++ b/internal/validitywindow/validitywindow.go
@@ -16,7 +16,6 @@ import (
 	"github.com/ava-labs/avalanchego/utils/set"
 	"go.uber.org/zap"
 
-	"github.com/ava-labs/hypersdk/consts"
 	"github.com/ava-labs/hypersdk/internal/emap"
 )
 
@@ -184,11 +183,10 @@ func (v *TimeValidityWindow[T]) calculateOldestAllowed(timestamp int64) int64 {
 	return max(0, timestamp-v.getTimeValidityWindow(timestamp))
 }
 
-func VerifyTimestamp(containerTimestamp int64, executionTimestamp int64, validityWindow int64) error {
+func VerifyTimestamp(containerTimestamp int64, executionTimestamp int64, divisor int64, validityWindow int64) error {
 	switch {
-	case containerTimestamp%consts.MillisecondsPerSecond != 0:
-		// TODO: make this modulus configurable
-		return fmt.Errorf("%w: timestamp=%d", ErrMisalignedTime, containerTimestamp)
+	case containerTimestamp%divisor != 0:
+		return fmt.Errorf("%w: timestamp (%d) %% divisor (%d) != 0", ErrMisalignedTime, containerTimestamp, divisor)
 	case containerTimestamp < executionTimestamp: // expiry: 100 block: 110
 		return fmt.Errorf("%w: timestamp (%d) < block timestamp (%d)", ErrTimestampExpired, containerTimestamp, executionTimestamp)
 	case containerTimestamp > executionTimestamp+validityWindow: // expiry: 100 block 10
--- a/internal/validitywindow/validitywindow.go
+++ b/internal/validitywindow/validitywindow.go
@@ -16,12 +16,16 @@ import (
 	"github.com/ava-labs/avalanchego/utils/set"
 	"go.uber.org/zap"
 
+	"github.com/ava-labs/hypersdk/consts"
 	"github.com/ava-labs/hypersdk/internal/emap"
 )
 
 var (
 	_                     Interface[emap.Item] = (*TimeValidityWindow[emap.Item])(nil)
 	ErrDuplicateContainer                      = errors.New("duplicate container")
+	ErrMisalignedTime                          = errors.New("misaligned time")
+	ErrTimestampExpired                        = errors.New("declared timestamp expired")
+	ErrFutureTimestamp                         = errors.New("declared timestamp too far in the future")
 )
 
 type GetTimeValidityWindowFunc func(timestamp int64) int64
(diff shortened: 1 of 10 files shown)


x/dsmr/node_test.go
```diff
@@ -511,12 +516,12 @@ func TestNode_GetChunkSignature_SignValidChunk(t *testing.T) {
...
 				1,
 				1,
-				testingDefaultValidityWindowDuration,
+				testRuleFactory,
 			),
 			wantErr:                   ErrInvalidChunk,
 			producerNode:              nodeID,
 			nodeLastAcceptedTimestamp: 1,
-			chunkExpiry:               1 + int64(testingDefaultValidityWindowDuration),
+			chunkExpiry:               2 + int64(testingDefaultValidityWindowDuration),
```
tsachiherman, line 524: not opposing to, but - was this change intentional ?

internal/validitywindow/validitywindow.go
```diff
@@ -179,3 +182,16 @@ func (v *TimeValidityWindow[T]) isRepeat(
...
 	return max(0, timestamp-v.getTimeValidityWindow(timestamp))
 }
+
+func VerifyTimestamp(containerTimestamp int64, executionTimestamp int64, divisor int64, validityWindow int64) error {
+	switch {
+	case containerTimestamp%divisor != 0:
+		return fmt.Errorf("%w: timestamp (%d) %% divisor (%d) != 0", ErrMisalignedTime, containerTimestamp, divisor)
+	case containerTimestamp < executionTimestamp: // expiry: 100 block: 110
+		return fmt.Errorf("%w: timestamp (%d) < block timestamp (%d)", ErrTimestampExpired, containerTimestamp, executionTimestamp)
+	case containerTimestamp > executionTimestamp+validityWindow: // expiry: 100 block 10
```
joshua-kim, line 192: These were in the old code but I feel like we should delete these comments as part of this diff anyways

x/dsmr/node_test.go
joshua-kim, line 524: Why is this diff needed?

internal/validitywindow/validitywindow.go
```diff
@@ -184,11 +183,10 @@ func (v *TimeValidityWindow[T]) calculateOldestAllowed(timestamp int64) int64 {
 	return max(0, timestamp-v.getTimeValidityWindow(timestamp))
 }
 
-func VerifyTimestamp(containerTimestamp int64, executionTimestamp int64, validityWindow int64) error {
+func VerifyTimestamp(containerTimestamp int64, executionTimestamp int64, divisor int64, validityWindow int64) error {
```
tsachiherman, line 186: the term `divisor` is mathematically correct, but a bit misleading, imho. I'd opt to use the term `gradient` or `frame` here.

joshua-kim: `gradient` doesn't make sense to me IMO...  I also don't like the `divisor` name but I don't have a good concrete suggestion... I would probably call this something like `boundary` but I also feel like this name is not amazing

internal/validitywindow/validitywindow.go
aaronbuchwald: I find these helpful, would prefer not to delete. Would you prefer they be moved in some way?

internal/validitywindow/validitywindow.go
aaronbuchwald: divisor is the right term, old code had referred to this as `modulus` which I don't think was right.

Happy to change if there's a better term, otherwise will leave as is since it seems the consensus here is that we don't love the term, but it is the most correct, and we don't have a better one. I do agree with both of you though haha.

internal/validitywindow/validitywindow.go
joshua-kim: IMO it's unnecessary to comment to explain how `<` and `>` work, but if you find it helpful feel free free to leave it in

x/dsmr/node_test.go
aaronbuchwald: Yes, this is needed to get the boundary test case passing and align with the behavior of the time validity window. These two were slightly different in how they enforced it.

Previous boundary verification: https://github.com/ava-labs/hypersdk/blob/main/x/dsmr/storage.go#L80
Tx boundary verification: https://github.com/ava-labs/hypersdk/blob/main/chain/base.go#L39

When we call `SetMin` on the validity window for timestamp `t`, we will clear out all containers that are strictly less than timestamp `t`: https://github.com/ava-labs/hypersdk/blob/main/internal/emap/emap.go#L101. This means that allowing a transaction in at the forward boundary ie. `blkTimestamp + validityWindow` at `blkTimestamp`, the transaction will be screened out and marked as invalid in any future block with a timestamp up to and including `blkTimestamp + validityWindow`. After that, it will be marked as invalid because it has expired.

I'll write up a comment and add a test case inside the `validitywindow` package.

aaronbuchwald: https://github.com/ava-labs/hypersdk/pull/1895#discussion_r1932877563

aaronbuchwald: Added a test here: https://github.com/ava-labs/hypersdk/pull/1895/commits/3278211f75b9e8963c5d9f4a41c9e9fa42a736f9

The previous code in chunk verification was overly defensive - it would mark a chunk as invalid if it was at the validity window duration in the future. This would be required if the emap removed any containers <= the current minimum timestamp.

internal/validitywindow/validitywindow.go
joshua-kim: Looks like we're now referring this as boundary anyways in this test, should we rename? https://github.com/ava-labs/hypersdk/pull/1895/commits/3278211f75b9e8963c5d9f4a41c9e9fa42a736f9

internal/validitywindow/validitywindow_test.go
```diff
@@ -298,6 +298,43 @@ func TestValidityWindowIsRepeat(t *testing.T) {
...
+	// and accepted parent.
+	// Processing:
+	r.ErrorIs(validityWindow.VerifyExpiryReplayProtection(context.Background(), blk2), ErrDuplicateContainer)
+
+	// Accepted:
+	validityWindow.Accept(blk1)
+	r.ErrorIs(validityWindow.VerifyExpiryReplayProtection(context.Background(), blk2), ErrDuplicateContainer)
+
+	// Verify that after passing the validity window, the timestamp is no longer valid
+	r.ErrorIs(VerifyTimestamp(validityWindowDuration, validityWindowDuration+1, 1, validityWindowDuration), ErrTimestampExpired)
```
joshua-kim, line 335: All of these look like separate test cases to me, can we make them separate tests? I don't think this is going to work well in a table-test which is why I suspect this test was written this way, but the cost of copy-and-paste is not that bad and I would prefer we did that over having a single monolithic test that tests multiple cases.

aaronbuchwald: I thought of this as the lifespan of a single tx being invalidated, but can break it up / see about making it a table test

aaronbuchwald: Added a separate table test for `VerifyTimestamp` and kept this test showing the lifespan of the tx (with an added comment).

The combination of the newly added test and the existing `VerifyExpiryReplayProtection` tests provides the same coverage (already covers processing/accepted at the boundary). Happy to remove this test if you prefer.

aaronbuchwald: https://github.com/ava-labs/hypersdk/pull/1895/commits/cef760574f00cc34eed55a9c9063090761d75f4b

internal/validitywindow/validitywindow.go
aaronbuchwald: I don't see any place in the added test where we are referring to the divisor as boundary. In that test, we are passing in `1` (ie. there's no such thing as a misaligned timestamp) in each instance.

tsachiherman (approved):

PR #1897, opened 2025-01-29
Add P-Chain ctx into inner block and verify within snow pkg

aaronbuchwald:
This PR adds the P-Chain context into the inner block.

This embeds it directly into the inner block, so that it's always available to the inner block. By adding this into the inner block interface, we can verify it against the consensus context in the snow package, so that the inner VM/chain does not need to.

### DIFF
--- a/chain/stateless_block.go
+++ b/chain/stateless_block.go
@@ -21,7 +21,7 @@ type StatelessBlock struct {
 	Tmstmp int64  `json:"timestamp"`
 	Hght   uint64 `json:"height"`
 
-	BlockContext *block.Context `json:"pChainHeight"`
+	BlockContext *block.Context `json:"blockContext"`
 
 	Txs []*Transaction `json:"txs"`
 
--- a/chain/stateless_block.go
+++ b/chain/stateless_block.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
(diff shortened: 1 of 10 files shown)


aaronbuchwald:
https://github.com/ava-labs/hypersdk/pull/1897/commits/b4ec9ce1ac14544ba9ef50ee777e983d2eb4a4cc fixes a bug where the block implements `block.WithVerifyContext`, but the VM did not implement `block.BuildBlockWithContextChainVM` so we'd build a block without the context and verify it with the P-Chain context.

Needed to add the covariant version of `BuildBlockWithContext` to implement the interface correctly and fixes the issue.

chain/stateless_block.go
```diff
@@ -20,6 +21,8 @@ type StatelessBlock struct {
 	Tmstmp int64  `json:"timestamp"`
 	Hght   uint64 `json:"height"`
 
+	BlockContext *block.Context `json:"pChainHeight"`
```
RodrigoVillar, line 24: Should we change the struct tag from `json:"pChainHeight"` to `json:"blockContext"`?

tsachiherman: same thought here. The inner BlockContext object would already have a json field name `PChainHeight`.

aaronbuchwald: Ya good callout

aaronbuchwald: It only contains P-Chain height, but this would be better

tsachiherman (approved):

Issue #1878, opened 2025-01-20
Add State Sync Transition Health Check to Snow VM


aaronbuchwald:
When transitioning from state sync to normal operation, the VM is in a position where it may have outstanding processing blocks that were vacuously verified (marked as verified successfully even though the VM did not have the information it needed to properly verify it ie. the current state).

We must also synchronize between the thread where state sync finishes and consensus (single threaded), which will continue to verify/accept/reject blocks in dynamic state sync mode until we mark the VM as ready.

When state sync finishes, it will complete at some block N (presumably the last accepted block of the network). Once it finishes, it will try to mark the VM as finished state syncing at that block, but if consensus is already holding the `chainLock` to accept a block, it's possible that state sync finishes with the state at block N, but observes that consensus actually ends up with a last accepted block of N+K when it successfully grabs the lock to mark itself as ready.

Additionally, when the merkle syncer terminates it will ignore future updates meaning that if it finishes before the block window syncer (which may occur on either small state or minimally active networks), then it may finish on an old block and have to re-process more blocks when the block window syncer completes.

To handle any difference between the actual consensus tip and the block that state sync finishes on, we re-process blocks to re-calculate the state of the last accepted block:

https://github.com/ava-labs/hypersdk/blob/main/snow/statesync.go#L59

We then re-process any currently processing blocks (verified, but not accepted or rejected yet):

https://github.com/ava-labs/hypersdk/blob/main/snow/statesync.go#L67

This ensures that after transitioning to normal operation, the output type is populated for every valid block in consensus and both the output and accepted types are populated for the last accepted block. This removes the need to handle recursively verifying/accepting blocks in the block code itself. Instead, if `parent.Output` is not populated, then we assume the parent is invalid and all of its children are considered invalid as well. We rely on consensus and a correct validator set to guarantee that if we allowed an invalid block into consensus during dynamic state sync, consensus will eventually reject it in favor of a valid block.

In the meantime, `GetPreferredBlock` needs to handle the case that the currently preferred block is invalid and not populated:

https://github.com/ava-labs/hypersdk/blob/main/snow/chain_index.go#L136

This ticket is to add a health check that will report unhealthy until dynamic state sync has completed and any such processing, but invalid blocks have been accepted/rejected.

PR #1898, opened 2025-01-29
feat(snow/vm): health check

Elvis339:
Closes: https://github.com/ava-labs/hypersdk/issues/1878

This patch set introduces a new health check mechanism for the Snow VM, which monitors VM readiness and tracks vacuously verified blocks.


snow/health.go
```diff
@@ -0,0 +1,130 @@
...
+// The health checker monitors blocks that were vacuously verified during state sync and reports unhealthy status if any remain unresolved
+//
+// Safety guarantee: We rely on consensus and correct validator set that any invalid blocks
+// accepted during state sync will eventually be rejected in favor of valid blocks.
+type UnresolvedBlocksCheck[I Block] struct {
+	unresolvedBlocks set.Set[ids.ID]
+	lock             sync.RWMutex
+}
+
+func NewUnresolvedBlocksCheck[I Block]() *UnresolvedBlocksCheck[I] {
```
joshua-kim, line 93: I don't think these types are necessary as part of the exported package api

snow/health.go
```diff
@@ -0,0 +1,130 @@
...
+// UnresolvedBlocksCheck
+// During state sync, blocks are vacuously marked as verified because the VM is missing the current state.
+// Consensus will eventually reject any invalid blocks, but this check ensures the VM waits to report healthy until it has cleared all invalid blocks from the processing set.
+// The health checker monitors blocks that were vacuously verified during state sync and reports unhealthy status if any remain unresolved
+//
+// Safety guarantee: We rely on consensus and correct validator set that any invalid blocks
+// accepted during state sync will eventually be rejected in favor of valid blocks.
+type UnresolvedBlocksCheck[I Block] struct {
+	unresolvedBlocks set.Set[ids.ID]
+	lock             sync.RWMutex
```
joshua-kim, line 90: Does this have to be safe against concurrent use? I thought that the event types had `Notify` called in serial... but I don't recall for `HealthCheck`...

snow/health.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func NewVMReadiness(isReady func() bool) *VMReadiness {
+	return &VMReadiness{isReady: isReady}
+}
+
+func (v *VMReadiness) HealthCheck(_ context.Context) (interface{}, error) {
+	ready := v.isReady()
+	if !ready {
+		return nil, errors.New("vm is not ready")
```
joshua-kim, line 72: Should we return this map in both cases but change the value to `false` if it's not ready?

snow/statesync.go
```diff
@@ -91,6 +92,15 @@ func (v *VM[I, O, A]) verifyProcessingBlocks(ctx context.Context) error {
 			return 1
 		}
 	})
+	healthChecker, err := v.HealthChecker(UnresolvedBlocksHealthChecker)
+	if err != nil {
+		return err
+	}
+
+	unresolvedBlkHealthChecker, typeConvOK := healthChecker.(*UnresolvedBlocksCheck[I])
```
joshua-kim, line 100: nit: can we just call this `ok`? This is idiomatic in Go

snow/vm.go
```diff
@@ -438,7 +444,34 @@ func (v *VM[I, O, A]) SetState(ctx context.Context, state snow.State) error {
 }
 
 func (v *VM[I, O, A]) HealthCheck(ctx context.Context) (interface{}, error) {
-	return v.healthChecker.HealthCheck(ctx)
+	var (
+		detailsMu sync.Mutex
+		details   = make(map[string]interface{})
+		errsMu    sync.Mutex
+		errs      []error
```
joshua-kim, line 451: Why do we need locks here? `syncmap.Range` is a sequential operation

snow/vm.go
```diff
@@ -143,16 +144,15 @@ type VM[I Block, O Block, A Block] struct {
 	tracer  trace.Tracer
 
 	shutdownChan chan struct{}
+
+	healthCheckers sync.Map
```
joshua-kim, line 148: Trying to figure out if we need a `sync.Map` here

snow/health.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// VMReadiness is concrete health check that should mark itself ready if isReady function returns true
+type VMReadiness struct {
+	isReady func() bool
+}
+
+func NewVMReadiness(isReady func() bool) *VMReadiness {
+	return &VMReadiness{isReady: isReady}
+}
```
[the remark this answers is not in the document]
Elvis339: I initially passed the VM instance since it simplified the implementation. When Aaron noted that unresolvedBlockHealthCheck shouldn't reference VM, I removed it from vmReadiness too. I kind of agree with Aaron because adding VM generics seems like over engineering for a basic health check. IMO there's a bit of decoupling benefit of keeping this health check generic-free.

snow/health.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/ava-labs/avalanchego/api/health"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/set"
+
+	"github.com/ava-labs/hypersdk/event"
+)
+
+const (
+	VMReadinessHealthChecker      = "snowVMReadiness"
+	UnresolvedBlocksHealthChecker = "snowUnresolvedBlocks"
```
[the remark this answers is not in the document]
Elvis339: Re on point 1.
The naming convention [here](https://github.com/ava-labs/hypersdk/blob/4915e261c632cca2da19c6e47c9960c972c5422d/vm/statesync.go#L23) is flatcase. I suggest using camelCase since heath checks responses typically get converted to JSON for structured logging in monitoring systems. [AvalancheGo](https://github.com/ava-labs/avalanchego/blob/f5feb053cccb7173ab92b35b1b2a07727b0de463/snow/consensus/snowman/topological.go#L311) is also following camelCase naming convention.

Re on point 2.
Updated.

snow/health.go
```diff
@@ -0,0 +1,130 @@
...
+func (v *VM[I, O, A]) RegisterHealthChecker(name string, healthChecker health.Checker) error {
+	if _, loaded := v.healthCheckers.LoadOrStore(name, healthChecker); loaded {
+		return fmt.Errorf("duplicate health checker for %s", name)
+	}
+
+	return nil
+}
+
+func (v *VM[I, O, A]) HealthChecker(name string) (interface{}, error) {
+	value, ok := v.healthCheckers.Load(name)
```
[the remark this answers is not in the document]
Elvis339: Agree, thanks for pointing it out.

snow/health.go
```diff
@@ -0,0 +1,130 @@
...
+type UnresolvedBlocksCheck[I Block] struct {
+	unresolvedBlocks set.Set[ids.ID]
+	lock             sync.RWMutex
+}
+
+func NewUnresolvedBlocksCheck[I Block]() *UnresolvedBlocksCheck[I] {
+	return &UnresolvedBlocksCheck[I]{
+		unresolvedBlocks: set.Set[ids.ID]{},
+	}
+}
```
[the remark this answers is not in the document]
Elvis339: Right, it's not. Updated, thanks.

snow/health.go
```diff
@@ -0,0 +1,130 @@
...
+// VMReadiness is concrete health check that should mark itself ready if isReady function returns true
+type VMReadiness struct {
+	isReady func() bool
+}
+
+func NewVMReadiness(isReady func() bool) *VMReadiness {
+	return &VMReadiness{isReady: isReady}
+}
+
+func (v *VMReadiness) HealthCheck(_ context.Context) (interface{}, error) {
```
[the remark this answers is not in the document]
Elvis339: Thanks, I replaced `interface{}` with `any`.

snow/vm.go
Elvis339: Looking at AvalancheGo [health package doc](https://github.com/ava-labs/avalanchego/tree/master/api/health#health-check-worker) it seems like we need it.

snow/vm.go
Elvis339: It is, removed, thanks for pointing it out.

snow/statesync.go
Elvis339: This has been removed.

snow/health.go
Elvis339: We do need sync primitive: https://github.com/ava-labs/avalanchego/tree/master/api/health#health-check-worker

snow/vm.go
joshua-kim: The link you described says that health checks run in their own goroutine, [ref](https://github.com/ava-labs/avalanchego/blob/576b3927c54b0e7ad91602572f4aa8f9fe24d768/api/health/worker.go#L215), but there aren't multiple concurrent health checks for a single type from what I can tell... it just means that multiple vms can have their health checks run concurrently, but each batch of health checks occurs in serial (i.e that a vm won't expect to have more than one health check occurring concurrently).

aaronbuchwald: This enables us to register health checks when AvalancheGo may call `HealthCheck` concurrently.

aaronbuchwald (approved):

snow/vm.go
joshua-kim: I don't think avalanchego calls `HealthCheck` concurrently in the way that this code is trying to guard against, refer to code I linked. VM have health checks are concurrent (i.e multiple vms can run health checks concurrently), but each VM's health check will only run once the previous one has completed.

PR #1901, opened 2025-01-30
Upgrade avalanchego dependency

tsachiherman:
## What ?

Update the avalanchego with the latest changes.

Include few fixes for the e2e testing, and in particular, avoid caching the URIs to prevent stale URIs.


x/dsmr/certificate.go
```diff
@@ -89,13 +89,11 @@ func (c *ChunkCertificate) Verify(
 	if err != nil {
 		return fmt.Errorf("failed to initialize unsigned warp message: %w", err)
 	}
-
+	vs, _ := warp.GetCanonicalValidatorSetFromChainID(ctx, pChainState, pChainHeight, msg.SourceChainID)
```
aaronbuchwald, line 92: Can we check the error value here?

x/dsmr/node.go
```diff
@@ -453,3 +453,10 @@ func (p pChain) GetValidatorSet(context.Context, uint64, ids.ID) (map[ids.NodeID
 
 	return result, nil
 }
+
+func (pChain) GetCurrentValidatorSet(
+	context.Context,
+	ids.ID,
+) (map[ids.ID]*snowValidators.GetCurrentValidatorOutput, uint64, error) {
+	return nil, 0, nil
+}
```
aaronbuchwald, line 462: why does this return empty?

tsachiherman: The method is currently not being used. I'll add an error here to indicate that.
( it will be removed in my subsequent PR ).

x/dsmr/certificate.go
```diff
@@ -89,13 +89,14 @@ func (c *ChunkCertificate) Verify(
 	if err != nil {
 		return fmt.Errorf("failed to initialize unsigned warp message: %w", err)
 	}
-
+	vs, err := warp.GetCanonicalValidatorSetFromChainID(ctx, pChainState, pChainHeight, msg.SourceChainID)
```
aaronbuchwald, line 92: Can we use a more descriptive variable name here, perhaps `canonicalValidatorSet` ?

aaronbuchwald (approved):

aaronbuchwald (approved):

Issue #1904, opened 2025-01-30
Random Block Compaction Offset Can Result in Divide by Zero


aaronbuchwald:
https://github.com/ava-labs/hypersdk/actions/runs/13060210653/job/36441114514?pr=1903#step:4:26

<img width="1057" alt="Image" src="https://github.com/user-attachments/assets/91bf213e-6e0d-4345-bfb3-7d8a35a75237" />

The chain index calls `Compact` at a regular `compactionOffset` interval when accepting blocks:

https://github.com/ava-labs/hypersdk/blob/39635525bed92ff2feb91295e4eb4d170dcf50af/chainindex/chain_index.go#L140

The compaction offset is calculated randomly here: https://github.com/ava-labs/hypersdk/blob/39635525bed92ff2feb91295e4eb4d170dcf50af/chainindex/chain_index.go#L85 and can be 0 with non-zero probability which will trigger a panic as seen in the test.

For context, we calculate a random offset locally to ensure that different nodes on the network trigger compaction at different points in time rather than all compacting simultaneously, which could trigger a network wide delay.

PR #1905, opened 2025-01-30
Fix chain index compaction offset

aaronbuchwald:
Fixes https://github.com/ava-labs/hypersdk/issues/1904

This also fixes an issue where the chain index compacted at an incorrect interval (randomly calculated frequency between every block and the actual requested block compaction frequency instead of the requested block compaction frequency).


RodrigoVillar (approved):
LGTM

PR #1914, opened 2025-02-04
Remove contracts

aaronbuchwald:
This PR removes contracts from the HyperSDK, which will no longer be supported.


aaronbuchwald:
CI actions are not running because it's using the GH Actions triggers from the base branch.

PR #1916, opened 2025-02-04
dsmr: implement chunk rate limiter

tsachiherman:
## What ? 
This PR adds chunk rate limiting.

## How ?
The `rules` contains now two additional methods:
```golang
GetChunkRateLimitingWindow() int64
GetAccumulatedChunkWeightWindowLimit() uint64
```

The `storage` have an additional function named `CheckRateLimit` that checks if adding a given chunk would breach the rate limit threshold in the following way:
- assume that we have a sorted slice of chunks, ordered by their expiry.
- within that slice, the accumulated weight of any contagious range of chunks within the `GetChunkRateLimitingWindow` range is always less than `GetAccumulatedChunkWeightWindowLimit`.
- If adding that chunk would result in a subset of chunks that meets the window range and exceed the limit, an error would be return.

## Where ?
The new `CheckRateLimit` is called from two places:
- During the ChunkBuilding, right after the chunk is being created, the new chunk is being validated. This would ensure that an honest node wouldn't generate a chunk that would exceed the rate limit threshold.
- During `ChunkSignatureRequestVerifier`, a call was added to ensure that inbound requests would not result in adding a chunk that would exceed the limit.


x/dsmr/node.go
```diff
@@ -183,6 +185,9 @@ func (n *Node[T]) BuildChunk(
 		// we have duplicates
 		return ErrDuplicateChunk
 	}
+	if err := n.storage.CheckRateLimit(chunk); err != nil {
+		return fmt.Errorf("failed to meet chunk rate limits threshold : %w", err)
+	}
```
aaronbuchwald, line 190: hmm makes sense that we rate limit ourselves and fail early rather than broadcast to peers that we know will rate limit us

x/dsmr/storage.go
```diff
@@ -168,7 +178,18 @@ func (s *ChunkStorage[T]) init() error {
...
+		s.pendingChunksProducers[chunk.Producer] = append(s.pendingChunksProducers[chunk.Producer], storedChunkSig)
+	}
+
+	// sort all the producer's chunks.
+	// no need to write back to the pendingChunksProducers since all the slice
+	// changes are in-place.
+	for _, chunks := range s.pendingChunksProducers {
+		sort.Slice(chunks, func(i, j int) bool {
+			return chunks[i].Chunk.Expiry < chunks[j].Chunk.Expiry
+		})
```
aaronbuchwald, line 192: Would it make sense to use the `ExpiryHeap` data structure here?

x/dsmr/storage.go
```diff
@@ -326,6 +378,46 @@ func (s *ChunkStorage[T]) GetChunkBytes(expiry int64, chunkID ids.ID) ([]byte, e
...
+	for i := 0; i < len(producerChunks); i++ {
+		accumulatedWeight := len(producerChunks[i].Chunk.bytes)
+		for j := i + 1; j < len(producerChunks) && producerChunks[i].Chunk.Expiry+window > producerChunks[j].Chunk.Expiry; j++ {
+			accumulatedWeight += len(producerChunks[j].Chunk.bytes)
+		}
+		if uint64(accumulatedWeight)+weight > weightLimit {
+			return ErrChunkRateLimitSurpassed
+		}
+	}
+	return nil
```
aaronbuchwald, line 418: Is it necessary to compute these on demand?

x/dsmr/storage.go
tsachiherman: I think that using a heap here isn't really what we're after, as we want to be able to dynamically "peek" into random windows. What we could do, is to store a slice of `StoredChunkWindow` as
```golang
type StoredChunkWindow struct {
    *StoredChunkSignature[T]
    windowWeight uint64
}
```
this would reduce the O(n^2) complexity in `CheckRateLimit` into `O(n)`.

Unfortunately, this is more theoretical than practical. Given a reasonable time limits on the time validity window, the number of chunks that would be in question above would be quite small, making this optimization less meaningful.

x/dsmr/storage.go
tsachiherman: see my response above..

x/dsmr/storage_test.go
```diff
@@ -369,3 +373,192 @@ func TestRestartSavedChunks(t *testing.T) {
...
+	require.NotContains(storage.pendingChunksProducers, chunkProducersNodes[1])
+
+	// delete some of the chunks for a specific producer.
+	require.NoError(storage.SetMin(6, []ids.ID{}))
+	require.Len(storage.pendingChunksProducers[chunkProducersNodes[2]], 2)
+	require.Equal(int64(6), storage.pendingChunksProducers[chunkProducersNodes[2]][0].Chunk.Expiry)
+	require.Equal(int64(7), storage.pendingChunksProducers[chunkProducersNodes[2]][1].Chunk.Expiry)
+}
+
+func TestChunkProducerRateLimiting(t *testing.T) {
```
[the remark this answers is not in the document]
tsachiherman: The impact of `SetMin` was covered in `TestChunkProducersOrdering`, but I don't mind adding it here as well.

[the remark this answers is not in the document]
tsachiherman: done.

x/dsmr/storage.go
aaronbuchwald: As discussed offline, it should be alright to allow the window to be the same as the validity window, which will allow us to only track a sum of total bandwidth consumed within that window and not worry about guaranteeing there's never a shorter interval where a validator consumes more bandwidth than they are allotted for that period of time.

tsachiherman:
deprecated in favor of https://github.com/ava-labs/hypersdk/pull/1922

PR #1920, opened 2025-02-07
Chain index test util

Elvis339:
This PR introduces a chain index utility for testing, making it easier to construct tests where chain index is dependency.

### DIFF
--- a/chainindex/chain_index_test.go
+++ b/chainindex/chain_index_test.go
@@ -7,17 +7,14 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"github.com/ava-labs/avalanchego/ids"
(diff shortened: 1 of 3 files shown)


chainindex/testing.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package chainindex
```
Elvis339, line 4: I initially had this utility in it's own package but `Config` and `Build` are creating cycle imports. I can make generic `Config`.

Elvis339:
Unit test is failing due to incompleteness - fixed in base branch.

PR #1922, opened 2025-02-10
dsmr: implement chunk rate limiter

tsachiherman:
## What ? 
This PR adds chunk rate limiting.

## How ?
The `rules` contains now one additional method:
```golang
GetMaxAccumulatedProducerChunkWeight() uint64
```

The `storage` have an additional function named `CheckRateLimit` that checks if adding a given chunk would breach the rate limit threshold in the following way:
- for each producer, we keep an accumulator of the total pending chunks weight.
- If adding the new chunk would cause this accumulator to exceed the max allowed weight, an error would get generated.

## Where ?
The new `CheckRateLimit` is called from two places:
- During the ChunkBuilding, right after the chunk is being created, the new chunk is being validated. This would ensure that an honest node wouldn't generate a chunk that would exceed the rate limit threshold.
- During `ChunkSignatureRequestVerifier`, a call was added to ensure that inbound requests would not result in adding a chunk that would exceed the limit.

## Note

This is a subtle version of https://github.com/ava-labs/hypersdk/pull/1916, which was time-window aware.


x/dsmr/storage.go
```diff
@@ -326,6 +356,15 @@ func (s *ChunkStorage[T]) GetChunkBytes(expiry int64, chunkID ids.ID) ([]byte, e
...
 }
 
+func (s *ChunkStorage[T]) CheckRateLimit(chunk Chunk[T]) error {
+	weightLimit := s.ruleFactory.GetRules(chunk.Expiry).GetMaxAccumulatedProducerChunkWeight()
+
+	if uint64(len(chunk.bytes))+s.pendingChunksSizes[chunk.Producer] > weightLimit {
+		return ErrChunkRateLimitSurpassed
+	}
+	return nil
+}
```
aaronbuchwald, line 366: Can we add this to complete the existing TODO here: https://github.com/ava-labs/hypersdk/blob/cdd2245a718539cccf4eb25a23058e086e243639/x/dsmr/storage.go#L84 rather than add this into storage?

x/dsmr/p2p.go
```diff
@@ -135,6 +135,12 @@ func (c ChunkSignatureRequestVerifier[T]) Verify(
 			Message: err.Error(),
 		}
 	}
+	if err := c.storage.CheckRateLimit(chunk); err != nil {
+		return &common.AppError{
+			Code:    p2p.ErrUnexpected.Code,
+			Message: err.Error(),
+		}
+	}
```
aaronbuchwald, line 143: Can we add this above `c.storage.GetChunkBytes(...)` ?

It's preferred to perform the lightest checks first, so that if a request will fail, we minimize the resources used to fail it. This is not a critical difference since a malicious peer can inspect our code and craft a response that will fail at a specific check, but it's a good practice.

aaronbuchwald: (unrelated to this PR) I think the above check if the bytes are present is unnecessary.

Since we ignore the result of the query, the only case where it changes our handling is if we hit an unexpected database error (ie. not `database.ErrNotFound`). I think we're just wasting a potential database read here.

x/dsmr/storage_test.go
```diff
@@ -369,3 +373,129 @@ func TestRestartSavedChunks(t *testing.T) {
...
+				for _, chunkExpiry := range testCase.acceptedChunksExpiry {
+					// find the chunk that corresponds to this expiry in the chunks slice.
+					chunkIndex := -1
+					for i, chunk := range chunks {
+						if chunk.Expiry == chunkExpiry {
+							// found.
+							chunkIndex = i
+							break
+						}
+					}
```
aaronbuchwald, line 479: Can we use `slices.Index` or `slices.IndexFunc` helper here, which does the same search for an equivalent value?

x/dsmr/storage_test.go
```diff
@@ -369,3 +373,129 @@ func TestRestartSavedChunks(t *testing.T) {
...
+					maxProducerChunkWeight: testCase.weightLimit,
+				},
+			}
+			storage, chunks, _, _, _ := createTestStorage(t, testCase.expiryTimes, []int64{}, rules)
+
+			for _, chunk := range chunks {
+				require.NoError(storage.AddLocalChunkWithCert(chunk, nil))
+			}
+
+			if testCase.minTime != 0 {
```
aaronbuchwald, line 468: Adding this check creates the possibility of a test case where a field is accidentally ignored ie. `minTime = 0`, but there's a non-zero number of accepted chunks, which will be ignored.

Could we remove this possibility by just removing this check and relying on the fact that `SetMin(0, nil)` will be a no-op that should be safe to call.

x/dsmr/storage_test.go
```diff
@@ -369,3 +373,129 @@ func TestRestartSavedChunks(t *testing.T) {
...
+		},
+		{
+			name:                 "success - accepted block clear previous limit",
+			expiryTimes:          []int64{0, 50, 100},
+			newChunkExpiry:       150,
+			weightLimit:          chunkSize * 3,
+			expectedErr:          nil,
+			minTime:              10,
+			acceptedChunksExpiry: []int64{50},
+		},
```
aaronbuchwald, line 443: Should accepting a block clear it from the limit within a window?

I think it should still count against the rate limit because it was consumed within that window.

tsachiherman: It's just a matter of framing, right ? Just think about this as the side of unaccepted chunks per producer.
Can you think of any clear advantage of maintaining a separate list of accepted-but-not-yet-expired chunks ?

x/dsmr/storage.go
```diff
@@ -228,20 +238,24 @@ func (s *ChunkStorage[T]) VerifyRemoteChunk(c Chunk[T]) (*warp.BitSetSignature,
 	return nil, nil
 }
 
+// putVerifiedChunk assumes that the given chunk is gurenteed not to surpass the producer's rate limit.
```
aaronbuchwald, line 241:
```suggestion
// putVerifiedChunk assumes that the given chunk is guaranteed not to surpass the producer's rate limit.
```

x/dsmr/storage_test.go
aaronbuchwald: Yup, I think it's preferable that accepting a chunk does not allow the same validator to consume the resources within that window of time, but this can be changed without a network upgrade, so think this is fine for now.

If we do this over a time window (irrespective of accepted/pending), then we have a clear upper bound on the max amount of chunk storage consumed by a single validator. If we clear the chunks when they are accepted, then we don't have a well defined limit because once the chunks are accepted the validator that produced it can consume more resources.

x/dsmr/storage.go
```diff
@@ -228,20 +238,24 @@ func (s *ChunkStorage[T]) VerifyRemoteChunk(c Chunk[T]) (*warp.BitSetSignature,
...
+// The rate limit is being checked via a call to CheckRateLimit from BuildChunk as well as by VerifyRemoteChunk.
```
aaronbuchwald, line 242: Could we make sure this comment is up to date? This is called to self-regulate within `BuildChunk` and it is called in the p2p handler `ChunkSignatureRequestVerifier` to rate limit peers, but this is not called in `VerifyRemoteChunk` atm.

Also, with the current code, it's not safe to call this within `VerifyRemoteChunk` because we call `VerifyRemoteChunk` within `Accept`. This means that if a peer consumed more bandwidth, this could fail within accept (which would fatal).

tsachiherman: updated.

x/dsmr/storage.go
```diff
@@ -228,20 +238,25 @@ func (s *ChunkStorage[T]) VerifyRemoteChunk(c Chunk[T]) (*warp.BitSetSignature,
 	return nil, nil
 }
 
+// putVerifiedChunk assumes that the given chunk is guaranteed not to surpass the producer's rate limit.
+// The rate limit is being checked via a call to CheckRateLimit from BuildChunk ( for local generate chunks )
+// as well as by ChunkSignatureRequestVerifier.Verify for incoming chunk signature requests.
```
aaronbuchwald, line 243:
```suggestion
// The rate limit is being checked via a call to CheckRateLimit from BuildChunk (for locally generated chunks)
// and ChunkSignatureRequestVerifier.Verify for incoming chunk signature requests.
```

aaronbuchwald (approved):

Issue #1923, opened 2025-02-11
Historical Read APIs


aaronbuchwald:
## Historical Read

We currently offer one method to fetch an accepted transaction (which does not return the full transaction, just the result) and three methods to fetch a block.

We should clean this up to offer a a single method to fetch a transaction and a single method to fetch a block that enables the caller to fetch a block based on (block hash, block number, or to simply request the latest block).

I'd propose that we define the following interface:

```golang
type BlockRequest struct {
	BlockID ids.ID
	BlockNumber uint64
}

type BlockResponse struct {
	Block      *chain.ExecutedBlock
	BlockBytes codec.Bytes         
}

type TxRequest struct {
	TxID ids.ID
}

type TxResponse struct {
	Tx *chain.Transaction
	TxBytes codec.Bytes
	Timestamp int64
	Result *chain.Result
}

type Indexer interface {
	GetBlock(req *http.Request, args *BlockRequest, reply *BlockResponse) error
	GetTx(req *http.Request, args *TxRequest, reply *TxResponse) error
}
```

To serve the block request, we can simply check the arguments in order to determine what block to fetch such as:
1. If `BlockID` is populated, fetch that block and ignore the rest of the request
2. If `BlockNumber` is max uint64, fetch the latest block
3. Otherwise, attempt to fetch the requested block number

Note: this API currently maintains separate indices for transaction data and blocks. To serve the relevant requests, we need some version of the indices:

txHash -> txDetails
blockNumber -> block details
blockHash -> block number

We may want to optionally disable blockHash -> blockNumber or map txHash -> block height to fetch the relevant details rather than duplicating the data. This would limit the on-disk footprint, but require fetching a potentially much larger value from disk.

Throughput tests should disable this API to reduce disk writes in favor of relying on the websocket API, so this should not impact throughput tests.

PR #1930, opened 2025-02-14
refactor historical API

tsachiherman:
## What ?

This PR is the first refactor ovehaul for the hyper sdk indexer.
The PR addresses consistency and interface functionality.

## Changes

- Unify the three grpc block request endpoints into a single endpoint (`GetBlockByHeight`, `GetBlockByHash`, `GetLatestBlock`) into `GetBlock`. The structure parameter for `GetBlock` would provide the same functionality previously provided by the individual endpoints.
- The query for the transaction information would now be capable of returning more details about the transaction.
- The indexer block storage and in-memory cache are now always aligned.
- The indexer internal variables are now correctly synchronized.

## Issue

https://github.com/ava-labs/hypersdk/issues/1923


api/indexer/indexer.go
```diff
@@ -87,28 +91,28 @@ func (i *Indexer) initBlocks() error {
 		}
 
 		i.blockIDToHeight.Put(blk.Block.GetID(), blk.Block.Hght)
-		i.blockHeightToBlock.Put(blk.Block.Hght, blk)
+		if _, removedBlk := i.blockHeightToBlock.Put(blk.Block.Hght, blk); removedBlk != nil {
+			// the following should not happen, as we maintain only blockWindow entries; however,
+			// if it does, the following would correct that.
```
aaronbuchwald, line 96: Since the block window can change on restart, this comment is not accurate ie. we may hit this when the block window shortens such that we should remove blocks

api/indexer/indexer.go
```diff
@@ -125,26 +129,41 @@ func (i *Indexer) storeBlock(blk *chain.ExecutedBlock) error {
...
+		}
 	}
 
 	i.blockIDToHeight.Put(blk.Block.GetID(), blk.Block.Hght)
-	i.blockHeightToBlock.Put(blk.Block.Hght, blk)
-	i.lastHeight.Store(blk.Block.Hght)
 
-	return nil
+	// ensure that lastHeight always contains the highest height we've seen.
+	i.lastHeight = max(i.lastHeight, blk.Block.Hght)
```
aaronbuchwald, line 150: Why do we use this here? We should only ever receive blocks in order.

api/indexer/indexer.go
```diff
@@ -87,28 +91,28 @@ func (i *Indexer) initBlocks() error {
...
-			return err
-		}
-	}
-
+	i.lastHeight = lastHeight
 	return nil
 }
 
 func (i *Indexer) Notify(_ context.Context, blk *chain.ExecutedBlock) error {
+	i.mu.Lock()
```
aaronbuchwald, line 114: Do we really need to grab the lock while writing to the database?

I believe using an atomic should be sufficient and removes the need to hold the lock during database operations.

We only read the last height during `GetLatestBlock`, if it's updated during that time, it should be fine to return a slightly outdated latest block. If a client performs sequential calls, this should still provide them a monotonically increasing view of the latest block.

api/indexer/indexer.go
```diff
@@ -87,28 +91,28 @@ func (i *Indexer) initBlocks() error {
...
+	defer i.mu.Unlock()
```
aaronbuchwald, line 115: let's add a blank line after grab and defer releasing the lock for consistent style

api/indexer/indexer.go
```diff
@@ -125,26 +129,41 @@ func (i *Indexer) storeBlock(blk *chain.ExecutedBlock) error {
...
+	// ensure that lastHeight always contains the highest height we've seen.
+	i.lastHeight = max(i.lastHeight, blk.Block.Hght)
+	return blkBatch.Write()
 }
 
 func (i *Indexer) GetLatestBlock() (*chain.ExecutedBlock, error) {
-	return i.GetBlockByHeight(i.lastHeight.Load())
+	i.mu.RLock()
+	defer i.mu.RUnlock()
+	return i.GetBlockByHeight(i.lastHeight)
```
aaronbuchwald, line 157: Same comment here for empty line after deferring the lock

api/indexer/indexer.go
```diff
@@ -125,26 +129,41 @@ func (i *Indexer) storeBlock(blk *chain.ExecutedBlock) error {
...
 }
 
 func (i *Indexer) GetBlockByHeight(height uint64) (*chain.ExecutedBlock, error) {
+	i.mu.RLock()
+	defer i.mu.RUnlock()
+	return i.getBlockByHeight(height)
```
aaronbuchwald, line 163: same comment to add an empty line here

api/indexer/indexer.go
```diff
@@ -153,11 +172,13 @@ func (i *Indexer) GetBlockByHeight(height uint64) (*chain.ExecutedBlock, error)
 }
 
 func (i *Indexer) GetBlock(blkID ids.ID) (*chain.ExecutedBlock, error) {
+	i.mu.RLock()
+	defer i.mu.RUnlock()
```
aaronbuchwald, line 176: same style comment here

api/indexer/indexer.go
```diff
@@ -200,31 +223,45 @@ func (*Indexer) storeTransaction(
...
+	}
+	result := &chain.Result{
+		Success: storageTx.Success,
+		Error:   []byte(storageTx.Error),
+		Outputs: storageTx.Outputs,
+		Units:   unpackedUnits,
+		Fee:     storageTx.Fee,
 	}
-	return true, storageTx.Timestamp, storageTx.Success, unpackedUnits, storageTx.Fee, storageTx.Outputs, storageTx.Error, nil
+	return true, tx, storageTx.TxBytes, storageTx.Timestamp, result, nil
```
aaronbuchwald, line 264: Can we return just the transaction instead of `storageTx.TxBytes` separately since a client can already retrieve the bytes from the `*chain.Transaction` type

api/indexer/server.go
```diff
@@ -72,46 +69,32 @@ func (s *Server) GetBlock(req *http.Request, args *GetBlockRequest, reply *GetBl
...
-func (s *Server) GetLatestBlock(req *http.Request, _ *struct{}, reply *GetBlockResponse) error {
-	_, span := s.tracer.Start(req.Context(), "Indexer.GetLatestBlock")
-	defer span.End()
-
-	block, err := s.indexer.GetLatestBlock()
 	if err != nil {
 		return err
 	}
-	return reply.setResponse(block)
+	return reply.setResponse(executedBlk)
```
aaronbuchwald, line 87: can we inline this since it's only used in one place with this change?

api/indexer/server.go
```diff
@@ -72,46 +69,32 @@ func (s *Server) GetBlock(req *http.Request, args *GetBlockRequest, reply *GetBl
...
-
-func (s *Server) GetBlockByHeight(req *http.Request, args *GetBlockByHeightRequest, reply *GetBlockResponse) error {
-	_, span := s.tracer.Start(req.Context(), "Indexer.GetBlockByHeight")
-	defer span.End()
-
-	block, err := s.indexer.GetBlockByHeight(args.Height)
-	if err != nil {
-		return err
+	var executedBlk *chain.ExecutedBlock
+	var err error
```
aaronbuchwald, line 73: nit: can we turn this into a single var block?

internal/cache/fifo.go
```diff
@@ -55,4 +63,5 @@ func (f *FIFO[K, V]) Get(key K) (V, bool) {
 // [WriteLock] is held when this is accessed.
 func (f *FIFO[K, V]) remove(key K) {
 	delete(f.m, key)
+	f.removedKey = &key
```
aaronbuchwald, line 66: Hmm it seems fairly odd to pass around this data via an internal pointer modified via the callback in the bounded buffer.

It seems that we added this to return the removed key for the indexer, but I don't think we should modify the fifo cache with this hack to support that specific case. Is there a reason not to use `DeleteRange` to make sure that we've removed everything below our desired height?

api/indexer/server.go
```diff
@@ -46,11 +46,8 @@ func (f *apiFactory) New(vm api.VM) (api.Handler, error) {
 }
 
 type GetBlockRequest struct {
-	BlockID ids.ID `json:"blockID"`
-}
-
-type GetBlockByHeightRequest struct {
-	Height uint64 `json:"height"`
+	BlockID     ids.ID `json:"blockID"`
+	BlockNumber uint64 `json:"blockNumber"`
```
aaronbuchwald, line 50: (out of scope for this PR) Can we add an `Encoding` argument that can be either `json` or `hex`, defaults to `hex`, and specifies which field in the response to fill out ie. we populate either the block bytes or block field, but not both?

Can we apply the same to the transaction and either populate the JSON of the transaction or the transaction bytes?

api/indexer/client.go
```diff
@@ -53,8 +56,8 @@ func (c *Client) GetBlockByHeight(ctx context.Context, height uint64, parser cha
 	resp := GetBlockClientResponse{}
 	err := c.requester.SendRequest(
 		ctx,
-		"getBlockByHeight",
-		&GetBlockByHeightRequest{Height: height},
+		"getBlock",
```
aaronbuchwald, line 59: Out of scope for this PR - we should move the parser from an argument to each function to an argument to construct the client instance, so that we don't need to pass it around each time

api/indexer/client.go
```diff
@@ -89,24 +92,34 @@ func (c *Client) GetTx(ctx context.Context, txID ids.ID) (GetTxResponse, bool, e
...
+		response, _, found, err := c.GetTx(ctx, txID, nil)
 		if err != nil {
 			return false, err
 		}
-		success = response.Success
-		fee = response.Fee
+		if found && response.Result != nil {
+			success = response.Result.Success
+			fee = response.Result.Fee
+		}
```
aaronbuchwald, line 122: Do we need the check for result to be nil? We should guarantee this is never the case

internal/cache/fifo.go
```diff
@@ -31,16 +32,23 @@ func NewFIFO[K comparable, V any](limit int) (*FIFO[K, V], error) {
...
 	f.l.Lock()
 	defer f.l.Unlock()
 
+	var removedKey *K
 	_, exists := f.m[key]
 	if !exists {
+		f.removedKey = nil
 		f.buffer.Push(key) // Push removes the oldest [K] if we are at the [limit]
+		removedKey = f.removedKey
+		f.removedKey = nil
```
aaronbuchwald, line 47: Why do we add `f.removedKey = nil` both before and after? It seems we should only ever need one.

api/indexer/indexer.go
tsachiherman: The lock no longer takes place during the database updates.

api/indexer/server.go
tsachiherman: yes, I'll tackle this on the next PR.

tsachiherman: I'd like to have some more unit testing for the server as well.

api/indexer/indexer.go
```diff
@@ -8,230 +8,252 @@ import (
...
+
+	if err := blkBatch.Put(latestBlockHeightKey, binary.BigEndian.AppendUint64(nil, i.lastHeight)); err != nil {
 		return err
 	}
 
-	i.blockIDToHeight.Put(blk.Block.GetID(), blk.Block.Hght)
-	i.blockHeightToBlock.Put(blk.Block.Hght, blk)
-	i.lastHeight.Store(blk.Block.Hght)
+	// clear the slice without allocation
+	i.evictedBlocks = i.evictedBlocks[:0]
```
aaronbuchwald, line 199: Do we need to grab the lock here?

I think the answer is just barely no because this is only accessed by the writer and not on reads and the writer is single threaded, but this is fairly non-obvious.

api/indexer/indexer.go
```diff
@@ -8,230 +8,252 @@ import (
...
+
+	i.cachedBlocks, err = buffer.NewBoundedQueue(int(blockWindow), i.evictBlock)
+	if err != nil {
+		return nil, err
 	}
+
 	return i, i.initBlocks()
 }
 
+func (i *Indexer) evictBlock(blockHeight uint64) {
```
aaronbuchwald, line 85: do we rely on the assumption that the lock is held whenever we insert into the bounded buffer, so that the lock is held when this is called?

The indirection here makes this a bit hard to follow, so I think we should either try and simplify this or clearly document what the invariants are here to ensure there's no race conditions.

tsachiherman: I don't think it's any different than FIFO, and I added a description so to reduce confusion.

api/indexer/indexer.go
```diff
@@ -8,230 +8,252 @@ import (
...
 }
 
 func (i *Indexer) GetLatestBlock() (*chain.ExecutedBlock, error) {
-	return i.GetBlockByHeight(i.lastHeight.Load())
+	i.mu.RLock()
+	defer i.mu.RUnlock()
+
+	if i.lastHeight == math.MaxUint64 {
+		return nil, database.ErrNotFound
+	}
```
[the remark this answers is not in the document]
tsachiherman: The error here is a bit different from the "regular" `errBlockNotFound` in that it means that the indexer never received even a single block. Maybe `fmt.Errorf("%w: indexer never bootstrapped", errBlockNotFound)` would be a better fit here ?

api/indexer/indexer.go
```diff
@@ -8,230 +8,252 @@ import (
...
-	unpackedUnits, err := fees.UnpackDimensions(storageTx.Units)
-	if err != nil {
-		return false, 0, false, fees.Dimensions{}, 0, nil, "", fmt.Errorf("failed to unpack units for storage tx %s: %w", txID, err)
+	cachedTx, ok := i.txCache[txID]
+	if !ok {
+		return false, nil, 0, nil, nil
 	}
-	return true, storageTx.Timestamp, storageTx.Success, unpackedUnits, storageTx.Fee, storageTx.Outputs, storageTx.Error, nil
+	tx := cachedTx.blk.Block.Txs[cachedTx.index]
+	result := cachedTx.blk.Results[cachedTx.index]
```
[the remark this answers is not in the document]
tsachiherman: I think that I'd rather checking that in `Notify` before storing (memory/disk) an invalid block. But I'll add that here for now.

api/indexer/server.go
```diff
@@ -46,72 +46,53 @@ func (f *apiFactory) New(vm api.VM) (api.Handler, error) {
...
 }
 
 type GetTxResponse struct {
-	Timestamp int64           `json:"timestamp"`
-	Success   bool            `json:"success"`
-	Units     fees.Dimensions `json:"units"`
-	Fee       uint64          `json:"fee"`
-	Outputs   []codec.Bytes   `json:"result"`
-	ErrorStr  string          `json:"errorStr"`
+	TxBytes   codec.Bytes   `json:"transactionBytes"`
```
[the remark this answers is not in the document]
tsachiherman: np, I'll do that in the next PR.

api/indexer/indexer.go
```diff
@@ -8,230 +8,252 @@ import (
...
-		}
-	}
-
 	return nil
 }
 
 func (i *Indexer) Notify(_ context.Context, blk *chain.ExecutedBlock) error {
-	if err := i.storeTransactions(blk); err != nil {
-		return err
+	if i.blockWindow == 0 {
```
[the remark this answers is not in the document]
aaronbuchwald: yes that's what I mean. There is already an `enabled: false` option in the config. This API is useless if the retention window is 0, so we could enforce in the constructor that `blockWindow == 0` is invalid, since it makes no sense to have this API enabled with that setting.

api/indexer/indexer.go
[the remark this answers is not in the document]
aaronbuchwald: Hmm so I guess the sides of this are:

- assume we only receive valid blocks, no need to handle this in notify or here
- we SHOULD only receive valid blocks, but this component should not assume that and protect itself on the write path
- assume that we receive only valid blocks, but apply defense in depth and protect ourselves on the read path

We should never hit an error here, so my monkey brain says that it makes the most sense to protect on the read path, so that it's very clear we don't panic.

chain/chaintest/test_balance_handler.go
RodrigoVillar, line 1: We already have a `mockBalanceHandler` that we use for testing TXs: 

https://github.com/ava-labs/hypersdk/blob/ee47a66b752e8fccb56d03eb0fb059ecaca912f2/chain/transaction_test.go#L48-L71

Instead of creating another `mockBalanceHandler`, would it be possible to create a `TestBalanceHandler` in `chaintest` which follows the following style? (this way, we can combine the two `mockBalanceHandler` implementations into one):

https://github.com/ava-labs/hypersdk/blob/ee47a66b752e8fccb56d03eb0fb059ecaca912f2/internal/validitywindow/validitywindowtest/test_validity_window.go#L17-L20

tsachiherman: yes, I've seen the other version, and was afraid of introducing a dependency cycle. But, reviewing this code now ( after @aaronbuchwald 's changes ), I see that `chain` is already depends on `chaintest`, and therefore won't create any further issue.

api/indexer/indexer.go
```diff
@@ -8,230 +8,262 @@ import (
...
 	blockDB            *pebble.Database // height -> block bytes
-	blockIDToHeight    *cache.FIFO[ids.ID, uint64]
-	blockHeightToBlock *cache.FIFO[uint64, *chain.ExecutedBlock]
-	blockWindow        uint64 // Maximum window of blocks to retain
-	lastHeight         atomic.Uint64
+	cachedBlocks       buffer.Queue[uint64]
+	blockIDToHeight    map[ids.ID]uint64
+	blockHeightToBlock map[uint64]*chain.ExecutedBlock
+	txCache            map[ids.ID]cachedTransaction
+	evictedBlocks      []uint64 // list of blocks that are queued for deletion.
```
aaronbuchwald, line 47: Since using the bounded buffer forces us to add to this slice via a callback function (bit harder to follow than ideal), could we switch to using a data structure that allows us to easily return the element that should be removed like a queue?

aaronbuchwald (changes requested):

chain/chaintest/test_balance_handler.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/ava-labs/hypersdk/state"
+)
+
+func NewEmptyBalanceHandler(canDeductError error) chain.BalanceHandler {
+	return &mockBalanceHandler{
+		canDeductError,
+	}
+}
+
+type mockBalanceHandler struct {
```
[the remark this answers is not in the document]
tsachiherman: yes. I've eliminate the mockBalanceHandler completely and used the `NewPrefixBalanceHandler` instead across all our testing.

chain/chaintest/block.go
```diff
@@ -42,3 +53,81 @@ func GenerateEmptyExecutedBlocks(
...
+	bh := NewEmptyBalanceHandler(nil)
+	rules := genesis.NewDefaultRules()
+	feeManager := internal_fees.NewManager(nil)
+	executedBlocks := make([]*chain.ExecutedBlock, numBlocks)
+	parser := NewTestParser()
+	ts := tstate.New(0)
+	db, err := merkledb.New(
+		context.Background(),
+		memdb.New(),
+		merkledb.Config{BranchFactor: 2},
```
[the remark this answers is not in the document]
tsachiherman: There is no "DefaultBranchFactor", so I'll switch to "merkledb.BranchFactor16" which is used elsewhere.

chain/chaintest/block.go
```diff
@@ -42,3 +53,81 @@ func GenerateEmptyExecutedBlocks(
...
+			tx, err := chain.NewTransaction(base, actions, auth)
+			require.NoError(err)
+			txs = append(txs, tx)
+		}
+		statelessBlock, err := chain.NewStatelessBlock(
+			parentID,
+			parentTimestamp+timestampOffset*int64(i),
+			parentHeight+1+uint64(i),
+			txs,
+			ids.Empty,
```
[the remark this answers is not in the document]
tsachiherman: This test currently doesn't use the root anywhere. I can definitely add that, but there isn't any immediate benefit of doing so that this moment ( unless I'm missing something )

chain/chaintest/block.go
```diff
@@ -42,3 +53,81 @@ func GenerateEmptyExecutedBlocks(
...
+		for _, tx := range txs {
+			res, err := tx.Execute(context.Background(), feeManager, bh, rules, tstateview, parentTimestamp+timestampOffset*int64(i))
+			require.NoError(err)
+			results = append(results, res)
+		}
+		// marshal and unmarshal the block in order to remove all non-persisting data members ( i.e. statekeys )
+		blkBytes := statelessBlock.MarshalCanoto()
+		require.NoError(err)
+		statelessBlock, err = chain.UnmarshalBlock(blkBytes, parser)
+		require.NoError(err)
```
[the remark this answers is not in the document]
tsachiherman: yes, it's needed. During the transaction execution, the statekeys field is being populated. We don't want to return this field since the UnmarshalBlock won't populate it.
i.e. this method is returning a series of blocks that are being used as the "expected". When comparing these to the api-decoded transactions, we need to create a match.

api/indexer/client.go
```diff
@@ -97,17 +101,54 @@ func (c *Client) GetTx(ctx context.Context, txID ids.ID) (GetTxResponse, bool, e
...
+		err := c.requester.SendRequest(
+			ctx,
+			"getTx",
+			&GetTxRequest{TxID: txID},
+			&resp,
+		)
+		var jsonErr *json2.Error
+		if errors.As(err, &jsonErr) && jsonErr.Message == ErrTxNotFound.Error() {
+			return false, nil
+		}
```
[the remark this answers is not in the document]
tsachiherman: we have it in multiple locations : 
- in client_test.go, we test the context.DeadlineExceeded condition.
- in morpheusvm, the generator test that in `confirmTx`
- in e2e testing, it's being tested in `Network.ConfirmTxs`

api/indexer/client_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/trace"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/hypersdk/api"
+	"github.com/ava-labs/hypersdk/chain/chaintest"
+)
+
+func TestIndexerClient(t *testing.T) {
```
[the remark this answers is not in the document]
tsachiherman: you mean testcase data driven ? yes, I can do that.

api/indexer/client_test.go
```diff
@@ -0,0 +1,182 @@
...
+	const (
+		numExecutedBlocks = 4
+		blockWindow       = 2
+		numTxs            = 3
+	)
+
+	testCases := []struct {
+		name     string
+		txBlkIdx int
+		txIdx    int
```
aaronbuchwald, line 94: Can we use `blockIndex` and `txIndex` for consistency? (we never use `Idx` for index in the codebase)

api/indexer/indexer.go
```diff
@@ -8,230 +8,256 @@ import (
...
+	errInvalidBlockWindowSize = errors.New("invalid indexer block window size specified")
+)
 
 var _ event.Subscription[*chain.ExecutedBlock] = (*Indexer)(nil)
 
 type Indexer struct {
 	blockDB            *pebble.Database // height -> block bytes
-	blockIDToHeight    *cache.FIFO[ids.ID, uint64]
-	blockHeightToBlock *cache.FIFO[uint64, *chain.ExecutedBlock]
+	cachedBlocks       []uint64
```
aaronbuchwald, line 47: This is unnecessary duplicate data since we can always check the length of `blockHeightToBlock` or the other data structures and figure out the latest one.

api/indexer/indexer.go
```diff
@@ -8,230 +8,256 @@ import (
...
+	txCache            map[ids.ID]cachedTransaction
 	blockWindow        uint64 // Maximum window of blocks to retain
-	lastHeight         atomic.Uint64
+	lastHeight         uint64
 	parser             chain.Parser
 
-	// ID -> timestamp, success, units, fee, outputs
-	txDB *pebble.Database
+	// synchronization mutex between r/w
+	mu sync.RWMutex
```
aaronbuchwald, line 56: Can we organize the fields in this struct more clearly so that the lock is above the fields that it is required to access and move the parser and blockWindow to the top next to the database which the lock is not needed to access?

api/indexer/indexer.go
```diff
@@ -8,230 +8,256 @@ import (
...
+}
+
+type cachedTransaction struct {
+	// blk is the block that contains the transaction
+	blk *chain.ExecutedBlock
```
aaronbuchwald, line 61: Could we just store the height rather than the block?

api/indexer/indexer.go
```diff
@@ -8,230 +8,256 @@ import (
...
+	hasLastHeight, err := i.blockDB.Has(latestBlockHeightKey)
+	if err != nil || !hasLastHeight {
+		return err
+	}
+
+	lastHeight, err := i.blockDB.Get(latestBlockHeightKey)
+	if err != nil {
+		return err
+	}
+
```
aaronbuchwald, line 103: Don't think this should be necessary

api/indexer/indexer.go
```diff
@@ -8,230 +8,256 @@ import (
...
 func (i *Indexer) Notify(_ context.Context, blk *chain.ExecutedBlock) error {
-	if err := i.storeTransactions(blk); err != nil {
-		return err
-	}
-	return i.storeBlock(blk)
+	i.mu.Lock()
+	evictedBlockHeights := i.insertBlockIntoCache(blk)
+	i.mu.Unlock()
+
+	return i.storeBlock(blk, evictedBlockHeights)
```
aaronbuchwald, line 139: Ik I commented that we should return this data to make the flow more clear, but looking now, it seems we should not need this at all because we can calculate what database entry (if any) should be deleted within `i.storeBlock(...)`

api/indexer/indexer.go
```diff
@@ -8,230 +8,256 @@ import (
...
-	if i.blockWindow == 0 {
-		return nil
+// insertBlockIntoCache add the given block and its transactions to the
+// cache. It returns a slice of the evicted block heights.
+// assumes the write lock is held
+func (i *Indexer) insertBlockIntoCache(blk *chain.ExecutedBlock) []uint64 {
+	var evictedBlockHeights []uint64
+	if len(i.cachedBlocks) == int(i.blockWindow) {
+		evictedBlockHeights = []uint64{i.cachedBlocks[0]}
+		i.cachedBlocks = i.cachedBlocks[1:]
```
aaronbuchwald, line 149: This is a memory leak since we maintain a reference to the subslice, always append, and never create an opportunity to garbage collect. We should just remove this field completely imo.

chain/chaintest/block.go
```diff
@@ -4,11 +4,24 @@
...
 
 	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/consts"
 	"github.com/ava-labs/hypersdk/fees"
+	"github.com/ava-labs/hypersdk/genesis"
+	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/state/balance"
+	"github.com/ava-labs/hypersdk/state/tstate"
+
+	internal_fees "github.com/ava-labs/hypersdk/internal/fees"
```
aaronbuchwald, line 24: Can we use `internalfees` for consistency with the rest of the codebase?

chain/chaintest/block.go
```diff
@@ -42,3 +55,81 @@ func GenerateEmptyExecutedBlocks(
...
+// these blocks are not semantically valid for execution, and are meant execlusively for serialization testing purposes.
+func GenerateTestExecutedBlocks(
+	require *require.Assertions,
+	parentID ids.ID,
+	chainID ids.ID,
+	parentHeight uint64,
+	parentTimestamp int64,
+	timestampOffset int64,
+	numBlocks int,
+	numTx int,
```
aaronbuchwald, line 69: nit: can this should be plural, can we make this `numTxsPerBlock`

chain/chaintest/block.go
```diff
@@ -42,3 +55,81 @@ func GenerateEmptyExecutedBlocks(
...
+	db, err := merkledb.New(
+		context.Background(),
+		memdb.New(),
+		merkledb.Config{BranchFactor: merkledb.BranchFactor16},
+	)
+	require.NoError(err)
+
+	tstateview := ts.NewView(state.CompletePermissions, db, 0)
+	require.NoError(tstateview.Insert(context.Background(), balance.NewPrefixBalanceHandler([]byte{0}).BalanceKey(NewDummyTestAuth().Sponsor()), binary.BigEndian.AppendUint64(nil, math.MaxUint64)))
+
```
aaronbuchwald, line 86: Can we create one instance of the balance handler and call the necessary functions on that rather than create an instance and immediately call a function on it and re-use the constructor in different places?

This is error prone if we change one instance, but not the others and less readable

chain/chaintest/block.go
```diff
@@ -42,3 +55,81 @@ func GenerateEmptyExecutedBlocks(
...
+	actions := NewDummyTestActions(numBlocks * numTx)
+	for i := range executedBlocks {
+		// generate transactions.
+		txs := []*chain.Transaction{}
```
aaronbuchwald, line 90: Can we use make when initializing slices that we know the final length of?

chain/chaintest/block.go
```diff
@@ -42,3 +55,81 @@ func GenerateEmptyExecutedBlocks(
...
+		base := chain.Base{
+			Timestamp: ((parentTimestamp + timestampOffset*int64(i)/consts.MillisecondsPerSecond) + 1) * consts.MillisecondsPerSecond,
+			ChainID:   chainID,
+			MaxFee:    math.MaxUint64,
+		}
+		for j := range numTx {
```
aaronbuchwald, line 96: Can we use meaningful local variable names instead of `i` and `j` to make this more readable? THis will make `action[blockIndex*numTx+txIndex]` much more readable

chain/chaintest/block.go
```diff
@@ -42,3 +55,81 @@ func GenerateEmptyExecutedBlocks(
...
+			actions := []chain.Action{actions[i*numTx+j]}
+			auth := NewDummyTestAuth()
+			tx, err := chain.NewTransaction(base, actions, auth)
+			require.NoError(err)
+			txs = append(txs, tx)
+		}
+		statelessBlock, err := chain.NewStatelessBlock(
+			parentID,
+			parentTimestamp+timestampOffset*int64(i),
```
aaronbuchwald, line 105: If we just calculate this above, then we can use this + some offset to set the expiry for the transactions we include in each block.

chain/chaintest/block.go
```diff
@@ -42,3 +55,81 @@ func GenerateEmptyExecutedBlocks(
...
+		parentID = statelessBlock.GetID()
+
+		results := []*chain.Result{}
+		for _, tx := range txs {
+			res, err := tx.Execute(context.Background(), feeManager, bh, rules, tstateview, parentTimestamp+timestampOffset*int64(i))
+			require.NoError(err)
+			results = append(results, res)
+		}
+		// marshal and unmarshal the block in order to remove all non-persisting data members ( i.e. statekeys )
+		blkBytes := statelessBlock.MarshalCanoto()
```
aaronbuchwald, line 121: Can we use `GetBytes` instead of `MarshalCanoto` ? marshal canoto produces the same result and is exported, but ideally we can remove it from the public interface so that we define the accessors used to access the block's fields.

chain/chaintest/block.go
```diff
@@ -42,3 +55,81 @@ func GenerateEmptyExecutedBlocks(
...
+
+	tstateview := ts.NewView(state.CompletePermissions, db, 0)
+	require.NoError(tstateview.Insert(context.Background(), balance.NewPrefixBalanceHandler([]byte{0}).BalanceKey(NewDummyTestAuth().Sponsor()), binary.BigEndian.AppendUint64(nil, math.MaxUint64)))
+
+	actions := NewDummyTestActions(numBlocks * numTx)
+	for i := range executedBlocks {
+		// generate transactions.
+		txs := []*chain.Transaction{}
+		base := chain.Base{
+			Timestamp: ((parentTimestamp + timestampOffset*int64(i)/consts.MillisecondsPerSecond) + 1) * consts.MillisecondsPerSecond,
```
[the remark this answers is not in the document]
tsachiherman: yes

chain/chaintest/block.go
```diff
@@ -42,3 +55,81 @@ func GenerateEmptyExecutedBlocks(
 	}
 	return executedBlocks
 }
+
+// GenerateTestExecutedBlocks creates [numBlocks] executed block for testing purposes, each with [numTx] transactions.
+// these blocks are not semantically valid for execution, and are meant execlusively for serialization testing purposes.
+func GenerateTestExecutedBlocks(
```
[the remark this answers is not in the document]
tsachiherman: I'm going to delete this function completely, as it's not being used.

api/indexer/indexer.go
```diff
@@ -8,230 +8,230 @@ import (
...
 	return i.storeBlock(blk)
 }
 
-func (i *Indexer) storeBlock(blk *chain.ExecutedBlock) error {
-	if i.blockWindow == 0 {
-		return nil
+// insertBlockIntoCache add the given block and its transactions to the
+// cache. It returns a slice of the evicted block heights.
+// assumes the write lock is held
+func (i *Indexer) insertBlockIntoCache(blk *chain.ExecutedBlock) {
```
aaronbuchwald, line 125: Can we update this comment since it's out of date with the latest change?

api/indexer/indexer.go
```diff
@@ -8,230 +8,230 @@ import (
...
+	if _, ok := i.blockHeightToBlock[blk.Block.Hght-i.blockWindow]; ok {
+		evictedBlockHeight := blk.Block.Hght - i.blockWindow
+
+		// find the block in the blocks cache.
+		evictedBlk := i.blockHeightToBlock[evictedBlockHeight]
```
aaronbuchwald, line 130: Can we avoid doing the same map lookup twice? We need this block, so we can just set this local variable in the if condition

api/indexer/indexer.go
```diff
@@ -8,230 +8,230 @@ import (
...
+		i.txCache[tx.GetID()] = cachedTransaction{
+			blkHeight: blk.Block.Hght,
+			index:     idx,
+		}
 	}
+	i.lastHeight = blk.Block.Hght
+}
 
+// storeBlock persist the given block to the database, and remove the
+// evicted blocks.
```
aaronbuchwald, line 155:
```suggestion
// storeBlock persist the given block to the database, and deletes a block
// if it surpasses the retention window
```

api/indexer/indexer.go
```diff
@@ -8,230 +8,230 @@ import (
...
-
-	unpackedUnits, err := fees.UnpackDimensions(storageTx.Units)
-	if err != nil {
-		return false, 0, false, fees.Dimensions{}, 0, nil, "", fmt.Errorf("failed to unpack units for storage tx %s: %w", txID, err)
+	tx := blk.Block.Txs[cachedTx.index]
+	if len(blk.ExecutionResults.Results) <= cachedTx.index {
+		return false, nil, 0, nil, fmt.Errorf("%w: block height %d, transaction index %d", errTxResultNotFound, blk.Block.Hght, cachedTx.index)
 	}
-	return true, storageTx.Timestamp, storageTx.Success, unpackedUnits, storageTx.Fee, storageTx.Outputs, storageTx.Error, nil
+	result := blk.ExecutionResults.Results[cachedTx.index]
```
aaronbuchwald, line 227: Can we include index out of bounds checks for both tx and result?

api/indexer/indexer_test.go
```diff
@@ -65,51 +68,83 @@ func checkBlocks(
...
+
+		// confirm all transactions are available
+		for blkTxIndex, blkTx := range expectedBlk.Block.Txs {
+			txID := blkTx.GetID()
+			found, tx, _, res, err := indexer.GetTransaction(txID)
+			require.NoError(err)
+			require.True(found)
+			require.Equal(tx, blkTx)
+			require.Equal(expectedBlk.ExecutionResults.Results[blkTxIndex], res)
+		}
```
aaronbuchwald, line 80:
```suggestion
		for txIndex, expectedTx := range expectedBlk.Block.Txs {
			txID := expectedTx.GetID()
			found, actualTx, _, res, err := indexer.GetTransaction(txID)
			require.NoError(err)
			require.True(found)
			require.Equal(expectedTx, actualTx)
			require.Equal(expectedBlk.ExecutionResults.Results[txIndex], res)
		}
```

api/indexer/indexer_test.go
```diff
@@ -65,51 +68,83 @@ func checkBlocks(
...
 		require.ErrorIs(err, errBlockNotFound, "height=%d", i)
+
+		// Confirm all transactions outside of the window are notretrievable
+		expectedBlk := expectedBlocks[i]
+		for _, blkTx := range expectedBlk.Block.Txs {
+			txID := blkTx.GetID()
+			found, _, _, _, err := indexer.GetTransaction(txID)
+			require.NoError(err)
+			require.False(found)
+		}
```
aaronbuchwald, line 95:
```suggestion
		expectedBlk := expectedBlocks[i]
		for _, tx := range expectedBlk.Block.Txs {
			txID := tx.GetID()
			found, _, _, _, err := indexer.GetTransaction(txID)
			require.NoError(err)
			require.False(found)
		}
```

aaronbuchwald (changes requested):

api/indexer/indexer.go
aaronbuchwald: The revision is now returning the same error that mentions tx result not found in both cases ie. tx and result not being found both result in the error "transaction result not found", could we separate these two?

aaronbuchwald (dismissed):
LGTM after final comment fixing the `errTxResultNotFound` error message

api/indexer/indexer.go
tsachiherman: np; please note that the case where the transaction index is missing is not possible, regardless of what input blocks are provided via Notify.

aaronbuchwald (approved):

PR #1932, opened 2025-02-18
refactor: use a more straightforward return value

hugehope:
use a more straightforward return value


aaronbuchwald:
> use a more straightforward return value

Good catch

aaronbuchwald (approved):

PR #1934, opened 2025-02-20
Separate TX Generation From API

RodrigoVillar:
This PR separates the TX generation logic from `JSONRPCClient` and puts it in the `chain` package. This PR also updates any dependency to properly use `GenerateTransaction()` and `GenerateTransactionManual()`.

### DIFF
--- a/chain/transaction.go
+++ b/chain/transaction.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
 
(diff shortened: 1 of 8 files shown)


chain/transaction.go
```diff
@@ -615,3 +616,48 @@ func EstimateUnits(r Rules, actions Actions, authFactory AuthFactory) (fees.Dime
...
+	// Construct transaction
+	now := time.Now().UnixMilli()
+	rules := parser.Rules(now)
+	base := &Base{
+		Timestamp: utils.UnixRMilli(now, rules.GetValidityWindow()),
+		ChainID:   rules.GetChainID(),
+		MaxFee:    maxFee,
+	}
+
+	// Build transaction
```
aaronbuchwald, line 656: (unrelated to this PR since it's just copy pasta from the original location) Can we change the comments here to construct and sign and put them in the corresponding places? Alternatively, we could just remove these as this code is simple enough to be self-documenting.

aaronbuchwald (approved):

PR #1936, opened 2025-02-20
Compute Time For Tracker

RodrigoVillar:
This PR updates the `issued/s` metric for the throughput test tracker to use a more precise calculation.

Previously, we were solely relying on `time.Ticker` to compute 1-second intervals. However, the time spent computing the stats for each 1-second interval is positive and therefore, `issued/second` was actually `issued/X seconds` where `X >= 1`.

This PR implements logic that computes `X` and applies it to the `issued/s` metric.

### DIFF
--- a/throughput/tracker.go
+++ b/throughput/tracker.go
@@ -51,6 +51,7 @@ func (t *tracker) logState(ctx context.Context, cli *jsonrpc.JSONRPCClient) {
 	var psent int64
 	go func() {
 		defer tick.Stop()
+		prevTime := time.Now()
 		for {
 			select {
 			case <-tick.C:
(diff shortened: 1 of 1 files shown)


throughput/tracker.go
```diff
@@ -51,6 +51,7 @@ func (t *tracker) logState(ctx context.Context, cli *jsonrpc.JSONRPCClient) {
 	var psent int64
 	go func() {
 		defer tick.Stop()
+		prevTime := time.Now()
```
aaronbuchwald, line 54: can we move `psent` and `prevTime` next to each other to make this easier to follow?

throughput/tracker.go
```diff
@@ -61,14 +62,17 @@ func (t *tracker) logState(ctx context.Context, cli *jsonrpc.JSONRPCClient) {
...
 					utils.Outf(
 						"{{yellow}}txs seen:{{/}} %d {{yellow}}success rate:{{/}} %.2f%% {{yellow}}inflight:{{/}} %d {{yellow}}issued/s:{{/}} %d {{yellow}}unit prices:{{/}} [%s]\n", //nolint:lll
 						t.totalTxs,
 						float64(t.confirmedTxs)/float64(t.totalTxs)*100,
 						t.inflight.Load(),
-						current-psent,
+						uint64(float64(current-psent)/diff),
 						unitPrices,
 					)
+					prevTime = currTime
```
aaronbuchwald, line 75: Can we put all of these next to each other so that it's more obviously correct?

It's a bit strange to grab `current` before grabbing the lock, but then grab the current time after we've grabbed the lock. This could be much simpler if we just set them next to each other and update `psent` and `prevTime` after we've used them.

throughput/tracker.go
```diff
@@ -48,30 +48,36 @@ func (t *tracker) logResult(txID ids.ID, result *chain.Result) {
...
-				current := t.sent.Load()
 				t.l.Lock()
 				if t.totalTxs > 0 {
 					unitPrices, err := cli.UnitPrices(ctx, false)
 					if err != nil {
 						continue
 					}
+					currSent := t.sent.Load()
+					currTime := time.Now()
+					diff := min(currTime.Sub(prevTime).Seconds(), 1)
```
aaronbuchwald, line 68: is it possible for less than 1s to elapse, such that taking the min with 1s will produce an incorrect result

aaronbuchwald (approved):

PR #1937, opened 2025-02-21
test multi arch build

Elvis339:
TEST


Elvis339:
Fixed in: https://github.com/ava-labs/hypersdk/pull/1929

PR #1938, opened 2025-02-21
Improve Pacer Error Handling

RodrigoVillar:
This PR improves the error handling of `pacer` by adding an `error` field which is written exactly once and utilizing the `done` channel as a signal to see whether if the pacer is still running (`done` is open) vs if the pacer has stopped run (`done` is closed).

### DIFF
--- a/throughput/pacer.go
+++ b/throughput/pacer.go
@@ -26,27 +26,19 @@ func (p *pacer) Run(ctx context.Context, max int) {
 
 	defer close(p.done)
 
-	for {
-		select {
-		case _, ok := <-p.inflight:
-			if !ok {
-				return
-			}
-			txID, result, err := p.ws.ListenTx(ctx)
-			if err != nil {
-				p.err = fmt.Errorf("error listening to tx %s: %w", txID, err)
-				return
-			}
-			if result == nil {
-				p.err = fmt.Errorf("tx %s expired", txID)
-				return
-			}
-			if !result.Success {
-				// Should never happen
-				p.err = fmt.Errorf("tx failure %w: %s", ErrTxFailed, result.Error)
-				return
-			}
-		case <-p.done:
+	for range p.inflight {
+		txID, result, err := p.ws.ListenTx(ctx)
+		if err != nil {
+			p.err = fmt.Errorf("error listening to tx %s: %w", txID, err)
+			return
+		}
+		if result == nil {
+			p.err = fmt.Errorf("tx %s expired", txID)
+			return
+		}
+		if !result.Success {
+			// Should never happen
+			p.err = fmt.Errorf("tx failure %w: %s", ErrTxFailed, result.Error)
 			return
 		}
 	}
--- a/throughput/pacer.go
+++ b/throughput/pacer.go
@@ -45,6 +45,7 @@ func (p *pacer) Run(ctx context.Context, max int) {
 }
 
 func (p *pacer) Add(tx *chain.Transaction) error {
+	// If Run failed, return the first error immediately, otherwise register the next tx
 	select {
 	case <-p.done:
 		return p.err
--- a/throughput/pacer.go
+++ b/throughput/pacer.go
@@ -16,31 +16,29 @@ type pacer struct {
 	ws *ws.WebSocketClient
 
 	inflight chan struct{}
-	done     chan error
+	err      error
 }
 
 func (p *pacer) Run(ctx context.Context, max int) {
 	p.inflight = make(chan struct{}, max)
-	p.done = make(chan error, 1)
 
 	// Start a goroutine to process transaction results
 	for range p.inflight {
 		txID, result, err := p.ws.ListenTx(ctx)
 		if err != nil {
-			p.done <- fmt.Errorf("error listening to tx %s: %w", txID, err)
+			p.err = fmt.Errorf("error listening to tx %s: %w", txID, err)
 			return
 		}
 		if result == nil {
-			p.done <- fmt.Errorf("tx %s expired", txID)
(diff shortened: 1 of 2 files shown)


throughput/pacer.go
```diff
@@ -6,57 +6,80 @@ package throughput
...
 
 func (p *pacer) Add(tx *chain.Transaction) error {
+	if p.err != nil {
+		return p.err
+	}
+
 	if err := p.ws.RegisterTx(tx); err != nil {
-		return err
+		p.setError(err)
+		return p.err
```
aaronbuchwald, line 63: Do we need to check `p.err` here or can we register the tx and read the err iff `p.done` has been closed?

Since this value is set at most once immediately before `p.done` is closed, we should technically check that the channel has been closed before reading this value.

throughput/pacer.go
```diff
@@ -6,57 +6,80 @@ package throughput
...
+	// Wait for all inflight transactions to finish
+	<-p.done
+	return p.err
+}
+
+func (p *pacer) setError(err error) {
+	p.errOnce.Do(func() {
+		p.err = err
+		close(p.done)
+	})
```
aaronbuchwald, line 84: Can we make this even simpler by removing `errOnce` and using `defer close(p.done)` within `Run` ?

This way there's only one place where `p.done` can be closed and it's explicit that it will be called once from that function.

If we handle it this way, it means that we can write the error value once within that function and rely on the deferred call to make sure it's closed before any reader accesses it (assuming we also remove reads of `p.err` within `Add`.

throughput/pacer.go
```diff
@@ -6,57 +6,80 @@ package throughput
...
-		if !result.Success {
-			// Should never happen
-			p.done <- fmt.Errorf("tx failure %w: %s", ErrTxFailed, result.Error)
+	for {
+		select {
+		case _, ok := <-p.inflight:
+			if !ok {
+				p.setError(nil)
+				return
+			}
```
aaronbuchwald, line 35: Why do we need this check?

throughput/pacer.go
```diff
@@ -6,57 +6,95 @@ package throughput
...
+
+	l       sync.RWMutex
+	errOnce sync.Once
+	err     error
 }
 
 func (p *pacer) Run(ctx context.Context, max int) {
 	p.inflight = make(chan struct{}, max)
-	p.done = make(chan error, 1)
+	p.done = make(chan struct{}, 1)
```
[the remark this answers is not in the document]
RodrigoVillar: Done

throughput/spam.go
```diff
@@ -341,6 +341,7 @@ func (s *Spammer) distributeFunds(ctx context.Context, parser chain.Parser, feeP
 			return nil, nil, err
 		}
 		if err := p.Add(tx); err != nil {
+			_ = p.Wait()
```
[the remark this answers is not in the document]
RodrigoVillar: Removed `p.Wait()` calls in spammer

throughput/pacer.go
RodrigoVillar: As per offline conversation, we need this `!ok` check as otherwise, `p.Run()` can block since it will be reading the `0` value even from a closed channel.

throughput/pacer.go
RodrigoVillar: Done

throughput/pacer.go
```diff
@@ -9,54 +9,70 @@ import (
...
-		return err
+	case <-p.done:
+		return p.err
 	}
 }
 
 func (p *pacer) Wait() error {
 	close(p.inflight)
-	return <-p.done
+	// Wait for all inflight transactions to finish
```
aaronbuchwald, line 75: This comment is not quite correct since we won't wait for all inflight transactions to finish if it hits an error.

We are waiting for `Run` to terminate and close the done channel here.

Could we replace this with a comment like ```// Wait for Run to complete```

throughput/pacer.go
```diff
@@ -9,54 +9,70 @@ import (
...
 func (p *pacer) Add(tx *chain.Transaction) error {
-	if err := p.ws.RegisterTx(tx); err != nil {
-		return err
+	select {
+	case <-p.done:
+		return p.err
+	default:
+		if err := p.ws.RegisterTx(tx); err != nil {
+			return err
+		}
```
aaronbuchwald, line 62:
```suggestion
	// If Run failed, return the first error immediately, otherwise register the next tx
	select {
	case <-p.done:
		return p.err
	default:
		if err := p.ws.RegisterTx(tx); err != nil {
			return err
		}
```

throughput/pacer.go
```diff
@@ -9,54 +9,70 @@ import (
...
+			if result == nil {
+				p.err = fmt.Errorf("tx %s expired", txID)
+				return
+			}
+			if !result.Success {
+				// Should never happen
+				p.err = fmt.Errorf("tx failure %w: %s", ErrTxFailed, result.Error)
+				return
+			}
+		case <-p.done:
```
aaronbuchwald, line 49: This case can never be hit since we only close `p.done` after this function has completed.

aaronbuchwald: If we wanted to add another case to support cancellation, we could use a context here and have a separate function to cancel.

I don't think we need that here and can wait for either the first error to occur or to finish reading all items from `inflight` and `Wait` to be called.

throughput/pacer.go
RodrigoVillar: Updated spec

throughput/pacer.go
RodrigoVillar: Removed the `p.done` case and switched to using a `for range` loop.

PR #1941, opened 2025-02-21
Avoid division by zero by skipping log if no time has passed

aaronbuchwald:
This PR changes from using `max(diff, smallIncrementOfTime)` to just skipping an interval log if the amount of time that elapsed is 0.

We should never observe `diff = 0` regardless.


RodrigoVillar (approved):
LGTM

PR #1944, opened 2025-02-25
Clean Up Tracker

RodrigoVillar:
This PR cleans up the `tracker` by doing the following:
- Adding a constructor for `tracker`
- Removing the `sync.WaitGroup` from the tracker and instead delegating it to the `Spammer`
    - The `tracker` is meant to track transactions, not coordinate issuers
- Adding a `Stop` function and making sure that `Start()` terminates prior to `Stop` returning.

### DIFF
--- a/throughput/spam.go
+++ b/throughput/spam.go
@@ -146,7 +146,7 @@ func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbo
 	}
 
 	// start logging
-	s.tracker.start(ctx, cli)
+	s.tracker.startPeriodicLog(ctx, cli)
 
 	// broadcast transactions
 	err = s.broadcast(cctx, sh, factories, issuers, feePerTx, terminate)
--- a/throughput/spam.go
+++ b/throughput/spam.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"sync"
 	"syscall"
 	"time"
 
@@ -59,13 +60,13 @@ type Spammer struct {
 	numAccounts int
 
 	// keep track of variables shared across issuers
-	tracker *tracker
+	tracker  *tracker
+	issuerWg *sync.WaitGroup
 }
 
 func NewSpammer(sc *Config, sh SpamHelper) (*Spammer, error) {
 	// Log Zipf participants
 	zipfSeed := rand.New(rand.NewSource(0)) //nolint:gosec
-	tracker := &tracker{}
 	balance, err := sh.LookupBalance(sc.authFactory.Address())
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


throughput/tracker.go
```diff
@@ -17,14 +17,23 @@ import (
...
 
-	sent atomic.Int64
+	sent   atomic.Int64
+	ticker *time.Ticker
+
+	cancel context.CancelFunc
+	done   chan struct{}
+}
+
+func NewTracker() *tracker {
```
aaronbuchwald, line 33: Could we unexport the constructor and all of the functions here?

We're inconsistent prior to this change, but think we can clean this up here as none of this is expected to be used outside of this package

throughput/tracker.go
```diff
@@ -45,21 +54,24 @@ func (t *tracker) logResult(txID ids.ID, result *chain.Result) {
 	}
 }
 
-func (t *tracker) logState(ctx context.Context, cli *jsonrpc.JSONRPCClient) {
+func (t *tracker) Start(ctx context.Context, cli *jsonrpc.JSONRPCClient) {
```
aaronbuchwald, line 57: maybe rename to `startPeriodicLog` ?

RodrigoVillar: Done

throughput/tracker.go
RodrigoVillar: Done

throughput/spam.go
```diff
@@ -143,19 +145,21 @@ func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbo
 		issuer.Start(cctx)
 	}
 
-	// set logging
-	s.tracker.logState(cctx, issuers[0].cli)
+	// start logging
+	s.tracker.Start(ctx, cli)
```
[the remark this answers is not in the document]
RodrigoVillar: Yes; `cancel()` is used to stop the issuers as seen here: https://github.com/ava-labs/hypersdk/blob/53e5d2cce28ad71a0bf1569ce5e1b527d6fa2ae3/throughput/spam.go#L151

I switched from passing in `cctx` to `ctx` to the tracker since we should continue tracking chain metrics even when the issuers are returning (i.e. tracker is unaffected by the `cancel()` call). I plan on cleaning up the issuer logic so we could remove `context.WithCancel` in a future PR.

throughput/spam.go
```diff
@@ -143,19 +145,21 @@ func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbo
 		issuer.Start(cctx)
 	}
 
-	// set logging
-	s.tracker.logState(cctx, issuers[0].cli)
+	// start logging
+	s.tracker.startPeriodicLog(ctx, cli)
```
aaronbuchwald, line 149: Do we still need `cctx` created above if we're not longer using it here to stop the tracker?

aaronbuchwald (approved):

PR #1945, opened 2025-02-27
Clean Up Issuer/Spammer

RodrigoVillar:
This PR achieves the following:

- Creates a constructor for the `pacer`
- Adds a `incrementInflight()` method to the tracker and updates `logResult()` so that the `tracker` updates `inflight` instead of the caller.
- Removes the `ws` recovery logic from the `issuer` and makes sure that metrics are updated only if a tx is successfully sent.
- Adds a constructor to `issuer`
- Moves `ctrl-c` logic from `broadcast` to `spam`  
- Removed `stop()` from `tracker`


throughput/spam.go
```diff
@@ -99,6 +101,10 @@ func NewSpammer(sc *Config, sh SpamHelper) (*Spammer, error) {
 // [terminate] if true, the spammer will stop after reaching the target TPS.
 // [symbol] and [decimals] are used to format the output.
 func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbol string) error {
+	// make sure we can exit gracefully & return funds
+	s.signals = make(chan os.Signal, 2)
+	signal.Notify(s.signals, syscall.SIGINT, syscall.SIGTERM)
```
aaronbuchwald, line 106: Rather than maintain a separate channel for this, we should create a child context using `WithCancel` that will be cancelled if receive a signal.

See examples of performing cleanup at the top level:
- https://github.com/ava-labs/avalanchego/blob/667c6b02315cfc86d9ff7a46440e891f47c04d2f/app/app.go#L101
- https://github.com/ava-labs/subnet-evm/blob/a6a3d338809940db47ffc94d418b5abb61427f1e/cmd/simulator/load/loader.go#L133

For almost every application, we want it to handle ctrl+c gracefully and terminate. Handling this at the top-level makes sure that the top-level context is cancelled and everything flowing from that should be informed to cancel and clean itself up. Then it's up to each component to make sure that it correctly respects the context and handles any necessary cleanup.

throughput/spam.go
```diff
@@ -142,21 +148,20 @@ func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbo
...
-	err = s.broadcast(cctx, sh, factories, issuers, feePerTx, terminate)
-	cancel()
-	if err != nil {
+	if err := s.broadcast(cctx, sh, factories, issuers, feePerTx, terminate); err != nil {
 		s.tracker.stop()
 		return err
 	}
 
 	// Wait for all issuers to finish
+	cancel()
```
aaronbuchwald, line 164: Could we add a comment explaining why we cancel here? Ik there's more to do to clean this up, but it's a bit non-obvious how this is used to clean up imo.

throughput/spam.go
```diff
@@ -259,7 +260,7 @@ func (s Spammer) broadcast(
 		case <-ctx.Done():
 			stop = true
 			utils.Outf("{{yellow}}context canceled{{/}}\n")
-		case <-signals:
+		case <-s.signals:
```
aaronbuchwald, line 263: We should be able to remove this and just rely on `<-ctx.Done()`

throughput/spam.go
```diff
@@ -282,23 +283,23 @@ func (s *Spammer) logZipf(zipfSeed *rand.Rand) {
...
-				i:       len(issuers),
-				cli:     cli,
-				ws:      webSocketClient,
-				parser:  parser,
-				uri:     s.uris[i],
-				tracker: s.tracker,
-				wg:      s.issuerWg,
-			}
+			issuer := newIssuer(
+				id,
```
aaronbuchwald, line 294: Why are we replacing `i: len(issuers)` with an `id` that is different for each issuer? Seems like `index` is an appropriate name, but maybe that's not quite right.

aaronbuchwald (changes requested):

throughput/spam.go
RodrigoVillar: Changed `id` to `index`

throughput/issuer.go
```diff
@@ -62,61 +65,32 @@ func (i *issuer) Start(ctx context.Context) {
...
-			utils.Outf("{{green}}re-created closed issuer:{{/}} %d\n", i.i)
-		}
-
-		// If issuance fails during retry, we should fail
-		return i.ws.RegisterTx(tx)
+		utils.Outf("{{orange}}failed to register tx:{{/}} %v\n", err)
+		return fmt.Errorf("failed to register tx: %w", err)
 	}
+	i.outstandingTxs.Add(1)
+	i.tracker.incrementInflight()
```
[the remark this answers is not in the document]
RodrigoVillar: A `tx` is inflight and outstanding only if it was registered via `ws`. If we updated the metrics prior to the `ws` call, then we could have the case where we mark a tx as inflight + outstanding but it was never registered (i.e. `i.ws.RegisterTx(tx) == nil`)

throughput/issuer.go
```diff
@@ -7,46 +7,49 @@ import (
...
-	"github.com/ava-labs/hypersdk/api/jsonrpc"
 	"github.com/ava-labs/hypersdk/api/ws"
 	"github.com/ava-labs/hypersdk/chain"
-	"github.com/ava-labs/hypersdk/pubsub"
 	"github.com/ava-labs/hypersdk/utils"
 )
 
 type issuer struct {
-	i      int
+	id     int
```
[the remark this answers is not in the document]
RodrigoVillar: Done

throughput/spam.go
RodrigoVillar: Done

throughput/spam.go
RodrigoVillar: Done

throughput/spam.go
RodrigoVillar: Updated PR to use the cleanup style mentioned above. Currently, if a user signals to exit, `cancel()` will be triggered and the process to return funds will begin. As a result, I decided to let `returnFunds()` use the parent context `ctx` instead of the context we create `cctx` since passing in `cctx` will result in `returnFunds()` doing nothing due to context expiration.

throughput/spam.go
aaronbuchwald: It seems we're using `cancel` to accomplish two different things here:
1. Cancel if we receive a user interrupt
2. Cancel to signal to the issuers/tracker to terminate

What do you think of separating the throughput logic that relies on this second cancellation into its own function with a locally scoped context? The separate function can create and cancel its own context to make sure it cleans up (ideally it would also ensure all goroutines exit).

This should make it more obvious what this second `cancel()` call is cancelling as opposed to treating it identically to a user interrupt.

throughput/spam.go
```diff
@@ -59,6 +59,8 @@ type Spammer struct {
 	// Number of accounts
 	numAccounts int
 
+	signals chan os.Signal
```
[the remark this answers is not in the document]
RodrigoVillar: Done

throughput/spam.go
```diff
@@ -97,16 +99,27 @@ func NewSpammer(sc *Config, sh SpamHelper) (*Spammer, error) {
...
+	// make sure we can exit gracefully & return funds
+	s.signals = make(chan os.Signal, 2)
+	signal.Notify(s.signals, syscall.SIGINT, syscall.SIGTERM)
+
+	cctx, cancel := context.WithCancel(ctx)
+	go func() {
+		<-s.signals
+		utils.Outf("{{yellow}}received interrupt signal{{/}}\n")
+		cancel()
+	}()
```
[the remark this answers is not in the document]
RodrigoVillar: Done

throughput/spam.go
RodrigoVillar: I added `run` which creates its own context, spins up the tracker and issuer, and starts `broadcast`. Upon returning, we have that the `tracker`/`issuer` start cleaning up.

throughput/spam.go
```diff
@@ -128,48 +140,56 @@ func (s *Spammer) Spam(ctx context.Context, sh SpamHelper, terminate bool, symbo
...
 
-	// broadcast transactions
-	err = s.broadcast(cctx, sh, factories, issuers, feePerTx, terminate)
-	cancel()
+	maxUnits, err = chain.EstimateUnits(ruleFactory.GetRules(time.Now().UnixMilli()), actions, s.authFactory)
 	if err != nil {
-		s.tracker.stop()
 		return err
 	}
+	return s.returnFunds(ctx, cli, maxUnits, sh, accounts, factories, symbol)
```
aaronbuchwald, line 159:
```suggestion
		// Use the original context, so that we attempt to return funds before exiting after
		// receiving user interrupt
		return s.returnFunds(ctx, cli, maxUnits, sh, accounts, factories, symbol)
```

aaronbuchwald (dismissed):

throughput/pacer.go
```diff
@@ -20,10 +20,15 @@ type pacer struct {
...
-	p.done = make(chan struct{})
+func newPacer(ws *ws.WebSocketClient, max int) *pacer {
+	return &pacer{
+		ws:       ws,
+		inflight: make(chan struct{}, max),
+		done:     make(chan struct{}),
+	}
+}
 
+func (p *pacer) Run(ctx context.Context) {
```
aaronbuchwald, line 31: (not necessary for this PR) un-export if not needed externally and for un-exported type

PR #1950, opened 2025-03-03
Add Processor Benchmarks

RodrigoVillar:
This PR adds a benchmark functions which tests the processor execution time of `N` blocks. This PR also adds a helper function which, when called, generates and returns `N` valid execution blocks.

### DIFF
--- a/chain/chaintest/block.go
+++ b/chain/chaintest/block.go
@@ -4,11 +4,19 @@
 package chaintest
 
 import (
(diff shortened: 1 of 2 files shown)


aaronbuchwald (commented):
We should benchmark a non-empty and non-trivial workload ie. real signatures + state diffs (ideally configurable size / workload)

PR #1954, opened 2025-03-05
Add Genesis To Chain

RodrigoVillar:
This PR moves the genesis state diff generation from the `vm` package to the `chain` package.

### DIFF
--- a/chain/dependencies.go
+++ b/chain/dependencies.go
@@ -7,7 +7,9 @@ import (
 	"context"
 
 	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/trace"
 	"github.com/ava-labs/avalanchego/utils/set"
+	"github.com/ava-labs/avalanchego/x/merkledb"
 
 	"github.com/ava-labs/hypersdk/codec"
 	"github.com/ava-labs/hypersdk/fees"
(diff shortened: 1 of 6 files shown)


chain/dependencies.go
```diff
@@ -31,6 +33,11 @@ type Mempool interface {
 	FinishStreaming(context.Context, []*Transaction) int
 }
 
+type Genesis interface {
+	InitializeState(ctx context.Context, tracer trace.Tracer, mu state.Mutable, balanceHandler BalanceHandler) error
+	GetStateBranchFactor() merkledb.BranchFactor
```
tsachiherman, line 38: I don't see where you're using `GetStateBranchFactor`. Did I miss anything ?

chain/genesis.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/ava-labs/hypersdk/fees"
+	"github.com/ava-labs/hypersdk/state"
+	"github.com/ava-labs/hypersdk/state/tstate"
+
+	internalfees "github.com/ava-labs/hypersdk/internal/fees"
+)
+
+// NewGenesisStateDiff returns both a block and a view that represents the
+// genesis state transition.
+func NewGenesisStateDiff(
```
aaronbuchwald, line 27: Can we rename this to `NewGenesisCommit` ?

RodrigoVillar: Done

chain/dependencies.go
RodrigoVillar: Removed this method from the `Genesis` interface

aaronbuchwald (approved):

Issue #1910, opened 2025-02-03
Benchmark Processor


RodrigoVillar:
Measure execution time for `N` blocks

PR #1955, opened 2025-03-05
Add Block Benchmark Suite

RodrigoVillar:
This PR adds a benchmarking test suite for benchmarking the processor execution time of a list of blocks. This test suite can be used for benchmarking both the HyperSDK itself and any custom VM which wants to benchmark its own block execution time. 

As part of the test suite, users are required to pass in a `BlockBenchmarkHelper` which when called, returns the following:

- The genesis to be used
- A `TxListGenerator` which, when called, produces a list of valid transactions.

An implementation of `TxListGenerator` can be used to benchmark the HyperSDK's pessimistic concurrency model against various state access distributions (i.e. TXs with no conflicting state accesses, TXs with all conflicting state accesses, TXs with state accesses under a Zipf distribution).

An example of how benchmarking would look like for the HyperSDK (with empty blocks) is as follows:

```golang
func BenchmarkProcessorExecuteEmptyBlocks(b *testing.B) {
	test := &chaintest.BlockBenchmark{
		MetadataManager:      metadataManager,
		BalanceHandler:       balance.NewPrefixBalanceHandler(metadataManager.FeePrefix()),
		AuthEngines:          auth.DefaultEngines(),
		RuleFactory:          &genesis.ImmutableRuleFactory{Rules: rules},
		BlockBenchmarkHelper: chaintest.NoopBlockBenchmarkHelper,
		Config: chain.Config{
			TargetBuildDuration:       100 * time.Millisecond,
			TransactionExecutionCores: 4,
			StateFetchConcurrency:     4,
			TargetTxsSize:             1.5 * units.MiB,
		},
		NumBlocks:      1_000,
		NumTxsPerBlock: 16,
	}
	test.Run(context.Background(), b)
}
```


chain/chaintest/block.go
```diff
@@ -42,3 +64,271 @@ func GenerateEmptyExecutedBlocks(
...
+	numBlocks uint64,
+	numOfTxsPerBlock uint64,
+) ([]*chain.ExecutionBlock, error) {
+	var timestampOffset int64
+	switch numOfTxsPerBlock {
+	case 0:
+		timestampOffset = rules.GetMinEmptyBlockGap()
+	default:
+		timestampOffset = rules.GetMinBlockGap()
+	}
```
aaronbuchwald, line 121: Can we handle this with a single if block instead of a switch? 

Not identical, but this is a case of "unnecessary else" - https://github.com/uber-go/guide/blob/master/style.md#unnecessary-else

chain/chaintest/block.go
```diff
@@ -42,3 +64,271 @@ func GenerateEmptyExecutedBlocks(
...
+		r.NoError(err)
+
+		factory := auth.NewED25519Factory(pk)
+		factories = append(factories, factory)
+
+		customAllocs = append(customAllocs, &genesis.CustomAllocation{
+			Address: factory.Address(),
+			Balance: uint64(allocBalance),
+		})
+	}
```
aaronbuchwald, line 271: Should we leave the auth factories behind the generator interface? This is a VM-level detail that will have a non-trivial performance impact.

chain/chaintest/block.go
```diff
@@ -42,3 +64,271 @@ func GenerateEmptyExecutedBlocks(
...
+		)
+		if err != nil {
+			return nil, err
+		}
+		parentID = statelessBlock.GetID()
+
+		blk := chain.NewExecutionBlock(statelessBlock)
+		executionBlocks[i] = blk
+	}
+	return executionBlocks, nil
```
aaronbuchwald, line 202: This code is getting duplicated across a growing number of places - can we make this more re-usable?

chain/chaintest/block.go
```diff
@@ -42,3 +64,271 @@ func GenerateEmptyExecutedBlocks(
...
+				view,
+				blocks[j],
+				false,
+			)
+			r.NoError(err)
+			view = outputBlock.View
+		}
+		view = db
+	}
+}
```
aaronbuchwald, line 334: Could we add a comment explaining the handling of views here?

chain/chaintest/block.go
```diff
@@ -42,3 +64,271 @@ func GenerateEmptyExecutedBlocks(
...
+	var view merkledb.View
+	view = db
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		for j := 0; j < int(test.NumOfBlocks); j++ {
+			outputBlock, err := processor.Execute(
+				ctx,
+				view,
+				blocks[j],
+				false,
```
aaronbuchwald, line 327: Why use `false` here? In normal operation, this will be set to true and we will need to check the expiry window

chain/chaintest/block.go
aaronbuchwald: For DSMR - we need the ability to assemble a block given a parent state / block, metadata for the next block (assigned from DSMR wrapper block), and a set of transactions that should be included. Maybe we can implement an `Assembler` and use that functionality here

chain/chaintest/block.go
```diff
@@ -42,3 +64,271 @@ func GenerateEmptyExecutedBlocks(
...
+
+	switch test.AuthVerificationCores {
+	case 0, 1:
+		processorWorkers = workers.NewSerial()
+	default:
+		processorWorkers = workers.NewParallel(
+			test.AuthVerificationCores,
+			numOfAuthVerificationJobs,
+		)
+	}
```
aaronbuchwald, line 256: Can we use a simple if condition instead of a switch since there are only two options?

chain/chaintest/block.go
aaronbuchwald, line 271: Hmm - this limits the usability of this package to VMs that use the default genesis / `CustomAllocation` types.

It seems like the caller should define their application logic including the genesis, transactions (auth/actions), and pass in the components they would have control over as opposed to making this a responsibility of the benchmark tool

aaronbuchwald (changes requested):

chain/chaintest/block.go
RodrigoVillar: Done

chain/chaintest/block.go
RodrigoVillar: Changed to `true`.

chain/chaintest/block.go
RodrigoVillar: Done

RodrigoVillar:
Regarding https://github.com/ava-labs/hypersdk/pull/1955#discussion_r1982287099 and https://github.com/ava-labs/hypersdk/pull/1955#discussion_r1982266901, I've decided to switch from users provided a tx-based helper (`TxListGenerator`) to a block-based helper (`BlockBenchmarkHelper`). `BlockBenchmarkHelper` still produces the TX lists required for `GenerateExecutionBlocks()` but in addition, it has the method `GenerateGenesis` which produces the genesis that the benchmark tests will use to generate the genesis commit. I found `GenerateGenesis` to be an improvement for the following reasons:

- Since `GenerateGenesis` returns an interface, we have that the tests are more VM agnostic (i.e. no longer using `genesis.DefaultGenesis`
- In `GenerateGenesis`, the test user can also instantiate any factories required for `GenerateTxList`, instead of having the benchmark tests produce the factories on their behalf.

chain/processor_test.go
```diff
@@ -509,6 +516,282 @@ func TestProcessorExecute(t *testing.T) {
 	}
 }
 
+func BenchmarkExecuteEmptyBlocks(b *testing.B) {
+	test := &chaintest.BlockBenchmark{
+		MetadataManager: metadata.NewDefaultManager(),
+		BalanceHandler:  &mockBalanceHandler{},
+		AuthVM: &chaintest.TestAuthVM{
+			GetAuthBatchVerifierF: func(authTypeID uint8, cores int, count int) (chain.AuthBatchVerifier, bool) {
```
tsachiherman, line 524: would it make sense to have a single copy of the `GetAuthBatchVerifierF` function and use it across these three locations ?

chain/processor_test.go
```diff
@@ -509,6 +516,282 @@ func TestProcessorExecute(t *testing.T) {
...
+		RuleFactory:          &genesis.ImmutableRuleFactory{Rules: genesis.NewDefaultRules()},
+		BlockBenchmarkHelper: &serialTxBlockHelper{},
+		Config:               chain.NewDefaultConfig(),
+		NumOfBlocks:          1_000,
+		NumOfTxsPerBlock:     16,
+	}
+	test.Run(context.Background(), b)
+}
+
+func BenchmarkExecuteBlocksZipfTxs(b *testing.B) {
```
tsachiherman, line 588: given that all these benchmarks are almost identical, I'd opt into having a single benchmark containing multiple sub-benchmarks.

RodrigoVillar: I agree, switched to using table benchmarks (looks much better as well!)

chain/processor_test.go
RodrigoVillar: Going to wait until #1960 is merged to address this.

chain/chaintest/block.go
RodrigoVillar: As per offline convo, it's not clear where an implementation of `Assembler` would reduce code duplication (besides  `GenerateExecutionBlocks()`; we can revisit this at a later time (perhaps after DSMR?)

chain/chaintest/block.go
```diff
@@ -42,3 +61,270 @@ func GenerateEmptyExecutedBlocks(
...
+				true,
+			)
+			r.NoError(err)
+			// we update [parentView] so that the view produced by [outputBlock]
+			// is the parent view of the next block
+			parentView = outputBlock.View
+		}
+		// we reset [parentView] to the genesis view
+		parentView = db
+	}
```
tsachiherman, line 329: Could you please add the following line after the loop ?
```golang
b.ReportMetric(float64(test.NumOfBlocks*test.NumOfTxsPerBlock*uint64(time.Second))/float64(int(b.Elapsed())*b.N), "tps")
```

this would give us the effective tps.

chain/chaintest/block.go
```diff
@@ -104,3 +111,270 @@ func GenerateTestExecutedBlocks(
...
+				actions,
+			)
+
+			return txData.Sign(factory)
+		}
+
+		txs, err := txListGenerator(numOfTxsPerBlock, txGenerator)
+		if err != nil {
+			return nil, err
+		}
```
aaronbuchwald, line 203: Why do we need both a `txGenerator` and `txListGenerator` that takes the original generator in?

chain/chaintest/block.go
```diff
@@ -104,3 +111,270 @@ func GenerateTestExecutedBlocks(
...
+		if err := tsv.Insert(ctx, chain.TimestampKey(metadataManager.TimestampPrefix()), binary.BigEndian.AppendUint64(nil, uint64(timestamp))); err != nil {
+			return nil, err
+		}
+		if err := tsv.Insert(ctx, chain.FeeKey(metadataManager.FeePrefix()), feeManager.Bytes()); err != nil {
+			return nil, err
+		}
+		tsv.Commit()
+
+		root, err := parentCtx.view.GetMerkleRoot(ctx)
```
aaronbuchwald, line 238: Can we rename this `parentRoot` for clarity? I thought this was a bug using the wrong root on first read

chain/chaintest/block.go
```diff
@@ -104,3 +111,270 @@ func GenerateTestExecutedBlocks(
 	}
 	return executedBlocks
 }
+
+// TXGenerator is a function that generates a valid TX that contains
+// [actions]. [factory] will be the signer of the TX.
+type TxGenerator func(actions []chain.Action, factory chain.AuthFactory) (*chain.Transaction, error)
```
aaronbuchwald, line 117: This seems to be more of a constructor than a generator. Generator implies that it generates some type of sequence whereas this is used to combine `chain.Base`, `[]chain.Action`, and `chain.AuthFactory` to create a valid HyperSDK transaction instance.

aaronbuchwald: Would it suffice instead to allow users to define a function that will provide `chain.Base` to be used at a given point in time and then have a single definition of how to combine these together? I think that would suffice and make it more clear what the caller has the opportunity to override.

chain/chaintest/block.go
```diff
@@ -104,3 +111,270 @@ func GenerateTestExecutedBlocks(
...
+	BlockBenchmarkHelper BlockBenchmarkHelper
+
+	Config                chain.Config
+	AuthVerificationCores int
+
+	// [NumOfBlocks] is set as a hyperparameter to avoid using [b.N] to generate
+	// the number of blocks to run the benchmark on.
+	NumOfBlocks      uint64
+	NumOfTxsPerBlock uint64
+}
```
aaronbuchwald, line 288:
```suggestion
	Config                chain.Config
	AuthVerificationCores int
	
	BlockBenchmarkHelper BlockBenchmarkHelper

	// [NumOfBlocks] is set as a hyperparameter to avoid using [b.N] to generate
	// the number of blocks to run the benchmark on.
	NumOfBlocks      uint64
	NumOfTxsPerBlock uint64
}
```

chain/chaintest/block.go
```diff
@@ -104,3 +111,270 @@ func GenerateTestExecutedBlocks(
...
+		test.BalanceHandler,
+		&validitywindowtest.MockTimeValidityWindow[*chain.Transaction]{},
+		metrics,
+		test.Config,
+	)
+
+	genesis, txListGenerator, err := test.BlockBenchmarkHelper(test.NumOfTxsPerBlock)
+	r.NoError(err)
+
+	genesisExecutionBlk, genesisView, err := chain.NewGenesisCommit(
```
aaronbuchwald, line 329: Looks like this is useful 🙏

chain/chaintest/block.go
aaronbuchwald: Not at all necessary for this PR, but agree with @tsachiherman some more detailed metrics would be great.

I think blocks/s and txs/s would be nice to add.

chain/dependencies.go
```diff
@@ -34,6 +35,7 @@ type Mempool interface {
 
 type Genesis interface {
 	InitializeState(ctx context.Context, tracer trace.Tracer, mu state.Mutable, balanceHandler BalanceHandler) error
+	GetStateBranchFactor() merkledb.BranchFactor
```
aaronbuchwald, line 38: Where is this newly needed in this PR?

chain/processor_test.go
```diff
@@ -516,6 +518,215 @@ func TestProcessorExecute(t *testing.T) {
...
+			return nil, nil, err
+		}
+		factory := auth.NewED25519Factory(pk)
+		factories[i] = factory
+		customAllocs[i] = &genesis.CustomAllocation{
+			Address: factory.Address(),
+			Balance: allocAmount,
+		}
+	}
+	return factories, genesis.NewDefaultGenesis(customAllocs), nil
```
aaronbuchwald, line 727: Could we rename to `numAccounts` ? I don't have a link, but think we generally avoid using preposition in variable names since it doesn't provide added clarity ie `numAccounts` instead of `numOfAccounts`

chain/processor_test.go
```diff
@@ -516,6 +518,215 @@ func TestProcessorExecute(t *testing.T) {
...
+				},
+				NumOfBlocks:      1_000,
+				NumOfTxsPerBlock: bm.numOfTxsPerBlock,
+			}
+			benchmark.Run(context.Background(), b)
+		})
+	}
+}
+
+func parallelTxsBlockBenchmarkHelper(numOfTxs uint64) (genesis.Genesis, chaintest.TxListGenerator, error) {
```
aaronbuchwald, line 570: Should we change this variable name from `numOfTxs` to `numAccounts` ?

Also same comment as previously to drop unnecessary prepositions from variable names

chain/processor_test.go
```diff
@@ -516,6 +518,215 @@ func TestProcessorExecute(t *testing.T) {
...
+
+			txs[i] = tx
+		}
+		return txs, nil
+	}
+
+	return genesis, txListGenerator, nil
+}
+
+func zipfTxsBlockBenchmarkHelper(numOfTxsPerBlock uint64) (genesis.Genesis, chaintest.TxListGenerator, error) {
```
aaronbuchwald, line 657: Providing each of these is awesome 🙏

examples/morpheusvm/tests/blocks/block_benchmark_helpers.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/ava-labs/hypersdk/auth"
+	"github.com/ava-labs/hypersdk/chain"
+	"github.com/ava-labs/hypersdk/chain/chaintest"
+	"github.com/ava-labs/hypersdk/codec"
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/examples/morpheusvm/actions"
+	"github.com/ava-labs/hypersdk/genesis"
+)
+
+func parallelTxsBlockBenchmarkHelper(numOfTxsPerBlock uint64) (genesis.Genesis, chaintest.TxListGenerator, error) {
```
aaronbuchwald, line 19: Could we move all of these helpers into `block_execution_test.go` ?

Since they are all used in a single place, I think it's clearer to put in a single file and use only there.

examples/morpheusvm/tests/blocks/block_benchmark_helpers.go
```diff
@@ -0,0 +1,132 @@
...
+
+			txs[i] = tx
+		}
+
+		return txs, nil
+	}
+	return gen, txListGenerator, nil
+}
+
+func createGenesis(numOfFactories uint64, allocAmount uint64) ([]chain.AuthFactory, genesis.Genesis, error) {
```
aaronbuchwald, line 116: same comment on the variable name

examples/morpheusvm/tests/blocks/block_benchmark_helpers.go
```diff
@@ -0,0 +1,132 @@
...
+				return nil, err
+			}
+
+			txs[i] = tx
+		}
+
+		return txs, nil
+	}
+	return gen, txListGenerator, nil
+}
```
aaronbuchwald, line 114: Not necessary for this PR / premature - this logic is very similar to what we use in the processor benchmarks. Even if there's a counterexample VM where it won't be possible to use, is there a portion of this code that we _can_ make re-usable, so parallelizable VMs can easily benchmark serial, parallel, and zipf distribution workloads?

aaronbuchwald (changes requested):

chain/dependencies.go
RodrigoVillar: Removed `GetStateBranchFactor()` from interface.

examples/morpheusvm/tests/blocks/block_benchmark_helpers.go
RodrigoVillar: Done

chain/processor_test.go
RodrigoVillar: Done

examples/morpheusvm/tests/blocks/block_benchmark_helpers.go
RodrigoVillar: Done

chain/chaintest/block.go
RodrigoVillar: `txListGenerator` is the only necessary component out of the two (VM-specific which needs to be injected by the test caller). `txGenerator` is a helper function provided by `BlockBenchmark` which allows users to create a valid TX without having to do repetitive logic (i.e. compute fees, create `txData`, etc.)

Since `txGenerator` is not necessary, I'm open to removing it and instead having each `txListGenerator` create its own valid tx.

chain/processor_test.go
RodrigoVillar: Changed to `numTxsPerBlock`. 

Also modified the `TxListGenerator` so that instead of taking in its own argument for the number of txs it should produce, it produces `numTxsPerBlock` for each call (i.e. we define how many txs it produces when calling `BlockBenchmarkHelper`). This makes it easier to see that there's a bijective relationship between txs and their signers.

chain/transaction_test.go
```diff
@@ -37,6 +39,32 @@ func init() {
...
+	return m.deductError
+}
+
+func (*mockBalanceHandler) GetBalance(_ context.Context, _ codec.Address, _ state.Immutable) (uint64, error) {
+	panic("unimplemented")
+}
+
+func (*mockBalanceHandler) SponsorStateKeys(_ codec.Address) state.Keys {
+	return state.Keys{}
+}
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/chaintest/block.go
RodrigoVillar: Done

RodrigoVillar:
Removing the milestone from this PR since we already have https://github.com/ava-labs/hypersdk/issues/1910.

chain/processor_test.go
```diff
@@ -516,6 +520,222 @@ func TestProcessorExecute(t *testing.T) {
...
+		{
+			name:                 "blocks with txs whose state keys are zipf distributed",
+			blockBenchmarkHelper: zipfTxsBlockBenchmarkHelper,
+			numOfTxsPerBlock:     16,
+		},
+	}
+
+	metadataManager := metadata.NewDefaultManager()
+	// modify default rules to avoid test failures due to fee spikes
+	rules := genesis.NewDefaultRules()
```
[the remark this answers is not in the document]
aaronbuchwald: Yup this would be a great standard function to add for tests, throughput, etc.

We should set reasonable defaults for production.

chain/chaintest/block.go
```diff
@@ -104,3 +111,267 @@ func GenerateTestExecutedBlocks(
...
+			if err != nil {
+				return chain.Base{}, err
+			}
+
+			return chain.Base{
+				Timestamp: utils.UnixRMilli(timestamp, rules.GetValidityWindow()),
+				ChainID:   rules.GetChainID(),
+				MaxFee:    maxFee,
+			}, nil
+		}
```
aaronbuchwald, line 192: Is there ever a time we need to populate this differently?

aaronbuchwald (changes requested):

examples/morpheusvm/tests/blocks/block_execution_test.go
```diff
@@ -0,0 +1,208 @@
...
+			blockBenchmarkHelper: parallelTxsBlockBenchmarkHelper,
+		},
+		{
+			name:                 "transfer txs that all send to the same recipient",
+			blockBenchmarkHelper: serialTxsBlockBenchmarkHelper,
+		},
+		{
+			name:                 "transfer txs whose recipient is sampled from the zipf distribution",
+			blockBenchmarkHelper: zipfTxsBlockBenchmarkHelper,
+		},
```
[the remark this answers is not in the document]
RodrigoVillar: Done

chain/chaintest/block.go
RodrigoVillar: Also, the `TxBaseConstructor` needs to be created with each new block since it relies on `unitPrices` and `timestamp` which are dynamic, hence the need for a function to create `TxBaseConstructor`.

examples/morpheusvm/tests/blocks/block_execution_test.go
```diff
@@ -0,0 +1,208 @@
...
+				NumBlocks:             10,
+				NumTxsPerBlock:        5_000,
+			}
+
+			benchmark.Run(context.Background(), b)
+		})
+	}
+}
+
+func parallelTxsBlockBenchmarkHelper(numTxsPerBlock uint64) (genesis.Genesis, chaintest.TxListGenerator, error) {
```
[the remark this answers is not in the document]
RodrigoVillar: Done

RodrigoVillar:
TODO: switch from `TxBaseGenerator` to passing in timestamp, fees, etc.

RodrigoVillar:
> TODO: switch from `TxBaseGenerator` to passing in timestamp, fees, etc.

Done

chain/chaintest/block.go
```diff
@@ -104,3 +109,249 @@ func GenerateTestExecutedBlocks(
 	}
 	return executedBlocks
 }
+
+// TxListGenerator is a function that should return a list of valid TXs of
```
aaronbuchwald, line 113:
```suggestion
// TxListGenerator is a function that should return a list of valid txs of
```

can we use clearly named parameters in the function signature? We are referencing `numTxsPerBlock` here, but there's no such variable

chain/chaintest/block.go
```diff
@@ -104,3 +109,249 @@ func GenerateTestExecutedBlocks(
...
+	view      state.View
+	id        ids.ID
+	height    uint64
+	timestamp int64
+}
+
+// GenerateExecutionBlocks generates numBlocks execution blocks with
+// txsPerBlock transactions per block.
+//
+// Block production is a simplified version of Builder; we execute
```
aaronbuchwald, line 141: Can we remove the semicolon?

chain/processor_test.go
```diff
@@ -516,6 +520,240 @@ func TestProcessorExecute(t *testing.T) {
...
+	rules.WindowTargetUnits = fees.Dimensions{20_000_000, consts.MaxUint64, consts.MaxUint64, consts.MaxUint64, consts.MaxUint64}
+	rules.MaxBlockUnits = fees.Dimensions{20_000_000, consts.MaxUint64, consts.MaxUint64, consts.MaxUint64, consts.MaxUint64}
+	rules.MinUnitPrice = fees.Dimensions{1, 1, 1, 1, 1}
+
+	for _, bm := range benchmarks {
+		b.Run(bm.name, func(b *testing.B) {
+			benchmark := &chaintest.BlockBenchmark{
+				MetadataManager:      metadataManager,
+				BalanceHandler:       balance.NewPrefixBalanceHandler(metadataManager.FeePrefix()),
```
aaronbuchwald, line 561: This is incorrect, we need to use `metadata.DefaultMinimumPrefix` when using the default metadata manager w/ the current code.

examples/morpheusvm/tests/blocks/block_execution_test.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+		factory := auth.NewED25519Factory(pk)
+		factories[i] = factory
+		customAllocs[i] = &genesis.CustomAllocation{
+			Address: factory.Address(),
+			Balance: allocAmount,
+		}
+	}
+	return factories, genesis.NewDefaultGenesis(customAllocs), nil
+}
```
aaronbuchwald, line 212: (not necessary for this PR) How can we de-duplicate this code across chain and morpheusvm?

examples/morpheusvm/tests/blocks/block_execution_test.go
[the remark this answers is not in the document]
aaronbuchwald: ref https://github.com/ava-labs/hypersdk/pull/2013

aaronbuchwald (changes requested):

examples/morpheusvm/tests/blocks/block_execution_test.go
RodrigoVillar: #2013

chain/processor_test.go
RodrigoVillar: 19b65c5c3

chain/chaintest/block.go
RodrigoVillar: bc52f9185

internal/validitywindow/validitywindowtest/test_chain_index.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/ava-labs/avalanchego/database"
+	"github.com/ava-labs/avalanchego/ids"
+
+	"github.com/ava-labs/hypersdk/internal/emap"
+	"github.com/ava-labs/hypersdk/internal/validitywindow"
+)
+
+type MockChainIndex[T emap.Item] struct {
+	blocks map[ids.ID]validitywindow.ExecutionBlock[T]
+}
```
RodrigoVillar, line 18: This is a exposed duplicate of https://github.com/ava-labs/hypersdk/blob/b1c49d1371477ae0d46f69da9e4d3d49971a5f2f/internal/validitywindow/validitywindow_test.go#L427-L429

Ideally, we would have `validitywindow_test.go` import this (and get rid of the duplicate), but this would result in a import cycle similar to what we have in `chain` (i.e. `chain`/`chaintest`/`chain_test`). Regardless, a mock implementation of `ChainIndex` needs to exist because this is a requirement of creating a `TimeValidityWindow`

chain/chaintest/block.go
```diff
@@ -104,3 +109,249 @@ func GenerateTestExecutedBlocks(
...
+
+	processor := chain.NewProcessor(
+		trace.Noop,
+		&logging.NoLog{},
+		test.RuleFactory,
+		processorWorkers,
+		test.AuthEngines,
+		test.MetadataManager,
+		test.BalanceHandler,
+		&validitywindowtest.MockTimeValidityWindow[*chain.Transaction]{},
```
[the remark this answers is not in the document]
RodrigoVillar: 75f0e2d9f

chain/chaintest/block.go
```diff
@@ -104,3 +110,265 @@ func GenerateTestExecutedBlocks(
...
+	r.NoError(err)
+
+	numAuthVerificationJobs := 100
+
+	processorWorkers := workers.NewSerial()
+	if test.AuthVerificationCores > 1 {
+		processorWorkers = workers.NewParallel(test.AuthVerificationCores, numAuthVerificationJobs)
+	}
+
+	chainIndex := &validitywindowtest.MockChainIndex[*chain.Transaction]{}
```
aaronbuchwald, line 286: This is fine because we never need to fetch a processing block during this execution benchmark.

internal/validitywindow/validitywindowtest/test_chain_index.go
aaronbuchwald: Can we export `blocks` rather than providing a `Set` function?

aaronbuchwald: Let's make a follow up GitHub issue to remove the duplicate

aaronbuchwald (approved):

PR #1960, opened 2025-03-07
Decouple `auth` Engines from VM

RodrigoVillar:
This PR removes the `chain.AuthVM` type and instead adds the `Engines` type to the `chain` package. By decoupling `auth` engines from the `VM`, this makes testing easier as we no longer need to implement duplicates of `GetAuthBatchVerifier`; that is, we can write the following:

```golang
&chaintest.TestAuthEngines{
  GetAuthBatchVerifierF: auth.DefaultEngines().GetAuthBatchVerifier,
}
```

instead of:

```golang
&chaintest.TestAuthEngines{
  GetAuthBatchVerifierF: func(authTypeID uint8, cores int, count int) (chain.AuthBatchVerifier, bool) {
    bv, ok := auth.Engines()[authTypeID]
    if !ok {
	    return nil, false
    }
    return bv.GetBatchVerifier(cores, count), ok
  },
}
```


aaronbuchwald (approved):

Elvis339 (approved):
LGTM, but it would be helpful to have more documentation. Without context, the term `AuthEngines` isn't intuitive to understand what it represents or does.

aaronbuchwald (approved):

Issue #1964, opened 2025-03-10
Add Go Generate Check to CI


aaronbuchwald:
Add `go generate` check to CI to ensure that it does not produce a diff

Ex. adding this for canoto in AvalancheGo: https://github.com/ava-labs/avalanchego/commit/4f3f5e7baa744295c20e40449ce051dca035031c#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

We should add a single check for all of `go generate` rather than on a case by case basis: https://github.com/ava-labs/avalanchego/issues/3779

Elvis339:
@aaronbuchwald did you mean to add more checks? AFAIK we support canoto, protobuf and custom codec.

aaronbuchwald:
Ideally we'd add a check for all auto-generated code

Issue #1963, opened 2025-03-10
Bump to latest commit / release of AvalancheGo after ACP-118 is Merged


aaronbuchwald:
HyperSDK depends on https://github.com/ava-labs/avalanchego/pull/3394

PR #1967, opened 2025-03-11
ci: add canotogen check

Elvis339:
This patch is closing: https://github.com/ava-labs/hypersdk/issues/1964 and https://github.com/ava-labs/hypersdk/issues/1963

Additionally:
- Bump go version from 1.22.8 to 1.23.7: https://github.com/ava-labs/hypersdk/blob/8d5e588e6fe2b1b241dfd296b939da602119ccee/go.mod#L3
- Bump canoto https://github.com/ava-labs/hypersdk/blob/9d13590bbcbebc22ec8ceaa8601fc36cb7230a35/go.mod#L6
- Annotate [Base](https://github.com/ava-labs/hypersdk/blob/9a86d59430f5bc96fd2eb1f440a486fcb13931ed/chain/base.go#L38) struct with `canoto:"noatomic"` field
- Exclude auto generated canoto files from codeql: https://github.com/ava-labs/hypersdk/blob/9a86d59430f5bc96fd2eb1f440a486fcb13931ed/.github/codeql/codeql-config.yml


.github/workflows/check-clean-branch.sh
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+# See the file LICENSE for licensing terms.
+
+set -euo pipefail
+
+git add --all
+git update-index --really-refresh >> /dev/null
+
+# Exits if any uncommitted changes are found.
+git diff-index --quiet HEAD
```
aaronbuchwald, line 11: nit: add newline at end of file

.github/workflows/hypersdk-ci.yml
```diff
@@ -56,6 +56,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Set up Go
+        uses: ./.github/actions/install-go
+      - shell: bash
+        run: go generate -run "github.com/StephenButtolph/canoto/canoto" ./...
+      - shell: bash
+        run: .github/workflows/check-clean-branch.sh
```
aaronbuchwald, line 70: Could we simply use `go generate ./...` directly so that this is not canoto specific ?

aaronbuchwald (changes requested):

.github/workflows/hypersdk-ci.yml
Elvis339: Just to verify this would run all `go generate` directives in the codebase. If we have directives that don't relate to canoto, or if they take a long time to run, this might be inefficient for this specific check or even produce unrelated diffs.

The current approach using go generate -run "github.com/StephenButtolph/canoto/canoto" ./... is more targeted - it only runs the `go generate` directives that match the specified pattern (those related to canoto).

.github/workflows/check-clean-branch.sh
Elvis339: db518267de2d3c6deb266e4e0e1794622960236d

chain/base.go
```diff
@@ -35,6 +35,7 @@ type Base struct {
 	// If the fee is too low to pay all fees, the transaction will be dropped.
 	MaxFee uint64 `canoto:"fint64,3" json:"maxFee"`
 
+	//nolint:revive // Required unexported field for canoto serialization
```
Elvis339, line 38: [struct-tag](https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#struct-tag) doesn't support exclusion by pattern. The linter reported:
```
Error: chain/base.go:38:2: struct-tag: tag on not-exported field canotoData (revive)
	canotoData canotoData_Base `canoto:"noatomic"`
```

Since `canotoData` must remain unexported for proper canoto functionality, and the `canoto:"noatomic"` tag is used infrequently, it's better to add a `//nolint:revive` directive in these specific cases rather than disabling the entire lint rule project-wide. This maintains our code quality standards while accommodating the necessary exceptions.

tsachiherman (commented):
all looks reasonable. One thing that I'm not a big fan of is the storage of the `blockHex` inside the go file.
I realize that we need this check in order to verify backward compatibility. However, before making an official release, this is going to break over and over, without any other meaningful changes to the stateless_block.go. I think it would be cleaner to place this in an embedded file, but I wouldn't feel too bad about that either way.

chain/transaction_test.go
```diff
@@ -496,3 +503,41 @@ func equalTxData(r *require.Assertions, expected chain.TransactionData, actual c
 	}
 	r.Equal(expected.UnsignedBytes(), actual.UnsignedBytes(), msgAndArgs...)
 }
+
+// updateReferenceBlockData regenerates the reference hex and JSON files with the current implementation
+// Only runs when UPDATE_TEST_DATA=1 is set eg: UPDATE_TEST_DATA=1 go test ./chain/...
+func updateReferenceTxData() {
```
aaronbuchwald, line 509: Creating separate functions for this duplicates the code we are comparing the test vector against.

Would it make sense to write to file directly in the tests comparing against the hardcoded types ie. if running a normal test, perform the `require.Equal(...)` check, if this environment variable is specified, then skip the require check and overwrite the file.

I think this reduce the code duplication and make it a bit easier to read ie. overwrite this test vector we're comparing against this way of re-generating it here.

Separately, it seems potentially error prone to use this environment variable to overwrite the test vectors during the test run.

This means that if the environment variable is set incorrectly, then our tests would pass, overwrite the test vector, and unless we have a diff check after running the tests, we may not notice.

What do you think of separating the command to update the test vectors from the tests themselves?

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,2 @@
+paths-ignore:
```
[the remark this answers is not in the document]
Elvis339: Screenshot 2025-03-14 at 9 16 46 PM

chain/base.go
```diff
@@ -33,7 +35,8 @@ type Base struct {
 	// If the fee is too low to pay all fees, the transaction will be dropped.
 	MaxFee uint64 `canoto:"fint64,3" json:"maxFee"`
 
-	canotoData canotoData_Base
+	// Required unexported field for canoto serialization
+	canotoData canotoData_Base `canoto:"noatomic"` //nolint:revive
```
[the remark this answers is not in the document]
Elvis339: Types from the `sync` and `atomic` packages contain a `noCopy` marker (implemented via a special `sync.noCopy` struct) that signals these types should never be copied. 

When `Base` is passed by value here:
https://github.com/ava-labs/hypersdk/blob/79bcbeda4f56b31ae499cdee30cb04ce8d3fc7cd/chain/transaction.go#L48

Any `sync` or `atomic` fields within it get copied, breaking their internal synchronization mechanisms. This violates the `noCopy` constraint, which the linter has successfully caught.

chain/transaction_test.go
Elvis339: I've been weighing between generating test data using commands or scripts because that's generally how this project automates things. I decided to have this function so that everything would be in one file, to avoid the 'internalized' knowledge that you need to run a specific script or command and which command. I relied on review and diff to avoid silent errors.

After this, I'm not sure what value we get with this change - it seems to me we're just adding complexity. I'd revert this change.

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,2 @@
...
+  - '**/*.canoto.go'  # Exclude all files ending with .canoto.go
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/1967/commits/3d055f1f68a9653a4a03b6651ac0addea3eeaaf9

chain/base.go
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/1967/commits/b3c4e79a36d4df44b1d2258801d49c40b564e0da

chain/chaintest/testdata/signedTransaction.json
```diff
@@ -0,0 +1 @@
+{"id":"4WM1BjXa5qS8atNqHc2mCL3ZyyGHvu3Q7RoqEwV9urUKES7rx","actions":["0000000000000000010000000000000000000000000000000000000000000000000000000000ffffffffffffffffffffffffffffffff"],"auth":"00000000000000000101020300000000000000000000000000000000000000000000000000000000000001020300000000000000000000000000000000000000000000000000000000000000ffffffffffffffffffffffffffffffff","base":{"timestamp":1724315246000,"chainId":"SkB92YpWkhEux4a79vjwrXCiQZqxoUuVEi2YsWFCmt7Ta1Ed","maxFee":1234567}}
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/1967/commits/ccc1f0858e564554f62a296fc02fb09e4441dc2b

chain/transaction.go
```diff
@@ -49,6 +50,12 @@ func NewTxData(base Base, actions []Action) TransactionData {
 		Base:    base,
 		Actions: actions,
 	}
+	// Call CalculateCanotoCache so that equivalent blocks pass an equals check.
+	// Without calling this function, canoto's required internal field will cause equals
+	// checks to fail on otherwise identical blocks.
+	// TODO: remove after Canoto guarantees this to be set correctly on the read path (unmarshal)
+	// rather than only setting it on the write path.
+	txData.Base.CalculateCanotoCache()
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/1967/commits/7d639ea85ee4cb68e825a7d6f1d300c7c00ddbe1

.github/workflows/hypersdk-ci.yml
aaronbuchwald: Yes, if we are relying on `go generate` in the codebase, we should guarantee that all of our auto-generated code is up to date, not just canoto. This may take longer, but when would we want auto-generated code that can no longer be generated from the existing source code to stick around?

I don't think this is needed for this PR, but I definitely think we'd be better off with an overall `go generate` check and aligning everything on it.

internal/validitywindow/handler_test.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// Use a fixed seed for deterministic test results
+var seed = rand.New(rand.NewSource(42)) //nolint:gosec
```
aaronbuchwald, line 19:
```suggestion
var seed = rand.New(rand.NewSource(0)) //nolint:gosec
```

why 42?

aaronbuchwald: I don't think this will do what you want. If tests are run in a non-deterministic order or if you run a different subset of tests, then the order in which we read from `seed` will result in different random values.

It seems we only use this in `generateBlockChain(...)`, what if we create and use a deterministic seed within that function, so that generating a chain of the same length is deterministic?

internal/validitywindow/handler_test.go
```diff
@@ -116,19 +119,19 @@ func TestBlockFetcherHandler_FetchBlocks(t *testing.T) {
...
+func generateBlockChain(n int, containersPerBlock int, r *rand.Rand) map[uint64]ExecutionBlock[container] {
 	genesis := newExecutionBlock(0, 0, []int64{})
 	blks := make(map[uint64]ExecutionBlock[container])
 	blks[0] = genesis
 
 	for i := 1; i < n; i++ {
 		containers := make([]int64, containersPerBlock)
 		for j := 0; j < containersPerBlock; j++ {
-			containers[j] = rand.Int63n(int64(n)) //nolint:gosec
+			containers[j] = r.Int63n(int64(n))
```
aaronbuchwald, line 130: Rather than use `r` here, can we replace the use of randomness with a deterministic function?

aaronbuchwald: Also - isn't this generating an invalid sequence of blocks since the containers may have an expiry below the execution time of the block we are including them in?

internal/validitywindow/client_test.go
```diff
@@ -366,3 +368,21 @@ func (t *testNodeSampler) Sample(_ context.Context, num int) []ids.NodeID {
 	}
 	return t.nodes[:num]
 }
+
+// receive collects blocks from resultChan until either:
+// 1. The context is canceled (ctx.Done), or
+// 2. The channel is closed by the sender
+func receive(ctx context.Context, resultChan <-chan ExecutionBlock[container], receivedBlocks map[uint64]ExecutionBlock[container]) {
+loop:
```
aaronbuchwald, line 376: Can we return instead of using a label for the for loop here?

internal/validitywindow/client_test.go
```diff
@@ -238,17 +229,28 @@ func TestBlockFetcherClient_FetchBlocksChangeOfTimestamp(t *testing.T) {
...
-	resultChan := fetcher.FetchBlocks(ctx, tip, &minTimestamp)
+	fetchCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	resultChan := fetcher.FetchBlocks(fetchCtx, tip, &minTimestamp)
 	receivedBlocks := make(map[uint64]ExecutionBlock[container])
 
 	recvCount := 0
-	for blk := range resultChan {
-		receivedBlocks[blk.GetHeight()] = blk
+loop:
```
aaronbuchwald, line 238: (not necessary for this PR) - can we re-use `receive` here with a small modification since this is near duplicate code?

internal/validitywindow/handler_test.go
aaronbuchwald: Great reason

internal/validitywindow/client_test.go
RodrigoVillar, line 1: Q (also applies to changes in `internal/validitywindow`): is this change related to the go version bump/canoto? If not, could we update the PR description? 

Also open to moving this diff into a separate PR since this seems different from the overall goal of this PR.

Elvis339: Types updates are related to AvalancheGo and canoto update. Flaky tests are skipped for now - will fix properly in a follow-up PR with better networking layer implementation (yolo).

internal/validitywindow/handler_test.go
```diff
@@ -124,11 +123,14 @@ func generateBlockChain(n int, containersPerBlock int) map[uint64]ExecutionBlock
 	for i := 1; i < n; i++ {
 		containers := make([]int64, containersPerBlock)
 		for j := 0; j < containersPerBlock; j++ {
-			containers[j] = rand.Int63n(int64(n)) //nolint:gosec
+			containers[j] = int64(j + 1)
+			if containers[j] < int64(i) {
+				containers[j] = int64(i) + 1
+			}
```
aaronbuchwald, line 129: Will this result in duplicates? I believe to make this data correct we need to create test containers that depend on more than just the expiry time.

aaronbuchwald (approved):

Issue #1970, opened 2025-03-13
Reduce runtime of proactive syncer tests


aaronbuchwald:
ref https://github.com/ava-labs/hypersdk/pull/1911#discussion_r1994019367

PR #1972, opened 2025-03-14
Reduce runtime of proactive syncer test

Elvis339:
Closing: https://github.com/ava-labs/hypersdk/issues/1970

~14s average on main
~8s average now

### DIFF
--- a/internal/validitywindow/handler_test.go
+++ b/internal/validitywindow/handler_test.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"math/rand"
 	"testing"
-	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils"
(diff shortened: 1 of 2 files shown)


internal/validitywindow/handler_test.go
```diff
@@ -18,20 +17,21 @@ import (
...
 func TestBlockFetcherHandler_FetchBlocks(t *testing.T) {
 	tests := []struct {
-		name         string
-		setupBlocks  func() map[uint64]ExecutionBlock[container]
-		blockHeight  uint64
-		minTimestamp int64
-		wantErr      bool
-		wantBlocks   int
+		name          string
+		setupBlocks   func() map[uint64]ExecutionBlock[container]
```
aaronbuchwald, line 21: idea for separate / follow up PR - all the parameters are only controlling an interface with a single function, can we just define that function as needed?

aaronbuchwald (approved):

PR #1985, opened 2025-03-18
Test: validitywindow package

Elvis339:
1. Introduced `P2PClient` interface to represent AvalancheGo's concrete peer-to-peer client
https://github.com/ava-labs/hypersdk/blob/e787b37e0d81bb8c5f5ee21444100b3d8701105b/internal/typedclient/client.go#L23
2. Replaced concrete `*p2p.Client` with a `NetworkBlockFetcher` interface, improving:
    1. Decoupling; less direct dependency on implementation details
    2. Testability; easier to create mocks for unit testing and fine-grained control
    3. Clearer responsibility
https://github.com/ava-labs/hypersdk/blob/e787b37e0d81bb8c5f5ee21444100b3d8701105b/internal/validitywindow/client.go#L24
3. Created [adapter](https://github.com/ava-labs/hypersdk/blob/e787b37e0d81bb8c5f5ee21444100b3d8701105b/internal/validitywindow/adapter.go) in validitywindow package providing concrete implementation of `NetworkBlockFetcher` using the p2p network client 
4. Reduced unit-test runtime to ~1.5s by replacing time-based waiting with channel communication, providing finer-grained control and more deterministic test execution 
5. Handle: https://github.com/ava-labs/hypersdk/pull/1967#discussion_r1999227356 https://github.com/ava-labs/hypersdk/pull/1985/commits/6875932c2c136c90bc449fba5ba34fba02d3814a


RodrigoVillar (changes requested):

internal/validitywindow/handler_test.go
```diff
@@ -10,28 +10,27 @@ import (
...
-		expectErr     bool
-		wantBlocks    int
-		controlBuffer func() chan struct{}
+		name           string
+		setupBlocks    func() map[uint64]ExecutionBlock[container]
+		blockHeight    uint64
+		minTimestamp   int64
+		expectErr      bool
+		expectedBlocks int
+		controlChannel chan struct{} // control the number of returned blocks
```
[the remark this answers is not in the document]
Elvis339: Yes, makes sense.

Updated: https://github.com/ava-labs/hypersdk/pull/1985/commits/5251d187b3251a43bf002aa9ab192a6ffc18f19e

internal/validitywindow/handler_test.go
```diff
@@ -123,10 +127,8 @@ func generateBlockChain(n int, containersPerBlock int) map[uint64]ExecutionBlock
 	for i := 1; i < n; i++ {
 		containers := make([]int64, containersPerBlock)
 		for j := 0; j < containersPerBlock; j++ {
-			containers[j] = int64(j + 1)
-			if containers[j] < int64(i) {
-				containers[j] = int64(i) + 1
-			}
+			expiry := int64((i * containersPerBlock) + j + 1)
+			containers[j] = expiry
```
aaronbuchwald, line 131: I believe this still depends on the validity window to determine whether or not these containers will be valid ie. if we want this to produce valid historical data, then we need the containers to include some data beyond just the expiry time.

internal/validitywindow/handler_test.go
```diff
@@ -62,16 +61,21 @@ func TestBlockFetcherHandler_FetchBlocks(t *testing.T) {
...
-			},
-			blockHeight:  9,
-			minTimestamp: 3,
-			wantBlocks:   1,
-			controlBuffer: func() chan struct{} {
-				return make(chan struct{}, 1)
+				// Generate one more block than buffer capacity (4 vs 3)
+				// This forces a 4th retrieval that will block indefinitely,
+				// triggering context timeout in GetBlockHeight's select
+				return generateBlockChain(4, 3)
```
aaronbuchwald, line 72: Suggestion: this still depends on time ie. this test's execution path will be:

- fetch 3 blocks and then block indefinitely
- observe context timeout at 50ms mark

This means the test takes at least 50ms to execute and the other tests must take less than 50ms (should be easy, but may see different results and flaky behavior on lower resource GitHub runners).

An alternative here would be that after fetching N blocks, the test cancels the context and ensures that context cancellation triggers an early exit and correct handling ie. send a response with the blocks fetched so far.

This would make it so other tests never fail due to hitting a 50ms timeout (50ms may be a long enough delay to make this unlikely, but it still makes the tests fragile) and ensure that this test does not need to wait at least 50ms to hit the context error.

internal/validitywindow/client.go
```diff
@@ -23,24 +21,28 @@ type BlockParser[T Block] interface {
 	ParseBlock(ctx context.Context, blockBytes []byte) (T, error)
 }
 
+type NetworkBlockFetcher interface {
+	FetchBlocksFromPeers(ctx context.Context, nodeID ids.NodeID, request *BlockFetchRequest) (*BlockFetchResponse, error)
+}
```
aaronbuchwald, line 26: This interface is still influenced by the AvalancheGo p2p client implementation ie. from the block fetcher's perspective, it doesn't care about the `nodeID`, it just needs to grab it in order to send a request.

Would it make sense to specify exactly the interface we need? I think that would be either:

```golang
type NetworkBlockFetcher interface {
	FetchBlocks(ctx context.Context, blockHeight uint64, minTimestamp int64) ([][]byte, error)
}

// This function could take responsibility for checking that the parentIDs form the expected chain
type NetworkBlockFetcherT[T Block] interface {
	FetchBlocks(ctx context.Context, blockHeight uint64, minTimestamp int64) ([]T, error)
}
```

The pros and cons here imo would be:
Pros:
- mocking this is trivial and allows testing the application level code w/o details of the p2p client
- decouples from the p2p client and removes implementation details that influence how we define our interface

Cons:
- premature interface?
- still want to test against the production implementation of that interface ie (p2p client adapter of this)
- separation of responsibilities may make it harder to see the full e2e check we're performing against data from the network

Elvis339: Thank you for the comment Aaron. I wanted to clarify why I don’t see a strong need for further decoupling in this particular case:

- Minimal P2P Dependency
The goal here wasn’t to move away from the p2p client entirely but rather to define a minimal interface (FetchBlocksFromPeer) so we can keep the dependency small. If you look at the tests, the only real p2p-related piece is node sampling—everything else is already abstracted away(networking/serializaiton/deserialization), which makes testing pretty straightforward imo.

- Splitting the logic across multiple files can sometimes help, but in this instance I feel it may add more cognitive load than it resolves. The core logic here isn’t especially intuitive, so the documentation is crucial. Spreading it across different files might make it harder to see the bigger picture, rather than easier.

- The tests demonstrate that our approach is already easy to test. We’re not seeing any direct p2p complexity leak in—beyond sampling—so decoupling further wouldn’t necessarily improve testability or maintainability.

I’m open to discussing if there’s a specific use case or scenario you think further decoupling would solve, but as of now, I’m not convinced it adds enough value to outweigh the extra overhead and fragmentation. Let me know what you think or if there’s another angle I might be missing.

internal/validitywindow/handler_test.go
Elvis339: Amazing insight! Thank you for providing alternative, will implement.

Elvis339: After closer examination, it seems like I'd have to violate the non-error principle here https://github.com/ava-labs/hypersdk/blob/6875932c2c136c90bc449fba5ba34fba02d3814a/internal/validitywindow/handler.go#L99 by returning both context error and partially fetched blocks & update this one: https://github.com/ava-labs/hypersdk/blob/6875932c2c136c90bc449fba5ba34fba02d3814a/internal/validitywindow/handler.go#L64
I'm going to remove the expectErr check for the timeout test and add a comment explaining why we don't expect an error in this case. The test would still validate that we correctly retrieve the 3 blocks before timeout, which is the important part of the test's intent.
What we're really testing is the behavior under timeouts - that we get partial results (the first 3 blocks) when we hit a timeout, not specifically that an error is returned.

internal/validitywindow/handler_test.go
Elvis339: Looking at the tests, `generateBlockChain` is only used for testing the block fetcher handler functionality and client behavior, they are not focused on validity window content validation itself.

The actual validity window tests in `validitywindow_test.go` use a different approach with more focused test cases for expiry-based replay protection and timestamp verification.

internal/validitywindow/client.go
aaronbuchwald: Sounds good - yup tradeoff is definitely cognitive load of splitting it vs making the dependency dumb simple.

internal/validitywindow/handler_test.go
aaronbuchwald: For that first case - as long as we return a non-zero number of blocks, we'll return the partial response as soon as we hit an error as we want. Am I misunderstanding here?

Since `fetchBlocks` drops the error, we should send the partial response back.

aaronbuchwald (approved):
LGTM - left a few more comments, but think this is good to merge in as is

PR #1991, opened 2025-03-20
[test] Add `load` Framework

### LINKED ISSUE #1978: Create Framework For Testing TPS
We should transition the load generation tests from using the `throughput` package to a well-documented and defensible load generation framework.

RodrigoVillar:
This PR removes the current `throughput` package from the HyperSDK and replaces it with `load`, a component-based framework designed to orchestrate load tests of different intents (i.e. short burst, gradual load, etc.)

Closes #1978.


vm/config.go
```diff
@@ -33,10 +34,10 @@ type Config struct {
 
 func NewConfig() Config {
```
RodrigoVillar, line 35: For testing purposes, we modify the mempool size since it's almost likely that, with the current default values, the test would fail to due the mempool being full.

load/issuer.go
```diff
@@ -0,0 +1,103 @@
...
+	client  *ws.WebSocketClient
+	tracker Tracker[ids.ID]
+
+	lock     sync.Mutex
+	numOfTxs uint64
+	heardTxs uint64
+	stopped  bool
+}
+
+func NewDefaultIssuer(uri string, tracker Tracker[ids.ID]) (*DefaultIssuer, error) {
```
aaronbuchwald, line 35: (brainstorm) Is it possible to define an issuer factory that only needs to take in these parameters? I think that's all they should ever need, so wonder if we can provide further syntactic sugar

load/issuer.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		i.lock.Lock()
+		i.heardTxs++
+		// if we've heard the status of all of our transactions, return nil
+		if i.stopped && i.numOfTxs == i.heardTxs {
+			i.lock.Unlock()
+			return nil
+		}
+		i.lock.Unlock()
```
aaronbuchwald, line 78: can we move this to its own function so we can use lock / defer unlock syntax making this easier to read?

load/issuer.go
```diff
@@ -0,0 +1,103 @@
...
+	ErrIssuedAlreadyStopped = errors.New("issuer already stopped")
+)
+
+type DefaultIssuer struct {
+	client  *ws.WebSocketClient
+	tracker Tracker[ids.ID]
+
+	lock     sync.Mutex
+	numOfTxs uint64
+	heardTxs uint64
```
aaronbuchwald, line 31: Can we rename these `issuedTxs` and `receivedTxs` ?

load/short_burst_orchestrator.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+var (
+	ErrMismatchedGeneratorsAndIssuers = errors.New("number of generators and issuers must match")
+
+	_ Orchestrator = (*ShortBurstOrchestrator[any, any])(nil)
+)
+
+type ShortBurstOrchestratorConfig struct {
+	N       uint64
```
aaronbuchwald, line 21: Can we use a more descriptive variable name here?

load/short_burst_orchestrator.go
```diff
@@ -0,0 +1,105 @@
...
+func (o *ShortBurstOrchestrator[T, U]) Execute(ctx context.Context) error {
+	observerCtx, cancel := context.WithCancel(ctx)
+	o.cancel = cancel
+
+	// command issuers to start listening to their transactions
+	for _, issuer := range o.issuers {
+		o.observerGroup.Go(func() error { return issuer.Listen(observerCtx) })
+	}
+
+	// command issuers to start sending transactions
```
aaronbuchwald, line 68:
```suggestion
	// start a goroutine to confirm each issuer's transactions
	for _, issuer := range o.issuers {
		o.observerGroup.Go(func() error { return issuer.Listen(observerCtx) })
	}

	// start issuing transactions sequentially from each issuer
```

load/short_burst_orchestrator.go
```diff
@@ -0,0 +1,105 @@
...
+	for i, issuer := range o.issuers {
+		generator := o.generators[i]
+		o.issuerGroup.Go(func() error {
+			defer issuer.Stop()
+
+			for range o.config.N {
```
aaronbuchwald, line 74: Should `N` be `TxsPerIssuer` ?

load/tracker.go
```diff
@@ -0,0 +1,55 @@
...
+	t.Lock()
+	defer t.Unlock()
+	t.txConfirmedCounter++
+}
+
+func (t *DefaultTracker[T]) ObserveFailed(T) {
+	t.Lock()
+	defer t.Unlock()
+	t.txFailedCounter++
+}
```
aaronbuchwald, line 37: follow up - we can track latency for each observed containers since we have the time it was issued and when we observe it marked as confirmed/failed

load/dependencies.go
```diff
@@ -0,0 +1,57 @@
...
+	// confirmed were accepted.
+	GetObservedConfirmed() uint64
+	// GetObservedFailed returns the number of transactions that the tracker has
+	// confirmed failed.
+	GetObservedFailed() uint64
+}
+
+type Orchestrator interface {
+	// Execute begins the load generation process and blocks until the
+	// orchestrator has finished or the context is done.
```
aaronbuchwald, line 55:
```suggestion
	// Execute the load test
```

aaronbuchwald: should be assumed that a function is blocking and also reasonable to assume that it correctly honors the context

genesis/rules.go
```diff
@@ -58,8 +58,10 @@ func NewDefaultRules() *Rules {
 		// Chain Fee Parameters
 		MinUnitPrice:               fees.Dimensions{100, 100, 100, 100, 100},
 		UnitPriceChangeDenominator: fees.Dimensions{48, 48, 48, 48, 48},
-		WindowTargetUnits:          fees.Dimensions{20_000_000, 1_000, 1_000, 1_000, 1_000},
-		MaxBlockUnits:              fees.Dimensions{1_800_000, 2_000, 2_000, 2_000, 2_000},
+		// WindowTargetUnits:          fees.Dimensions{20_000_000, 1_000, 1_000, 1_000, 1_000},
```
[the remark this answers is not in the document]
aaronbuchwald: Can we modify this where we create the genesis instead of in DefaultRules? Setting the targets to max uint64 may not be a good default

load/tracker.go
```diff
@@ -0,0 +1,57 @@
...
+
+	txIssuedCounter    uint64
+	txConfirmedCounter uint64
+	txFailedCounter    uint64
+}
+
+func (t *DefaultTracker) Issue(ids.ID, time.Time) {
+	t.Lock()
+	t.txIssuedCounter++
+	t.Unlock()
```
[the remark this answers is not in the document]
RodrigoVillar: Done

load/issuer.go
RodrigoVillar: Done

load/issuer.go
RodrigoVillar: Done

load/short_burst_orchestrator.go
RodrigoVillar: Done

load/short_burst_orchestrator.go
RodrigoVillar: Changed to `TxsPerIssuer`

genesis/rules.go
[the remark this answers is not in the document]
RodrigoVillar: Reverted diff since `DefaultTestNetworkConfiguration` already maxes out units.

tests/e2e/e2e.go
```diff
@@ -31,18 +32,32 @@ var (
...
 
-func SetWorkload(networkConfigImpl workload.TestNetworkConfiguration, generator workload.TxGenerator, abi abi.ABI, sh throughput.SpamHelper, key chain.AuthFactory) {
+// ShortBurstComponentsGenerator returns the components necessary to instantiate
+// a ShortBurstOrchestrator.
+// We use a generator here since the node URIs are not known until runtime.
+type ShortBurstComponentsGenerator func(
+	ctx context.Context,
+	uris []string,
+	authFactories []chain.AuthFactory,
+) ([]load.TxGenerator[*chain.Transaction], []load.Issuer[*chain.Transaction], load.Tracker, error)
```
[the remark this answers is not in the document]
RodrigoVillar: Good point, I changed it so that a URI is passed in to the client (i.e. to be used for account balance queries) and the HyperSDK `e2e` test will create the issuer instead.

load/issuer.go
RodrigoVillar: Will defer until I have integrate the `GradualLoadOrchestrator` to see if this makes sense

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,240 @@
...
+		o.issuerGroup.Go(func() error {
+			for {
+				select {
+				case <-issuerCtx.Done():
+					return issuerCtx.Err()
+				default:
+				}
+				currTime := time.Now()
+				txsPerIssuer := uint64(math.Ceil(float64(currTargetTPS.Load())/float64(len(o.issuers))) * o.config.TxMultiplier)
+				for range txsPerIssuer {
```
tsachiherman, line 152: I think that the generation and the issuance should not be blocking eachother.
i.e.
one goroutine would call `GenerateTx` and feed the transactions into a channel.
second goroutine would pick up the transaction from the channel and send it to `IssueTx`.

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,240 @@
...
+					tx, err := o.generators[i].GenerateTx(issuerCtx)
+					if err != nil {
+						return err
+					}
+					if err := issuer.IssueTx(ctx, tx); err != nil {
+						return err
+					}
+				}
+				diff := time.Second - time.Since(currTime)
```
tsachiherman, line 161: why do we need to sleep here ?

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,240 @@
...
+					time.Sleep(diff)
+				}
+			}
+		})
+	}
+
+	// blocks until either 1) the max TPS target has been reached, 2) we've
+	// maxed out the number of attempts, or 3) an issuer has errored.
+	for {
+		time.Sleep(o.config.SustainedTime)
```
tsachiherman, line 172: use `time.After` here so that you could quickly abort in case of context cancelation.

load/short_burst_orchestrator.go
```diff
@@ -0,0 +1,105 @@
...
+		tracker:    tracker,
+		config:     config,
+	}, nil
+}
+
+// Execute orders issuers to send a fixed number of transactions and then waits
+// for all of their statuses to be confirmed or for a timeout to occur.
+func (o *ShortBurstOrchestrator[T, U]) Execute(ctx context.Context) error {
+	observerCtx, cancel := context.WithCancel(ctx)
+	o.cancel = cancel
```
tsachiherman, line 61: same as before. you don't need to store `o.cancel`

load/short_burst_orchestrator.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// start a goroutine to confirm each issuer's transactions
+	for _, issuer := range o.issuers {
+		o.observerGroup.Go(func() error { return issuer.Listen(observerCtx) })
```
tsachiherman, line 65: why do we need to observe all the issuers ? It should be enough to monitor a single node, as the accepted blocks are guaranteed to be the same.

tsachiherman (changes requested):
see comments/requests.

load/default_tracker.go
```diff
@@ -0,0 +1,61 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
```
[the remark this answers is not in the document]
RodrigoVillar: Yes; however, the linter adds `2024` since it hasn't been updated yet. Going to keep this until the linter is updated so that it updates *all* files in the HyperSDK codebase to have `CURRENT-YEAR` ([ref](https://github.com/ava-labs/hypersdk/pull/1848#discussion_r1900549151))

load/README.md
```diff
@@ -0,0 +1,130 @@
...
+for each issuer: // Async
+    send txsPerIssuer txs per second
+end for
+
+for
+    wait for SustainedTime
+
+    tps := number of accepted TXs in this time period
+    if tps >= currTargetTPS:
+        increase currTargerTPS by step
```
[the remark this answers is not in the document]
RodrigoVillar: Could you elaborate on both points (that is, why do we need a search algorithm and why do we need to wait between iterations?)

load/gradual_load_orchestrator.go
RodrigoVillar: Each issuer sends `txsPerIssuer` TXs per second. If they finish earlier, they sleep for `diff` amount of time so that they have a TPS of `txsPerIssuer`.

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -4,29 +4,49 @@
...
 	he2e "github.com/ava-labs/hypersdk/tests/e2e"
 	ginkgo "github.com/onsi/ginkgo/v2"
 )
 
 const owner = "morpheusvm-e2e-tests"
 
-var flagVars *e2e.FlagVars
+const (
+	metricsURI      = "localhost:8080"
+	metricsFilePath = ".tmpnet/prometheus/file_sd_configs/hypersdk-e2e-metrics.json"
```
Elvis339, line 43: The name should probably be: hypersdk-e2e-load-generator-metrics.json or something similar.

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +66,140 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+	tc := e2e.NewTestContext()
+	r := require.New(tc)
+
+	// Start metrics server
+	mux := http.NewServeMux()
+	mux.Handle("/ext/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{
+		Registry: registry,
+	}))
+
+	metricsServer := &http.Server{ //nolint:gosec
```
Elvis339, line 107: Seems like "//nolint:gosec" is typo?

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +66,140 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
 
 	// Initialize the local test environment from the global state
 	e2e.InitSharedTestEnvironment(ginkgo.GinkgoT(), envBytes)
+
+	tc := e2e.NewTestContext()
+	r := require.New(tc)
+
+	// Start metrics server
+	mux := http.NewServeMux()
+	mux.Handle("/ext/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{
```
Elvis339, line 103: Usually prometheus metrics are registered on `/metrics` route is there a reason for `/ext/metrics` route?

load/prometheus_tracker.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const namespace = "hypersdk_prometheus_tracker"
```
Elvis339, line 14: There's no need to prefix with hypersdk and prometheus, simple tracker or load namespace should be sufficient afaik load generation is agnostic to the VM?

examples/morpheusvm/tests/e2e/e2e_test.go
RodrigoVillar: With the current implementation, removing this `nolint` directive yields the following linter error:

```
tests/e2e/e2e_test.go:107:20: G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (gosec)
        metricsServer := &http.Server{
                Addr:    metricsURI,
                Handler: mux,
        }
```

I'm open to removing this `nolint` directive if there's a "correct" way of setting up this server.

examples/morpheusvm/tests/e2e/e2e_test.go
Elvis339: I see, this is so collector from tmpnet can scrape metrics from CI?

examples/morpheusvm/tests/e2e/e2e_test.go
Elvis339:
```go
&http.Server{
    Addr:              metricsURI,
    Handler:           mux,
    ReadHeaderTimeout: 10 * time.Second
}
```
10 seconds seems good enough

load/prometheus_tracker.go
RodrigoVillar: Done

examples/morpheusvm/tests/e2e/e2e_test.go
RodrigoVillar: Done

examples/morpheusvm/tests/e2e/e2e_test.go
RodrigoVillar: This works, thanks!

load/short_burst_orchestrator.go
RodrigoVillar: The main reason for this is with our WS implementation; tx senders are always notified about the status of their transaction (i.e. confirmed in a block or expired). If we were to switch to listening to blocks, we would lose the ability to check the status of our transaction in the case that its expired. 

The load generation in subnet-evm works similarly, where each issuer is responsible for listening to their own TX. Given that we might rewrite the WS implementation in the future, I'm open to switching from issuers listen to a "listener" listening once we reach that point.

load/gradual_load_orchestrator.go
RodrigoVillar: Below is a sketch of the design I think you're suggesting:

```mermaid
graph
    A1["Tx Generator 1"]
    A2["Tx Generator 2"]
    A3["Tx Generator 3"]


    subgraph chans_list["Chan List"]
        chan1["chan 1"]
        chan2["chan 2"]
        chan3["chan 3"]
    end

    B1["Issuer 1"]
    B2["Issuer 2"]
    B3["Issuer 3"]

    A1 --> chan1
    A2 --> chan2
    A3 --> chan3

    chan1 --> B1
    chan2 --> B2
    chan3 --> B3
```

If this is the case, then I agree that this eliminates the tx generator from blocking the issuer. However, I'm not sure if this is worth the tradeoff of having to dealing with another group of goroutines (along with a channel DS that needs to be thread-safe). The TX generator, in theory, blocks for an almost-zero amount of time (if we are not making any API calls, which MorpheusVM doesn't) and therefore you could make the case that the optimization we're attempting to make here is overshadowed by the increased technical complexity of the orchestrator.

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +66,141 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
 	generator := workload.NewTxGenerator(authFactories[1])
-	tc := e2e.NewTestContext()
-	he2e.SetWorkload(testingNetworkConfig, generator, expectedABI, &spamHelper, spamKey)
 
-	return fixture.NewTestEnvironment(tc, flagVars, owner, testingNetworkConfig, consts.ID).Marshal()
+	reg := prometheus.NewRegistry()
+	tracker, err := hload.NewPrometheusTracker[ids.ID](reg)
+	require.NoError(err)
+
+	registry = reg
```
aaronbuchwald, line 76: Why not just use `registry = prometheus.NewRegistry()` to assign the variable?

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +66,141 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
 }, func(envBytes []byte) {
 	// Run in every ginkgo process
 
 	// Initialize the local test environment from the global state
 	e2e.InitSharedTestEnvironment(ginkgo.GinkgoT(), envBytes)
+
+	tc := e2e.NewTestContext()
+	r := require.New(tc)
+
+	// Start metrics server
```
aaronbuchwald, line 101: Why do we setup the metrics server in the second function within `SynchronizedBeforeSuite` ?

The first function is run in the first process, whereas the second is run in every process, so it appears to me that this code will run multiple servers for a single registry when we only need one.

aaronbuchwald: It also seems this could cause a problem given we are listening on the hardcoded endpoint `localhost:8080`, it may be that this does not cause a problem now if ginkgo is only starting a single process.

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +66,141 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+	uri string,
+	authFactories []chain.AuthFactory,
+) ([]hload.TxGenerator[*chain.Transaction], error) {
+	lcli := vm.NewJSONRPCClient(uri)
+	ruleFactory, err := lcli.GetRuleFactory(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	numOfFactories := len(authFactories)
```
aaronbuchwald, line 148: nit: remove preposition from variable names ie. `numFactories` instead of `numOfFactories`

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +66,141 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+		return err
+	}
+	defer file.Close()
+
+	if _, err := file.Write(config); err != nil {
+		return err
+	}
+
+	return nil
+}
```
aaronbuchwald, line 206: Is this specific to MorpheusVM? If not, would it make sense to move to `e2e.go` so that other VMs do not need to re-implement?

examples/morpheusvm/tests/e2e/e2e_test.go
aaronbuchwald: Is the added metrics setup specific to MorpheusVM? If not, should this be added somewhere else?

load/README.md
```diff
@@ -0,0 +1,130 @@
...
+between implementations (e.g. short burst vs gradual load).
+
+When applicable, the orchestrator can also query the tracker to make a decision
+(e.g. increase TPS or return).
+
+### Transaction Generator
+
+The TX generator is responsible for generating a valid transaction which any
+issuer can send. Each TX generator is assumed to have an associated account that
+will sign transactions and whose balance is monitored to err in the case of fund exhaustion.
```
aaronbuchwald, line 56: There's a difference in the assumptions documented in the README vs. on the interfaces themselves. What are the minimum assumptions we need to make?

load/README.md
```diff
@@ -0,0 +1,130 @@
...
+The TX issuer has two responsibilities:
+- Sending transactions to the network
+- Listening for the status of transactions it has sent
+
+The TX issuer also has the following obligations to the tracker:
+
+- Upon sending a TX, the issuer should call `Issue()`
+- Upon confirming an accepted TX, the issuer should call `ObserveConfirmed()`
+- Upon confirming a failed TX, the issuer should call `ObserveFailed()`
+
```
aaronbuchwald, line 69:
```suggestion
The Tx issuer is responsible for sending and confirming transactions. As it issues/confirms transactions, it notifies the tracker, so that it can maintain metrics across the generated load.
```

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,234 @@
...
+	// the step size to increase the TPS by.
+	Step uint64
+
+	// the factor by which to pad the number of TXs an issuer sends per second
+	// for example, if targetTPS = 1000 and numOfIssuers = 10, then each issuer
+	// will send (1000/10)*TxRateMultiplier transactions per second.
+	//
+	// this is useful when the network can handle the current target TPS but it
+	// hasn't received enough transactions to achieve that TPS (as a result of
+	// latency + other factors).
```
aaronbuchwald, line 34: What other factors?

load/gradual_load_orchestrator.go
aaronbuchwald: Agree the channel approach makes sense if we are performing an operation that represents a potential bottleneck.

I think we can make an optimization like this if tx generation is a bottleneck, but don't see a need to before.

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,234 @@
...
+				diff := time.Second - time.Since(currTime)
+				if diff > 0 {
+					time.Sleep(diff)
+				}
+			}
+		})
+	}
+
+	// blocks until either 1) the max TPS target has been reached, 2) we've
+	// maxed out the number of attempts, or 3) an issuer has errored.
```
aaronbuchwald, line 160: can we break 1, 2, and 3 onto multiple lines?

load/issuer.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/ava-labs/hypersdk/pubsub"
+)
+
+var (
+	_ Issuer[*chain.Transaction] = (*DefaultIssuer)(nil)
+
+	ErrIssuedAlreadyStopped = errors.New("issuer already stopped")
+)
+
+type DefaultIssuer struct {
```
aaronbuchwald, line 25: (not necessary for this PR) This isn't so much a default as a HyperSDK specific issuer. We may want to put this in a different package or rename it within this one.

load/prometheus_tracker.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const namespace = "load"
+
+var _ Tracker[any] = (*PrometheusTracker[any])(nil)
+
+type PrometheusTracker[T comparable] struct {
```
aaronbuchwald, line 18: Do we need both trackers or can we just maintain this one?

load/short_burst_orchestrator.go
aaronbuchwald: Potential alternative - separate issuance from confirmation. This may require connecting them ie. issuers notify "confirmers" of the outstanding set via a channel or similar.

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,234 @@
...
+
+		tps := computeTPS(prevConfirmed, currConfirmed, currTime.Sub(prevTime))
+		if tps >= currTargetTPS.Load() {
+			if currTargetTPS.Load() >= o.config.MaxTPS {
+				o.log.Info(
+					"max TPS reached",
+					zap.Uint64("target TPS", currTargetTPS.Load()),
+					zap.Uint64("average TPS", tps),
+				)
+				achievedMaxTPS = true
```
aaronbuchwald, line 183: Should we be using the word max or target? It seems like we are talking about target here

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,234 @@
...
+	currTargetTPS := atomic.Uint64{}
+	currTargetTPS.Store(o.config.MinTPS)
+
+	issuerGroup, issuerCtx := errgroup.WithContext(ctx)
+	o.issuerGroup = issuerGroup
+
+	// start a goroutine to each issuer to continuously send transactions
+	// if an issuer errors, all other issuers will stop as well.
+	for i, issuer := range o.issuers {
+		o.issuerGroup.Go(func() error {
```
aaronbuchwald, line 133: Can we separate this into its own function?

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,234 @@
...
+				if diff > 0 {
+					time.Sleep(diff)
+				}
+			}
+		})
+	}
+
+	// blocks until either 1) the max TPS target has been reached, 2) we've
+	// maxed out the number of attempts, or 3) an issuer has errored.
+	for {
```
aaronbuchwald, line 161: Same comment to separate into its own function here?

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// block until both the observers and issuers have stopped
+	return errors.Join(o.issuerGroup.Wait(), o.observerGroup.Wait())
+}
+
+// run the gradual load test by continuously increasing the rate at which
+// transactions are sent
+//
+// run stops when the network can no longer make progress or if an issuer errors.
+func (o *GradualLoadOrchestrator[T, U]) run(ctx context.Context) {
```
aaronbuchwald, line 116: Can we support an option to run indefinitely at target after hitting target?

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,234 @@
...
+
+		prevConfirmed = currConfirmed
+		prevTime = currTime
+	}
+
+	o.maxObservedTPS = currTargetTPS.Load()
+	if !achievedMaxTPS {
+		o.maxObservedTPS -= o.config.Step
+	}
+}
```
aaronbuchwald, line 220: Rather than performing this calculation at the end, can we set `maxObservedTPS` when we observe it for the first time if we are surpassing the previous threshold? I think putting this closer to the observation makes it more obviously correct.

aaronbuchwald: This has the added benefit of making it so `GetMaxObservedTPS` works throughout instead of only at the end.

Speaking of.. when is `GetMaxObservedTPS` expected to return a valid result? Is this strictly after `Execute` has returned?

aaronbuchwald (changes requested):

load/gradual_load_orchestrator.go
RodrigoVillar: Done

load/README.md
```diff
@@ -0,0 +1,130 @@
...
+When applicable, the orchestrator can also query the tracker to make a decision
+(e.g. increase TPS or return).
+
+### Transaction Generator
+
+The TX generator is responsible for generating a valid transaction which any
+issuer can send. Each TX generator is assumed to have an associated account that
+will sign transactions and whose balance is monitored to err in the case of fund exhaustion.
+
+### TX Issuer
```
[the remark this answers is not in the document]
RodrigoVillar: Done

vm/config.go
```diff
@@ -33,10 +34,10 @@ type Config struct {
...
 	return Config{
-		MempoolSize:                      2_048,
+		MempoolSize:                      consts.MaxInt,
 		AuthVerificationCores:            1,
 		RootGenerationCores:              1,
-		MempoolSponsorSize:               32,
+		MempoolSponsorSize:               consts.MaxInt,
```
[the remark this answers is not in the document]
RodrigoVillar: Reverted the diff and instead modified `NewHyperVMSubnet()` so that it takes in a VM config string that the caller needs to provide. This makes it so that we define the VM config (and therefore, the mempool sizes) in the `e2e` test.

tests/e2e/e2e.go
```diff
@@ -111,26 +133,99 @@ var _ = ginkgo.Describe("[HyperSDK Tx Workloads]", ginkgo.Serial, func() {
...
+				require.NoError(err)
+				issuers[i] = issuer
+			}
+		} else {
+			for i := 0; i < len(txGenerators); i++ {
+				issuer, err := load.NewDefaultIssuer(uris[i], tracker)
+				require.NoError(err)
+				issuers[i] = issuer
+			}
+		}
```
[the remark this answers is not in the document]
RodrigoVillar: My intuition was that each issuer should be targeting a different node and that if that wasn't possible, to have them all target a single node. 

Changed if-else to use modulus for URI assignment.

tests/e2e/e2e.go
```diff
@@ -111,26 +133,99 @@ var _ = ginkgo.Describe("[HyperSDK Tx Workloads]", ginkgo.Serial, func() {
...
```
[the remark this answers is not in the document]
RodrigoVillar: https://github.com/ava-labs/hypersdk/pull/1991#discussion_r2014969007

examples/morpheusvm/tests/e2e/e2e_test.go
RodrigoVillar: Done

load/gradual_load_orchestrator.go
RodrigoVillar: `achievedMaxTPS` is true if the observed TPS is greater than `maxTPS`. Therefore, I think the name makes sense here.

tests/e2e/e2e.go
```diff
@@ -31,18 +33,38 @@ var (
...
 
-func SetWorkload(networkConfigImpl workload.TestNetworkConfiguration, generator workload.TxGenerator, abi abi.ABI, sh throughput.SpamHelper, key chain.AuthFactory) {
+// LoadTxGenerator returns the components necessary to instantiate an
+// Orchestrator.
+// We use a generator here since the node URIs are not known until runtime.
+type LoadTxGenerator func(
+	ctx context.Context,
+	uri string,
+	authFactories []chain.AuthFactory,
+) ([]load.TxGenerator[*chain.Transaction], error)
```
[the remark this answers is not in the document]
RodrigoVillar: `authFactories` are the accounts that will be used for each TxGenerator (i.e. 1 account => 1 TxGenerator). We pass in the URI because we need this to query the `ruleFactory`, `unitPrices`, and the balance of each account.

examples/morpheusvm/tests/e2e/e2e_test.go
RodrigoVillar: This was due to me not fully understanding `SynchronizedBeforeSuite`. I've changed the code so that the metric server setup is done in the the first process. 

Furthermore, I've also changed it so that `e2e.go` does the majority of the metrics server setup, with the MorpheusVM `e2e` test passing in the test network and the registry.

load/issuer.go
```diff
@@ -0,0 +1,110 @@
...
+	return nil
+}
+
+// isFinished increments the number of transactions heard and returns true if all transactions have been heard
+func (i *DefaultIssuer) isFinished() bool {
+	i.lock.Lock()
+	defer i.lock.Unlock()
+
+	i.receivedTxs++
+	return i.stopped && i.issuedTxs == i.receivedTxs
```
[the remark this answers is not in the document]
RodrigoVillar: Done

examples/morpheusvm/tests/e2e/e2e_test.go
RodrigoVillar: Removed the global variable altogether.

load/prometheus_tracker.go
RodrigoVillar: Removed `DefaultTracker`; can definitely add this back in if the prometheus metrics introduces a problematic amount of overhead.

load/gradual_load_orchestrator.go
RodrigoVillar: Done

load/gradual_load_orchestrator.go
RodrigoVillar: I think delegating the for-loop would introduce an unnecessary amount of function nesting (we would have `Execute() => run() => search()` where `search()` is the proposed new function). Looking at the for-loop that `search()` would encapsulate, I would argue that the for-loop itself is relatively simple (with the logs making it seem larger than it is) and therefore can be kept in `run()`.

In the case of `issueTxs()`, the code is async and somewhat complicated so it made sense to split it into its own function.

load/gradual_load_orchestrator.go
RodrigoVillar: Done

RodrigoVillar: > Speaking of.. when is GetMaxObservedTPS expected to return a valid result? Is this strictly after Execute has returned?

`GetMaxObservedTPS` returns the max observed TPS that the orchestrator has seen, regardless if the test passes or fails (only caveat is if the context is cancelled, which in that case the tps number returned is from the previous iteration).

load/gradual_load_orchestrator.go
RodrigoVillar: Done

tests/e2e/e2e.go
```diff
@@ -111,26 +133,99 @@ var _ = ginkgo.Describe("[HyperSDK Tx Workloads]", ginkgo.Serial, func() {
...
+			issuers,
+			tracker,
+			tc.Log(),
+			gradualLoadConfig,
+		)
 		require.NoError(err)
+
+		if err := orchestrator.Execute(ctx); err != nil {
+			require.True(errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded))
+		}
```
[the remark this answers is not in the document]
RodrigoVillar: I believe that this is not an issue of "why are we requiring an error" but rather "how should we handle the error".  IIRC, we previously discussed that the callee should not interpret the context error but rather the caller (and so, the callee shouldn't set err = nil if it sees a context-related error).

Walking through the gradual load orchestrator, we have that if we reach the max TPS, then we'll cancel the context that the issuers (the listener and tx issuance goroutines) are relying on. As a result, both error groups will return a `context.Cancelled` error which then gets combined and returned as one error:

https://github.com/ava-labs/hypersdk/blob/ebca116d0aa2e7f95767cbafd25245df1f95ba3b/load/gradual_load_orchestrator.go#L227-L228

https://github.com/ava-labs/hypersdk/blob/ebca116d0aa2e7f95767cbafd25245df1f95ba3b/load/issuer.go#L56-L57

The straightforward fix here would be to modify the gradual load orchestrator to overwrite `context.Cancelled` errors with `nil` (and I am open to it).

load/README.md
RodrigoVillar: Synced `README` and `dependencies.go` so that they match up on component descriptions.

tests/e2e/e2e.go
```diff
@@ -111,26 +133,99 @@ var _ = ginkgo.Describe("[HyperSDK Tx Workloads]", ginkgo.Serial, func() {
...
+		ctx := tc.DefaultContext()
 		blockchainID := e2e.GetEnv(tc).GetNetwork().GetSubnet(networkConfig.Name()).Chains[0].ChainID
 		uris := getE2EURIs(tc, blockchainID)
-		key := spamKey
 
-		err := spamHelper.CreateClient(uris[0])
+		txGenerators, err := loadTxGenerator(
+			ctx,
+			uris[0],
+			networkConfig.AuthFactories(),
```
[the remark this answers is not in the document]
RodrigoVillar: Added https://github.com/ava-labs/hypersdk/pull/1991/commits/24f1fe4371c860bfb3638db2c70d02c14a1470a2, which adds a VM-specific `FundDistributor` and `FundConsolidator` which distributes/consolidates funds to/from a generated set of test accounts.

Also happy to move this commit to this PR if you think it makes more sense, given that this might be considered outside the scope of this PR.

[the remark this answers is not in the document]
aaronbuchwald: I'd consider this in scope for this PR because we are adding the load package and need to see how it will be used as well.

I think we can further clean up the fund distribution / consolidation, but will move that to another comment and resolve this one.

load/default_tracker.go
```diff
@@ -0,0 +1,41 @@
...
+	t.txIssuedCounter.Add(1)
+}
+
+func (t *DefaultTracker[T]) ObserveConfirmed(T, time.Time) {
+	t.txConfirmedCounter.Add(1)
+}
+
+func (t *DefaultTracker[T]) ObserveFailed(T, time.Time) {
+	t.txFailedCounter.Add(1)
+}
```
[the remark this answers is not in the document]
aaronbuchwald: I think it's simpler if we remove time as a parameter and we can add it later if needed.

load/gradual_load_orchestrator.go
aaronbuchwald: It is true within this context, but it implies that it's the maximum TPS we can achieve, which does not seem accurate. We are really talking about a target here.

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,265 @@
...
+			)
+			attempts++
+		}
+
+		prevConfirmed = currConfirmed
+		prevTime = currTime
+	}
+}
+
+// GetObservedIssued returns the max TPS the orchestrator observed (modulus the step size).
```
aaronbuchwald, line 215: Could we make this comment more clear? "modulus the step size" sounds like we are doing maxTPS % step, which is not quite right.

tests/e2e/e2e.go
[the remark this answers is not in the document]
aaronbuchwald: If the caller provides a context and cancels it, the callee code should rarely if ever imo overwrite/ignore that context cancellation unless it is really an expected condition.

If the callee provides its own internal contexts and hits a context cancellation, that may or may not be separate. Since the caller passed in a parent context, if the callee is using an internal context and wants to use cancellation to signal internally then it should imo try to differentiate an internal cancellation condition vs. an external cancellation condition.

For example, if we receive a top-level context and create a child context that uses cancellation internally to signal between threads that they should stop, we should not mis-interpret a top-level context cancellation as a sign that we completed work, when we did not and mis-report that we completed our work. It's ofc fine to still interrupt and terminate that work, but that component should make sure not to report that it completed everything expected of it.

load/gradual_load_orchestrator.go
RodrigoVillar: Removed modulus comment (remanent of previous implementation).

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,265 @@
...
+
+	issuerGroup, issuerCtx := errgroup.WithContext(ctx)
+	o.issuerGroup = issuerGroup
+
+	o.issueTxs(issuerCtx, currTargetTPS)
+
+	// blocks until either:
+	// 1. the max TPS target has been reached
+	// 2. we've maxed out the number of attempts
+	// 3. an issuer has errored.
```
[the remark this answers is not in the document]
RodrigoVillar: Great style advice; applied 65e24a9

tests/e2e/e2e.go
```diff
@@ -13,36 +18,136 @@ import (
...
+	fundDistributor = distributor
+	fundConsolidator = consolidator
+}
+
+// ExposeMetrics begins a HTTP server that exposes the metrics of registry and
+// writes the collector config to a file which tmpnet uses to discover the
+// server and scrape its metrics.
+// Returns a cleanup function which, when called, stops the server and removes
+// the collector config file.
+// ExposeMetrics his should be called at most once
```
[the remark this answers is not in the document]
RodrigoVillar: At most once per test run; update documentation to make it clear that calling `ExposeMetrics` deletes any existing collector files d4872a1

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,265 @@
...
+		// if maxTPS has been reached and we don't terminate, then continue so
+		// that we do not manipulate TPS target
+		if achievedMaxTPS && !o.config.Terminate {
+			o.log.Info(
+				"current network state",
+				zap.Uint64("current TPS", tps),
+				zap.Uint64("max observed TPS", o.maxObservedTPS.Load()),
+			)
+			continue
+		}
```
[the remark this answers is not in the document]
RodrigoVillar: I think 65e24a9 clarifies this. Regardless, we need the `!o.config.Terminate` condition because otherwise, we will return prematurely.

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +51,93 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+		load.FundConsolidator,
+	)
+
+	vmConfig := `{
+		"statesync": {"minBlocks": 128},
+		"vm": {
+			"mempoolSize"       : 9223372036854775807,
+			"mempoolSponsorSize": 9223372036854775807
+		}
+	}`
```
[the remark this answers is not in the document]
RodrigoVillar: Great point; keeping the overriden mempool values but moving it to `subnet.go` 8166ab0

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +51,93 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+		loadTxGenerators,
+		tracker,
+		hload.ShortBurstOrchestratorConfig{
+			TxsPerIssuer: 1_000,
+			Timeout:      20 * time.Second,
+		},
+		hload.DefaultGradualLoadOrchestratorConfig(),
+		load.FundDistributor,
+		load.FundConsolidator,
+	)
```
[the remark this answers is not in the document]
RodrigoVillar: Switch to using `CreateTransfer` since we can query all other information we need to create + submit TX in `e2e.go`; 785925b

examples/morpheusvm/load/funds.go
```diff
@@ -0,0 +1,210 @@
...
+	ErrInsufficientFunds = errors.New("insufficient funds")
+	ErrTxFailed          = errors.New("transaction failed")
+)
+
+func FundDistributor(
+	ctx context.Context,
+	uri string,
+	funder chain.AuthFactory,
+	numAccounts uint64,
+) ([]chain.AuthFactory, error) {
```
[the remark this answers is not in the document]
RodrigoVillar: `CreateTransfer` is now the smallest possible component that is specific to the VM; 785925b

examples/morpheusvm/load/funds.go
```diff
@@ -0,0 +1,210 @@
...
+
+		success, _, err := indexerCli.WaitForTransaction(ctx, txCheckInterval, txID)
+		if err != nil {
+			return nil, err
+		}
+		if !success {
+			return nil, ErrTxFailed
+		}
+	}
+	return accounts, nil
```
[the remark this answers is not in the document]
RodrigoVillar: Reintroduced the `pacer` and switched to submitting/confirming txs async: 6ebac27e97fa0f711c502f54c274164788c1dae9

tests/e2e/e2e.go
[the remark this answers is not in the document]
RodrigoVillar: Understood; in this case, since we know that we'll be cancelling the child context, the orchestrator will ignore ctx cancellations from the async tx issuers and the issuer will ignore ctx cancellations as well. Also made it so that if the orchestrator fails to hit the TPS target, it will return an error abcabd7

load/gradual_load_orchestrator.go
RodrigoVillar: Updated `achievedMaxTPS`+ TPS language to make this clearer; 9728c89c3

load/README.md
```diff
@@ -0,0 +1,120 @@
...
+The current TPS in the gradual load orchestrator is determined by taking the
+number of transactions confirmed in a given time window (`SustainedTime`) and
+diving it by `SustainedTime` (in terms of seconds) Furthermore, the orchestator
+has `maxAttempt` tries to try and achieve a given TPS before determining that
+the given TPS is not achievable.
+
+Below is the pseudocode for how the gradual load orchestrator determines TPS and
+for how it increases TPS:
+
+```
```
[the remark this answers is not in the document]
RodrigoVillar: Originally, a mix of golang and LaTeX. Modified this to remove most of the LaTeX style bb26945cc

tests/e2e/e2e.go
```diff
@@ -13,36 +18,136 @@ import (
...
-	"github.com/ava-labs/hypersdk/throughput"
 	"github.com/ava-labs/hypersdk/utils"
 
 	ginkgo "github.com/onsi/ginkgo/v2"
 )
 
+const (
+	metricsURI = "localhost:8080"
+	// relative to the user home directory
+	metricsFilePath = ".tmpnet/prometheus/file_sd_configs/hypersdk-e2e-metrics.json"
```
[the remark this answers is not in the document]
aaronbuchwald: Makes sense, will resolve this comment and we can fix / improve as needed. Thanks for explaining!

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +54,91 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+	}
+
+	// Create tx generator
+	txGenerators := make([]hload.TxGenerator[*chain.Transaction], numFactories)
+	for i := 0; i < numFactories; i++ {
+		txGenerators[i] = load.NewTxGenerator(authFactories[i], ruleFactory, balances[i], unitPrices)
+	}
+
+	return txGenerators, nil
+}
```
aaronbuchwald, line 144: (not necessary for this PR) - I think we'll end up moving this next to `load.NewTxGenerator` in the future to re-use outside of e2e tests as @Elvis339 has requested

load/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+The Tx issuer is responsible for tracking the status of all sent transactions. 
+As it issues/confirms transactions, it notifies the tracker, so that it can 
+maintain metrics across the generated load.
+
+### Tracker
+
+The tracker is responsible for maintaining metrics for all sent txs. Since the 
+tracker is used by both the issuers and the orchestrator, all methods of the
+tracker must be thread safe.
```
aaronbuchwald, line 64: Can we use consistent capitalization throughout here? Should `Orchestrator`, `Transaction Generator`, etc be proper nouns?

aaronbuchwald: Same comment for the names of the orchestrator strategies below

load/README.md
```diff
@@ -0,0 +1,118 @@
...
+once. This orchestrator is parameterizable via the following:
+
+- `N`: the number of transactions an issuer will send to the network.
+- `timeout`: the maximum amount of time which, after all transactions have been sent,
+  the orchestrator will wait to hear the confirmation of all outstanding
+  transactions. 
+
+### Gradual Load
+
+The gradual load orchestrator sends transactions at a initial rate (TPS) and
```
aaronbuchwald, line 82:
```suggestion
The gradual load orchestrator sends transactions at an initial rate (TPS) and
```

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,277 @@
...
+	// the step size to increase the TPS by.
+	Step uint64
+
+	// the factor by which to pad the number of txs an issuer sends per second
+	// for example, if targetTPS = 1000 and numIssuers = 10, then each issuer
+	// will send (1000/10)*TxRateMultiplier transactions per second.
+	//
+	// this is useful when the network can handle the current target TPS but it
+	// hasn't received enough transactions to achieve that TPS (as a result of
+	// latency between individual issuers and the API nodes).
```
aaronbuchwald, line 38: "as a result of latency between individual issuers and the API nodes" - there's probably a number of other cases that we could run into this and using the tx rate multiplier provides some breathing room to make sure that we overshoot instead of undershoot.

For example, if we target exactly targetTPS outstanding, then we will probably get bursty behavior ie. we hit the following sequence:

1. We have enough txs outstanding
2. We stop issuing txs
3. Network accepts the txs
4. We have now delayed issuing further transactions
5. It takes us longer to issue the next batch

Could we modify this to:

"Maintaining a multiplier above target provides a toggle to keep load persistently above instead of below target. This ensures load generation does not pause issuance at the target and persistently under-issue and fail to account for the time it takes to add more load."

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,277 @@
...
+
+	for {
+		// wait for the sustained time to pass or for the context to be cancelled
+		select {
+		case <-time.After(o.config.SustainedTime):
+		case <-issuerCtx.Done(): // the parent context was cancelled or an issuer errored
+		}
+
+		if issuerCtx.Err() != nil {
+			break // Case 1
```
aaronbuchwald, line 158: Nice, way more readable!

load/gradual_load_orchestrator.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+
+		currConfirmed := o.tracker.GetObservedConfirmed()
+		currTime := time.Now()
+
+		tps := computeTPS(prevConfirmed, currConfirmed, currTime.Sub(prevTime))
+		o.setMaxObservedTPS(tps)
+
+		// if max TPS target has been reached and we don't terminate, then continue so
+		// that we do not manipulate TPS target
```
aaronbuchwald, line 168:
```suggestion
		// if max TPS target has been reached and we don't terminate, then continue here
		// so we do not keep increasing the target TPS
```

tests/e2e/e2e.go
```diff
@@ -257,6 +427,196 @@ var _ = ginkgo.Describe("[Custom VM Tests]", ginkgo.Serial, func() {
 	}
 })
 
+func distributeFunds(
+	ctx context.Context,
+	tc *e2e.GinkgoTestContext,
+	funder chain.AuthFactory,
+	numAccounts uint64,
+) ([]chain.AuthFactory, error) {
```
aaronbuchwald, line 435: can we pass in `createTransfer` as a parameter? We can still set it within this file via `SetWorkload`, but passing it in as a parameter makes this function cleaner and easier to read. We will likely want to separate this later.

load/gradual_load_orchestrator.go
RodrigoVillar: W

tests/e2e/e2e.go
RodrigoVillar: f0c0fbb45

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -46,22 +54,91 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+		consts.ID,
+	)
+
+	cleanUpFunc, err := he2e.ExposeMetrics(context.Background(), env, registry)
+	require.NoError(err)
+
+	ginkgo.DeferCleanup(func() {
+		require.NoError(cleanUpFunc())
+	})
+
```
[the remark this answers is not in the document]
RodrigoVillar: Moved metrics server instantiation + clean-up to `e2e`. Two caveats:

1. Caller still needs to pass in registry
2. Since `ginkgo.BeforeAll()` requires the scope its called in to be ordered, the workload `e2e` tests are now ordered.

f2da46f8b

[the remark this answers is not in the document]
aaronbuchwald: Why does the caller need to pass in the registry?

tests/e2e/e2e.go
aaronbuchwald: Can we also rename the global variable so that they are not as easily confused?

examples/morpheusvm/tests/e2e/e2e_test.go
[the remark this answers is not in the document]
RodrigoVillar: The registry is a sub-component of the `tracker`; as a result, we have:

```
create registry -> create tracker with registry -> call SetWorkload()
```

We could create the registry in `e2e` but this would require making `e2e` creating the tracker as well.

load/README.md
RodrigoVillar: 39c150f

tests/e2e/e2e.go
RodrigoVillar: 39c150f

examples/morpheusvm/tests/e2e/e2e_test.go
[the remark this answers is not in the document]
RodrigoVillar: 5c7ce60

aaronbuchwald (approved):

PR #1996, opened 2025-03-21
test(validitywindow): make validity window timestamp update test deteterministic

Elvis339:
The test was flaky because it relied on unpredictable timing between goroutines when updating minTimestamp. Fixed by using controlled node failure as a deterministic trigger for timestamp updates, ensuring the test produces consistent results.

Fixes: https://github.com/ava-labs/hypersdk/actions/runs/13979063622/job/39140000054?pr=1992

### DIFF
--- a/internal/validitywindow/client_test.go
+++ b/internal/validitywindow/client_test.go
@@ -232,11 +232,6 @@ func TestBlockFetcherClient_FetchBlocksChangeOfTimestamp(t *testing.T) {
 		},
 	}
 
-	test := newTestEnvironment(nodeSetups)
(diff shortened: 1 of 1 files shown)


internal/validitywindow/client_test.go
```diff
@@ -284,7 +278,8 @@ func generateTestChain(n int) []ExecutionBlock[container] {
 }
 
 type nodeSetup struct {
-	nodeID ids.NodeID
+	cancelFunc context.CancelFunc
```
RodrigoVillar, line 281: nit: `cancelFunc` => `cancel`

Elvis339: https://github.com/ava-labs/hypersdk/pull/1996/commits/a3d7757d52b9b50505becb95435a24cf6f70405d

aaronbuchwald (approved):

Issue #1942, opened 2025-02-21
Add json error codes for expected indexer rpc endpoints


tsachiherman:
The indexer server currently returns go errors as a result an endpoint error.

Instead, we should return application errors (common.AppError)

aaronbuchwald:
`common.AppError` are specific to the AvalancheGo engine and used in p2p networking. I don't think we should re-use this type here (although could certainly define our own with the same error code + message).

PR #2000, opened 2025-03-24
add indexer client unit tests

tsachiherman:
## What ? 

This PR extends the indexer client api unit tests.


api/indexer/client_test.go
```diff
@@ -59,26 +70,36 @@ func TestIndexerClientBlocks(t *testing.T) {
 
 			client := NewClient(httpServer.URL)
 
-			expectedBlock := executedBlocks[tt.blkHeight]
-			executedBlock, err := client.GetBlockByHeight(ctx, expectedBlock.Block.Hght, parser)
-			if tt.err == nil {
+			executedBlock, err := client.GetBlockByHeight(ctx, tt.blkHeight, parser)
+			if tt.blkHeightErr == nil {
```
RodrigoVillar, line 74: Could we simplify this if-condition? I think it's sufficient to check that the errors are "equal" before invoking the if-condition rather than doing an if-else:

```go
r.ErrorIs(tt.blkHeightErr, err)
if tt.blkHeightErr == nil {
    expectedBlock := executedBlocks[tt.blkHeight-1]
    r.Equal(expectedBlock.Block, executedBlock.Block)
}
```

api/indexer/client_test.go
```diff
@@ -59,26 +70,36 @@ func TestIndexerClientBlocks(t *testing.T) {
...
+			expectedBlkID := ids.GenerateTestID()
+			if tt.blkHeight > 0 {
+				expectedBlock := executedBlocks[tt.blkHeight-1]
+				expectedBlkID = expectedBlock.Block.GetID()
 			}
 
-			executedBlock, err = client.GetBlock(ctx, expectedBlock.Block.GetID(), parser)
-			if tt.err == nil {
+			executedBlock, err = client.GetBlock(ctx, expectedBlkID, parser)
+			if tt.blkHeightErr == nil {
```
RodrigoVillar, line 89: Same comment regarding simplifying if-else.

api/indexer/client_test.go
```diff
@@ -127,33 +183,45 @@ func TestIndexerClientTransactions(t *testing.T) {
...
 			txResponse, found, err := client.GetTxResults(ctx, executedTx.GetID())
-			r.Equal(tt.err, err)
-			r.Equal(tt.found, found)
-			if tt.found {
-				r.Equal(GetTxResponse{
-					TxBytes:   executedTx.Bytes(),
-					Timestamp: executedBlock.Block.Tmstmp,
-					Result:    executedBlock.ExecutionResults.Results[tt.txIndex],
-				}, txResponse)
+			if tt.getTxResultsErr != nil {
```
RodrigoVillar, line 198: Same comment regarding simplifying if-else.

api/indexer/client_test.go
```diff
@@ -90,26 +111,61 @@ func TestIndexerClientTransactions(t *testing.T) {
 		numTxs            = 3
 	)
 
+	parser := chaintest.NewTestParser()
+	badparser := &chain.TxTypeParser{
```
RodrigoVillar, line 115: nit: `badParser`

api/indexer/client_test.go
```diff
@@ -90,26 +111,61 @@ func TestIndexerClientTransactions(t *testing.T) {
...
+			blockIndex:      0,
+			txIndex:         0,
+			found:           false,
+			getTxResultsErr: nil,
+			getTxErr:        nil,
+			parser:          parser,
+			malformedBlock:  false,
+		},
+		{
+			name:            "badparser",
```
RodrigoVillar, line 151: nit: `"bad parser"`

RodrigoVillar (changes requested):

api/indexer/client_test.go
tsachiherman: Unfortunately, that won't work ( as is.. ). The returned error is an rpc error that contains a string representing the internal server error. Hence, we need to compare it by string (`ErrorContains`).

We do have a PR to update this and create RPC specific error codes, which would resolve this issue (https://github.com/ava-labs/hypersdk/issues/1942)

RodrigoVillar: Ah understood, thanks for the clarification!

api/indexer/client_test.go
```diff
@@ -127,33 +183,45 @@ func TestIndexerClientTransactions(t *testing.T) {
...
-
 			client := NewClient(httpServer.URL)
+
 			executedBlock := executedBlocks[tt.blockIndex]
 			executedTx := executedBlock.Block.Txs[tt.txIndex]
 
+			if tt.malformedBlock {
+				// "damage" the last executed block by removing the last result.
+				executedBlock.ExecutionResults.Results = executedBlock.ExecutionResults.Results[1:]
+				r.NoError(indexer.Notify(context.Background(), executedBlock))
```
RodrigoVillar, line 194: Could we use the `ctx` already declared at the beginning of the subtest?

api/indexer/client_test.go
```diff
@@ -59,26 +70,36 @@ func TestIndexerClientBlocks(t *testing.T) {
...
+			if tt.blkHeightErr == nil {
 				r.NoError(err)
+				expectedBlock := executedBlocks[tt.blkHeight-1]
 				r.Equal(expectedBlock.Block, executedBlock.Block)
 			} else {
-				r.ErrorContains(err, tt.err.Error())
+				r.ErrorContains(err, tt.blkHeightErr.Error())
+			}
+
+			expectedBlkID := ids.GenerateTestID()
```
RodrigoVillar, line 82: Could we use `ids.Empty{}` here? This way, it's clear that if we're not targeting a particular block (i.e. `numExecutedBlocks >= height > 0`), then it's clear that we're querying with a invalid block ID (since a valid block cannot have `id = ids.Empty{}`).

RodrigoVillar (changes requested):
Left two comments but afterwards, LGTM.

Could we create an issue afterwards to address https://github.com/ava-labs/hypersdk/pull/2000#discussion_r2012350523? With the current if-else usage, the table tests are hard to read ([ref](https://github.com/uber-go/guide/blob/master/style.md#avoid-unnecessary-complexity-in-table-tests)) but am willing to defer this to a future PR.

RodrigoVillar (approved):
LGTM

api/indexer/client_test.go
```diff
@@ -9,43 +9,54 @@ import (
...
+			blkHeight:         0,
+			blkHeightErr:      errBlockNotFound,
+			numExecutedBlocks: 4,
 		},
 		{
-			name:      "missing block",
-			blkHeight: 0,
-			err:       errBlockNotFound,
+			name:              "no blocks",
+			blkHeight:         0,
```
aaronbuchwald, line 49: In production, we should be able to serve the genesis block, should we generate a sequence of blocks that includes a genesis block?

api/indexer/client_test.go
```diff
@@ -59,26 +70,36 @@ func TestIndexerClientBlocks(t *testing.T) {
...
 			} else {
-				r.ErrorContains(err, tt.err.Error())
+				r.ErrorContains(err, tt.blkHeightErr.Error())
 			}
 
 			executedBlock, err = client.GetLatestBlock(ctx, parser)
-			r.NoError(err)
-			r.Equal(executedBlocks[numExecutedBlocks-1].Block, executedBlock.Block)
+			if tt.numExecutedBlocks == 0 {
+				r.ErrorContains(err, database.ErrNotFound.Error())
```
aaronbuchwald, line 98: Handling the special case this way makes it seem like we default to not including the genesis. Could we include at least the genesis block (we should receive a notification of at least the genesis block on startup)

api/indexer/client_test.go
```diff
@@ -127,33 +183,45 @@ func TestIndexerClientTransactions(t *testing.T) {
...
 
-			parser := chaintest.NewTestParser()
-
 			client := NewClient(httpServer.URL)
+
 			executedBlock := executedBlocks[tt.blockIndex]
 			executedTx := executedBlock.Block.Txs[tt.txIndex]
 
+			if tt.malformedBlock {
+				// "damage" the last executed block by removing the last result.
```
aaronbuchwald, line 192: Could we fix this comment? This removes the first result not the last result

api/indexer/client_test.go
```diff
@@ -127,33 +183,45 @@ func TestIndexerClientTransactions(t *testing.T) {
...
+			} else {
+				r.Equal(tt.found, found)
+				if tt.found {
+					r.Equal(GetTxResponse{
+						TxBytes:   executedTx.Bytes(),
+						Timestamp: executedBlock.Block.Tmstmp,
+						Result:    executedBlock.ExecutionResults.Results[tt.txIndex],
+					}, txResponse)
+					r.Equal(executedTx, tx)
+				}
```
aaronbuchwald, line 224: There's a lot of conditional nesting in these table tests. Could we try and simplify them so that they have narrow scope and reduce unnecessary nesting as much as possible?

For example, this section could be re-written as:

```golang
			txResponse, tx, found, err := client.GetTx(ctx, executedTx.GetID(), tt.parser)
			if tt.getTxErr != nil {
				r.ErrorContains(err, tt.getTxErr.Error())
			}
			r.Equal(tt.found, found)
			if !tt.found {
				return
			}
			r.Equal(GetTxResponse{
				TxBytes:   executedTx.Bytes(),
				Timestamp: executedBlock.Block.Tmstmp,
				Result:    executedBlock.ExecutionResults.Results[tt.txIndex],
			}, txResponse)
			r.Equal(executedTx, tx)
```

style guide refs:
- https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting
- https://github.com/uber-go/guide/blob/master/style.md#avoid-unnecessary-complexity-in-table-tests

aaronbuchwald (changes requested):

api/indexer/client_test.go
aaronbuchwald: Hmm putting this in place fails on the tests: `badParser` and `malformed_block`

it seems this is because `found` is re-used across both `GetTxResults` and `GetTx`, so we require the error in `GetTx` to short circuit the test at the error.

api/indexer/client_test.go
tsachiherman: this is not intended to cover the genesis block; it's really about covering the no-blocks-of-any-kind usecaswe.
Adding genesis is ok of course, but I don't feel that it's needed since it wouldn't be treated any differently from the indexer perspective.

api/indexer/client_test.go
```diff
@@ -144,17 +210,22 @@ func TestIndexerClientTransactions(t *testing.T) {
 				}, txResponse)
 			}
 
-			txResponse, tx, found, err := client.GetTx(ctx, executedTx.GetID(), parser)
-			r.Equal(tt.err, err)
+			txResponse, tx, found, err := client.GetTx(ctx, executedTx.GetID(), tt.parser)
+			if tt.getTxErr != nil || err != nil {
```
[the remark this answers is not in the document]
tsachiherman: I've updated the testing code to ensure it won't panic in that condition.

api/indexer/client_test.go
```diff
@@ -127,14 +183,24 @@ func TestIndexerClientTransactions(t *testing.T) {
...
+				// create a malformed block and have the indexer add it.
+				// removing of the first result makes the length of the ExecutionResults differ from the
+				// transaction index within the block.
+				executedBlock.ExecutionResults.Results = executedBlock.ExecutionResults.Results[1:]
+				r.NoError(indexer.Notify(ctx, executedBlock))
+			}
+
 			txResponse, found, err := client.GetTxResults(ctx, executedTx.GetID())
-			r.Equal(tt.err, err)
+			if tt.getTxResultsErr != nil || err != nil {
```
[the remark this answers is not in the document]
tsachiherman: I've updated the testing code to ensure it won't panic in that condition.

PR #2004, opened 2025-03-25
[load] Add UTs for Gradual Load Orchestrator

RodrigoVillar:
This PR adds UTs for the gradual load orchestrator.

This PR also makes the following modifications:
- Register the `txLatency` metric
- Update gradual load orchestrator `attempts` to use one-based numbering.
- Update `computeTPS` to use a more precise TPS calculation
    - Previously, we were rounding down the total number of seconds passed before computing the rate change. We now round down after we've made the rate of change computation.


load/gradual_load_orchestrator.go
```diff
@@ -133,12 +133,12 @@ func (o *GradualLoadOrchestrator[T, U]) Execute(ctx context.Context) error {
 // 3. the maximum number of attempts to reach a target TPS has been reached
 func (o *GradualLoadOrchestrator[T, U]) run(ctx context.Context) bool {
 	var (
-		prevConfirmed = o.tracker.GetObservedConfirmed()
-		prevTime      = time.Now()
-		currTargetTPS = new(atomic.Uint64)
+		prevConfirmed        = o.tracker.GetObservedConfirmed()
+		prevTime             = time.Now()
+		currTargetTPS        = new(atomic.Uint64)
+		attempts      uint64 = 1
```
aaronbuchwald, line 139: Why do we need to start at 0 instead of 1?

load/gradual_load_orchestrator_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+func (*mockTracker) GetObservedIssued() uint64 {
+	return 0
+}
+
+func (*mockTracker) Issue(any) {}
+
+func (*mockTracker) ObserveConfirmed(any) {}
+
+func (*mockTracker) ObserveFailed(any) {}
```
aaronbuchwald, line 228: Why do we need a mock tracker instead of just using our production code?

load/gradual_load_orchestrator.go
RodrigoVillar: Do you mean why we need to start at 1 instead of 0? If so, then the change is purely cosmetic (i.e. if `MaxAttempts = 3`, then this change makes it so that the gradual orchestrator prints: `attempt number 1`, `attempt number 2`, ... instead of `attempt number 0`, `attempt number 1`, ... and exits with `number of attempts: 3` instead of `number of attempts: 2`

PR #2005, opened 2025-03-25
Backfill chain's validity window on startup

Elvis339:
This PR addresses a non-deterministic issue that occurs during node restart. When a node restarts and finds that it is only a few blocks behind (or, in rare cases, slightly ahead) of the network, it may decide to skip the state sync process due to small block gap. This behavior causes the node’s validity window—which tracks recent transactions—to remain incomplete. This can cause the node to accept a transaction that the rest of the network already accepted (it doesn’t realize it’s a duplicate). As a result, the network will reject the transaction, creating a mismatch


vm/vm.go
```diff
@@ -697,3 +702,47 @@ func (vm *VM) Submit(
...
+		}
+		currentBlock = parentBlock
+	}
+
+	for i := len(executionBlocks) - 1; i >= 0; i-- {
+		vm.chainTimeValidityWindow.Accept(executionBlocks[i])
+	}
+
+	return nil
+}
```
aaronbuchwald, line 748: Why move this here into the VM rather than the validity window itself?

This duplicates the code in the validity window syncer, so I wonder if we can better share this or even just move it into the constructor.

Elvis339: The chain validity window seemed to be highlighted so I tough it's only relevant there, I'll move this in the constructor than. Thanks.

Elvis339: Resolved privately.

**Summary:**

- We cannot prefill the validity window during the VM constructor because of a circular dependency:
  - The validity window is needed by the chain,
  - The chainStore (which fetches blocks from disk) requires the chain,
  - And the validity window itself needs the chainStore to backfill its state.
  
- Additionally, the ChainIndex interface used in the validity window (referenced in vm/resolutions.go#L79) is different from the one in chainindex/chain_index.go, which creates confusion.

- At initialization, the consensus index is nil (it’s only set later via SetConsensusIndex), so we must rely on the chainStore (which only contains accepted blocks) for backfilling the validity window.

- Aaron pointed out that we need the consensus index to fetch blocks still in consensus, since the chainStore only contains blocks that have been accepted. This led to the idea of exporting the backfill logic to avoid duplication.

- In summary, due to these dependency challenges, the current design requires backfilling from disk via the chainStore, and the consensus index will only support fetching blocks from both disk and live consensus once it’s set later via SetConsensusIndex.

vm/vm.go
```diff
@@ -697,3 +701,29 @@ func (vm *VM) Submit(
...
+// Therefore, we must use the chainStore (which reads blocks from disk) to backfill the validity window.
+// This prepopulation ensures the validity window is complete, even if state sync is skipped.
+func (vm *VM) populateValidityWindow(ctx context.Context) error {
+	lastAcceptedBlkHeight, err := vm.chainStore.GetLastAcceptedHeight(ctx)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			return nil
+		}
+		return err
+	}
```
aaronbuchwald, line 721:
```suggestion
	if err == database.ErrNotFound {
		return nil
	}
	if err != nil {
		return fmt.Errorf("failed to fetch last accepted height to populate validity window: %w", err)
	}
```

vm/vm.go
```diff
@@ -334,6 +334,10 @@ func (vm *VM) Initialize(
 		return nil, nil, nil, false, err
 	}
 
+	if err := vm.populateValidityWindow(ctx); err != nil {
+		return nil, nil, nil, false, err
+	}
+
```
aaronbuchwald, line 340: proposed alternative - we can also execute this within `startNormalOp` or `AddNormalOpStarter` immediately below.

`AddNormalOpStarter` adds a callback that will be executed when the engine invokes `SetState(snow.NormalOp)` ie. transitioning from state sync / bootstrapping to normal operation.

`vm.startNormalOp` is invoked when the VM _actually_ transitions into normal operation. Ie. the AvalancheGo consensus engine sets the state to normal operation when starting dynamic state sync even though we are not actually in a normal mode of operation.

We define `startNormalOp` to perform all operations that should be invoked when transitioning into normal operation either via the normal flow ie. initialize and skip state sync or via state sync completing.

This has the added advantage of being invoked after `SetConsensusIndex` has been called, so we can defer setup code like `populateValidityWindow` that depends on it.

internal/validitywindow/validitywindow.go
```diff
@@ -200,6 +200,39 @@ func (v *TimeValidityWindow[T]) calculateOldestAllowed(timestamp int64) int64 {
...
+
+		seenValidityWindow = block.GetTimestamp()-parent.GetTimestamp() > validityWindow
+		if seenValidityWindow {
+			break
+		}
+	}
+
+	for i := len(parents) - 1; i >= 0; i-- {
+		v.Accept(parents[i])
+	}
```
aaronbuchwald, line 232: (not necessary for this PR) use `slices.Reverse(blocks)` for simpler to read code

aaronbuchwald (approved):
LGTM - left some optional suggestions

PR #2007, opened 2025-03-26
Update docs for prerequisites and golang version maintenance

maru-ava:
- Document how to update the golang version
 - Document the requirement for modern bash on macos

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,9 @@ Thank you for your interest in contributing to `hypersdk`! By contributing to hy
 
 To contribute to `hypersdk`, you'll need:
 
-- [Go](https://golang.org/dl/) 1.22.8 or higher
+- [Go](https://golang.org/dl/) 1.23.7 or higher
+
+On MacOS, a modern version of bash is required (e.g. via [homebrew](https://brew.sh/) with `brew install bash`). The version installed by default is not compatible with HyperSDK's [shell scripts](scripts).
 
 ### Setting up your development environment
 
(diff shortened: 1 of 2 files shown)


RodrigoVillar (dismissed):
LGTM

CONTRIBUTING.md
```diff
@@ -8,7 +8,9 @@ Thank you for your interest in contributing to `hypersdk`! By contributing to hy
 
 To contribute to `hypersdk`, you'll need:
 
-- [Go](https://golang.org/dl/) 1.22.8 or higher
+- [Go](https://golang.org/dl/) 1.23.7 or higher
+
+On MacOS, a modern version of bash is required (e.g. via [homebrew](https://brew.sh/) with `brew install bash`). The version installed by default is not compatible with HyperSDK's [shell scripts](scripts).
```
aaronbuchwald, line 13: Could we mention and link the optional dev shell here as well? 

imo it's preferred to treat this as a convenient option rather than a requirement and specify (as we do here) the installation requirements for using / contributing to HyperSDK.

go.mod
```diff
@@ -1,5 +1,12 @@
 module github.com/ava-labs/hypersdk
 
+// - Changes to the minimum golang version must also be replicated in:
+//   - CONTRIBUTING.md
+//   - examples/morpheusvm/go.mod
+//   - go.mod (here)
```
aaronbuchwald, line 6: Quick search through the repo, there are a few further places that we should mention here:

- Dockerfile
- devcontainer.json (only for minor version bumps)

aaronbuchwald (approved):

PR #2010, opened 2025-03-27
[chain] Add Timestamp as Parameter to TX Generation Functions

RodrigoVillar:
This PR allows for the timestamp to be passed into the TX generation functions, which makes it possible for applications like #1955 to create TXs while using their own timestamp (instead of `time.Now().UnixMilli()`.

### DIFF
--- a/chain/transaction.go
+++ b/chain/transaction.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
-	"time"
 
 	"github.com/StephenButtolph/canoto"
 	"github.com/ava-labs/avalanchego/ids"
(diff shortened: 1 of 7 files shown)


chain/transaction.go
```diff
@@ -574,10 +573,11 @@ func (*Transaction) CanotoSpec(types ...reflect.Type) *canoto.Spec {
 func GenerateTransaction(
 	ruleFactory RuleFactory,
 	unitPrices fees.Dimensions,
+	timestamp int64,
```
tsachiherman, line 576: this is the block timestamp, right ?

tsachiherman (approved):
this PR looks good to me.
One comment though; I'm not sure if we always want to use the current time or whether we want to use the previous block's timestamp.
Given that this PR doesn't change any of that logic, I think that we can defer that decision to another time.

chain/transaction.go
RodrigoVillar: No. `timestamp` is the current time (which is padded in `Base` to follow validity window rules). You could use the block timestamp but that isn't necessary.

aaronbuchwald: You cannot safely use the last block's timestamp. If you did, then when the next block timestamp had increased, your transaction would already be marked as expired.

aaronbuchwald: (not necessary for this PR) - could we add comments explaining the parameters to `GenerateTransaction` and `GenerateTransactionManual` - `timestamp` is a bit confusing here because it could be the current time or the expiry used in `chain.Base`, but we actually pad it as you mentioned with the validity window duration to maximize the chance that it's included.

We may also want to rename the functions for clarity `-Manual` does not really explain the difference between these two functions

aaronbuchwald (approved):

PR #2011, opened 2025-03-27
Add package level godoc for snow

aaronbuchwald:
This PR adds a `doc.go` with a high level description of the snow package.

### DIFF
--- a/snow/doc.go
+++ b/snow/doc.go
@@ -7,9 +7,9 @@
 //
 // What does the original interface require that this package simplifies?
 //
-// 1. Maintaining a [Processing Block Tree]
-// 2. Vacuous block verification and acceptance during [Dynamic State Sync]
-// 3. [common.VM] interface that we can replace with good, customizable defaults
+//  1. Maintaining a [Processing Block Tree]
+//  2. Vacuous block verification and acceptance during [Dynamic State Sync]
+//  3. [common.VM] interface that we can replace with good, customizable defaults
 //
 // This package defines a simpler to use interface for use within the HyperSDK and to eventually iterate from
 // an initial small, well-defined public API backwards to more layers of customizability including building directly on top of this package.
--- a/snow/doc.go
+++ b/snow/doc.go
@@ -2,10 +2,10 @@
 // See the file LICENSE for licensing terms.
 
 // Snow provides an adapter from the AvalancheGo [block.ChainVM] interface (the interface AvalancheGo VM developers must implement)
-// to provide a more ergonomic interface for VM developers.
+// to provide a more ergonomic interface to VM developers.
 //
-// The existing interface was largely defined for ease of use by the AvalancheGo consensus engine and as a single all access type for
-// anything required by AvalancheGo (APIs, health check, shutdown, etc.)
+// The existing interface was largely defined for ease of use by the AvalancheGo consensus engine. As a result,
+// all responsibilities are combined into a single, all-access type for anything required by AvalancheGo (APIs, health check, shutdown, etc.).
 //
 // # What does the original interface require that this package simplifies?
 //
--- a/snow/doc.go
+++ b/snow/doc.go
@@ -1,0 +1,74 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// Snow provides an adapter from the AvalancheGo [block.ChainVM]() interface to provide a more ergonomic interface for chain developers.
+// The existing interface was largely defined for ease of use by the AvalancheGo consensus engine and as a single all access type for
+// anything required by AvalancheGo (APIs, health check, shutdown, etc.)
+//
+// What does the original interface require that this package simplifies?
+// 1. Maintaining a tree of processing blocks and proposed state diffs [ref](https://github.com/ava-labs/avalanchego/tree/a353b45944904ce56b5eddf961bf83430892dc32/vms#implementing-the-snowman-vm-block)
+// 2. Vacuous block verification and acceptance during [Dynamic State Sync](https://github.com/ava-labs/avalanchego/blob/a353b45944904ce56b5eddf961bf83430892dc32/snow/engine/snowman/block/README.md#state-sync-engine-role-and-workflow)
+// 3. [common.VM](https://github.com/ava-labs/avalanchego/blob/a353b45944904ce56b5eddf961bf83430892dc32/snow/engine/common/vm.go#L17) interface that we can replace with good, customizable defaults
+//
+// This package defines a simpler to use interface for use within the HyperSDK and to eventually iterate from
+// an initial small, well-defined public API backwards to more layers of customizability including building directly on top of this package.
+//
+// This package simplifies the existing ChainVM interface by implementing the [snowman.Block]() interface and allowing VM developers to provide
+// generic type parameters for each possible state of a block within consensus (bytes, input/un-verified, output/verified, accepted).
+// Developers implement a single function to transition between each of these states with all of the information guaranteed to be available
+// at that point in time.
+//
+// For example, when a block is verified, its parent is guarnateed to have been verified already, but may still be in processing. Therefore,
+// chain developers implement `VerifyBlock` as follows:
+//
+// ```go
+//
+//	VerifyBlock(
+//		ctx context.Context,
+//		parent O,
+//		block I,
+//	) (O, error)
+//
+// ```
+// 
+// See [Chain] for complete documentation of the interface.
+//
+// Further, the package provides two functions `Initialize` and `SetConsensusIndex` as the effective "entrypoint" of the VM.
+// This needs to be split into two parameters, because the consensus index depends on a fully initialized version of the chain.
+// This provides all of the information available to the VM at the time of initialization. See [Chain.Initialize] and
+// [Chain.]
+// This package implements the [snowman.Block] type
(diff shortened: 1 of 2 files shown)


aaronbuchwald:
Screenshot 2025-03-28 at 09 34 04

Elvis339:
I think we can improve reading flow and format if we were to use Markdown supported by `godoc`. 
Additionally this creates anchors which can be shared.

For example
```
// Package snow simplifies VM development on AvalancheGo by providing adapter 
// between VM logic and AvalancheGo's consensus engine.
//
// # Why Snow Exists
//
// Building directly on the AvalancheGo [block.ChainVM] interface requires deep knowledge of consensus internals. 
// Snow exists to abstract away these complexities, allowing developers to focus on their VMs unique logic rather than
// consensus integration details.
```

snow/doc.go
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// Snow provides an adapter from the AvalancheGo [block.ChainVM] interface to provide a more ergonomic interface for chain developers.
+// The existing interface was largely defined for ease of use by the AvalancheGo consensus engine and as a single all access type for
+// anything required by AvalancheGo (APIs, health check, shutdown, etc.)
+//
```
Elvis339, line 7: GoLand is reporting this:

Screenshot 2025-03-28 at 6 36 49 PM

Potentail improvement:
```
// Package snow simplifies custom VM development by providing adapter from the AvalancheGo [block.ChainVM] between custom VM logic and AvalancheGo's consensus engine. 
```

snow/doc.go
```diff
@@ -0,0 +1,43 @@
...
+// What does the original interface require that this package simplifies?
```
Elvis339, line 8:
```suggestion
// # What does the original interface require that this package simplifies?
```

snow/doc.go
```diff
@@ -0,0 +1,43 @@
...
+//
+//  1. Maintaining a [Processing Block Tree]
+//  2. Vacuous block verification and acceptance during [Dynamic State Sync]
+//  3. [common.VM] interface that we can replace with good, customizable defaults
+//
+// This package defines a simpler to use interface for use within the HyperSDK and to eventually iterate from
+// an initial small, well-defined public API backwards to more layers of customizability including building directly on top of this package.
```
Elvis339, line 15: I think this does not belong to the documentation, it describes internal development strategy rather than information about the package.

IMO documentation should focus on what the package currently does, how to use it, and what problems it solves. Future development plans would be better communicated in a separate roadmap document or project README, not in the package's API documentation.

snow/doc.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// For example, when a block is verified, its parent is guaranteed to have been verified successfully. However, it may not be accepted. Therefore,
+// chain developers implement `VerifyBlock` and receive the verified version of the parent (but not accepted type) and the input block to be verified:
+//
+//	VerifyBlock(
+//		ctx context.Context,
+//		parent O,
+//		block I,
+//	) (O, error)
+//
```
Elvis339, line 30: A combination of what you wrote with this should answer the "why snow" question, imo. 
```
// # Snow egonomics (or something similar, this doesn't need to be h1)
// Snow simplifies VM development through a type-safe block state system
//
// Instead of managing block state machine, Snow introduces
// **generic** type parameters that represent each distinct block state:
//
//   - B: Raw block bytes
//   - I: Input/unverified blocks ready for processing
//   - O: Output/verified blocks that passed validation
//   - A: Accepted blocks committed to the chain
//
// This type system ensures you always have access to the appropriate block
// representation at each stage of consensus, preventing common errors and
// simplifying implementation.
```

snow/doc.go
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// Snow provides an adapter from the AvalancheGo [block.ChainVM] interface to provide a more ergonomic interface for chain developers.
```
Elvis339, line 4: It might be too much but it'd be good to explain block.ChainVM:

```
ChainVM is the required interface that any Snowman-compatible blockchain must implement to work with Avalanche's consensus engine. 
```

snow/doc.go
```diff
@@ -0,0 +1,43 @@
...
+//		parent O,
+//		block I,
+//	) (O, error)
+//
+// See [Chain] for complete documentation of the interface.
+//
+// Further, the package provides two functions [chain.Initialize] and [chain.SetConsensusIndex] as the effective "entrypoint" of the VM.
+// This needs to be split into two parameters, because the consensus index depends on a fully initialized version of the chain.
+// This provides all of the information available to the VM at the time of initialization. See [Chain.Initialize] and
+// [Chain.SetConsensusIndex] for details on the parameters provided to the VM at initialization.
```
Elvis339, line 36: This is very useful! 
nit: I don't see the links

Screenshot 2025-03-28 at 7 02 34 PM

snow/doc.go
aaronbuchwald: Do you know what specific linter or rule that's coming from? I'd prefer to add the linter and start to follow it.

snow/doc.go
aaronbuchwald: Changing to:

```
// This package implements these tough to get right and tedious components and provides
// a slimmed down interface for VM developers to implement.
```

snow/doc.go
aaronbuchwald: Struggling to get these links to work within the same package based off these docs https://tip.golang.org/doc/comment#links ... 🤔

snow/doc.go
aaronbuchwald: Going with a bit shorter: `(the interface AvalancheGo VM developers must implement)`

snow/doc.go
aaronbuchwald: Great call on breaking out the block types this way, incorporated this feedback with a bit more modification

snow/doc.go
aaronbuchwald: Maybe I'm too close to it, but a long explanation of what `block.ChainVM` is seems like overkill or re-summarizing the same information to me. Tried to keep it short, let me know if you have another suggestion!

snow/doc.go
aaronbuchwald: Great call to use markdown formatting - godoc is a whole new world for me

snow/doc.go
```diff
@@ -0,0 +1,55 @@
...
+//   - [Accepted]: accepted block
+//
+// This provides a customizable block representation and ensures that VM developers
+// have access to the exact set of information guaranteed to be available in each
+// state of block handling.
+//
+// For example, when a block is verified, its parent is guaranteed to have been verified successfully. However, it may not be accepted. Therefore,
+// chain developers implement `VerifyBlock` and receive the verified version of the parent (but not accepted type) and the input block to be verified:
+//
+//	VerifyBlock(
```
RodrigoVillar, line 36: nit: change to `Ouput, Input`

snow/doc.go
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) 2024, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+// Snow provides an adapter from the AvalancheGo [block.ChainVM] interface (the interface AvalancheGo VM developers must implement)
+// to provide a more ergonomic interface to VM developers.
+//
+// The existing interface was largely defined for ease of use by the AvalancheGo consensus engine. As a result,
+// all responsibilities are combined into a single, all-access type for anything required by AvalancheGo (APIs, health check, shutdown, etc.).
```
RodrigoVillar, line 8: > As a result, all responsibilities are combined into a single, all-access type for anything required by AvalancheGo (APIs, health check, shutdown, etc.).

Is this the problem with using the current `block.ChainVM` interface? If so, can we make it so that the problem is evident? It seems that this is a statement rather than describing the actual problem.

snow/vm.go
```diff
@@ -95,12 +95,12 @@ type VM[I Block, O Block, A Block] struct {
...
 
-	verifiedSubs         []event.Subscription[O]
-	rejectedSubs         []event.Subscription[O]
-	acceptedSubs         []event.Subscription[A]
-	preReadyAcceptedSubs []event.Subscription[I]
+	verifiedSubs         []event.Subscription[Output]
+	rejectedSubs         []event.Subscription[Output]
+	acceptedSubs         []event.Subscription[Accepted]
+	preReadyAcceptedSubs []event.Subscription[Input]
 	// preRejectedSubs handles rejections of I (Input) during/after state sync, before they reach O (Output) state
```
RodrigoVillar, line 102: Can we remove references to the former `I` generic name?

snow/doc.go
RodrigoVillar: I think the following:

> This package implements these tough to get right and tedious components and provides a slimmed down interface for VM developers to implement.

is actual describing the problem.

RodrigoVillar (commented):
Seeing `chain.Initialize` and `chain.SetConsensusIndex` hurts me, but everything else looks great!

snow/doc.go
Elvis339: Found this lint rule: https://staticcheck.dev/docs/checks/#ST1020

Issue #2003, opened 2025-03-25
Initialize Validity Window outside of State Sync


aaronbuchwald:
We initialize the time validity window here:

https://github.com/ava-labs/hypersdk/blob/d296eddf29a33329d7a6a1b6f359ff59df2f7585/vm/vm.go#L300

This relies on populating the `seen` emap to contain the transactions (or containers) of all blocks within the validity window's range ie.

If the last accepted block has timestamp 1000 and the validitywindow is 120, then it must contain all transactions included in blocks with a timestamp in the range [880, 1000].

The chain's time validity window is updated while accepting blocks during normal operation of consensus (after dynamic state sync) and from the state syncer.

The state syncer will backfill the validity window from blocks that are already on disk: https://github.com/ava-labs/hypersdk/blob/d296eddf29a33329d7a6a1b6f359ff59df2f7585/internal/validitywindow/syncer.go#L72, add blocks as they are accepted during dynamic state sync consensus, and proactively backfill them by fetching from the network.

However, this assumes that the VM decides to start dynamic state sync and executes this line:

https://github.com/ava-labs/hypersdk/blob/d296eddf29a33329d7a6a1b6f359ff59df2f7585/statesync/client.go#L105

This is not guaranteed to execute if we start up and see that we are less than min blocks behind the proposed new tip. Additionally, there's an edge case where the ProposerVM will accept a dynamic state sync summary and skip calling accept on the inner VM if it has already accepted up to that height (assumes if the outer VM has accepted up to block 100, then there's no need for the inner VM to accept a state sync summary targeting block 100) (https://github.com/ava-labs/hypersdk/issues/1993).

If we don't initialize the the validity window correctly, then we can see non-deterministic results in e2e test runs due to nodes disagreeing about the validity of a block.

To fix this, we should:

- populate the `seen` emap of the validity window in the constructor, so that it does not require a state sync summary to be accepted
- force the ProposerVM to call `Accept` on the inner state summary if invoking it on the outer summary (https://github.com/ava-labs/avalanchego/pull/3831)
- document the current initialization paths with normal operation/bootstrapping/state sync from AvalancheGo (simplify if possible)

PR #2012, opened 2025-03-28
Followup pre populate chain validity window

Elvis339:
This PR is closing https://github.com/ava-labs/hypersdk/issues/2003 by addressing several issues related to the validity window initialization flow.

1. **Improved Validity Window API**: 
   - Added `NewPopulatedTimeValidityWindow` as a factory method to create and populate a validity window in one step
   - Made `populateValidityWindow` private (formerly `PopulateValidityWindow`) to reduce API surface area
   - Simplified timestamp comparison logic within `populateValidityWindow`

2. **Cleaner VM Initialization Flow**:
   - Removed the circular dependency between chain store and validity window
   
3. **Validity Window Initialization**
    - Enforce populated validity window before starting normal operation
    - Save historical blocks on disk


Elvis339:
https://github.com/ava-labs/hypersdk/pull/2005#discussion_r2014825899 -> https://github.com/ava-labs/hypersdk/blob/5d398413cebfb02e3c16625c82d20b7b027a4d50/internal/validitywindow/syncer.go#L138

https://github.com/ava-labs/hypersdk/pull/2005#discussion_r2014828861 -> https://github.com/ava-labs/hypersdk/blob/5d398413cebfb02e3c16625c82d20b7b027a4d50/internal/validitywindow/validitywindow.go#L218

https://github.com/ava-labs/hypersdk/pull/2005#discussion_r2014917234 -> https://github.com/ava-labs/hypersdk/blob/5d398413cebfb02e3c16625c82d20b7b027a4d50/vm/resolutions.go#L83

https://github.com/ava-labs/hypersdk/pull/2005#discussion_r2014921116 -> https://github.com/ava-labs/hypersdk/blob/5d398413cebfb02e3c16625c82d20b7b027a4d50/vm/vm.go#L727

https://github.com/ava-labs/hypersdk/pull/2005#discussion_r2014930810 -> https://github.com/ava-labs/hypersdk/blob/5d398413cebfb02e3c16625c82d20b7b027a4d50/vm/vm.go#L718

https://github.com/ava-labs/hypersdk/pull/2005#discussion_r2014932635 -> https://github.com/ava-labs/hypersdk/blob/5d398413cebfb02e3c16625c82d20b7b027a4d50/internal/validitywindow/validitywindow.go#L241

chain/chain.go
```diff
@@ -16,25 +16,29 @@ import (
...
+type ExecutionBlockParser interface {
+	ParseBlock(ctx context.Context, bytes []byte) (*ExecutionBlock, error)
+}
+
 type Chain struct {
 	builder     *Builder
 	processor   *Processor
 	preExecutor *PreExecutor
-	blockParser *BlockParser
+	blockParser ExecutionBlockParser
```
RodrigoVillar, line 27: nit: since `NewChain` has an argument `executionBlockParser`, can we also switch from `blockParser` to `executionBlockParser` in the `Chain` struct?

internal/validitywindow/validitywindow.go
```diff
@@ -196,16 +210,12 @@ func (v *TimeValidityWindow[T]) isRepeat(
 	}
 }
 
-func (v *TimeValidityWindow[T]) calculateOldestAllowed(timestamp int64) int64 {
-	return max(0, timestamp-v.getTimeValidityWindow(timestamp))
-}
-
-func (v *TimeValidityWindow[T]) PopulateValidityWindow(ctx context.Context, block ExecutionBlock[T]) ([]ExecutionBlock[T], bool) {
+func (v *TimeValidityWindow[T]) populateValidityWindow(ctx context.Context, block ExecutionBlock[T]) ([]ExecutionBlock[T], bool) {
```
RodrigoVillar, line 213: Could we add some documentation here describing the order of the blocks returned (i.e. "youngest-to-oldest")?

RodrigoVillar (changes requested):

chain/chain.go
Elvis339: I'm going to switch it back to `blockParser`, `executonBlockParser` is too definitive since `snow` uses generic type parameters to represent different block states.

internal/validitywindow/validitywindow.go
Elvis339: Sure, we have it few lines below: https://github.com/ava-labs/hypersdk/blob/5d398413cebfb02e3c16625c82d20b7b027a4d50/internal/validitywindow/validitywindow.go#L225

RodrigoVillar (dismissed):
LGTM

vm/vm.go
```diff
@@ -297,9 +297,16 @@ func (vm *VM) Initialize(
...
-	})
+	executionBlockParser := chain.NewBlockParser(vm.Tracer(), vm.txParser)
+
+	if err := vm.initChainStore(executionBlockParser); err != nil {
+		return nil, nil, nil, false, err
+	}
+
+	if err := vm.initValidityWindow(ctx); err != nil {
+		return nil, nil, nil, false, err
+	}
```
aaronbuchwald, line 308: We cannot guarantee within `vm.Initialize` that the validity window is fully populated.

If we move this into `startNormalOp`, then we CAN guarantee that some code we define ie. `initValidityWindow` executes and indicates that we successfully filled the validity window before entering normal operation.

vm/vm.go
```diff
@@ -702,28 +704,52 @@ func (vm *VM) Submit(
...
+// a node restarts with an incomplete validity window that could allow duplicate transactions.
+func (vm *VM) initValidityWindow(ctx context.Context) error {
+	getWindowSize := func(timestamp int64) int64 {
+		return vm.ruleFactory.GetRules(timestamp).GetValidityWindow()
+	}
+
+	// If we don't have a last accepted block or we're not in normal operation,
+	// initialize with empty validity window to be populated later
+	lastAcceptedHeight, err := vm.chainStore.GetLastAcceptedHeight(ctx)
+	normalOp := vm.normalOp.Load()
```
aaronbuchwald, line 726: If we move this into `startNormalOp` then we can guarantee that at least the genesis or some last accepted block has been accepted, so that we do not need to handle the error this way.

aaronbuchwald, line 726: Currently `vm.normalOp` will always be false when this function is called

internal/validitywindow/validitywindow.go
```diff
@@ -79,6 +80,19 @@ func NewTimeValidityWindow[T emap.Item](
...
 }
 
+func NewPopulatedTimeValidityWindow[T emap.Item](
+	ctx context.Context,
+	log logging.Logger,
+	tracer trace.Tracer,
+	chainIndex ChainIndex[T],
+	lastAcceptedBlock ExecutionBlock[T],
+	getTimeValidityWindowF GetTimeValidityWindowFunc,
+) *TimeValidityWindow[T] {
```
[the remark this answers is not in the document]
Elvis339: I was thinking to keep the intent clear for example https://github.com/ava-labs/hypersdk/blob/db938f585be2269d703eb59845532653d4decfba/vm/vm.go#L723, I can remove it if it creates confusion.

internal/validitywindow/validitywindow.go
```diff
@@ -62,21 +63,33 @@ type TimeValidityWindow[T emap.Item] struct {
...
-	return &TimeValidityWindow[T]{
+	t := &TimeValidityWindow[T]{
 		log:                   log,
 		tracer:                tracer,
 		chainIndex:            chainIndex,
 		seen:                  emap.NewEMap[T](),
 		getTimeValidityWindow: getTimeValidityWindowF,
+		populated:             false,
+	}
+	if lastAcceptedBlock != nil {
```
Elvis339, line 85: Allowing `nil` `lastAcceptedBlock` for testing purposes. While not ideal for production use, passing `nil` signals the creation of an empty validity window. The constructor handles this case gracefully. I'm open to improvements. 
```go
vw := NewTimeValidityWindow(ctx, log, tracer, chainIndex, nil, ...)
```

vm/resolutions.go
```diff
@@ -82,9 +82,10 @@ func (vm *VM) GetExecutionBlock(ctx context.Context, blkID ids.ID) (validitywind
 
 	if vm.consensusIndex != nil {
 		blk, err := vm.consensusIndex.GetBlock(ctx, blkID)
-		if err == nil {
-			return blk, nil
+		if err != nil {
+			return nil, err
 		}
+		return blk, nil
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/83b5138bf35a501b4c68366e17f867840abe4d82

vm/vm.go
```diff
@@ -544,8 +554,23 @@ func (vm *VM) startNormalOp(ctx context.Context) error {
...
+	}
+	executionBlk, err := vm.GetExecutionBlock(ctx, lastAccepted.GetID())
+	if err != nil {
+		return err
+	}
+
+	isValidityWindowComplete := vm.chainTimeValidityWindow.Complete(ctx, executionBlk)
+	if !isValidityWindowComplete {
+		// Fail loudly, a validity window is critical part to the vm
+		panic("critical error: validity window's partial state may lead to inconsistencies")
```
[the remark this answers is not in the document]
Elvis339: In both cases panic or error, AvalancheGo shut's down consensus engine and subsequent services.

```
[04-18|15:40:23.650] INFO <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snowman/engine.go:436 shutting down consensus engine
[04-18|15:40:23.650] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:465 Shutting down VM
[04-18|15:40:23.650] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:470 Shutting down service {"service": "statedb"}
[04-18|15:40:23.650] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> gossiper/target.go:323 stopping gossip loop
[04-18|15:40:23.662] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:473 Finished shutting down service {"service": "statedb", "duration": "12.051709ms"}
[04-18|15:40:23.662] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:470 Shutting down service {"service": "results"}
[04-18|15:40:23.667] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:473 Finished shutting down service {"service": "results", "duration": "5.166125ms"}
[04-18|15:40:23.667] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:470 Shutting down service {"service": "auth verifiers"}
[04-18|15:40:23.667] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:473 Finished shutting down service {"service": "auth verifiers", "duration": "2.666µs"}
[04-18|15:40:23.667] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:470 Shutting down service {"service": "chainindex"}
[04-18|15:40:23.672] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:473 Finished shutting down service {"service": "chainindex", "duration": "4.82075ms"}
[04-18|15:40:23.672] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:470 Shutting down service {"service": "syncer"}
[04-18|15:40:23.676] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:473 Finished shutting down service {"service": "syncer", "duration": "4.056708ms"}
[04-18|15:40:23.676] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:470 Shutting down service {"service": "builder"}
[04-18|15:40:23.676] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:473 Finished shutting down service {"service": "builder", "duration": "1.458µs"}
[04-18|15:40:23.676] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:470 Shutting down service {"service": "gossiper"}
[04-18|15:40:23.676] [0;0mINFO[0;0m <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> snow/vm.go:473 Finished shutting down service {"service": "gossiper", "duration": "1.125µs"}
[04-18|15:40:23.677] DEBUG <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> subprocess/non_linux_stopper.go:25 subprocess was killed
[04-18|15:40:23.677] INFO <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> router/chain_router.go:676 chain shutdown {"shutdownDuration": "26.829ms"}
[04-18|15:40:23.679] INFO <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> subprocess/runtime.go:111 stdout collector shutdown
[04-18|15:40:23.679] INFO <27aFNxNZDqDnNnrn6ync3fgFXhxLp6JE2EizCToApjfKTR3snY Chain> subprocess/runtime.go:124 stderr collector shutdown
```

internal/validitywindow/syncer.go
```diff
@@ -83,6 +90,15 @@ func (s *Syncer[T, B]) Start(ctx context.Context, target B) error {
...
 		for blk := range resultChan {
+			err := s.blockStore.SaveHistorical(blk)
+			if err != nil {
+				s.errChan <- fmt.Errorf(
+					"%w; aborting to prevent inconsistencies %w",
+					errSaveHistoricalBlocks,
+					err,
+				)
+				return
+			}
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/3581f267ab44d723a4e449cf3aa0ac657a514c0e

internal/validitywindow/validitywindow.go
```diff
@@ -53,30 +54,57 @@ type Interface[T emap.Item] interface {
...
 		getTimeValidityWindow: getTimeValidityWindowF,
 	}
+	if tip != nil {
+		t.populate(ctx, tip)
+	}
+	return t
+}
+
+// Complete will attempt to complete a validity window it will return boolean
+// as a signal if it's ready to reliably prevent replay attacks
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/d2639033d9724123963d9dd0a3442312560b5796

internal/validitywindow/validitywindow.go
```diff
@@ -53,30 +54,57 @@ type Interface[T emap.Item] interface {
 	IsRepeat(ctx context.Context, parentBlk ExecutionBlock[T], currentTimestamp int64, containers []T) (set.Bits, error)
 }
 
+// TimeValidityWindow is a time-based transaction uniqueness mechanism that prevents replay attacks.
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/d2639033d9724123963d9dd0a3442312560b5796

internal/validitywindow/validitywindow.go
```diff
@@ -53,30 +54,57 @@ type Interface[T emap.Item] interface {
...
+//     they would disagree about the state of the blockchain.
+//
+// TimeValidityWindow builds up on assumption of ChainIndex being up to date
 type TimeValidityWindow[T emap.Item] struct {
-	log    logging.Logger
-	tracer trace.Tracer
-
+	log                     logging.Logger
+	tracer                  trace.Tracer
 	lock                    sync.Mutex
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/d2639033d9724123963d9dd0a3442312560b5796

vm/vm.go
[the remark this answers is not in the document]
RodrigoVillar: Could we use an error here/maybe use `log.Fatal` instead? If I understand correctly, `panic()` should be reserved for the top-level caller (i.e. propagate error to caller who then panics). Also open to leaving this as is if there's a strong benefit to the current code.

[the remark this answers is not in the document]
aaronbuchwald: Can we log and return an error instead? Returning an error here will be treated as fatal, if we want to include more detail we can include a stacktrace in the error message. We should not panic here.

vm/vm.go
```diff
@@ -354,21 +359,19 @@ func (vm *VM) Initialize(
...
 	stateReady := !vm.SyncClient.MustStateSync()
-	var lastAccepted *chain.OutputBlock
-	if stateReady {
-		lastAccepted, err = vm.initLastAccepted(ctx)
-		if err != nil {
-			return nil, nil, nil, false, err
-		}
+	// The branch is executed when the VM must preform state sync
+	if !stateReady {
+		lastAccepted = nil
```
aaronbuchwald, line 364: What's the reason for this change?

internal/validitywindow/validitywindow.go
```diff
@@ -53,36 +54,63 @@ type Interface[T emap.Item] interface {
...
 func NewTimeValidityWindow[T emap.Item](
+	ctx context.Context,
 	log logging.Logger,
 	tracer trace.Tracer,
 	chainIndex ChainIndex[T],
+	tip ExecutionBlock[T],
 	getTimeValidityWindowF GetTimeValidityWindowFunc,
 ) *TimeValidityWindow[T] {
-	return &TimeValidityWindow[T]{
+	t := &TimeValidityWindow[T]{
```
aaronbuchwald, line 90: Do we ever need to pass in and handle nil in our production code? It seems like we do not, so I'd prefer to avoid introducing to the interface the need to handle (and documenting the handling of) a nil last tip.

aaronbuchwald (changes requested):

vm/vm.go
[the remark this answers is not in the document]
Elvis339: Return error instead of panic, seems logging is not needed on this level. 

https://github.com/ava-labs/hypersdk/pull/2012/commits/580cb648471e7c726b6fa28cc924cd5c5511e996

vm/vm.go
Elvis339: https://github.com/ava-labs/hypersdk/blob/72c2b907196bb458026aa892fda63993b918de8a/vm/vm.go#L306
Has been moved to accommodate validity window's API: https://github.com/ava-labs/hypersdk/blob/72c2b907196bb458026aa892fda63993b918de8a/vm/vm.go#L311

chainindex/chain_index.go
```diff
@@ -151,6 +143,18 @@ func (c *ChainIndex[T]) UpdateLastAccepted(ctx context.Context, blk T) error {
...
+// which will delete expired blocks
+func (c *ChainIndex[T]) SaveHistorical(blk T) error {
+	batch := c.db.NewBatch()
+	if err := c.writeBlock(batch, blk); err != nil {
+		return err
+	}
+
+	return batch.Write()
+}
+
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/2951d4ee15aa3cf224d13c2855f86d781454d224

internal/validitywindow/validitywindow.go
aaronbuchwald: I'm more concerned about the interface clarity and minimizing what can go wrong. I also think that given we never use nil in our production code, writing our tests with nil passed in allows us to keep the existing tests, but is less representative of the actual bugs we can hit. I think we'd be better off requiring that it's non-nil and updating our tests.

chainindex/chain_index.go
```diff
@@ -183,6 +187,95 @@ func (c *ChainIndex[T]) GetBlockByHeight(ctx context.Context, blkHeight uint64)
...
+		if height%c.config.BlockCompactionFrequency == c.compactionOffset {
+			go func() {
+				start := time.Now()
+				if err := c.db.Compact([]byte{blockPrefix}, prefixBlockKey(height)); err != nil {
+					c.log.Error("failed to compact block store", zap.Error(err))
+					return
+				}
+				c.log.Info("compacted disk blocks", zap.Uint64("end", height), zap.Duration("t", time.Since(start)))
+			}()
+		}
```
aaronbuchwald, line 269: Rather than start goroutines to compact, could we compact once at the end of the iteration?

vm/vm.go
aaronbuchwald: I see, I'm actually not sure that we need to set `lastAccepted` to nil here. I think we may only need to make sure to return `stateReady` correctly. Will double check and add another comment.

aaronbuchwald: The return values are only used here: https://github.com/ava-labs/hypersdk/blob/f3700b7a67bcb5885dfc950d2bddc6b36be9c704/snow/vm.go#L302

and the last accepted value is only used if `stateReady` is set to true: https://github.com/ava-labs/hypersdk/blob/348491718dd944a13ed8a21e303a3a34d9173adc/snow/chain_index.go#L64

However, this is not a clearly documented part of the interface at the moment. Definitely reasonable to assume we should set it to nil, a further improvement imo would be to make sure this behavior is documented, but don't think that's necessary.

chainindex/chain_index.go
```diff
@@ -101,19 +105,9 @@ func (c *ChainIndex[T]) GetLastAcceptedHeight(_ context.Context) (uint64, error)
...
+	height := blk.GetHeight()
 	heightBytes := binary.BigEndian.AppendUint64(nil, height)
-	err := errors.Join(
-		batch.Put(lastAcceptedKey, heightBytes),
-		batch.Put(prefixBlockIDHeightKey(blkID), heightBytes),
-		batch.Put(prefixBlockHeightIDKey(height), blkID[:]),
-		batch.Put(prefixBlockKey(height), blkBytes),
-	)
-	if err != nil {
+	if err := errors.Join(batch.Put(lastAcceptedKey, heightBytes), c.writeBlock(batch, blk)); err != nil {
```
aaronbuchwald, line 110: nit: can we re-format this to putting each statement that returns an error on a separate line? I find this format to be much easier to read

chainindex/chain_index.go
```diff
@@ -79,15 +80,18 @@ func New[T Block](
...
 		// Offset by random number to ensure the network does not compact simultaneously
 		compactionOffset: rand.Uint64() % config.BlockCompactionFrequency, //nolint:gosec
 		metrics:          metrics,
 		log:              log,
 		db:               db,
 		parser:           parser,
-	}, nil
+	}
+
+	ctx := context.Background()
```
aaronbuchwald, line 93: If we need a context, can we add the parameter to the constructor?

vm/vm.go
```diff
@@ -544,8 +554,22 @@ func (vm *VM) startNormalOp(ctx context.Context) error {
...
+		return err
+	}
+	executionBlk, err := vm.GetExecutionBlock(ctx, lastAccepted.GetID())
+	if err != nil {
+		return err
+	}
+
+	isValidityWindowComplete := vm.chainTimeValidityWindow.Complete(ctx, executionBlk)
+	if !isValidityWindowComplete {
+		return errors.New("critical error: validity window's partial state may lead to inconsistencies")
```
aaronbuchwald, line 569:
```suggestion
		return errors.New("cannot start normal operation with incomplete validity window")
```

can we declare this error at the top of the file in a var block to match the existing style?

chainindex/chain_index.go
Elvis339: Done: https://github.com/ava-labs/hypersdk/blob/a5f691b1d1ee5873a7ec293389d89e24765ab1bc/chainindex/chain_index.go#L110

chainindex/chain_index.go
Elvis339: Done: https://github.com/ava-labs/hypersdk/blob/a5f691b1d1ee5873a7ec293389d89e24765ab1bc/chainindex/chain_index.go#L69

vm/vm.go
Elvis339: Done: https://github.com/ava-labs/hypersdk/blob/94538b1c8919e9bde96ef52ae888b3c42d0d22c2/vm/vm.go#L66

https://github.com/ava-labs/hypersdk/blob/94538b1c8919e9bde96ef52ae888b3c42d0d22c2/vm/vm.go#L576

chainindex/chain_index.go
Elvis339: Done: https://github.com/ava-labs/hypersdk/blob/a5f691b1d1ee5873a7ec293389d89e24765ab1bc/chainindex/chain_index.go#L277

internal/validitywindow/validitywindow.go
```diff
@@ -53,36 +55,66 @@ type Interface[T emap.Item] interface {
...
+// It maintains a record of transactions that have been seen within a
+// configurable time window and rejects any duplicates.
+//
+// This component is critical as it:
+//  1. Prevents transaction replay attacks
+//  2. Enforces double-spend protection
+//  3. Provides temporal validation boundaries
+//  4. Maintains consensus safety across nodes i.e.;
+//     if different nodes had different rules for transaction uniqueness,
+//     they would disagree about the state of the blockchain.
```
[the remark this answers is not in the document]
aaronbuchwald:
```suggestion
// Maintains a window of all transactions included onchain, which have an expiry time that would
// allow them to be included a second time. Once their expiry time passes, transactions can be pruned
// from the window, since their expiry will mark them as invalid.
```

RodrigoVillar (changes requested):

internal/validitywindow/validitywindow.go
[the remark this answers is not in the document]
Elvis339: I updated the comment 

https://github.com/ava-labs/hypersdk/pull/2012/commits/fcf1b4f9325c555ecf52785275786e18794ec3eb

internal/validitywindow/validitywindow.go
```diff
@@ -53,36 +55,66 @@ type Interface[T emap.Item] interface {
...
+//
+// TimeValidityWindow builds up on assumption of ChainIndex being up to date
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/fcf1b4f9325c555ecf52785275786e18794ec3eb

internal/validitywindow/validitywindow.go
```diff
@@ -53,36 +55,66 @@ type Interface[T emap.Item] interface {
 	IsRepeat(ctx context.Context, parentBlk ExecutionBlock[T], currentTimestamp int64, containers []T) (set.Bits, error)
 }
 
+// TimeValidityWindow is a time-based transaction mechanism that prevents replay attacks.
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/fcf1b4f9325c555ecf52785275786e18794ec3eb

chain/chain.go
```diff
@@ -27,14 +27,14 @@ type Chain struct {
 func NewChain(
 	tracer trace.Tracer,
 	registerer *prometheus.Registry,
-	parser Parser,
```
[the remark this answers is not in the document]
Elvis339: I nade `BlockParser` an interface in one of my previous commits, here's one of the comments from Aaron regarding that: https://github.com/ava-labs/hypersdk/pull/2012#discussion_r2033684365 I tend to agree with him at this point in time.

internal/validitywindow/syncer.go
```diff
@@ -83,6 +90,14 @@ func (s *Syncer[T, B]) Start(ctx context.Context, target B) error {
 	go func() {
 		resultChan := s.blockFetcherClient.FetchBlocks(syncCtx, s.oldestBlock, &s.minTimestamp)
 		for blk := range resultChan {
+			if err := s.blockStore.SaveHistorical(blk); err != nil {
+				s.errChan <- fmt.Errorf(
+					"%w; aborting to prevent inconsistencies %w",
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/248dac97e8a9f0e1ecb9d67b86fd3e06949ad6dc

internal/validitywindow/syncer_test.go
```diff
@@ -18,209 +19,245 @@ type testCases struct {
...
-
-					// the last accepted height should be the last accepted height from the cache, since historical blocks should not update the last accepted field
-					req.Equal(blkChain[len(blkChain)-1].GetHeight(), syncer.timeValidityWindow.lastAcceptedBlockHeight)
-					req.Equal(blkChain[15].GetHeight(), syncer.oldestBlock.GetHeight())
-
-					// verify the oldest allowed block in time validity window
-					req.Equal(blkChain[4].GetTimestamp(), syncer.timeValidityWindow.calculateOldestAllowed(target.GetTimestamp()))
-					req.NotEqual(blkChain[3].GetTimestamp(), syncer.timeValidityWindow.calculateOldestAllowed(target.GetTimestamp()))
-				},
+			verifyFunc: func(ctx context.Context, req *require.Assertions, blkChain []ExecutionBlock[container], syncer *Syncer[container, ExecutionBlock[container]], _ *testChainIndex) {
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/248dac97e8a9f0e1ecb9d67b86fd3e06949ad6dc

chainindex/chain_index_test.go
```diff
@@ -128,3 +129,174 @@ func TestChainIndexExpiry(t *testing.T) {
...
+	// Verify lastAccepted points to the new block
+	confirmLastAcceptedHeight(r, ctx, chainIndex, blk1.GetHeight())
+}
+
+func TestChainIndex_Cleanup(t *testing.T) {
+	tests := []struct {
+		name                   string
+		config                 Config
+		setupFunc              func(ctx context.Context, r *require.Assertions, chainIndex *ChainIndex[*testBlock])
+		verifyAfterCleanupFunc func(ctx context.Context, r *require.Assertions, chainIndex *ChainIndex[*testBlock])
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/248dac97e8a9f0e1ecb9d67b86fd3e06949ad6dc

chainindex/chain_index.go
```diff
@@ -79,15 +81,17 @@ func New[T Block](
...
 		// Offset by random number to ensure the network does not compact simultaneously
 		compactionOffset: rand.Uint64() % config.BlockCompactionFrequency, //nolint:gosec
 		metrics:          metrics,
 		log:              log,
 		db:               db,
 		parser:           parser,
-	}, nil
+	}
+
+	return ci, ci.cleanupOnStartup(ctx)
```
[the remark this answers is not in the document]
Elvis339: https://github.com/ava-labs/hypersdk/pull/2012/commits/248dac97e8a9f0e1ecb9d67b86fd3e06949ad6dc

chain/chain.go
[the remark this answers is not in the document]
RodrigoVillar: Sounds good; going to resolve this then

RodrigoVillar (approved):
LGTM

chain/chain.go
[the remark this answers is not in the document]
aaronbuchwald: Can we keep this as a concrete type since we already define one to use in the chain package and don't have a need to pass a different definition in via an interface?

internal/validitywindow/validitywindow.go
```diff
@@ -53,36 +55,66 @@ type Interface[T emap.Item] interface {
...
 	chainIndex ChainIndex[T],
+	head ExecutionBlock[T],
 	getTimeValidityWindowF GetTimeValidityWindowFunc,
-) *TimeValidityWindow[T] {
-	return &TimeValidityWindow[T]{
+) (*TimeValidityWindow[T], error) {
+	if head == nil {
+		return nil, fmt.Errorf("cannot construct time validity window: %w", ErrNilInitialBlock)
+	}
+
```
aaronbuchwald, line 94: Can we remove this check in favor of requiring that the caller passes in a fully populated value?

The nil check won't work reliably for generic types and we commonly make this assumption rather than include defensive checks, so I'd lean towards removing it.

Elvis339: I agree with removing the nil check - it makes sense to follow the common practice of assuming callers pass valid values rather than adding defensive checks.

My original thinking was that having an explicit check with a clear error message would make debugging easier if we ever had an inconsistency. When something goes wrong in a system like this, having specific error messages that point directly to the source of the problem (like explicitly calling out a nil head block) can save time during troubleshooting.

But you're right that for generic types, nil checks aren't reliable, and it's better to be consistent with our approach of requiring callers to pass fully populated values. I'll remove this check.

aaronbuchwald (approved):

RodrigoVillar (approved):

PR #2013, opened 2025-03-28
[chain] Use State Key Sets Instead of `TxListGenerator`

### LINKED ISSUE #2009: Refactor `BlockBenchmarkHelper`s To Avoid Duplicate Code
Ref: https://github.com/ava-labs/hypersdk/pull/1955#discussion_r2014964151

RodrigoVillar:
Currently, the `TxListGenerator` is responsible for generating a list of transactions which mainly differ only in their state accesses. As a result, implementations of the `TxListGenerator` (and the `BlockBenchmarkHelper`) contain mostly duplicate code. 

This PR removes the `BlockBenchmarkHelper`/`TxListGenerator` into the following sub-components:

- `GenesisGenerator`: responsible for generating the chain genesis, the set of accounts to use for the benchmarks, and a set of keys that will be used for the state access distribution.
- `ActionConstructor`: given a key and a nonce, this function will return an action (in HyperSDK, this would be a `TestAction` which for MorpheusVM, this would be `Transfer`).
- `StateAccessDistributor`: responsible for generating a transaction list where the state access of the txs vary based on the implementation. For example, in the case of parallel state accesses, each TX would touch a different part of state (and therefore, can be parallelized).

This PR reduces the amount of duplicate code shared between the MorpheusVM and HyperSDK block benchmark tests as now, each test suite just needs to implement the `GenesisGenerator` and the `ActionConstructor` while `chaintest` provides implementations for `StateAccessDistributor`.

For example, consider parallel txs: here's what MorpheusVM/the HyperSDK would each need to provide:
- MorpheusVM: 
  - A `GenesisGenerator` which returns a set of unique addresses
  - An `ActionConstructor` which takes in an address and returns the `Transfer` action
- HyperSDK:
  - A `genesisGenerator` which returns a set of unique database keys
  - An `ActionConstructor` which takes in a database key `x` and returns a `TestAction` that writes at `x`.

Closes #2009.


chain/chaintest/block.go
```diff
@@ -372,3 +376,165 @@ func (test *BlockBenchmark) Run(ctx context.Context, b *testing.B) {
...
+// allows for the transactions to be executed in parallel.
+func ParallelDistribution[T any](
+	numTxs int,
+	factories []chain.AuthFactory,
+	keys []T,
+	ac ActionConstructor[T],
+	ruleFactory chain.RuleFactory,
+	unitPrices fees.Dimensions,
+	timestamp int64,
+	nonce uint64,
```
tsachiherman, line 430: this nonce here is used as a base nonce; maybe name it `baseNonce` ?

chain/chaintest/block.go
```diff
@@ -372,3 +376,165 @@ func (test *BlockBenchmark) Run(ctx context.Context, b *testing.B) {
...
+) ([]*chain.Transaction, error) {
+	if numTxs != len(keys) || numTxs != len(factories) {
+		return nil, ErrMismatchedKeysAndFactoriesLen
+	}
+
+	txs := make([]*chain.Transaction, numTxs)
+	for i := range numTxs {
+		action := ac(keys[i], nonce)
+
+		nonce++
```
tsachiherman, line 440: I think that it would be cleaner to write
```golang
nonce := baseNonce + i
action := ac(keys[i], nonce)
```

chain/chaintest/genesis.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/ava-labs/hypersdk/crypto/ed25519"
+	"github.com/ava-labs/hypersdk/genesis"
+)
+
+func CreateGenesis(
+	numAccounts uint64,
+	allocAmount uint64,
+	factory func() (chain.AuthFactory, error),
+) ([]chain.AuthFactory, genesis.Genesis, error) {
+	factories := make([]chain.AuthFactory, numAccounts)
```
Elvis339, line 18: Lines 16, 18 and 21 are shadows declaration, let's consider changing them so we avoid potential shadowing and hard to trace bugs that may arise.

chain/chaintest/block.go
```diff
@@ -372,3 +376,165 @@ func (test *BlockBenchmark) Run(ctx context.Context, b *testing.B) {
...
+	factories []chain.AuthFactory,
+	keys []T,
+	ac ActionConstructor[T],
+	ruleFactory chain.RuleFactory,
+	unitPrices fees.Dimensions,
+	timestamp int64,
+	nonce uint64,
+) ([]*chain.Transaction, error) {
+	if numTxs != len(keys) || numTxs != len(factories) {
+		return nil, ErrMismatchedKeysAndFactoriesLen
```
tsachiherman, line 433: could you add a unit test that mimics this case and test this value ?

chain/chaintest/block.go
```diff
@@ -372,3 +376,165 @@ func (test *BlockBenchmark) Run(ctx context.Context, b *testing.B) {
...
+	numTxs int,
+	factories []chain.AuthFactory,
+	keys []T,
+	ac ActionConstructor[T],
+	ruleFactory chain.RuleFactory,
+	unitPrices fees.Dimensions,
+	timestamp int64,
+	nonce uint64,
+) ([]*chain.Transaction, error) {
+	if len(keys) != 1 {
```
tsachiherman, line 471: I'm not sure that checking that the len(keys) != 1 is needed.
I can't see any harm if the caller provided 2 keys. On the other hand, this code would crash if len(keys) == 0, which isn't covered.

tsachiherman (commented):
overall, looks good. I've added few comments.

Elvis339 (commented):
I noticed we're using the function types like `ActionConstructor[T any]`. Have you considered using an interface instead?

Something like:
```go
type ActionConstructor[T any] interface {
    Generate(key T, nonce uint64) chain.Action
}
``` 
Potential advantages:
- Better IDE support for discovering functionality
- More explicit and easier to understand what it does
- Clearer semantics with the named method

Downsides:
- Requires a bit more boilerplate code

It could help break down the generators with this approach:
We'd have one for string as I see that being used as well. It does add a bit more boilerplate but GoLand's support for discoverability makes it much more cognitive ease to work with.
```go
type actionGenerator[T codec.Address] struct {
}

// goland discovers actionGenerator T implements ActionConstructor
func (*actionGenerator[T]) Generate(key codec.Address, nonce uint64) chain.Action {
    return &actions.Transfer{
       To:    key,
       Value: 1,
       Memo:  binary.BigEndian.AppendUint64(nil, nonce),
    }
}
```
This pattern would let us have specialized generators for different types while maintaining a consistent interface.
This isn't a blocker for me, just something to consider. Feel free to keep it as is if you prefer the function approach.

chain/chaintest/genesis.go
RodrigoVillar: Renamed `factory` to `factoryF`: 61afe13

chain/chaintest/block.go
RodrigoVillar: Ideally, `keys` would be a single value but unfortunately, generics cannot be both a single value and an array/slice (even if the underlying types are the same). The implicit invariant here is that there's a single state access key used for all the txs (as this will force serial execution) and so enforcing that `len(keys) == 1` does that.

RodrigoVillar:
> I noticed we're using the function types like `ActionConstructor[T any]`. Have you considered using an interface instead?
> 
> Something like:
> 
> ```go
> type ActionConstructor[T any] interface {
>     Generate(key T, nonce uint64) chain.Action
> }
> ```
> 
> Potential advantages:
> 
> * Better IDE support for discovering functionality
> * More explicit and easier to understand what it does
> * Clearer semantics with the named method
> 
> Downsides:
> 
> * Requires a bit more boilerplate code
> 
> It could help break down the generators with this approach: We'd have one for string as I see that being used as well. It does add a bit more boilerplate but GoLand's support for discoverability makes it much more cognitive ease to work with.
> 
> ```go
> type actionGenerator[T codec.Address] struct {
> }
> 
> // goland discovers actionGenerator T implements ActionConstructor
> func (*actionGenerator[T]) Generate(key codec.Address, nonce uint64) chain.Action {
>     return &actions.Transfer{
>        To:    key,
>        Value: 1,
>        Memo:  binary.BigEndian.AppendUint64(nil, nonce),
>     }
> }
> ```
> 
> This pattern would let us have specialized generators for different types while maintaining a consistent interface. This isn't a blocker for me, just something to consider. Feel free to keep it as is if you prefer the function approach.

This is a great point. I think especially for the `ActionConstructor` and the `StateAccessDistributor`, there is a benefit to having these types be interfaces, especially since we can then store items like the `RuleFactory` and the `nonce` in them.  Going to take a stab at rewriting these into interfaces before committing to a switch.

chain/chaintest/block.go
RodrigoVillar: https://github.com/ava-labs/hypersdk/pull/2013#discussion_r2047025489

chain/chaintest/block.go
RodrigoVillar: 81776cc9f

Elvis339 (approved):
Approving on behalf of my comments. LGTM

aaronbuchwald (approved):

PR #2018, opened 2025-04-08
[chain] Apply `exported` lint rule

RodrigoVillar:
This PR applies the `exported` lint rule to the `chain` package, which forces documentation on exported functions and interfaces. 

Long term, we will want to apply `exported` to all public facing packages, but this is a good starting point.


.golangci.yml
```diff
@@ -190,6 +190,42 @@ linters-settings:
...
+          - "**/snow/**/*.go"
+          - "**/state/**/*.go"
+          - "**/statesync/**/*.go"
+          - "**/storage/**/*.go"
+          - "**/tests/**/*.go"
+          - "**/throughput/**/*.go"
+          - "**/utils/**/*.go"
+          - "**/vm/**/*.go"
+          - "**/x/**/*.go"
+          - "TEST"
```
RodrigoVillar, line 228: Normally, a regex that includes all packages except `chain` would've been better here. However, since we plan on applying the `exported` rule to all public-facing packages, it might be better to leave it in this format.

.golangci.yml
```diff
@@ -190,6 +190,42 @@ linters-settings:
       # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#useless-break
       - name: useless-break
         disabled: false
+      - name: exported
```
Elvis339, line 193:
```suggestion
# https://github.com/mgechev/revive/blob/HEAD/RULES_DESCRIPTIONS.md#exported
      - name: exported
```

I'd add link to the rule for reference.

chain/dependencies.go
```diff
@@ -17,43 +17,64 @@ import (
...
 	ParseAction([]byte) (Action, error)
+	// Parse marshaled auth
 	ParseAuth([]byte) (Auth, error)
 }
 
 type Mempool interface {
-	Len(context.Context) int  // items
-	Size(context.Context) int // bytes
+	// Number of items
+	Len(context.Context) int
```
Elvis339, line 28: Linter doesn't seem to apply `exported` rule correctly.

> The Go documentation guidelines specify that documentation comments for exported identifiers should start with the name being declared. This helps with generated documentation, maintains a consistent style, and makes it clear what's being described, especially when multiple declarations are being documented in sequence.

```go
// This lacks a comment
func HandleRequest(w http.ResponseWriter, r *http.Request) {
    // ...
}

// Processes incoming HTTP requests
func HandleRequest(w http.ResponseWriter, r *http.Request) {
    // ...
}
```

Both of these functions should be targeted.
GoLand IDE is reporting invalid comment pattern as well.

Screenshot 2025-04-09 at 6 27 37 PM

Elvis339 (commented):
I'm not sure why isn't `exported` rule picking up on this... Will play a bit and see what I find out.

RodrigoVillar:
Putting this back into draft; realized that `revive` isn't being enforced in its entirety by our linter due to [exclude-use-default](https://golangci-lint.run/usage/false-positives/#exclusion-presets)

.golangci.yml
```diff
@@ -16,6 +16,37 @@ run:
...
+    stutters; consider calling this'
+    - '(possible misuse of unsafe.Pointer|should have signature)'
+    - 'SA4011'
+    - 'G103: Use of unsafe calls should be audited'
+    - 'G204: Subprocess launched with variable'
+    - 'G104'
+    - '(G301|G302|G307): Expect (directory permissions to be 0750|file
+    permissions to be 0600) or less'
+    - 'G304: Potential file inclusion via variable'
+    - '(ST1000|ST1020|ST1021|ST1022)'
```
RodrigoVillar, line 42: @Elvis339 these are the list of regex exclusions that were previously enforced by `exclude-use-defaults`, but now have to be manually excluded.

To see the full list of default regex exclusions (including the `revive` ones we're now enforcing), you can query via `golangci-lint run --help` and scroll to the `--exclude-use-default` flag

RodrigoVillar:
https://github.com/ava-labs/hypersdk/actions/runs/14362530871/job/40267461117?pr=2018#step:4:1201 ❤️

RodrigoVillar:
Closing this in favor in a future PR which targets a smaller package. The goal of this PR was to commit to eventually adding documentation for all PRs (via `revive`/`staticcheck`) but its clear that choosing the `chain` package was not ideal due to its size.

A future PR should target the `context` package, since its much smaller in scope.

Issue #1993, opened 2025-03-20
Possible Sync E2E Test Flake


aaronbuchwald:
Seeing a bug that I first saw while reviewing https://github.com/ava-labs/hypersdk/pull/1911 in CI on what appears to be an unrelated change. From the downloaded artifact https://github.com/ava-labs/hypersdk/actions/runs/13978775473/job/39139034651?pr=1992#step:4:3009, I see the following log in `./networks/20250320-203333.453062-morpheusvm-e2e-tests/NodeID-4WAF6eR6n3cbZDfnPqaeKzGQ1693ys4KQ/logs/2A4cQ7ZxBBWU4Kdso1SxMLLQHi4mnnEecienFA7nUCTPiJYbSr.log`:

```
[03-20|20:48:29.533] DEBUG <2A4cQ7ZxBBWU4Kdso1SxMLLQHi4mnnEecienFA7nUCTPiJYbSr Chain> proposervm/block.go:158 verified post-fork block {"blkID": "22rRgDAjSJ2jxoQNCYRgETUp2iffXyCkgw2xNjnf6f9gBoAJZH", "parentTimestamp": "[03-20|20:35:24.000]", "blockTimestamp": "[03-20|20:35:25.000]"}
[03-20|20:48:29.533] INFO <2A4cQ7ZxBBWU4Kdso1SxMLLQHi4mnnEecienFA7nUCTPiJYbSr Chain> snow/block.go:200 Verifying block {"block": "((BlockID=2cywQQqxBLfw55bdAi2pxbUzndJS6pePDaEr31Ub7WFPsSSraR, Height=215, ParentRoot=R2YLg1ryD7su7QDZUjyNLefrohdCNSA83cr1wohAZ2ynqgze4, NumTxs=0, Size=90), verified = false, accepted = false)"}
[03-20|20:48:29.533] INFO <2A4cQ7ZxBBWU4Kdso1SxMLLQHi4mnnEecienFA7nUCTPiJYbSr Chain> snow/block.go:214 parent was already verified
[...]

PR #2022, opened 2025-04-10
Update avalanchego to latest commit

aaronbuchwald:
Updates AvalancheGo dependency to the latest commit on master branch: https://github.com/ava-labs/avalanchego/commit/c8b3f57d2a2556d41ca6a0b2da11561b1d7bab75

The primary reason for bumping this now is to ensure that we are using a commit from master that incorporates https://github.com/ava-labs/avalanchego/pull/3831, which fixes a bug in HyperSDK e2e tests due to an interesting case of how the ProposerVM manages state sync summaries explained here https://github.com/ava-labs/hypersdk/issues/1993.


scripts/dev_shell.sh
```diff
@@ -16,4 +16,4 @@ source "$SCRIPT_DIR"/constants.sh
 # Start a dev shell with the avalanchego flake
 FLAKE="github:ava-labs/avalanchego?ref=${AVALANCHE_VERSION}"
 echo "Starting nix shell for ${FLAKE}"
-nix develop "${FLAKE}" "${@}"
+nix develop "${FLAKE}"
```
aaronbuchwald, line 19:
```suggestion
nix develop "${FLAKE}" "${@}"
```

Optional to revert this change. This is needed to start the dev shell on my mac because MacOS treats "${@}" as an unset variable when no arguments are passed in, whereas other platforms treat it as an empty string. Since we use `set -euo pipefail` in all of our bash scripts, this causes the script to fail.

tests/fixture/subnet.go
```diff
@@ -24,9 +25,9 @@ func NewHyperVMSubnet(name string, vmID ids.ID, genesisBytes []byte, nodes ...*t
 				}`,
 			},
 		},
-		Config: tmpnet.FlagsMap{
-			"proposerMinBlockDelay":       0,
-			"proposerNumHistoricalBlocks": 50_000,
+		Config: &subnets.Config{
+			ProposerMinBlockDelay:       0,
+			ProposerNumHistoricalBlocks: 50_000,
 		},
```
aaronbuchwald, line 31: Changed in https://github.com/ava-labs/avalanchego/pull/3857

.github/workflows/hypersdk-ci.yml
```diff
@@ -152,7 +152,7 @@ jobs:
             go.sum
             examples/morpheusvm/go.sum
       - name: Run e2e tests
-        uses: ava-labs/avalanchego/.github/actions/run-monitored-tmpnet-cmd@50f1601bf39a106d066012b4ef470c1058f5005a
+        uses: ava-labs/avalanchego/.github/actions/run-monitored-tmpnet-cmd@c8b3f57d2a2556d41ca6a0b2da11561b1d7bab75
```
RodrigoVillar, line 155: q: why are we using commit `c8b3f57d2a2556d41ca6a0b2da11561b1d7bab75` even though the PR description says we're bumping to `5ea90b6f2484112cec7949a63feaa186c8d8f568`?

aaronbuchwald: Had to bump it again and didn't update the PR description. Nice catch, will update.

PR #2026, opened 2025-04-14
[ci] Update `golangci-lint` to `v2.0.2`

RodrigoVillar:


.golangci.yml
```diff
@@ -71,151 +47,182 @@ linters:
...
+    spancheck:
+      checks:
+        - end
+    staticcheck:
+      # https://staticcheck.io/docs/options#checks
+      checks:
+        - all
+        - -SA1019 # Using a deprecated function, variable, constant or field
+        - -SA6002 # Storing non-pointer values in sync.Pool allocates memory
+        - -QF1008
```
aaronbuchwald, line 178: Could we add a comment for this check as well?

aaronbuchwald: Same comment in the other

aaronbuchwald (changes requested):

PR #2029, opened 2025-04-15
[e2e] Add Dedicated Load Factory

RodrigoVillar:
This PR updates the `e2e` tests relating to `load` by adding a dedicated `loadFactory` which the `load` tests will use for funds. This improves the current `e2e` framework as, in the case of a `load` test failing, we have that any other tests using a different account will not fail due to fund exhaustion errors originating from the failed `load` test.


examples/morpheusvm/tests/workload/genesis.go
```diff
@@ -28,8 +28,12 @@ const (
 
 // hardcoded initial set of ed25519 keys. Each will be initialized with InitialBalance
 var ed25519HexKeys = []string{
+	// integration test key
 	"323b1d8f4eed5f0da9da93071b034f2dce9d2d22692c172f3cb252a64ddfafd01b057de320297c29ad0c1f589ea216869cf1938d88c9fbd70d6748323dbf2fa7", //nolint:lll
+	// txGenerator keys
 	"8a7be2e0c9a2d09ac2861c34326d6fe5a461d920ba9c2b345ae28e603d517df148735063f8d5d8ba79ea4668358943e5c80bc09e9b2b9a15b5b15db6c1862e88", //nolint:lll
+	// load generation key
```
aaronbuchwald, line 35: These comments are a huge help 🙏

tests/e2e/e2e.go
```diff
@@ -257,7 +260,7 @@ var _ = ginkgo.Describe("[HyperSDK Load Workloads]", ginkgo.Ordered, ginkgo.Seri
 		blockchainID := e2e.GetEnv(tc).GetNetwork().GetSubnet(networkConfig.Name()).Chains[0].ChainID
 		uris := getE2EURIs(tc, blockchainID)
 
-		accounts, err := distributeFunds(ctx, tc, createTransferF, networkConfig.AuthFactories()[0], uint64(len(uris)))
```
aaronbuchwald, line 260: Would it be simpler if we use `networkConfig.AuthFactories()` throughout and choose a separate index for each one? We could also add a check at the start that the config includes a sufficient number of keys for the tests.

RodrigoVillar: If we were to enforce indices, I think this would lead to inconsistent code since in the case of `keys = 3`, we would be requiring 3 keys while only using one of them (the first two keys are managed by the VM and are never accessed by us). 

As it currently stands, we're asking the client to pass in a load key and we access that key directly; we are passed in the other two keys via `networkConfig` but we never use those/the client maintains those. From an API perspective, I think this is better since we are asking less from the client (i.e. "please give us a key that we will use" vs. "please give us three keys, but we'll only use one").

PR #2036, opened 2025-04-25
chore(load): minor changes to API and implementation

qdm12:
(I would suggest reviewing commit by commit)

A few minor changes brought whilst reading the load code


load/dependencies.go
```diff
@@ -9,13 +9,14 @@ import (
...
 	// GenerateTx returns a valid transaction.
-	GenerateTx(ctx context.Context) (T, error)
+	GenerateTx() (T, error)
 }
 
 type Issuer[T comparable] interface {
 	// Listen for the final status of transactions and notify the tracker
 	// Listen stops if the context is done, an error occurs, or if the issuer
 	// has sent all their transactions.
+	// Listen MUST return a nil error if the context is canceled.
```
qdm12, line 19: Shouldn't this be the opposite?? If the context is canceled, we usually reached the final timeout without collecting all the transactions back, that seems like we should error?

load/dependencies.go
```diff
@@ -9,22 +9,23 @@ import (
...
 	Listen(ctx context.Context) error
 
-	// Stop notifies the issuer that no further transactions will be issued.
-	// If a transaction is issued after Stop has been called, the issuer should error.
-	Stop()
-
-	// Issue sends a tx to the network, and informs the tracker that its sent
-	// said transaction.
-	IssueTx(ctx context.Context, tx T) error
+	RegisterIssued(T)
```
qdm12, line 28: This Stop -> RegisterIssued change is to track "in flight" txs in order to mark them as failed if the listener exits before receiving all remaining transactions

load/gradual_load_orchestrator_test.go
```diff
@@ -91,80 +98,53 @@ func TestGradualLoadOrchestratorTPS(t *testing.T) {
 	}
 }
 
-func TestGradualLoadOrchestratorConstructor(t *testing.T) {
```
qdm12, line 94: Test seems unneeded now that we use agents forcing the same number of issuers, listeners and trackers

load/dependencies.go
```diff
@@ -9,22 +9,23 @@ import (
 
 type TxGenerator[T comparable] interface {
 	// GenerateTx returns a valid transaction.
-	GenerateTx(ctx context.Context) (T, error)
+	GenerateTx() (T, error)
```
michaelkaplan13, line 12: What's the motivation for removing the context here? Generating a transaction could likely depend on making some number of RPC requests, so it'd be good to have a context available to use in that case.

load/dependencies.go
michaelkaplan13: Does "failed" in this context mean "not included in an accepted block"?

For general purpose VMs, we'll want to easily distinguish between that case, and the case of transaction reverting (due to being out of gas, hitting a revert statement, etc).

load/dependencies.go
qdm12: This is changing to be a `GenerateAndIssueTx(context.Context)` anyway, so no worry on this, a context will be available.

Now the previous rationale was just that generating txs should be quick, and we don't need any IO (here or in geth-based-evms), so there was no need for a context. Something we could change further down the line though, of course.

load/gradual_load_orchestrator_test.go
RodrigoVillar: Yup I agree

load/dependencies.go
RodrigoVillar: > Does "failed" in this context mean "not included in an accepted block"?

Yes

> For general purpose VMs, we'll want to easily distinguish between that case, and the case of transaction reverting (due to being out of gas, hitting a revert statement, etc).

Yup, assuming the EVM-listener can access the type of failure status of a tx, we can make this distinction for any dependency (e.g. metrics tracker).

load/listener.go
```diff
@@ -0,0 +1,93 @@
...
+
+	lock          sync.RWMutex
+	issuedTxs     uint64
+	receivedTxs   uint64
+	inFlightTxIDs map[ids.ID]struct{}
+}
+
+// NewDefaultListener creates a new DefaultListener instance.
+// Set txTarget to 0 to never stop the [DefaultListener.Listen] method unless the context is cancelled.
+func NewDefaultListener(client *ws.WebSocketClient, tracker Tracker[ids.ID], txTarget uint64) *DefaultListener {
```
RodrigoVillar, line 31: Not sure how I feel about the `txTarget` parameter. One of the original motivations for having `Stop()` was that, w.r.t. the number of transactions an issuer would send, there was a single source of truth - the `TxsPerIssuer` field in the config. With this, there's now two sources of truth that the developer needs to account for.

After pondering about this for some time, I think this is **cleaner** than using `Stop()`, but wanted to hear your thoughts regarding the above.

load/short_burst_orchestrator.go
```diff
@@ -25,30 +20,18 @@ type ShortBurstOrchestratorConfig struct {
 // ShortBurstOrchestrator tests the network by sending a fixed number of
 // transactions en masse in a short timeframe.
 type ShortBurstOrchestrator[T, U comparable] struct {
-	generators []TxGenerator[T]
-	issuers    []Issuer[T]
-	tracker    Tracker[U]
-
-	issuerGroup   errgroup.Group
-	observerGroup errgroup.Group
```
RodrigoVillar, line 33: Great call localizing the `errgroup`s

load/dependencies.go
michaelkaplan13: Yeah, for any interesting/complex transaction types in generic VMs, I think a context will be needed. 

For instance, when calling a smart contract, the only good way to estimate the gas needed is to use the `eth_estimateGas` RPC. I'd image in the future there could be similar cases in other VMs for things such as fetching UTXOs to spend, etc.

michaelkaplan13: I think we should probably leave the context in the interface for now to reduce the changes here, and potential changes needed down the line

load/dependencies.go
```diff
@@ -49,9 +50,9 @@ type Tracker[T comparable] interface {
 	GetObservedFailed() uint64
 }
 
-// Orchestrator executes the load test by coordinating the issuers to send
+// orchestrator executes the load test by coordinating the issuers to send
 // transactions, in a manner depending on the implementation.
-type Orchestrator interface {
+type orchestrator interface {
```
michaelkaplan13, line 55: I very well could be missing something here, but is this interface actually used anywhere other than for the compile time checks? An un-exported interface seems odd to me

tests/e2e/e2e.go
```diff
@@ -228,24 +228,29 @@ var _ = ginkgo.Describe("[HyperSDK Load Workloads]", ginkgo.Ordered, ginkgo.Seri
...
+		for i := range agents {
+			client, err := ws.NewWebSocketClient(
+				uris[i%len(uris)],
+				ws.DefaultHandshakeTimeout,
+				pubsub.MaxPendingMessages,
+				pubsub.MaxReadMessageSize,
+			)
+			require.NoError(err, "creating websocket client")
+			issuer := load.NewDefaultIssuer(client, tracker)
+			listener := load.NewDefaultListener(client, tracker, shortBurstConfig.TxsPerIssuer)
```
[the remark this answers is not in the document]
qdm12: That bite me originally, I did have 2 clients which drove me mad for a good 20 minutes until I realized the client had to be shared 😄

load/listener.go
```diff
@@ -0,0 +1,93 @@
...
+
+type DefaultListener struct {
+	client   *ws.WebSocketClient
+	tracker  Tracker[ids.ID]
+	txTarget uint64
+
+	lock          sync.RWMutex
+	issuedTxs     uint64
+	receivedTxs   uint64
+	inFlightTxIDs map[ids.ID]struct{}
```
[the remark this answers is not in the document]
qdm12: Hmmm perhaps, although

- native language/std library is more "reading portable"
- set.Set seems to have quite a few features, which we won't use, this feels a bit odd to use when only needing a `map[ids.ID]struct{}` 😄 

If `set.Set` is used quite a bit elsewhere in the codebase though, I'll use it to match the rest of the codebase though!

load/listener.go
qdm12: I don't totally like it to be fair 😄 
For example for the gradual orchestrator, `Stop()` isn't used, and here we set the `txTarget` to `0`, which kind of bothers me too!
Another way would be to have `RegisterIssued(tx, last bool)`. That way it would signal to the listener this is the last tx issued.

load/dependencies.go
qdm12: This is just for the compile time check that the 2 orchestrator implementations implement that interface.
When moving this to avalanchego, this can also stay unexported, given orchestrators will be defined locally there too.

What's wrong with unexported interfaces? We don't accept any argument with that interface, nor do we need to import it anywhere else

load/listener.go
qdm12: I did change it now to pass a `lastIssued` boolean to RegisterIssued, and it does make things simpler I think, what do you think?

load/listener.go
```diff
@@ -0,0 +1,93 @@
...
+
+		l.lock.Lock()
+		delete(l.inFlightTxIDs, txID)
+		l.receivedTxs++
+		if l.txTarget > 0 &&
+			l.issuedTxs == l.txTarget &&
+			l.issuedTxs == l.receivedTxs &&
+			len(l.inFlightTxIDs) == 0 {
+			l.lock.Unlock()
+			break
```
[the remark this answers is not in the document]
qdm12: I think this is now unneeded, given it's only 2 conditions now.

load/dependencies.go
michaelkaplan13: Nothing wrong with it, and I agree that having it unexported rather than exported if it's not needed anywhere.

I guess my question is more turning that around more generally: if an interface is not used anywhere (internal or external to the package it's defined in), then what's the purpose of having it at all?

I'm fine with keeping it as is either way though.

examples/morpheusvm/load/dependencies.go
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package load
+
+import "github.com/ava-labs/avalanchego/ids"
+
+// Tracker keeps track of the status of transactions.
+// This must be thread-safe, so it can be called in parallel by the issuer(s) or orchestrator.
+type Tracker interface {
```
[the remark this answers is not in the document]
qdm12: It was such that the examples/morpheusvm/load package does not depend on the load package, to reduce package dependencies. I also changed to be narrower to just

```go
type Tracker interface {
	Issue(ids.ID)
}
```

I can import it from load if you prefer

load/dependencies.go
qdm12: I'm fine with removing it, it's just more of a documentation for future orchestrators showing what minimal implementation an orchestrator should have.

examples/morpheusvm/load/dependencies.go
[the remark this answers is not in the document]
RodrigoVillar: I think we should have just a single `Tracker` interface which example VMs can rely on.

load/dependencies.go
```diff
@@ -47,11 +47,18 @@ type Tracker[T comparable] interface {
 	// GetObservedFailed returns the number of transactions that the tracker has
 	// confirmed failed.
 	GetObservedFailed() uint64
+
+	// String returns a string representation of the tracker state.
+	String() string
+}
+
+type Logger interface {
```
michaelkaplan13, line 55: Lets not redefine the Logger interface here, and use the standard one from github.com/ava-labs/avalanchego/utils/logging

Generally, lets not proliferate interface definitions across packages just to try to slim down a few functions. It can lead to confusion about where the true interface definition lives, what functions should be available for future code changes, and cause more code changes in the future if things change.

load/gradual_load_orchestrator.go
```diff
@@ -67,11 +66,9 @@ func DefaultGradualLoadOrchestratorConfig() GradualLoadOrchestratorConfig {
 // the network can no longer make progress (i.e. the rate at the network accepts
 // transactions is less than currTargetTPS).
 type GradualLoadOrchestrator[T, U comparable] struct {
-	generators []TxGenerator[T]
-	issuers    []Issuer[T]
-	tracker    Tracker[U]
+	agents []Agent[T, U]
 
-	log logging.Logger
+	log Logger
```
RodrigoVillar, line 71: Could we defer to using the AvalancheGo logger?

load/prometheus_tracker.go
```diff
@@ -16,7 +17,7 @@ const namespace = "load"
 var _ Tracker[any] = (*PrometheusTracker[any])(nil)
 
 type PrometheusTracker[T comparable] struct {
-	sync.RWMutex
+	mutex sync.RWMutex
```
RodrigoVillar, line 20: Could we be consistent with our naming of `sync.RWMutex` values (i.e. `mutex` => `lock`)?

load/short_burst_orchestrator.go
```diff
@@ -59,31 +45,38 @@ func (o *ShortBurstOrchestrator[T, U]) Execute(ctx context.Context) error {
...
-				if err := issuer.IssueTx(ctx, tx); err != nil {
-					return err
+				lastIssued := i == o.config.TxsPerIssuer-1
+				agent.Listener.RegisterIssued(tx, lastIssued)
+
+				select {
+				case <-logTicker.C:
+					o.log.Info(agent.Tracker.String())
+				default:
 				}
```
RodrigoVillar, line 72: I think we should remove this. If there's multiple agents, this will flood the user's console and will therefore not be of much use.

tests/e2e/e2e.go
```diff
@@ -63,14 +63,16 @@ var (
 	ErrTxFailed          = errors.New("transaction failed")
 )
 
-// LoadTxGenerator returns the components necessary to instantiate an
+// LoadIssuers returns the components necessary to instantiate an
 // Orchestrator.
-// We use a generator here since the node URIs are not known until runtime.
-type LoadTxGenerator func(
+// We use an issuer here since the node URIs are not known until runtime.
```
RodrigoVillar, line 68: Could we update the documentation here to make it clear that LoadIssuers is a function/generator/initializer, rather than a `Issuer` type?

load/prometheus_tracker.go
```diff
@@ -31,7 +32,12 @@ type PrometheusTracker[T comparable] struct {
 	txLatency           prometheus.Histogram
 }
 
-func NewPrometheusTracker[T comparable](reg *prometheus.Registry) (*PrometheusTracker[T], error) {
+type PrometheusRegistry interface {
+	prometheus.Registerer
+	prometheus.Gatherer
+}
```
RodrigoVillar, line 38: Similar to the `Logger` comment, could we defer to using an already existing `prometheus` type?

examples/morpheusvm/tests/e2e/e2e_test.go
```diff
@@ -114,9 +118,9 @@ func loadTxGenerators(
 	}
 
 	// Create tx generator
```
RodrigoVillar, line 120: Could we update the comment?

load/short_burst_orchestrator.go
qdm12: Very true, removed. I was asked to add some logs I'm wondering how we could best do that 🤔

tests/e2e/e2e.go
qdm12: Changed to

```go
// LoadIssuers is an initializer function since the node URIs are not known until runtime.
```

load/prometheus_tracker.go
qdm12: Sure, it's just good practice to accept exported function arguments as locally defined interfaces generally (`Logger`, `PrometheusRegistry`), to have narrower interfaces and decouple implementations.
But yes, it doesn't matter much in that case I suppose, so changed back.

examples/morpheusvm/tests/e2e/e2e_test.go
qdm12: Done, and variable below renamed from `txGenerators` -> `issuers`

examples/morpheusvm/load/dependencies.go
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package load
+
+import "github.com/ava-labs/avalanchego/ids"
+
+// Tracker keeps track of the status of transactions.
+type Tracker interface {
```
RodrigoVillar, line 9: Could we move the `Tracker` interface to the top level `load/` package?

qdm12: It's already in the load package :slightly_smiling_face: The point is to have narrower locally defined interfaces for exported symbols, especially exported constructor arguments, this allows some flexibility for the implementation plugged in and reduces package inter-dependencies. But this seems unpopular, so reverting this to import `load.Tracker[ids.ID]`!

RodrigoVillar (dismissed):
LGTM

michaelkaplan13 (dismissed):

RodrigoVillar (approved):

michaelkaplan13 (approved):

load/short_burst_orchestrator.go
qdm12: Solution is to handle this outside the orchestrator with another goroutine logging out the tracker state, if needed.

load/dependencies.go
```diff
@@ -7,24 +7,22 @@ import (
...
-	// Stop notifies the issuer that no further transactions will be issued.
-	// If a transaction is issued after Stop has been called, the issuer should error.
-	Stop()
-
-	// Issue sends a tx to the network, and informs the tracker that its sent
-	// said transaction.
-	IssueTx(ctx context.Context, tx T) error
+	// RegisterIssued informs the listener that a transaction was issued,
+	// and specifies if it was the last transaction to be issued.
+	RegisterIssued(tx T, last bool)
```
aaronbuchwald, line 25: Could we create a separate function to signal `last`? It seems that we're using a boolean to combine two distinct steps into a single function

load/listener.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (l *DefaultListener) markRemainingAsFailed() {
+	l.lock.RLock()
+	defer l.lock.RUnlock()
+	for txID := range l.inFlightTxIDs {
+		l.tracker.ObserveFailed(txID)
+	}
+	l.inFlightTxIDs.Clear()
+}
```
aaronbuchwald, line 91: Why are we grabbing a read lock for what looks like a write operation?

load/listener.go
```diff
@@ -0,0 +1,91 @@
...
+
+	lock          sync.RWMutex
+	issuedTxs     uint64
+	lastIssued    bool
+	receivedTxs   uint64
+	inFlightTxIDs set.Set[ids.ID]
+}
+
+// NewDefaultListener creates a new DefaultListener instance.
+// Set txTarget to 0 to never stop the [DefaultListener.Listen] method unless the context is cancelled.
```
aaronbuchwald, line 31: What is `txTarget` ?

tests/e2e/e2e.go
```diff
@@ -221,31 +223,43 @@ var _ = ginkgo.Describe("[HyperSDK Load Workloads]", ginkgo.Ordered, ginkgo.Seri
...
 		require.NoError(err)
 
-		issuers := make([]load.Issuer[*chain.Transaction], len(txGenerators))
-		for i := 0; i < len(txGenerators); i++ {
-			issuer, err := load.NewDefaultIssuer(uris[i%len(uris)], tracker)
-			require.NoError(err)
-			issuers[i] = issuer
+		agents := make([]load.Agent[*chain.Transaction, ids.ID], len(uris))
+		for i := range agents {
+			require.NoError(err, "creating websocket client")
```
aaronbuchwald, line 248: Duplicate error check?

load/gradual_orchestrator_test.go
```diff
@@ -192,47 +162,40 @@ func TestGradualLoadOrchestratorExecution(t *testing.T) {
...
-// To simulate TPS, the number of txs IssueTx can issue/confirm is capped by maxTxs
-func (m *mockIssuer) IssueTx(_ context.Context, id ids.ID) error {
 	if m.issueTxErr != nil {
-		return m.issueTxErr
+		return ids.ID{}, m.issueTxErr
 	}
 
 	if m.currTxsIssued >= m.maxTxs {
-		return nil
+		return ids.ID{}, nil
```
aaronbuchwald, line 186: Should this return a non-nil error as opposed to returning an empty ID? Unless an empty ID is specifically checked (which it won't be since this is a generic type from the perspective of the upstream component), then this will look like successful generation of a new tx.

If I'm misunderstanding, unless it's for a very simple reason I'm missing, would it make sense to add a comment in the code explaining as it seems non-obvious to me?

load/README.md
```diff
@@ -2,12 +2,12 @@
 
 This package provides generic utilities for blockchain load testing. We break load generation down into the following components:
 
-- tx generator(s)
 - tx issuer(s)
-- tracker
+- tx listener(s)
+- tracker(s)
```
aaronbuchwald, line 7: Should tracker stay singular here and in the graph below? Seems we still keep this to a single tracker in the code

load/gradual_orchestrator_test.go
RodrigoVillar: This should return a nil error. The point of `maxTxs` is to "simulate" the TPS of the network and so to simulate the case of there being an upper bound on TPS, we mock a successful TX issuance to the orchestrator. Errors are mocked in the `if`- statement above.

load/listener.go
qdm12: Good catch! Outdated comment, it's removed!

tests/e2e/e2e.go
qdm12: Good catch, removed!

load/listener.go
RodrigoVillar: Isn't `Clear()` a write operation though?

load/README.md
qdm12: Interesting point. I thought it might be useful to have a tracker per issuer, if we have N issuers issuing N different kind of transactions, to have metrics separated by issuer (=transaction kind). We can always share the same tracker across all "agents" (issuer+listener+tracker).

load/dependencies.go
qdm12: Added a `Listener` method `IssuingDone()` for this (similar to how it was with `Stop()`)

aaronbuchwald: Could we just keep the original name `Stop` ?

load/README.md
aaronbuchwald: The tracker is also currently used explicitly for load test wide metrics ie. total transactions confirmed across all of the agents. This means that if we use the same tracker, the orchestrator is expected to call a single function to get the sum across all agents as opposed to needing to explicitly sum all of the trackers together.

Could we simplify this by keeping it to a single tracker instance that is shared across all of them? Alternatively, how should we make sure not to create confusion in the future if we support both.

load/gradual_orchestrator_test.go
aaronbuchwald: Hmm seems this is not a change from before, but could produce confusing behavior to the orchestrator

load/dependencies.go
qdm12: IMO `Stop` is misleading since it does not actually stop the listener (at least in a blocking way), it tells the listener to start a timer to wait for the last outstanding transactions and then stop. Any other idea is welcome though

load/README.md
qdm12: Good point, we can have a single tracker for now, which could later be a global tracker and have a tracker per agent on top of this global tracker later. I don't think it would be that confusing to have, but let's not do it now. Changed back to a single global tracker per orchestrator!

aaronbuchwald (approved):

